PR #265, opened 2011-11-16
Fix for files and directories with + as a character in the file name b0rking Rack::Directory

VictorLowther:
Having a + in file and path names is a perfectly cromulent thing, and unescaping the name magically turns those + es into  es


jeremy:
giphy

PR #457, opened 2012-11-14
Configuration options for Rack::Deflate

jakubpawlowicz:
Rack::Deflater currently does not support any configuration options which makes it cumbersome to control.

This pull request adds support for the following options:
- **include** - an array of content types enabled for compression
- **if** - a lambda for choosing whether to deflate based on current execution scope (request, status, body, or headers)

Examples:
`use Rack::Deflater, include: %w(text/json application/json)`
`use Rack::Deflater, if: lambda { |env, status, headers, body| body.length > 512 }`

**EDIT**: description was update based on a PR discussion

### DIFF
--- a/lib/rack/deflater.rb
+++ b/lib/rack/deflater.rb
@@ -17,19 +17,26 @@ module Rack
   # directive of 'no-transform' is present, or when the response status
   # code is one that doesn't allow an entity body.
   class Deflater
-    def initialize(app)
+    ##
+    # Creates Rack::Deflater middleware.
+    #
+    # [app] rack app instance
+    # [options] hash of deflater options, i.e.
+    #           'if' - a lambda enabling / disabling deflation based on returned boolean value
+    #                  e.g use Rack::Deflater, :if => lambda { |env, status, headers, body| body.length > 512 }
+    #           'include' - a list of content types that should be compressed
+    def initialize(app, options = {})
       @app = app
+
+      @condition = options[:if]
+      @compressible_types = options[:include]
     end
 
     def call(env)
       status, headers, body = @app.call(env)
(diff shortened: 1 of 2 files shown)


jjb:
in case someone finds this thread, here's what I'm using for mime-types:

```ruby
include: Rack::Mime::MIME_TYPES.select{|k,v| v =~ /text|json|javascript/ }.values.uniq
```

campbecf:
I had to do `use Rack::Deflater, if: lambda { |env, status, headers, body| body.body.length > 512 }` to get the example to work.

jjb:
@campbecf What is "the example"?

jjb:
@campbecf looks like the example got fixed here :) : https://github.com/rack/rack/commit/c987ffa268f7f2e11ce4115296d9dc37b3a9e9c1

axelson:
Note: the example has been updated again in https://github.com/rack/rack/pull/1211

Now it is:

```
use Rack::Deflater, :if => lambda { |*, body| sum=0; body.each { |i| sum += i.length }; sum > 512 }
```

AlessandroMinali:
Even though I made that last update I was looking again at this today and this is a better solution:

```ruby
use Rack::Deflater, :if => lambda { |*, body| body.each { |i| return i.bytesize > 512 } }
```

The original change was due to the need to handle `Rack::File::Iterator`. Since the first chunk of the iterator is buffered then we can exit early vs wasting time iterating.

https://github.com/rack/rack/blob/a05f8d56f9ac4da14dddb8f312a3b43644f73397/lib/rack/files.rb#L177

As long as your threshold is below the buffer size then we can use this quicker evaluation.

I won't update the code documentation as this is not a general solution.

PR #479, opened 2013-01-02
Rack::File serves gzipped assets when available

jbaudanza:
@raggi, After poking around a bit, I think Rack::File is the cleanest place to put this change. 

This is a very small addition to Rack::File. I think adding it to Rack::Static would be slightly more complex. For example, since Rack::File would set all compressed content-types to "application/gzip", we would need to duplicate the Rack::File mime type logic in Rack::Static to correct this.

I looked at the header_rule stuff in Rack::Static, but couldn't see any graceful way to reuse it.

LMK what you think.


vipulnsward:
Looks like this got lost. Anything needed to get this past the line?

jbaudanza:
@vipulnsward This functionality now exists in Rails, which is probably where it belongs. Is there any interest in having this in Rack? I was waiting for feedback from a maintainer before doing any more work on it.

vipulnsward:
@jbaudanza maybe I missed this, where in Rails?

vipulnsward:
Ah, got it. Thanks.

schneems:
I added it here https://github.com/rails/rails/pull/16466

gumatias:
Are there still any interest in merging this PR? It's been open for over 3 years and no activity for the past 8 months.

jjb:
I have no significant interest, because it's now in rails 4.2+ (if i'm reading https://github.com/rails/rails/commit/90006acb09605756aa3914874102bd20971c422c correctly). @jbaudanza, what do you think?

In general it does seems appropriate for this functionality to live in Rack instead of Rails, but I'm saying that without first examining if there is anything special about the Rails implementation and/or what the history is of other Rails middleware being developed independent of Rack or rack-contrib.

schneems:
I wrote the Rails implementation, mostly cribbed from a few other implementations such as this one. I'm torn, I would love to move that behavior to Rack so other frameworks can get the benefit. 

I see one downside of making it a default and one complicating factor. The downside is that there is a perf cost for the extra file lookup in the Rails case we can safely assume that you'll have gzipped files since we also generate gzipped files via asset pipeline. That's an assumption we can't make with all other frameworks. The complicating factor is that different people may desire custom functionality for example they may want to support browsers that support `brotli` in addition to those that support `gzip`. I'm guessing in the future there will be even more file formats and potentially even more logic that we want to support.

jjb:
It could be a standalone middleware or a configurable feature on Rack::File.

multiple file types would perhaps be easy to support with a simple mime type / extension mapping

probably the more important question is why does Rails maintain ActionDispatch::Static instead of using Rack::File. Seems like it would be a noble goal to either

1. use it and contribute changes to it
2. subclass it
3. use it alongside another complementary middleware

(that's a big topic and i'm not proposing we dive into that here, but i'm just noting it's probably a more useful discussion to have than the one about this particular PR/functionality).

schneems:
> why does Rails maintain ActionDispatch::Static instead of using Rack::File

ActionDispatch::Static does use Rack::File however Rails supports things that Rack::File doesn't. The biggest thing that comes to mind is the html index rendering support. For example if you go to http://localhost:3000/home then it will render `public/home.html` if it exists or `public/home/index.html` if that exists.

jjb:
> ActionDispatch::Static does use Rack::File

o rite 😊

raggi:
I thought that was added to file a while back, or maybe that was dir?

raggi:
The bigger problem trying to merge rails static and racks file here is that rails makes certain design assumptions that turn up in many places. This isn't just limited to gzip serving, it's also encoding decisions (rails can assert utf-8 defaults, but rack does not limit itself this way). The rails gzip serving is actually limited to specific content types, which is not the approach one would take for a generic pre-computed-gzip serving implementation (such as the one in nginx). These concerns are subtle, and a glance at the code won't make them obvious, but they're distinct differences in implementation goals that end up as bugs, most frequently for non-US users.

rmacklin:
Is this PR still needed after https://github.com/rack/rack/pull/863 ?

schneems:
Doesn't look like it. Can someone with commit close?

PR #559, opened 2013-05-19
Support multiple byte-range requests

rakoo:
As stated, support multiple byte-range requests for serving static files. The drawback is that the `each` method yields the headers and the `\r\n` after the content separately, leading in a less efficient use of the socket.


zzak:
I guess this needs to be rebased :/

jeremyevans:
Closing this as I think we are probably going with the implementation in #1420.

PR #561, opened 2013-05-21
Ensure request.POST always raises EOFError on bad input

jdelStrother:
Before this fix, if you had a bad multipart request, request.POST would only raise EOFError the first time it was called, and would then return nil on subsequent invocations.  This would typically result in the cryptic "can't convert nil into Hash" as a result of calling request.params.

WDYT?


aselder:
@spastorino @raggi Will Rack 1.4.6 ever be released? Those of us poor souls stuck in Rails 3 would appreciate this fix. :)

raggi:
@tenderlove is maintainer now

PR #644, opened 2014-01-16
Fix default Content-Type (RFC2616 7.2.1)

yegortimoshenko:
This patch fixes the behaviour of `Content-Type` header if it was not included in the request according to [RFC 2616 section 7.2.1](http://www.w3.org/Protocols/rfc2616/rfc2616-sec7.html#sec7.2.1). It also fixes sinatra/sinatra#804 issue.

### DIFF
--- a/test/spec_request.rb
+++ b/test/spec_request.rb
@@ -26,7 +26,7 @@ describe Rack::Request do
     req.port.should.equal 8080
 
     req.content_length.should.equal "0"
-    req.content_type.should.be.nil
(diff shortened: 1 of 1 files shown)


yegortimoshenko:
What's the status?

raggi:
It's a breaking change and it missed 2.0 @tenderlove

yegortimoshenko:
I don't think that is the problem with Rack anymore. I think it should be defined in server's internals.

PR #647, opened 2014-01-19
Replace direct comparison with Range#cover

ArtemPyanykh:
It seems more idiomatic to me to use Range#cover?


tenderlove:
Please send this [upstream](https://github.com/kr/okjson). We don't maintain okjson, we just embed it.  I don't want to carry custom patches against it.

raggi:
I don't see any point in this either, it will make the code significantly slower.

PR #649, opened 2014-01-21
Refactor URLMap

artempyanykh:
URLMap mixes two different concerns: application mapping and request matching. I believe separating 'matching' concern into other class will lead to better code structure and easier maintanance.

This PR contains just minor 'Extract Class' refactoring. It conforms URLMap spec and doesn't change routes ordering.


artempyanykh:
Closing due to inactivity.

PR #666, opened 2014-03-02
return "Accept-Ranges" header when making file requests

nuex:


nuex:
Closing since it was only for my specific use case for testing large file transfers using Rack.

PR #696, opened 2014-06-07
Rename files to follow ruby naming conventions

teoljungberg:
Rename files to follow ruby naming conventions

If a class is named NullLogger, it’s file would be named null_logger.rb


tenderlove:
Rebase this please?

teoljungberg:
Done @tenderlove

Issue #658, opened 2014-02-12
app error: deadlock; recursive locking (ThreadError) with Rack::Deflater


jonhyman:
Last week, we moved some servers over to Rackspace's cloud and started getting a bunch of deadlocks (once every few minutes) in our application. Before I had seen this very very infrequently, and had read through various issues which describe people running into this when enabling `Rack::Deflater` (e.g., #349, https://gist.github.com/zporter/2936818).

We're using rack 1.4.5, I dug into the source code of the gem and confirmed that the body is getting closed.

However, I can now consistently reproduce a deadlock in our application when we `use Rack::Deflater` in our config.ru. If I remove it, I can no longer reproduce the deadlocks. I imagine that there might be something different about the Rackspace environment since the timing was too coincidental (and nothing else changed), but removing `Rack::Deflater` and gzipping via nginx fixes this for me.

```
Feb 12 17:37:16 dashboard-app-st-02 stderr.log:  E, [2014-02-12T22:37:16.239518 #32444] ERROR -- : app error: deadlock; recursive locking (ThreadError)
Feb 12 17:37:16 dashboard-app-st-02 stderr.log:  E, [2014-02-12T22:37:16.239683 #32444] ERROR -- : /var/www/dashboard/shared/bundle/ruby/2.0.0/gems/rack-1.4.5/lib/rack/lock.rb:14:in `lock'
Feb 12 17:37:16 dashboard-app-st-02 stderr.log:  E, [2014-02-12T22:37:16.239726 #32444] ERROR -- : /var/www/dashboard/shared/bundle/ruby/2.0.0/gems/rack-1.4.5/lib/rack/lock.rb:14:in `call'
Feb 12 17:37:16 dashboard-app-st-02 stderr.log:  E, [2014-02-12T22:37:16.239760 #32444] ERROR -- : /var/www/dashboard/shared/bundle/ruby/2.0.0/gems/actionpack-3.2.10/lib/action_dispatch/middleware/static.rb:62:in `call'
Feb 12 17:37:16 dashboard-app-st-02 stderr.log:  E, [2014-02-12T22:37:16.239793 #32444] ERROR -- : /var/www/dashboard/shared/bundle/ruby/2.0.0/gems/rack-cache-1.2/lib/rack/cache/context.rb:136:in `forward'
[...]

PR #721, opened 2014-08-11
Rack 1.5: app error: deadlock; recursive locking (ThreadError) with Rack::Deflater

thoughtless:
Based on the comments at https://github.com/rack/rack/issues/658 the fix already on master seems to work. This backports the relevant commits to the 1.5 branch.

Note, I only ran `rake test` on `ruby 1.9.3p392`. I didn't run the tests on other ruby versions. I didn't run the "full" test suite with `rake fulltest`.

### DIFF
--- a/SPEC
+++ b/SPEC
@@ -207,7 +207,7 @@ but only contain keys that consist of
 letters, digits, <tt>_</tt> or <tt>-</tt> and start with a letter.
 The values of the header must be Strings,
 consisting of lines (for multiple header values, e.g. multiple
-<tt>Set-Cookie</tt> values) seperated by "\n".
+<tt>Set-Cookie</tt> values) separated by "\n".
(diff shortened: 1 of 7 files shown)


tenderlove:
I think 1.6 doesn't have this bug.  1.5.x is only receiving security updates at this point.

thoughtless:
The fix was back-ported from the 1.6 branch, so 1.6 is fine. If you don't plan on merging this, you might as well close the PR.

Now that Rails 4.2 is out (and using Rack 1.6) this isn't as important. It still affects those of us on Rails 4.1, but we should be making plans to upgrade to Rails 4.2 anyway :)

tenderlove:
Okay, closing.  @thoughtless thanks for backporting anyway.  I appreciate the work. Sorry it took so long to get around to this.

PR #722, opened 2014-08-11
Rack 1.4: app error: deadlock; recursive locking (ThreadError) with Rack::Deflater

thoughtless:
Based on the comments at https://github.com/rack/rack/issues/658 the fix already on master seems to work. This backports the relevant commits to the 1.4 branch.

Note, I only ran `rake test` on `ruby 1.9.3p392` and `ruby 1.9.2p326`. I didn't run the tests on other ruby versions. I didn't run the "full" test suite with `rake fulltest`.


nextofsearch:
Has this pull request ever merged to master? If so, which version of 1.4.x would it be?

PR #741, opened 2014-10-02
Follow-up for the pull request #737

soulim:
In the pull request #737 I've noticed that some of strings weren't replaced with constants (e.g. `"HEAD"` [here](https://github.com/rack/rack/pull/737/files#diff-894804c816a11584eb28499cb8bbe396R73) or `"REQUEST_PATH"` [here](https://github.com/rack/rack/pull/737/files#diff-0a1c17672485b0f62a7f6a89a1f8de42R40)).

So I decided to fix that, and also converted other commonly used strings into constants. Hope it will add more consistency.


soulim:
Thank you! :heart:

PR #749, opened 2014-10-24
Memoizing Request#POST

Wardrop:
I've added logic to my application so when JSON data is POST'd to it, it's automatically parsed and merged with Request#POST. The processing is meant to transparent to the rest of the application. This is how I deal with JSON data posted to the server by JavaScript code on the client.

  request.POST.merge!(JSON.parse('...'))

The problem is, if no form data has been POST'd, Request#POST returns a new empty hash every time it's called, so my `merge!` doesn't stick. Looking at the source code for Request#POST, it's easy to see the problem. I'm wondering if we can modify Request#POST so it memoizes the empty hash it returns if no form data exists, instead of returning a new hash every time.

Otherwise, I'm open to suggestions on how I could go about this another way that achieves the same result, but Request#POST returning a new hash every time seems wrong either way.

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -199,28 +199,27 @@ module Rack
     # This method support both application/x-www-form-urlencoded and
     # multipart/form-data.
     def POST
-      if @env["rack.input"].nil?
-        raise "Missing rack.input"
-      elsif @env["rack.request.form_input"].equal? @env["rack.input"]
-        @env["rack.request.form_hash"]
-      elsif form_data? || parseable_data?
-        unless @env["rack.request.form_hash"] = parse_multipart(env)
-          form_vars = @env["rack.input"].read
+      raise "Missing rack.input" if @env["rack.input"].nil?
+      @env["rack.request.form_hash"] ||= {}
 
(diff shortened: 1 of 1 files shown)


ioquatix:
We could do this, but I think we should also freeze the cached data.

jeremyevans:
We already cache the `params`, so caching `POST` seems reasonable.  For consistency, we should probably also cache `GET` as well if we do this.  Note that `params` caching occurs as a separate `Request` instance variable (not in `env`), so presumably `POST` and `GET` caching should operate the same, which is a different implementation than this PR.

ioquatix:
@jeremyevans do you still agree with your assessment? I agree with it too. However, I think keeping all state within the `Request` instance without modifying `env` would be the most appropriate implementation. WDYT?

jeremyevans:
Yep, I still think it's reasonable to cache in a Request instance variable.

jeremyevans:
I did a more detailed review of this and I think the only issue is we already always cache GET results in env, and already cache POST results in env, except for the corner case where we don't try parsing the input data due to it not having a recognized content type.  We should cache in that case as well, using the same approach as other GET/POST caching.  I've submitted #1795 to implement that.

PR #765, opened 2014-12-14
avoid ; to separate statements or expressions

arturocadena:
I think that this is a little more clear and it follows the ruby style guide.


tenderlove:
Can you change it to multiple lines?

PR #766, opened 2014-12-14
Make WEBrick handler use Tempfile for large input

vais:
WEBrick handler is reading the entire request body into a giant string in memory. Usually this is avoided by setting an upper limit for input held in memory and switching to a Tempfile once that limit is exceeded. Puma and Thin both have this maximum value set to 1024 * (80 + 32) #=> 114688 bytes, so I cargo-culted that number as well.

### DIFF
--- a/lib/rack/handler/webrick.rb
+++ b/lib/rack/handler/webrick.rb
@@ -60,7 +60,16 @@ module Rack
         env = req.meta_vars
         env.delete_if { |k, v| v.nil? }
 
-        rack_input = StringIO.new(req.body.to_s)
+        rack_input = StringIO.new
+        read_chunk = lambda do |chunk|
+          rack_input << chunk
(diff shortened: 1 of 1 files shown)


vais:
@raggi @spastorino guys is there a technical reason this cannot be merged? If I'm doing it wrong, I'd like to know. Thanks!

badosu:
Does this make any difference on issues like this? http://stackoverflow.com/questions/24639701/random-slow-rackmethodoverridecall-on-rails-app-on-heroku

vais:
@tenderlove, please allow me to put forth this, the gentlest of bumps. It's been a while, but I still would like to get this fix (or something like this) into rack to prevent denial of service on WEBrick. Could you please take a look and offer your feedback?

P.S. It looks like Travis build failed for unrelated reasons, so re-running it should take care of the problem.

PR #774, opened 2015-01-02
a little more clear Rack::Server.default_middleware_by_environment

itsNikolay:


itsNikolay:
still can not understand why jruby fails

Issue #779, opened 2015-01-05
Backporting multipart_part_limit on the 1.5.X branch


byroot:
Is there any chance the `multipart_part_limit` set of commit could be backported on the `1.5` branch?

We have to maintain a fork of rack because of this and to use rack `1.6` we need to upgrade to actionpack 4.2 which won't happen before months.

To not break compatibility the feature should probably be disabled by default.

Would you guys be ok with this? Alternatively a minor version of actionpack 4.1 could allow the use of rack 1.6.

What do you think @tenderlove or @spastorino?

tenderlove:
I think it might be ok to backport.  Would you mind sending a PR so we can see the diffs? (If it's not too much work)

byroot:
I would do it with please, I just don't know how to create a PR against another branch than master `/o\`. I need to check that up.

tenderlove:
@byroot there should just be a dropdown when you send the PR.  Just select the branch from that.

spastorino:
I was also talking with Aaron about relaxing Rack dependency on Rails side. At least we could do that in master ~> 1.x then ~> 2.x it should be that way and Rack shouldn't break Rails.

byroot:
See: https://github.com/rack/rack/pull/783

PR #783, opened 2015-01-08
Multipart part limit backport

byroot:
Fixes: https://github.com/rack/rack/issues/779

I disabled the feature by default so it won't create bad surprises for people that upgrade.

@tenderlove @spastorino for review please.

cc @boourns @orenmazor

### DIFF
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -2,6 +2,8 @@ require 'rack/utils'
 
 module Rack
   module Multipart
+    class MultipartLimitError < Errno::EMFILE; end
+
     class Parser
       BUFSIZE = 16384
 
@@ -14,7 +16,13 @@ module Rack
 
         fast_forward_to_first_boundary
 
+        opened_files = 0
         loop do
+          if Utils.multipart_part_limit > 0
+            raise MultipartLimitError, 'Maximum file multiparts in content reached' if opened_files >= Utils.multipart_part_limit
+            opened_files += 1
+          end
+
           head, filename, content_type, name, body =
             get_current_head_and_filename_and_content_type_and_name_and_body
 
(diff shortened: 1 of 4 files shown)


tenderlove:
I'm fine with this.  @spastorino wdyt?

spastorino:
I don't think we should backport features in Rack. Given the careful way releases are made Rack users should just be able to easily upgrade from 1.4 to 1.5 or even to 1.6.

This is a Rails issue and I think we should relax Rack dependency on Rails side.

I know that given the things Rails does with Request, Session, etc that may not be trivial but I still think that trying that is maybe a better idea.

@tenderlove I will leave the decision up to you bro :) :heart: :heart: :heart:

tenderlove:
@spastorino if we can easily relax the dependency on the Rails side, it seems better if we do it there (IMO).

byroot:
I also think relaxing the dependencies on the Rails side is a better idea.

spastorino:
Done in master https://github.com/rails/rails/commit/da175a2b46c4db59cd8c4bd0b9c6c613651e3db0 and 4-2-stable https://github.com/rails/rails/commit/d5e329e8ce1e725405f096a44fc4a14e25c1c785

This won't have any impact anyway because it won't be a 1.7 and we shouldn't do this kind of aggressive change in 4-1-stable. So your issue is not fixed by any of this stuff :(.

I guess your alternatives now are 

1. upgrade to 4.2
2. monkey patch Rack's/Rails' stuff to provide what you want

Ping me if I can help you in any way ...

byroot:
We're already working on upgrading to 4.2 but it will take a while.

But I totally understand the issue, we'll keep our fork for a little while. Thanks again!

byroot:
cc @arthurnn

PR #785, opened 2015-01-16
Add support for custom hmac.

frodsan:


tenderlove:
This change doesn't seem to be equivalent.  Can you pass a string instead of an object?  Also, could you add a test?

frodsan:
I think it iss supported: https://github.com/ruby/ruby/blob/master/ext/openssl/ossl_digest.c#L40..L52.

But I will revert it. I will pass the class instead. I will add a test. Ty.

tenderlove:
> But I will revert it to avoid confusion. I will pass the class instead. I will add a test. Ty.

Sounds good. Let me know when you've updated and I'll merge this.

frodsan:
Closing this in favor of https://github.com/rack/rack/pull/958

PR #790, opened 2015-01-28
Update SPEC

yeonhoyoon:
fix typo.


spastorino:
:+1:, BUT you should update `lib/rack/lint.rb` and then run `rake SPEC` and SPEC file will be updated accordingly.

yeonhoyoon:
okay. created another pull request #793

PR #791, opened 2015-01-28
Add woff2 mime type

tagliala:
Reference:
http://dev.w3.org/webfonts/WOFF2/spec/#IMT (still a draft)

Browser support:
* Google Chrome 36
* Opera 23
* Firefox 35 ([disabled by default on stable and beta channels, enabled in aurora and nightly](https://developer.mozilla.org/en-US/docs/Web/Guide/WOFF))

I'm temporary using

```rb
Rack::Mime::MIME_TYPES['.woff2'] = 'application/font-woff2'
```


tagliala:
Travis fails should not be caused by my PR

PR #798, opened 2015-02-15
Prioritise CLI options over comment options

smcgivern:
I'll start with my use case: I want to provide a default port for an app, but still be able to override it when launching from the command-line.

I found four sources of options for `Rack::Server`:

1. In the constructor, as a hash: `Rack::Server.new(server: 'thin')` (if this is provided, 4 and 2 won't be used.)
2. On the command line: `bundle exec rackup --server thin`
3. In a magic comment on the first line of config.ru (or whichever config file was used): `#\ --server thin`
4. `Rack::Server#default_options`

Previously, the order of precedence - earlier 'wins' - was 3, 2, 4. This changes it to 2, 3, 4. That way you can have default options in the magic comment, but still let them be overridden by passing options on the command line. This way makes far more sense to me, but I appreciate it's a breaking change.

### DIFF
--- a/lib/rack/handler.rb
+++ b/lib/rack/handler.rb
@@ -46,10 +46,6 @@ module Rack
     def self.default(options = {})
       # Guess.
       if ENV.include?("PHP_FCGI_CHILDREN")
-        # We already speak FastCGI
-        options.delete :File
-        options.delete :Port
-
         Rack::Handler::FastCGI
       elsif ENV.include?(REQUEST_METHOD)
         Rack::Handler::CGI
--- a/lib/rack/server.rb
+++ b/lib/rack/server.rb
@@ -99,7 +99,7 @@ module Rack
           abort opt_parser.to_s
         end
 
-        options[:config] = args.last if args.last
+        options[:config] = args.last if args.last && !args.last.empty?
         options
       end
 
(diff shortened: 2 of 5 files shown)


smcgivern:
As a side note, if anyone else has my use case, they can do this by using `Module#prepend` in Ruby 2, or by aliasing methods in earlier versions, but it's verbose and feels like a hack:

```ruby
module DefaultPort
  def options
    if super[:Port] == default_options[:Port]
      super.merge(Port: 2929)
    else
      super
    end
  end
end

class Rack::Server
  prepend DefaultPort
end
```

rkh:
:+1: on the feature

smcgivern:
Whoops, should have installed lightppd and run those specs first. They are fixed now (`Rack::Handler.default` was trying to modify the options, which won't work now).

Some specs seem to be intermittently failing during constant lookup in `Rack::Cascade`, but I don't think I caused that. I don't know what to make of the [Rubinius failure](https://travis-ci.org/rack/rack/jobs/50830599), I'll have to take a look later.

smcgivern:
The Rubinius specs were failing because I was modifying `ARGV` in the specs. Fixed now by allowing `Rack::Server` to use `SPEC_ARGV` instead if it's defined.

The build is green except for JRuby failing on the constant loading issue I mentioned above.

tenderlove:
> I found four sources of options for Rack::Server:

Any way we can get that up to 5 sources? I kid tho, it seems fine.  The 1.7 ship has already sailed, so

PR #800, opened 2015-02-17
Fix bug when `qs` is `nil`

eileencodes:
In c4596b3 I failed to test Rails tests :cry: and in working on other issues discovered some Rails ActionPack tests were erroring. I found that sometimes `qs` is nil so I changed this to check for `qs.nil? || qs.empty?`.

I'm going to investigate as to why we're calling this method when `qs` is nil but this will fix it for now until I figure out if we can skip this call. :grin:

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -109,7 +109,7 @@ module Rack
     # ParameterTypeError is raised. Users are encouraged to return a 400 in this
     # case.
     def parse_nested_query(qs, d = nil)
-      return {} if qs.empty?
+      return {} if qs.nil? || qs.empty?
       params = KeySpaceConstrainedParams.new
 
       (qs || '').split(d ? /[#{d}] */n : DEFAULT_SEP).each do |p|


spastorino:
It expect a query string so I'd say that Rails is doing the wrong thing ... please do investigate :)

tenderlove:
@spastorino the method already supports nil query strings, so @eileencodes change is backwards compatible.  We should figure out why rails is calling with nil though.

spastorino:
@tenderlove indeed it is backwards compatible, it may just be adding unneeded behavior.
The method didn't support nil query strings on Rack side, it expects a string that's mentioned on the doc and also on code first thing ir does is qs.empty? that doesn't work with nil.

Agreed with your last phrase we need to figure out why Rails is calling with nil and then maybe revert this merge?.

tenderlove:
@spastorino we need to keep this merge because the method has always handled nil values for parameters.  We also need to figure out why Rails is calling it with `nil` so we can deprecate the behavior.

PR #806, opened 2015-02-19
Update documentation of Rack::Response defaults

tomstuart:
The “containing HTML” remark stopped being true in 3623d04526b953a63bfb3e72de2d6920a042563f, as reflected by https://github.com/rack/rack/blob/master/test/spec_response.rb#L5-L21.


rkh:
:+1: thanks

PR #810, opened 2015-02-27
Use yield instead of block.call

DamirSvrtan:
yielding is faster than performing block.call 

[Reference](https://www.omniref.com/ruby/2.2.0/symbols/Proc/yield?#annotation=4087638&line=711)

### DIFF
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -15,7 +15,7 @@ module Rack
 
     class LintError < RuntimeError; end
     module Assertion
-      def assert(message, &block)
+      def assert(message)
         unless yield
           raise LintError, message
         end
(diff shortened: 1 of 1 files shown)


lib/rack/lint.rb
```diff
@@ -16,7 +16,7 @@ def initialize(app)
     class LintError < RuntimeError; end
     module Assertion
       def assert(message, &block)
```
davydovanton: i think that you need delete excess `&block` argument in method definition

DamirSvrtan:
Thnx @davydovanton for the suggestion! 

It seems that this build is broken not due to changes. It happened also on another pull request. Does this happen alot?

PR #811, opened 2015-02-27
Prefer attr_reader over self-written getters.

DamirSvrtan:
attr_reader is somewhat faster than writing plain getters, but this also increases readability.

### DIFF
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -86,9 +86,8 @@ module Rack
       end
 
       private
-      def full_boundary; @full_boundary; end
 
-      def rx; @rx; end
+      attr_reader :full_boundary, :rx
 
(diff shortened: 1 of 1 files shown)


tenderlove:
Won't this cause warnings?

DamirSvrtan:
Ough, I guess in Ruby 1, but not in 2 as i see. Will close.

chastell:
FWIW (note: not sure whether it _is_ worth it), the warning can be worked around with

```ruby
# …public/protected methods…

attr_reader :full_boundary, :rx
private     :full_boundary, :rx

private

# …rest of private methods
```

or using [private_attr](https://github.com/jswanner/private_attr). Hopefully this warning [goes away in Ruby 2.3](https://github.com/ruby/ruby/pull/849).

PR #812, opened 2015-02-28
Remove redundant nil check and use Symbol#to_proc.

DamirSvrtan:


lib/rack/tempfile_reaper.rb
```diff
@@ -14,7 +14,7 @@ def call(env)
       env['rack.tempfiles'] ||= []
       status, headers, body = @app.call(env)
       body_proxy = BodyProxy.new(body) do
-        env['rack.tempfiles'].each { |f| f.close! } unless env['rack.tempfiles'].nil?
+        env['rack.tempfiles'].each(&:close!)
```
tenderlove: Since the env hash is mutable, doesn't it mean that `@app.call` could delete the key?

DamirSvrtan: I saw that it might be mutable, but didn't see the call mutating it.

However, if the nil check is left in, what's the purpose of setting `env['rack.tempfiles'] ||= []` at the beginning of the method?

ioquatix:
This has been improved elsewhere. Thanks for your contribution.

PR #813, opened 2015-03-01
Remove check for `nil` on `qs`

eileencodes:
Fixed by brynary/rack-test@4a4b2c1. rack-test was turning the query
string to nil if there was no argument passed in the params. This extra
`nil` check is no longer necessary after the changes made to the
rack-test gem. Fixes issue brought up in #800.

I found Rails wasn't doing anything wrong as we had originally thought. Since rack-test update has been released should I update rack-test to use 0.6.3 instead of 0.6.2 in Rails master as well? 

cc/ @tenderlove @spastorino

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -109,7 +109,7 @@ module Rack
     # ParameterTypeError is raised. Users are encouraged to return a 400 in this
     # case.
     def parse_nested_query(qs, d = nil)
-      return {} if qs.nil? || qs.empty?
+      return {} if qs.empty?
       params = KeySpaceConstrainedParams.new
 
       (qs || '').split(d ? /[#{d}] */n : DEFAULT_SEP).each do |p|


spastorino:
@eileencodes :+1: bump rack-test on Rails side too and /cc me

eileencodes:
Hey @spastorino - I realized that rack-test with the fix for this hasn't been released yet. Rails' tests passed and my local app passed because I was depending on Rails master - but all of those are using `~> 0.6.3` from Rails.

I totally somehow missed the date of the rack-test release was in January and that I was grabbing changes from rack-test master. So technically it's "fixed" but if anyone is using rack-test 0.6.3 and below and rack master with rails they'll get the error I caused earlier `NoMethodError: undefined method 'empty?' for nil:NilClass`. David found this on a Basecamp app and I realized the gem dependency is set to exactly `0.6.3` which breaks, not `~>` 0.6.3 which works fine. Setting our Gemfile to ~> 0.6.3 fixes the issue and it works fine for brand new applications btw.

I'm really sorry I wasn't paying more attention and that we have to do this...again. :disappointed: I'm not sure what the best route is here. To tell anyone with rack-test and rack master to make sure it's `~> 0.6.3` or to revert. Let me know what you think.

spastorino:
Can't rack-test be released? AFAIK rack-test already had a couple of worth releasing changes.

PR #814, opened 2015-03-02
Only count files (not all form elements) against the Multipart File Limit

johnnaegle:
We were getting "Too many open files - Maximum file multiparts in content reached".  However, our form only had one file input.  It had several hundred other form elements.  Each was counted as a file. 

Only increment the opened_files counter for file inputs.  If you have a large form, you can quickly hit the MultipartPartLimitError, even though you aren't actually opening files

### DIFF
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -54,13 +54,14 @@ module Rack
 
         opened_files = 0
         loop do
-          if Utils.multipart_part_limit > 0
-            raise MultipartPartLimitError, 'Maximum file multiparts in content reached' if opened_files >= Utils.multipart_part_limit
-            opened_files += 1
-          end
 
           head, filename, content_type, name, body =
             get_current_head_and_filename_and_content_type_and_name_and_body
+
+          if Utils.multipart_part_limit > 0
+            opened_files += 1 if filename
+            raise MultipartPartLimitError, 'Maximum file multiparts in content reached' if opened_files >= Utils.multipart_part_limit
+          end
 
           # Save the rest.
           if i = @buf.index(rx)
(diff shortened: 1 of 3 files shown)


johnnaegle:
Has anybody looked at this? It seems like quite the bug to count all fields as files when parsing multipart requests.

tenderlove:
Hesitant to pull this in because I know the limit is to do with a security issue.  I'm not sure where the file handles are actually created, so I don't know if this is an OK change to make.  @spastorino can you take a look?  Seems like you signed off on b0b5fb94

spastorino:
Can you squash your commits?

johnnaegle:
squashed

ZachBeta:
We're running into this issue as well. Any thoughts on when 1.6.1 will be released?

sjain:
The fix seems to be in master, but not in version 1.6.1. When is next version expected?

paneq:
Could you backport this to 1.5 and release?

kuahyeow:
We have just hit this issue when Nginx (with Passenger) started erroring with a `502` with the error message:

```
upstream prematurely closed connection while reading response header from upstream
```

Turning on debug log level pointed to the Rack::Multipart::MultipartLimitError, which was announced recently as a security release for 1.5.4

We could workaround by setting the RACK_MULTIPART_PART_LIMIT env var for now, but it would be nice to release a proper fix for this which doesn't count non-files towards the limit ?

gchan:
Thanks for resolving this issue and releasing new versions so promptly! @johnnaegle @tenderlove  :heart:

wnm:
I'm also getting a  "Too many open files - Maximum file multiparts in content reached" error.

I'm using rails's `accepts_nested_attributes_for` to edit multiple entries in one huge form. 

So **there are** multiple input fields on one page, but even if none of them is beeing changed (an image uploaded) I still get the error.

Any ideas on how to fix that?

PR #817, opened 2015-03-04
Pass files under 4 MB to webrick thought memory

vladimir-tikhonov:
As described in https://github.com/rails/rails/issues/18828 passing IO object instead of file content to Webrick causes huge speed decrease (426 MB/s vs 907 KB/s on my machine).


vladimir-tikhonov:
Strange ci fail, seems to be not related to changes

oskarpearson:
@vladimir-tikhonov There's some discussion (https://github.com/rails/rails/issues/18828#issuecomment-84577902) in the rails ticket about fixing an underlying problem in the webrick config.

It's probably best to hold of merging this for now, I think.

PR #819, opened 2015-03-06
Expand initialize ivars instead of using 1 line

eileencodes:
Flattening the ivars here causes them to be allocated as an array.
This commit expands them to avoid these extra allocations.

Total allocations before: 2913000
Total allocations after:  2892000

(I promise this one didn't break rails :grin:)


spastorino:
Is this a bottleneck? anyway, merging because of aesthetics :smile:

PR #820, opened 2015-03-08
Allow specifying a params hash class to use when parsing

jeremyevans:
Currently, Rack is hardcoded to use KeySpaceConstrainedParams
as the params hash class, and this commit doesn't change that
behavior.

Rails and Sinatra and other web frameworks that want to allow
indifferent access to the params generally end up making a deep
copy of the params.  This is inefficient.  If they could overide
the params hash class to use the type of indifferent hash they
want, they wouldn't need to make a deep copy, they could use the
hash Rack provides.  This commit makes this possible without
monkey patching.

Note that this isn't really a perfect solution, since changing
the params_hash_class for the Utils class is still a global
change.  Ideally, this code would be in a class separate from
Rack::Utils, which you could subclass and then inject that
dependency into whatever Request subclass you use for the
framework.

I'm certainly willing to do the work of making it possible for
local overrides of the params hash class, while keeping
backward compatibility.  But as that's a bit more involved,
I'd like to get your feedback on this first part before doing
more work.

### DIFF
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -10,7 +10,7 @@ module Rack
 
       DUMMY = Struct.new(:parse).new
 
-      def self.create(env)
+      def self.create(env, params = QueryParser::DEFAULT.params_class.new)
         return DUMMY unless env['CONTENT_TYPE'] =~ MULTIPART
 
         io = env['rack.input']
@@ -23,13 +23,13 @@ module Rack
           lambda { |filename, content_type| Tempfile.new(["RackMultipart", ::File.extname(filename)]) }
         bufsize = env['rack.multipart.buffer_size'] || BUFSIZE
 
-        new($1, io, content_length, env, tempfile, bufsize)
+        new($1, io, content_length, env, tempfile, bufsize, params)
       end
 
-      def initialize(boundary, io, content_length, env, tempfile, bufsize)
+      def initialize(boundary, io, content_length, env, tempfile, bufsize, params = QueryParser::DEFAULT.params_class.new)
         @buf            = "".force_encoding(Encoding::ASCII_8BIT)
 
-        @params         = Utils::KeySpaceConstrainedParams.new
+        @params         = params
         @boundary       = "--#{boundary}"
         @io             = io
         @content_length = content_length
--- a/lib/rack/multipart.rb
+++ b/lib/rack/multipart.rb
@@ -21,8 +21,8 @@ module Rack
     MULTIPART_CONTENT_ID = /Content-ID:\s*([^#{EOL}]*)/ni
 
     class << self
-      def parse_multipart(env)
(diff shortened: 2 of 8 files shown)


spastorino:
@jeremyevans :+1:

spastorino:
@jeremyevans just in case because my :+1: is to vague ... I agree with the current PR and I second the idea of moving this out of Utils and allowing to subclass.

jeremyevans:
@spastorino  Thanks. I should be able to work on that soon.  I'll push some additional commits when it is ready.

jeremyevans:
Travis failure is only for ruby 2.0 (all other implementations pass), and appears to be unrelated:

    /home/travis/.rvm/gems/ruby-2.0.0-p598/gems/eventmachine-1.0.7/lib/eventmachine.rb:1478:in `event_callback': received ConnectionUnbound for an unknown signature: 11 (EventMachine::ConnectionNotBound)

rkh:
:+1: and having this a non-global setting would be amazing

jeremyevans:
@spastorino @tenderlove I think this is ready to be committed, could one of you please review and merge or let me know if you'd like any changes?

spastorino:
@jeremyevans thanks for the work you've done. I will try to review as soon as I have some free time.

lib/rack/multipart/parser.rb
```diff
@@ -22,18 +22,19 @@ def self.create(env)
         tempfile = env['rack.multipart.tempfile_factory'] ||
           lambda { |filename, content_type| Tempfile.new(["RackMultipart", ::File.extname(filename)]) }
         bufsize = env['rack.multipart.buffer_size'] || BUFSIZE
+        params = (env['rack.multipart.params_class'] || QueryParser::DEFAULT.params_class).new
```
tenderlove: I don't really want to encourage sticking more stuff in `env`.  Can we just change the signature for `create` to have a default like `initialize` below?

jeremyevans: I'm fine with changing the API, I just wanted to keep API changes to the minimum.  I've pushed another commit that changes the API as you suggested.

tenderlove:
Can you make `parse_multipart` and `self.parse` both default to `QueryParser::DEFAULT.params_class.new`?  Sorry to be a pain, but I don't want to support `params` as `nil`.  I'm positive about merging this, I just don't want to deal with nil checks.

jeremyevans:
Sounds good, I've pushed another commit that makes that change.  Thanks!

jeremyevans:
@tenderlove There were commits merged recently that conflicted with this pull request, so I've fixed the merge conflicts and squash rebased the branch against master. It should be safe to merge now.

spastorino:
Thanks @jeremyevans I haven't had the time to go through this will try to do ASAP.

spastorino:
@jeremyevans thanks for your work.

PR #823, opened 2015-03-12
remove 2 unnecessary attr_readers

jackxxu:
removing these two attributes, as `original_headers` is not referenced anywhere except `MockResponse`, which has its own `original_headers`, and `headers` already aliased to `@header`


deepj:
I think it can be merged freely /cc @spastorino or @tenderlove

spastorino:
There is something weird about original_headers, it was added here 7f889ec35dfb7dd90828e12af5e1b2d7859f9384 in this module to be used by Response and reused by MockResponse but MockResponse is currently not using it but it is defining original_headers anyway.

jackxxu:
@spastorino, I agree that that `original_headers` is currently not being used, at least not in this repo -- it seems to be intended to hold the original headers, while @headers hold the more transformed header values. as it is now, i believe `original_headers` should be the same as `headers.to_hash`, so  `original_headers` is no longer necessary.

ioquatix:
It looks like this was already implemented. Please feel free to check it and let me know if that is not the case. Thanks for your contribution.

PR #826, opened 2015-03-19
Added media_type methods in Rack::Response

AMekss:
Hi, It would be really cool and convenient to have `#media_type` method on `Rack::Response` as well (similar to the one is already implemented in `Rack::Request`). This would help a lot in [this PR](https://github.com/interagent/committee/pull/70)
This is my attempt to achieve that. Suggestions and comments are welcome.

### DIFF
--- a/lib/rack/media_type.rb
+++ b/lib/rack/media_type.rb
@@ -1,0 +1,38 @@
+module Rack
+  # Rack::MediaType parse media type and parameters out of content_type string
+
+  class MediaType
+    SPLIT_PATTERN = %r{\s*[;,]\s*}
+
+    class << self
+      # The media type (type/subtype) portion of the CONTENT_TYPE header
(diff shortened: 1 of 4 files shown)


tenderlove:
I'm fine adding `media_type` to `Rack::Response`, but I don't see the value in keeping a reference to the `MediaType` object.  Can you change the instance methods to class methods, then cache the result of parsing on the Response instance?

e.g.:

```ruby
def self.type content_type
  return nil unless content_type
  content_type.split(SPLIT_PATTERN, 2).first.downcase
end
```

Then in Response:

```ruby
def media_type
  MediaType.type(content_type)
end
```

We can cache the return value of `MediaType.type` in `def media_type` if it's a performance issue.

AMekss:
It's a good point, I did the changes @tenderlove

PR #830, opened 2015-03-25
MockRequest should include REQUEST_PATH as part of env

umtrey:
The Rack handlers add in an additional key for REQUEST_PATH on each Request, but this key is not present in the MockRequest object. This pull request includes this additional key.


tenderlove:
> The Rack handlers add in an additional key for REQUEST_PATH on each Request, but this key is not present in the MockRequest object.

Doesn't that mean that you're not going through some middleware that you should be going through?  Why should the mock object add this and not the middleware?

PR #831, opened 2015-03-26
RACK_ENV=production should be the same as deployment

wuputah:
For a few years, Heroku has been recommending setting RACK_ENV to the same as RAILS_ENV: to "production". This has been the default for countless Rails apps on Heroku, and even using RACK_ENV in place of RAILS_ENV had become somewhat popular. These settings have now become common outside the use of Heroku.

It's come to our attention that [this was bad advice](http://www.hezmatt.org/~mpalmer/blog/2013/10/13/rack_env-its-not-for-you.html), and can have real production impact for our customers (particularly those running Unicorn) due to not including [the middlewares that RACK_ENV=deployment provide](https://github.com/defunkt/unicorn/blob/master/lib/unicorn.rb#L70-L76). We are working on changing this for our customers and spreading the word, such as [this PR to Rails](https://github.com/rails/rails/pull/19404) and [Sinatra](https://github.com/sinatra/sinatra/pull/984) that reverses the use of RACK_ENV.

Of course, there are many apps that have RACK_ENV=production set, and they might have written software that depends on RACK_ENV=production. Sinatra users are in a particularly bad way here since it does not have a separate setting (yet). So, it seems reasonable to propose that RACK_ENV=production work the same as RACK_ENV=deployment, provided they update their software. Since Unicorn looks to rack as a reference for this behavior, I'm proposing changing rack first.

I believe the semver impact for this would be a bump to a minor release (i.e. 1.7.0), which I believe is currently on master. At some point in the future, deprecating either "deployment" or "production" might make sense, but that would be a major change.

Thanks for considering this patch!

cc/ @zzak @dmathieu @hone @schneems


schneems:
We'll need to do this in 2 parts. Give a warning to anyone who is using `RACK_ENV=production` to tell them that behavior is changing. It's a minority of cases but i'm sure more than one person is using that to indicate they don't want **any** default middleware. This would be pretty noisy so ideally we could release a version change at the same time so people can upgrade right away to get rid of the errors. 

After that i'm :+1: on allowing "production" and "deployment". It would be nice to mirror options of the most popular rack frameworks (sinatra, rails). On a less scientific note, at a glance `deployment` looks visual similar to `development` where `production` is more obviously different possibly leading to less mis-reading bugs.

This is one of those few cases where I believe common usage may be more significant than the specification/docs. Once we decide on a direction for this, we should also make to explicitly call out behavior in the README, as the current behavior is unexpected and fairly well hidden unless you seek out the method documentation.

This PR is failing but, it doesn't look to be a direct cause of this code

```ruby
 uninitialized constant Rack::Handler::WEBrick
```

jeremy:
The blog post makes a strong stand, but it doesn't capture the history or usage of `RACK_ENV`. It's a convention that grew *outside* of Rack, in Passenger, Mongrel, Unicorn and other servers, and in early Rack-friendly frameworks, like Merb and Sinatra. Linking it to Rack::Server's `-E` reflects Rack::Server's role as *just another user of the env*.

Retrospectively recasting its role into a "you're doing it wrong" post is appealing, but misleading. `RACK_ENV` has consistently meant the environment you're running the *app* in, not the "Rack environment." It's caused some confusion, for sure. That's worth addressing. But nobody has been doing it wrong.

If we're going to break compatibility, a stronger path forward could be to remove the default middleware decision-making in Rack::Server. Double down on `RACK_ENV`'s well-established role instead of reinventing it.

jeremy:
Additionally: Introducing a new middleware stack to `RACK_ENV=production` apps is more than a compatibility break. It's dangerously unsafe for production Rack users. It's a submerged behavior change that'll break compatibility *at production deploy* and may cause an app to run aground. Nope nope nope 🙉

tenderlove:
> If we're going to break compatibility, a stronger path forward could be to remove the default middleware decision-making in Rack::Server.

master is 2.0.0, and I prefer this.  I don't think Rack should make any decision about the middleware you have.  IMO Rack has one job: be an adapter for webservers.  Of course it's a little more nuanced than just that, but I think picking default middleware is outside Rack's scope.

wuputah:
This is only about `rackup`, which I suppose you could say is a reference implementation for HTTP servers. Hopefully `rackup` is not actually used widely in production. But I think having a reasonable reference implementation is important, and since Unicorn follows the pattern of `rackup`, I figured this might be a reasonable place to start.

The difference about `RACK_ENV=deployment` is it, effectively, enables most modern features of HTTP/1.1, like keep-alive (rather than running with `Connection: close` and relying on EOF). Without it, there's no Content-Length, chunked encoding, etc, which means keep-alive cannot be used. Note that these middleware are actually present when `RACK_ENV=development`, but then disappear when in `production`, which I think is equally bad behavior. But, yes, perhaps we should also remove all middleware injection from development as well?

Note that the choice to implement those things as middlewares is the HTTP server's choice. For instance, `rackup` and Unicorn do it this way, but puma does not (and thus, puma does not need a patch for this issue).

This isn't really about the blog post or "doing it wrong" -- and yes, that blog post only highlights a bit of the issue. We discovered this issue with production applications using Unicorn who did not have Content-Length headers or chunked encoding, even though it was using HTTP/1.1. This was leading to issues downstream with HTTP routers (since fixed) and Internet Explorer that did not normally see this type of traffic.  Further research noted that no one was adding these middlewares to the stack (e.g. Unicorn nor Rails) since `RACK_ENV` was set to `production`. I did further research to determine the related behavior in rack and puma.

I think that this is about whether `rackup` is a reference implementation for a well-behaved HTTP server, and if that server should support these HTTP/1.1 features all the time. That said, there are also times when you might not want some of these behaviors, such as when doing streaming responses--though I'm not sure that there'd be any negative impact of having these middlewares in place in those cases (e.g. `Rack::ContentLength` [would handle this case fine](https://github.com/rack/rack/blob/master/lib/rack/content_length.rb#L20) if you set an transfer encoding).

raggi:
> master is 2.0.0, and I prefer this.

You are aware of the Tempfile reaper right? The semantic there is actually pretty important.

Also if we're delegating the responsibility of setting up content-length and chunking middleware to non-servers, but frameworks, then we really need to document their use cases. It took some folks in the rails community years to understand those issues, even though they were relative experts, and we never improved the docs or anything that would help others repeating the process.

raggi:
As for this change, it should be fine to apply in most cases. The change isn't particularly breaking. The middleware involved should be spec conforming.

ioquatix:
Getting content length right is the responsibility of the server protocol layer and application, not the middleware. Falcon already gets this correct for HTTP/1.0, HTTP/1.1 & HTTP/2.0 so it's certainly possible to do this correctly without any additional middleware and also usually more efficient.

While I agree with @tenderlove, I also believe that rackup is not for production. So, as it stands, maybe it should default to a reasonable stack (could be empty) for development purposes, and remove RACK_ENV and environment selection via rackup entirely.

jeremyevans:
I strongly feel that `production` environment should not add middleware, and would like to close this.  I don't think we should break backwards compatibility here.  If people want the deployment middleware added by rackup, they should run with the `deployment` environment.  

As this is solely regarding rackup, I don't think it is a SPEC issue.

ioquatix:
I think the only way we could realistically bring this into alignment is remove ensure that `RACK_ENV=deployment` does the same thing as `RACK_ENV=production`. Because I don't know anyone who sets `RACK_ENV=deployment`. And, probably, it's the right thing. We already removed `Rack::Chunked` from the default stack. Maybe now it's a good time to only special case `RACK_ENV=development` or maybe even better, remove RACK_ENV special case completely.

fedsta (approved):

mmizutani:
Could anyone with a push permission resolve the blocking conflict?

ioquatix:
Yes I can take a look.

ioquatix:
By the way, I'd personally like to simply get rid of `RACK_ENV`. I understand why it exists and what it does but I don't think it adds much value and adds significant confusion.

jeremyevans:
I'm still strongly against this.  `production` should continue to mean no middleware added.  Adding  middleware by default breaks backwards compatibility and will make applications using `production` slower.

ioquatix:
@jeremyevans Yes, so do you agree to get rid of it completely?

jeremyevans:
I don't think we should get rid of `RACK_ENV` completely.  Doing so would break backwards compatibility for users who expect it to add middleware in the cases it currently adds middleware.  I think the cost of backwards compatibility breakage is higher than the benefit from removing it.

fedsta:
Ok

ioquatix:
I'm going to close this but I'm going to open a new PR if I think there is an obvious way we can reduce our dependence on `RACK_ENV`.

PR #834, opened 2015-04-01
cache result of to_app to avoid re-traversal of maps

lucaong:
This prevents re-traversing the whole middleware tree and re-instantiating
Rack::Builder instances on each request.

I guess that this commit was in the same direction, except it does not solve the
problem as far as I can see:

https://github.com/rack/rack/commit/b51f3036fbdb911eac6b5cf97ebf9388fb1f14a9

This passes tests. Of course it assumes that it's fine to reuse the same
Rack::Builder instance for more than one request (I don't see why that should
not be the case).

I already use this change in a production app with a large number of mappings,
where we experienced massive improvements on routing time and object allocation.

Here is a quick benchmark (to be run on the current Rack::Builder, before
applying this change): https://gist.github.com/lucaong/ec70d8059adb48e4ca8a

### DIFF
--- a/lib/rack/builder.rb
+++ b/lib/rack/builder.rb
@@ -150,7 +150,8 @@ module Rack
     end
 
     def call(env)
-      to_app.call(env)
+      @to_app ||= to_app
+      @to_app.call(env)
     end
 
     private


lucaong:
Any opinions on this? I understand that the charts are only of limited use when out of context, but this is the order of magnitude of the improvement that I am experienced (on a production Rack app with roughly a hundred mapped routes, mostly fast but high-throughput endpoints where routing time is relevant):

Average response time

GC load and object allocation

The sharp decrease in response time and object allocation correspond to the introduction of this patch in our codebase.

tenderlove:
Wow, I didn't know this happened. I take it you're using a straight rack app, and not Rails or something else?  I *think* we cache that builder in Rails.  I'll try to read through the code and think of stuff that could go wrong, but on the surface this change seems fine.

lucaong:
Yeah our app is rack + Sinatra (basically mapping several Sinatra applications on different endpoints with `Rack::Builder`), so it is possible that Rails is fine. Thanks for taking time to review this change.

zotherstupidguy:
fascinating :)

spastorino:
Wouldn't be better to cache on `to_app` method and invalidate the cache where it makes sense?

lucaong:
Not sure... I personally think it is good to be able to (re)generate a fresh instance of the app by calling `to_app`, and leave the responsibility of caching it to callers of this method.

Uncaching might be interesting, @spastorino would you uncache it whenever `use`, `map`, `run` or `warmup` are called?

tenderlove:
@lucaong what happens if you change `Rack::Builder.new` to `Rack::Builder.app`?

ioquatix:
What is the logic behind having `Rack::Builder#call` it seems like the wrong place to have that functionality? Perhaps this method should simply be removed?

raggi:
This was left like this for Rack::Reloader use cases. The intent is you will call to_app deliberately when appropriate.

rafaelfranca:
I merged https://github.com/rack/rack/pull/953 since it includes a test. Thank you for the pull request.

PR #835, opened 2015-04-08
Ignore param parsing errors in MethodOverride

greysteil:
If unparseable parameters are passed through `Rack::MethodOverride` it will bomb whilst looking for the `_method` parameter. Because `Rack::MethodOverride` is usually included low down in the stack this exception often goes uncaught, causing a 500.

This PR silently ignores parameter parsing errors that occur in `Rack::MethodOverride`. If the params are used later the same exception will be raised, but this will typically be much higher up the stack, and after any exception handling middleware.

See https://github.com/rails/rails/pull/19632 for a discussion of this in Rails.

### DIFF
--- a/lib/rack/methodoverride.rb
+++ b/lib/rack/methodoverride.rb
@@ -37,6 +37,7 @@ module Rack
 
     def method_override_param(req)
       req.POST[METHOD_OVERRIDE_PARAM_KEY]
+    rescue Utils::InvalidParameterError, Utils::ParameterTypeError
     end
   end
 end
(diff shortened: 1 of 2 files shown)


greysteil:
@spastorino - looks like you reviewed the last change to `MethodOverride` - any chance you could take a quick look over this?

greysteil:
@spastorino - would be great to get your opinion on whether this should be handled here or if the error should be caught in the framework.

spastorino:
Sorry, I'm right now with zero time to check this.
But we should take a look at all the open issues and PRs soon.

greysteil:
Thanks @spastorino. Would be great to get your opinion when you have a moment :octocat:

greysteil:
Hate to nag but I'd love an opinion on this vs. https://github.com/rails/rails/pull/19632

greysteil:
:sparkling_heart:

PR #836, opened 2015-04-08
Lazily instantiate ERB template object

Wardrop:
ShowExceptions currently parses it's ERB template on instantiation. This can add significant overhead, especially for frameworks and applications that may add middleware dynamically on each request, such as Scorched, where caching an application instance is non-trivial.


tenderlove:
Since this is essentially static, can you change it to assign the template to a constant so that the constructor isn't doing any work?

PR #837, opened 2015-04-11
Secure request.referer

ppworks:
Sanitize `request.referer` for preventing XSS


tenderlove:
Escaping the referrer is outside Rack's scope.  The request object will return to you exactly what the client sent.  It's up to you to escape it before using it.

PR #839, opened 2015-04-14
Add HTTP response helpers for 204, 301, and 412.

BrentWheeldon:


filipegiusti:
This seems good to be merged, the CI failure seems to be something else on master.

PR #841, opened 2015-04-15
Use the default HTTP output buffer, instead of 5

oskarpearson:
Setting the output buffer size to 5 means that we will do thousands of
small writes when serving a large file, which is inefficient. This fix
improves performance of the HTTP server.

This resolves the slow-down referenced at https://github.com/rails/rails/issues/18828

### DIFF
--- a/lib/rack/handler/webrick.rb
+++ b/lib/rack/handler/webrick.rb
@@ -28,7 +28,6 @@ module Rack
 
         options[:BindAddress] = options.delete(:Host) || default_host
         options[:Port] ||= 8080
-        options[:OutputBufferSize] = 5
         @server = ::WEBrick::HTTPServer.new(options)
         @server.mount "/", Rack::Handler::WEBrick, app
         yield @server  if block_given?


lvonk:
Nice find! I am also experiencing the same slow down.

chneukirchen:
Can anyone familiar with WEBrick check whether this breaks c5621e93 which introduced this line?

oskarpearson:
@dahakawang Any comments here? I see you authored the original commit.

@tenderlove has some plans for removing that bit of code entirely... but I think this is a simple change that fixes an immediate performance regression. Perhaps this is a good interim solution?

tenderlove:
The default buffer size [is `65536`](https://github.com/ruby/ruby/blob/daed91295479ff3e981bcc9e395721ba4057c79e/lib/webrick/config.rb#L53).  I guess that means streaming stuff won't write to the socket until that much data is in the buffer.  I don't do any streaming stuff with WEBrick in production, so this is probably fine.

Honestly I wish we had our own WEBrick handler in Rails though.  I'll merge this for now!  Thanks!

lfittl:
@tenderlove Is there a chance this could be back-ported into the `1-6-stable` branch for the next minor release?

This fixed a major slowdown in our test suite (about 3x difference) due to a large asset that took ~20 seconds to load without this fix.

We currently have a monkey patch that allows us to use this patch with the 1.6.1 release, but would be nice to not have to do that.

tenderlove:
@lfittl ya, I'm fine with that.  I'll cherry pick it over!

PR #845, opened 2015-04-24
Add backported bug fix.

aq1018:
This bug is causing confusing errors such as:

{"error" => "Failed to access params -- no implicit conversion of nil into Hash"}

This code is provided in this PR: https://github.com/rack/rack/pull/561 and is never officially released to the 1.4.x branch.

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -197,7 +197,6 @@ module Rack
       elsif @env["rack.request.form_input"].eql? @env["rack.input"]
         @env["rack.request.form_hash"]
       elsif form_data? || parseable_data?
-        @env["rack.request.form_input"] = @env["rack.input"]
         unless @env["rack.request.form_hash"] = parse_multipart(env)
           form_vars = @env["rack.input"].read
 
@@ -210,6 +209,7 @@ module Rack
 
           @env["rack.input"].rewind
         end
+        @env["rack.request.form_input"] = @env["rack.input"]
         @env["rack.request.form_hash"]
       else
         {}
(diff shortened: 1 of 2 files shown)


lib/rack/request.rb
```diff
@@ -118,25 +118,25 @@ def path_info=(s);   @env["PATH_INFO"] = s.to_s               end
...
-    
+
     # Checks the HTTP request method (or verb) to see if it was of type POST
     def post?;    request_method == "POST"    end
-    
+
     # Checks the HTTP request method (or verb) to see if it was of type PUT
     def put?;     request_method == "PUT"     end
-    
+
```
zzak: I don't think we should commit whitespace changes, can you revert these?

joevandyk: once white space is added to a blank line, it can never be removed? :(

joevandyk: Oh, is this only for backporting? I notice the whitespace is gone in master.

aselder:
:+1: Would love to see this back ported and 1.4.6 released.

lib/rack/request.rb
aq1018: Ah... My editor strips trailing whitespaces automatically. I'll make the changes.

aq1018:
@zzak I corrected the whitespace issues. Please let me know if there is anything that I need to change. Thanks!

cadorett:
Any update behind the scenes on if this branch will be merged?

zzak:
I guess this depends on whether t.lo wants to backport for 1.4 series still

mikegee:
Since there has been no release of Rack 1.4.x since 2015, I suspect this isn't going to happen. Can we close it?

PR #850, opened 2015-05-06
adding a security policy

tenderlove:
This commit adds a security policy.  It's mostly stolen from the Rails security policy.

If it seems good, then lets go with it.

/cc @spastorino

### DIFF
--- a/SECURITY_POLICY.md
+++ b/SECURITY_POLICY.md
@@ -33,7 +33,7 @@ When a release series is no longer supported, it’s your own responsibility to 
 
 ## Reporting a bug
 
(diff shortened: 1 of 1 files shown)


chneukirchen:
+1, not sure you want to do the pre-release dance tho.

tenderlove:
I don't think the security policy ties us to any pre releases.  In fact, I want to avoid pre-releases since they would typically mean that we're shipping something other than sec fixes in the release.

spastorino:
:+1:

chneukirchen:
Sorry, I was not clear.  I meant this:

> 3. Fixes are prepared for all releases which are still supported. These fixes are not committed to the public repository but rather held locally pending the announcement.
> 4. A suggested embargo date for this vulnerability is chosen and distros@openwall is notified. This notification will include patches for all versions still under support and a contact address for packagers who need advice back-porting patches to older versions.

tenderlove:
@chneukirchen ah, I don't mind that dance.  The first one we have to do anyway (since the fixes will have to be secret before the release), and the second one is just sending an email (which is the same email I usually send to announcement lists).

PR #853, opened 2015-05-08
Add support for multiple, semicolon-separated X-Accel-Mappings

nbibler:
This updates the header-defined mapping functionality to match that of the explicitly-instantiated mappings.

In all current versions of `Rack::Sendfile`, a developer may define multiple mappings if they instantiate Sendfile with an Array of mappings. The ability to define multiple mappings is unavailable, however, when parsing the request environment's `X-Accel-Mapping` header.

This pull request introduces the ability to define multiple mappings in the `X-Accel-Mapping` request header.

### DIFF
--- a/lib/rack/sendfile.rb
+++ b/lib/rack/sendfile.rb
@@ -53,6 +53,15 @@ module Rack
   # that it maps to. The middleware performs a simple substitution on the
   # resulting path.
   #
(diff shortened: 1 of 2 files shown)


907th:
+1 on this!

nbibler:
Closing this PR since #1187 meets the need (although uses a slightly different data structure).

PR #855, opened 2015-05-14
API to extract cookies from mock response

petercline:
The Rack::MockResponse class did not expose any API to provide access to
cookies.  This change adds that to facilitate testing and prevent all
clients using the MockResponse class from having to parse response
headers in order to get at the cookies present in the response.  Inspiration derived from the Spring Framework's MockHttpServletResponse class which exposes an API method getCookie which accepts a String representing the cookie name and returns an object representing the cookie corresponding to that name.  I considered adding this functionality to Rack::Test::CookieJar but it exposes a conflicting API which supplies just cookie name and value and hides all other attributes.  The Rack::Test module has a Cookie class that includes all the relevant attributes but all accessor methods for those are marked :api: private and hence it seems like my conception of how a cookie should be represented didn't jive with the existing API.


tenderlove:
Please don't add a licenses file. Rack is MIT licensed, not Apache licensed.

astratto:
This patch is actually quite useful.

@tenderlove do you think it could be merged or do you have other concerns?

tenderlove:
@astratto yep, thanks for bumping this.  I'll merge it, though it's conflicting so I may do it in a day or two.

petercline:
And to think I had given this patch set up for dead.  I'm excited to see it exhumed from the heap of forgotten pull requests and perhaps offered a reprieve from its formerly lackluster fate. :-)

Thanks all!  Please let me know if there's anything I can do to help shepherd this through to completion.

astratto:
Cool, thank you guys.

ignatiusreza:
friendly bump? maybe PR need to be rebased?

PR #856, opened 2015-05-15
Duplicate the @names hash on dup/clone to avoid shared references.

jrafanie:
We need to manually dup the `@names` hash since a HeaderHash.dup is
shallow, leading to multiple HeaderHash objects sharing the same
`@names` hash.

Very similar to https://github.com/lostisland/faraday/pull/478

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -480,6 +480,12 @@ module Rack
         hash.each { |k, v| self[k] = v }
       end
 
+      # on dup/clone, we need to duplicate @names hash
+      def initialize_copy(other)
+        super
+        @names = other.names.dup
+      end
+
       def each
         super do |k, v|
           yield(k, v.respond_to?(:to_ary) ? v.to_ary.join("\n") : v)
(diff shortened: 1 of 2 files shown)


lib/rack/utils.rb
```diff
@@ -480,6 +480,12 @@ def initialize(hash={})
         hash.each { |k, v| self[k] = v }
       end
 
+      # on dup/clone, we need to duplicate @names hash
+      def initialize_copy(other)
+        super
+        @names = @names.dup
```
tenderlove: We should be getting the names off the `other` object.  Is there a way to get them?  If not, can you add a protected method?

jrafanie: sure, via instance_variable_get

jrafanie: I thought that was really ugly, but will do it if that's desired.

jrafanie:
Thanks @tenderlove, I made your suggested changes.

jrafanie:
Note, I indented the `protected` method to keep similarity with style elsewhere in rack.

jrafanie:
Thanks @spastorino :smile:

spastorino:
you're welcome, thanks for helping :).

PR #859, opened 2015-05-20
Constantize Transfer-Encoding header where possible

deepj:


spastorino:
A rebase is needed

deepj:
@spastorino Done

deepj:
@spastorino I've made also a huge refactoring env variables and other similar stuff (e.g. rack variables such as `rack.errors` etc) to frozen constants (not pushed on GitHub yet). Would there be any interest about that? It has some nice speed and objects allocation improvements. But I don't want to spam here by this kind of changes. On the other hand I believe it can help to improve a speed of Rack.

spastorino:
@deepj go ahead and send a PR. Benchmarking it would be great too.

PR #863, opened 2015-05-21
Add support for serving gzipped static files

jeremyevans:
This makes it simple to served static files that have already been
gzipped.  This is significantly faster than using Rack::Deflater
to serve the same files, as it results in less IO and it avoids
the need to gzip at runtime.

This provides a similar feature to nginx's gzip_static directive.


jeremyevans:
This fails only on rbx-2 due to an unrelated issue (pid file writing for Rack::Server), it passes on all other implementations

PR #865, opened 2015-05-22
Small code improvement.

gfvcastro:


lib/rack/utils/okjson.rb
```diff
@@ -278,7 +278,7 @@ def numtok(s)
 
   def strtok(s)
     m = /"([^"\\]|\\["\/\\bfnrt]|\\u[0-9a-fA-F]{4})*"/.match(s)
-    if ! m
+    unless m
```
tenderlove: We imported okjson, so we shouldn't make this change.  The rest look fine though.

PR #870, opened 2015-05-26
Initial clean up and preparation for Ruby 2.2+ only support

deepj:
I've started working on some small improvements. I mostly remove some dead code and some obsolete things such as Mongrel. A question here is if Rack is going to support things like Memcached sessions (based on legacy library), lighttp and so on. Any suggestions?


tenderlove:
@deepj I'm really sorry. I didn't see this PR. Can you rebase it and I'll apply the rest of your changes? Thank you so much for this work!

deepj:
@tenderlove done

tenderlove:
Thanks!

spastorino:
:+1: I was going to merge this one thanks t.lo

lib/rack/utils.rb
```diff
@@ -360,19 +360,6 @@ def delete_cookie_header!(header, key, value = {})
...
-      def bytesize(string)
-        string.bytesize
-      end
-    else
-      def bytesize(string)
-        string.size
-      end
-    end
-    module_function :bytesize
-
```
jeremy: For the sake of gracefully deprecating breaking changes to public API, suggest keeping this as
```ruby
def bytesize(string)
  warn 'Rack::Utils.bytesize is deprecated in Rack 2.0 and will be removed in Rack 2.1. Use String#bytesize instead.'
  string.bytesize
end
```

Ref. https://github.com/rack/rack-contrib/pull/124 which used this public API.

glaucocustodio: Please take a look at https://github.com/rack/rack/pull/1103 cc/ @jeremy

jeremy: Thanks @glaucocustodio.

PR #874, opened 2015-06-04
request.referer checks HTTP_REFERRER as well.

joevandyk:
I had a request come in that had HTTP_REFERRER set (HTTP_REFERER was blank).

```
  "HTTP_REFERRER": "https://www.lolshirts.com/",
  "HTTP_USER_AGENT": "Mozilla/5.0 (Windows NT 6.3; WOW64) AppleWebKit/549.36 (KHTML, like Gecko) Chrome/41.0.2382.89 Safari/549.16",
```

`request.referer` was `nil`. This fixes that.

[Review on Reviewable](https://reviewable.io/reviews/rack/rack/874)

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -289,7 +289,7 @@ module Rack
 
     # the referer of the client
     def referer
-      @env['HTTP_REFERER']
+      @env['HTTP_REFERER'] || @env['HTTP_REFERRER']
     end
     alias referrer referer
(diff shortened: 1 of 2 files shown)


ioquatix:
As this is non-standard, I don't believe it's in scope for Rack. If you can find a relevant standard which shows we should support it, I'm against merging this. However, thanks for your contribution.

Issue #768, opened 2014-12-17
complex object parsing issues


guilherme:
The nested query builder aren't  generating a format that is parsed to its original.

``` ruby
Rack::Utils.parse_nested_query(Rack::Utils.build_nested_query({simple_test: [{ 1 => [{ 2 =>  [1,2,3,4,5] }] }] }))
# {"simple_test"=>[{"1"=>[{"2"=>[nil, nil, nil, nil, nil]}]}]}
```

``` ruby
Rack::Utils.parse_nested_query({simple_test: [{ 1 => [{ 2 =>  [1,2,3,4,5] }] }] }.to_param) # with rails generating the parameters
# {"simple_test"=>[{"1"=>[{"2"=>["1", "2", "3", "4", "5"]}]}]}
```

I've noted another issue with the parser but i'm not sure what is the problem: with the data format or the builder/parser:

``` ruby
Rack::Utils.parse_nested_query(Rack::Utils.build_nested_query({ simple_test: [[1, [[2, [3,2,3,4]]]], [2]] }))
# {"simple_test"=>[nil, nil, nil, nil, nil, nil, nil]}
Rack::Utils.parse_nested_query({ simple_test: [[1, [[2, [3,2,3,4]]]], [2]] }.to_param)
# {"simple_test"=>[nil, nil, nil, nil, nil, nil, nil]}
```

The CGI::parse gives me another output:

``` ruby
CGI::parse({ simple_test: [[1, [[2, [3,2,3,4]]]], [2]] }.to_param)
# {"simple_test[][]"=>["1", "2"], "simple_test[][][][]"=>["2"], "simple_test[][][][][]"=>["3", "2", "3", "4"]}
CGI::parse(Rack::Utils.build_nested_query({ simple_test: [[1, [[2, [3,2,3,4]]]], [2]] }))
# {"simple_test[][]"=>[], "simple_test[][][][]"=>[], "simple_test[][][][][]"=>[]}
```

rack-1.4.5
rails-3.2.11

jeremyevans:
The first example in the original post is now handled correctly by rack:

```ruby
Rack::Utils.parse_nested_query(Rack::Utils.build_nested_query({simple_test: [{ 1 => [{ 2 =>  [1,2,3,4,5] }] }] }))
# => {"simple_test"=>[{"1"=>[{"2"=>["1", "2", "3", "4", "5"]}]}]}
```

The second case still is not parsed correctly.  Here's a simplified case:

```ruby
[...]

PR #880, opened 2015-06-10
Remove old ruby code - < 1.9 -

fabianrbz:
Hi guys,

This PR removes some Ruby < 1.9 code.
It also improves the legibility of a test, in Ruby 1.8 Hash was unordered so the test compared the output of ``` #build_nested_query ``` agains the output of ``` #parse_nested_query ```, now it is compared against the result string instead.

### DIFF
--- a/test/spec_utils.rb
+++ b/test/spec_utils.rb
@@ -258,9 +258,45 @@ describe Rack::Utils do
     Rack::Utils.build_nested_query('foo' => 'bar', 'baz' => {}).
       should.equal 'foo=bar'
 
-    # The ordering of the output query string is unpredictable with 1.8's
-    # unordered hash. Test that build_nested_query performs the inverse
-    # function of parse_nested_query.
+    Rack::Utils.build_nested_query('foo' => nil, 'bar' => '').
+      should.equal 'foo&bar='
+    Rack::Utils.build_nested_query('foo' => 'bar', 'baz' => '').
+      should.equal 'foo=bar&baz='
+    Rack::Utils.build_nested_query('foo' => ['1', '2']).
+      should.equal 'foo[]=1&foo[]=2'
+    Rack::Utils.build_nested_query('foo' => 'bar', 'baz' => ['1', '2', '3']).
+      should.equal 'foo=bar&baz[]=1&baz[]=2&baz[]=3'
+    Rack::Utils.build_nested_query('foo' => ['bar'], 'baz' => ['1', '2', '3']).
+      should.equal 'foo[]=bar&baz[]=1&baz[]=2&baz[]=3'
+    Rack::Utils.build_nested_query('foo' => ['bar'], 'baz' => ['1', '2', '3']).
+      should.equal 'foo[]=bar&baz[]=1&baz[]=2&baz[]=3'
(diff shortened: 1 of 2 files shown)


test/spec_utils.rb
```diff
@@ -257,30 +257,38 @@ def initialize(*)
...
-      {"x" => {"y" => [{"z" => ["1"]}]}},
-      {"x" => {"y" => [{"z" => "1", "w" => "2"}]}},
-      {"x" => {"y" => [{"v" => {"w" => "1"}}]}},
-      {"x" => {"y" => [{"z" => "1", "v" => {"w" => "2"}}]}},
-      {"x" => {"y" => [{"z" => "1"}, {"z" => "2"}]}},
-      {"x" => {"y" => [{"z" => "1", "w" => "a"}, {"z" => "2", "w" => "3"}]}}
-    ].each { |params|
-      qs = Rack::Utils.build_nested_query(params)
-      Rack::Utils.parse_nested_query(qs).should.equal params
-    }
```
tenderlove: I'm not sure that I want this change.  I know that we can predict hash ordering in 1.9+, but I really like that this is testing that `build_nested_query` and `parse_nested_query` work well together.  If I take this change, then we lose coverage for `parse_nested_query`.

spastorino: @tenderlove I agree, anyway I like the change as a way of documenting what happen when you call `build_nested_query`. What about adding tests calling `parse_nested_query`?

spastorino: I mean all the list of calls back calling `parse_nested_query`

fabianrbz: @tenderlove I agree with what you are saying, but I think that this way the test acts as documentation. -and it also test against the actual result which can detect bugs that might be in ``` parse_nested_query ```-.
Also, the opposite scenario of each of those tests is tested for ``` parse_nested_query ```. Maybe we can add one test case that documents how ``` build_nested_query ``` and ``` parse_nested_query ```work together if we want to be explicit. what do you think?

Anyway, I can remove that commit from the PR if you think that is not worth it and just keep the other one :smiley:

tenderlove: > detect bugs that might be in parse_nested_query-

You mean `build_nested_query`?

> Maybe we can add one test case that documents how build_nested_query and parse_nested_querywork together if we want to be explicit.

That sounds good to me.  We have some issue where `build` and `parse` don't play well together (#768), so having explicit "round tripping" tests would be good.

> Anyway, I can remove that commit from the PR if you think that is not worth it and just keep the other one :smiley:

TBH I'm mainly worried about being coupled to the internals.  Maybe someday we want to change the algorithm for the function and don't want to be coupled to hash ordering.  But maybe I shouldn't worry about that.  ;)

I think if we keep this, then just continue to add round tripping tests going forward, that would probably be best.

tenderlove:
Can you add some tests for roundtripping, and I'll merge this?

fabianrbz:
sure!

fabianrbz:
@tenderlove something along those lines? Or do you prefer to iterate over an array of hashes like before?

tenderlove:
@fabianrbz just do the array of hashes like before please.  Thank you!

fabianrbz:
@tenderlove I added back the tests and reworded the commits a bit.

PR #881, opened 2015-06-10
Minitest conversion

zenspider:
This is just the first phase and is not meant for long-term use. It brings in minitest + minitest-bacon to help with the conversion. Phase 2 will be removing minitest-bacon and doing plain minitest/spec. Phase 3 will be plain minitest/test.

### DIFF
--- a/rack.gemspec
+++ b/rack.gemspec
@@ -29,6 +29,8 @@ EOF
   s.rubyforge_project = 'rack'
   s.required_ruby_version = '>= 2.2.2'
 
-  s.add_development_dependency 'bacon'
+  s.add_development_dependency 'minitest-bacon'
+  s.add_development_dependency 'minitest', "~> 5.0"
+  s.add_development_dependency 'minitest-sprint'
   s.add_development_dependency 'rake'
 end
--- a/rack.gemspec
+++ b/rack.gemspec
@@ -1,6 +1,6 @@
 Gem::Specification.new do |s|
   s.name            = "rack"
-  s.version         = eval File.readlines('lib/rack.rb').grep(/RELEASE\s*=/)[0]
+  s.version = File.read('lib/rack.rb')[/RELEASE += +([\"\'])([\d][\w\.]+)\1/, 2]
   s.platform        = Gem::Platform::RUBY
   s.summary         = "a modular Ruby webserver interface"
   s.license         = "MIT"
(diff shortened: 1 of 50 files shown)


rack.gemspec
```diff
@@ -1,6 +1,6 @@
 Gem::Specification.new do |s|
   s.name            = "rack"
-  s.version         = eval File.readlines('lib/rack.rb').grep(/RELEASE\s*=/)[0]
+  s.version = File.read('lib/rack.rb')[/RELEASE += +([\"\'])([\d][\w\.]+)\1/, 2]
```
baweaver: I'd mentioned it on IRC earlier, but why not just ``Rack::RELEASE``?

test/spec_rewindable_input.rb
```diff
@@ -81,38 +88,43 @@
     }.should.not.raise
   end
 
+  after do
```
teoljungberg: Couldn't we use `teardown` instead of `after`? Since the `before` blocks are changed to `def setup` further down?

test/spec_server.rb
```diff
@@ -1,3 +1,4 @@
+require 'minitest/bacon'
```
teoljungberg: There seems to be a need for this to be required in a `spec_helper` of sorts, thoughts?

test/spec_webrick.rb
```diff
@@ -1,11 +1,13 @@
...
 require 'rack/mock'
 require File.expand_path('../testrequest', __FILE__)
 
 Thread.abort_on_exception = true
 
 describe Rack::Handler::WEBrick do
-  extend TestRequest::Helpers
+  include TestRequest::Helpers
 
+  before do
```
tenderlove: The original tests only spin up webrick once.  This change makes webrick start on every test and causes `Errno::ECONNREFUSED` exceptions.  Try `ruby -I lib:test test/spec_webrick.rb`.

tenderlove: Ugh, now I see the after block....  hmmm

tenderlove:
```diff
diff --git a/test/spec_webrick.rb b/test/spec_webrick.rb
index aa44813..f3d6fb0 100644
--- a/test/spec_webrick.rb
+++ b/test/spec_webrick.rb
@@ -14,7 +14,7 @@ describe Rack::Handler::WEBrick do
                                     :AccessLog => [])
   @server.mount "/test", Rack::Handler::WEBrick,
     Rack::Lint.new(TestRequest.new)
-  Thread.new { @server.start }
+  @thread = Thread.new { @server.start }
   trap(:INT) { @server.shutdown }
   end
 
@@ -185,5 +185,6 @@ describe Rack::Handler::WEBrick do
 
   after do
   @server.shutdown
+  @thread.join
   end
 end
```

Needs this ^^

PR #883, opened 2015-06-12
Tests should be quiet

zenspider:


tenderlove:
I think I fixed this in 809a752093015398274b40dd04f3bad536160446, can you check?

zenspider:
*shrug*

PR #885, opened 2015-06-12
Constantize all rack environment variables and make them frozen

deepj:
Some of these variables was already constantized in the past. So this PR unifies the rest and move all these constants under one place (`lib/rack.rb`).

### DIFF
--- a/lib/rack.rb
+++ b/lib/rack.rb
@@ -44,6 +44,7 @@ module Rack
   TRANSFER_ENCODING = 'Transfer-Encoding'.freeze
   HTTP_COOKIE       = 'HTTP_COOKIE'.freeze
 
+  # HTTP method verbs
   GET     = 'GET'.freeze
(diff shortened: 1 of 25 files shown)


deepj:
There are other things which could be done the same way. E.g. something like this https://github.com/rack/rack/blob/131d9fdcc7cdd1d090f0905401abcc07e849740a/lib/rack/recursive.rb#L20

What do you think? /cc @tenderlove @spastorino

tenderlove:
> I believe it's good to make these things consistent across the whole codebase.

Consistency for the sake of consistency worries me.  It makes backporting more difficult.  An argument for "performance" could be made, but IDK if any of these places are hotspots.  That said, master is 2.0, and using a hash for everything (the env hash) really annoys me.  I'll leave this up to @spastorino

spastorino:
@tenderlove we can remove all that when we change the hash. :+1:

PR #887, opened 2015-06-13
Close body set via initializer, matching direct #body accessor

### LINKED ISSUE #877: Rack::Response.new(body) does not close original body on #close
These two lines of code seem to have slightly different behavior:

```
response = Rack::Response.new
response.body = body
response.close
```

```
response = Rack::Response.new body
response.close
```

In the first example, `body` is closed properly, but in the second example, `body` is left open, which caused deadlock issues in my Rack application. Is there a reason for this major difference, or should #close be explicitly called on the initializer body after iterating through its contents?

wjordan:
This PR makes `Request.new(body)` behave the same as `Request.new.body = body`, ensuring that a body object passed via the class initializer will be closed properly by `Response#finish`. Besides eliminating that strange difference in behavior, it has the additional benefit of making it possible to mix #write and direct #body access without problems, so I've removed that warning comment and added a test to verify it works.

Fixes #877.

### DIFF
--- a/lib/rack/response.rb
+++ b/lib/rack/response.rb
@@ -103,9 +103,6 @@ module Rack
     end
 
     # Append to body and update Content-Length.
-    #
-    # NOTE: Do not mix #write and direct #body access!
-    #
     def write(str)
       s = str.to_s
       @length += s.bytesize unless @chunked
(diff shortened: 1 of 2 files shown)


wjordan:
This also fixes a bug where the content-length header was not updated when the body was set via accessor, so I added a test for this as well.

wjordan:
closing this PR in favor of #891- though feel free to reopen if you think this one is more appropriate for 1.6.x.

PR #888, opened 2015-06-14
Add "Body has not been closed" detection to Rack::Lint

wjordan:
A Lint implementation of the following Rack spec requirement has been missing from Rack::Lint [forever](https://github.com/rack/rack/blame/7b5820f8de7f9dffeb03d86588bba13fb0e91df1/lib/rack/lint.rb#L725):

> If the Body responds to +close+, it will be called after iteration. If the body is replaced by a middleware after action, the original body must be closed first, if it responds to close.

This PR implementation takes two Lint instances surrounding the application under test (which is the [recommended setup](https://github.com/rack/rack/blob/master/SPEC#L3-L4)), communicating via a class variable object using the Rack environment as the key. Tests included.

### DIFF
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -41,6 +41,7 @@ module Rack
       ## It takes exactly one argument, the *environment*
       assert("No env given") { env }
       check_env env
+      @env = env
 
       env['rack.input'] = InputWrapper.new(env['rack.input'])
       env['rack.errors'] = ErrorWrapper.new(env['rack.errors'])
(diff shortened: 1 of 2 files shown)


ioquatix:
I like this but it might require a little bit more work. However, I'll merge as is, because fundamentally it's working.

lib/rack/lint.rb
```diff
@@ -719,12 +721,22 @@ def each
...
       ## If the Body responds to +close+, it will be called after iteration. If
       ## the body is replaced by a middleware after action, the original body
       ## must be closed first, if it responds to close.
-      # XXX howto: assert("Body has not been closed") { @closed }
+      open_bodies.delete self
+      assert("Body has not been closed") { !open_bodies.one? {|b| !b.closed?} }
+    end
 
+    def open_bodies
+      ((@@open_bodies ||= {})[@env] ||= [])
```
ioquatix: Unfortunately this is not thread safe.

PR #891, opened 2015-06-15
Refactor Rack::Response to accept unbuffered body content

wjordan:
This refactor follows the discussion in issue #877, and is a further improvement upon PR #887 based on feedback from @tenderlove to allow buffered/unbuffered body contents to be explicitly configurable in the class constructor and to allow some breaking API changes for the 2.0 branch.

Summary:
- make the behavior among the available interfaces to update body contents consistent
  - v2 API change: the `body=` setter will become private
- allow configurable buffering of body contents (through `buffered` constructor parameter, or `Response.buffered`/`Response.unbuffered` factory methods)
  - v2 API change: `unbuffered` will become the default setting (on the fence about this one- while unbuffered will be more efficient, it also will not support updating the `Content-Length` header)
- v2 API change: raise error when `#write` or `#each` are called after `#close` has already been called
- update class documentation

### DIFF
--- a/lib/rack/mock.rb
+++ b/lib/rack/mock.rb
@@ -165,7 +165,7 @@ module Rack
       @errors           = errors.string if errors.respond_to?(:string)
       @body_string      = nil
 
-      super(body, status, headers)
+      super(body, status, headers, true)
     end
 
     def =~(other)
@@ -187,7 +187,7 @@ module Rack
       #     ...
       #     res.body.should == "foo!"
       #   end
-      super.join
+      @body_string ||= [].tap{|str|super.each{|part|str << part}}.join
     end
 
     def empty?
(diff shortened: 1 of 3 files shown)


matthewd:
> enable breaking v2 API changes for Rack.release.to_i > 1 only (so the same code changes may be included in v1.6.x without modification)

:-1:

wjordan:
Thanks for feedback @matthewd, the breaking API changes are now hard-coded into this PR so they are no longer keyed on `Rack.release`.

ioquatix:
We implemented this without remembering your effort here. So it was released in 2.1.0 - if you have time, do you think you can check the latest implementation and bring forward any tests and/or changes that are still relevant? Thanks for your contribution.

PR #726, opened 2014-08-23
Fixes for RFC 2231

sigmavirus24:

PR #892, opened 2015-06-16
Omit unnecessary ?s in EXTENDED_INITIAL_VALUE

sigmavirus24:
The ? outside the groups is redundant and Ruby's regular expression engine
rightfully produces a warning when compiling the regular expression.

Removing the ?s prevents the warning from being printed to stderr.

See also: http://rubular.com/r/J4CWeIHQTK

---

This is a quick fix up of the warnings produced by #726


matthewd:
There seem to be several things amiss here

sigmavirus24:
@matthewd can you elaborate?

matthewd:
* the `?` is not unnecessary
* `EXTENDED_OTHER_PARAMETER` will only match single-byte values
* `DISPPARM` is clearly wrong
* there are no tests that demonstrate any of the above problems
* the removal of the anchors on `RFC2183` is extremely suspicious
* your rubular example is not doing what you think it is

jeremy:
#913

Issue #419, opened 2012-08-23
[response].flatten causes infinite loop


dchelimsky:
- `Rack::Response#finish` is aliased with `to_ary` [1]
- `Rack::Response#finish` returns an array including `self`when the response code is anything other than 204, 205, 304 [2]

The result is an infinite loop if the result of `app.call(env)` (assigned to `response` in Rails' testing framework) is included in an array that is flattened. The reason is that `Array#flatten` invokes `to_ary` on each element that responds to `to_ary`, then does the same to the contents of the result of `to_ary`, and so on.

Because `self` is in the array returned by `to_ary`, it gets `to_ary` invoked and returns the same array, including `self`, which then gets `to_ary` invoked on it, which returns the same array, including `self`, and so on.

[1] https://github.com/rack/rack/blob/ab67e70d82911c0c86dce737c7064d0bd6daf3d4/lib/rack/response.rb#L83
[2] https://github.com/rack/rack/blob/ab67e70d82911c0c86dce737c7064d0bd6daf3d4/lib/rack/response.rb#L79

ioquatix:
@jeremy we will implement your solution.

PR #895, opened 2015-06-18
Support BodyProxy objects in ContentLength middleware

### LINKED ISSUE #734: Rack::BodyProxy with Rack::ContentLength middleware
I ran into this issue using the Rack::ContentLength middleware with Rails 4.1.5. The middleware checks  body.respond_to?(:to_ary) before calculating the Content-Length header. This fails if the body is a Rack::BodyProxy. The line has been added in 5b251a986962f3577e14c01a698fab9dd21383ba to fix https://github.com/rack/rack/issues/419 . Is there another solution for 419 or should the middleware be changed? It seems there is also no spec for this.

wjordan:
Fixes #734 by localizing the workaround added to `BodyProxy` from #419 to the specific API issue within `Response#to_ary`.

This patch is compatible with 1.6-stable. For 2.x, I would suggest deprecating/removing `Response#to_ary` entirely as [suggested](https://github.com/rack/rack/issues/419#issuecomment-55509509) by @jeremy, then all of the `NoAryBody` code can just be removed entirely.

### DIFF
--- a/lib/rack/body_proxy.rb
+++ b/lib/rack/body_proxy.rb
@@ -4,9 +4,8 @@ module Rack
       @body, @block, @closed = body, block, false
     end
 
-    def respond_to?(*args)
-      return false if args.first.to_s =~ /^to_ary$/
-      super or @body.respond_to?(*args)
+    def respond_to?(method_name, include_all=false)
+      super or @body.respond_to?(method_name, include_all)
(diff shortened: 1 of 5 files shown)


jeremyevans:
I merged a simpler approach for this in #1509.

PR #899, opened 2015-06-18
[1-6-stable] Fix GET semicolons without breaking API compatibility

matthewd:
Well.. without breaking compatibility in a way that affects Rails.

I'm not sure whether we should go with this, or just stick with the revert in cb9a68493b5a17a35c31b2c8cbacd81d5b0e4fae and declare semicolons to be a problem for future-Rack.   ¯\\\_(ツ)_/¯

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,3 +1,7 @@
+Fri Jun 19 07:14:50 2015  Matthew Draper <matthew@trebex.net>
+
+	* Work around a Rails incompatibility in our private API
+
 Fri Jun 12 11:37:41 2015  Aaron Patterson <tenderlove@ruby-lang.org>
 
 	* Prevent extremely deep parameters from being parsed. CVE-2015-3225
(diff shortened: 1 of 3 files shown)


ericboehs:
If my understanding is correct, when rack is loaded with actionpack, `parse_query` is getting overrode by action_dispatch. Try raising in this method with actionpack loaded.

godfat:
We're overriding Rack::Request#parse_query to avoid `Utils.parse_nested_query` and the application broke due to this change.

    undefined method `split' for {:query=>"", :separator=>"&;"}:Hash

``` ruby
class Request < Rack::Request
  def parse_query(qs)
    Rack::Utils.parse_query(qs, '&')
  end
end
```

godfat:
I think you should do something like:

``` ruby
def parse_query(qs, d='&')
   # ...
end
```

and call it with:

``` ruby
parse_query(qs, '&;')
```

Instead of trying to make qs be a String or Hash and detect it.

PR #900, opened 2015-06-19
Allow users to disable the secure cookie warning for custom coders

dmcinnes:
Hello,

Since we've recently updated Rack we've been getting the "No secret option provided to Rack::Session::Cookie" warning in our logs. However we are providing a custom coder to `Rack::Session::Cookie` that handles all of the encryption:
https://github.com/heroku/identity/blob/master/lib/identity/main.rb#L12-L19

This pull request adds an option to `Rack::Session::Cookie` to disable this warning, but only if a coder is supplied in the cases were the user wants to manage the cookie encryption themselves:

```ruby
use Rack::Session::Cookie,
  coder: MyFancyEncryptionCoder.new,
  let_coder_handle_secure_encoding: true
```

Does this approach make sense? If not I'm happy to hear alternatives.

Thanks!

### DIFF
--- a/lib/rack/session/cookie.rb
+++ b/lib/rack/session/cookie.rb
@@ -105,7 +105,7 @@ module Rack
 
       def initialize(app, options={})
         @secrets = options.values_at(:secret, :old_secret).compact
-        warn <<-MSG unless @secrets.size >= 1
+        warn <<-MSG unless secure?(options)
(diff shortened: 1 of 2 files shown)


brandur:
Nice! Thanks @tenderlove!

dmcinnes:
Thanks!

Issue #849, opened 2015-05-04
Should cookie parsing follow RFC6265 strictly?


al2o3cr:
The cookie parser accepts both `;` and `,` as delimiters:

https://github.com/rack/rack/blob/7e0833d38c79d7d2fdf5de2a511a1a0cf9561f97/lib/rack/request.rb#L313

This follows the original specifications in [RFC 2109](http://www.rfc-editor.org/info/rfc2109) and [RFC 2965](http://www.rfc-editor.org/info/rfc2965), but both of these have been obsoleted by [RFC 6265](http://www.rfc-editor.org/info/rfc6265) which does not include the possibility of `,` as a delimiter in `Cookie:` headers. Note that `,` is still not allowed in cookie values (it is explicitly excluded from `cookie-octets` in RFC6265).

I ran across this today investigating whether Rails / Rack could be subject to the parsing error mentioned here:
https://hackerone.com/reports/14883

tenderlove:
Huh. Honestly I have no idea if we should be strict.  I'd like to know what browsers send today.  Other than that, I'm planning on making master 2.0, so we could experiment with making it strict and see what happens.

tenderlove:
@al2o3cr master is 2.0.  Would you mind putting together a PR that does this?

Adam89:
@tenderlove and @al2o3cr I can put together a pr for this, would either of you mind?

tenderlove:
Please do it!

--
Aaron Patterson
http://tenderlovemaking.com/
I'm on an iPhone so I apologize for top posting.

Adam89:
I did not see much value in adding a test that makes sure the cookie parser only accepts `;`. Do you think I should add one and is there anything I might have missed in this pr?

Adam89:
cc @tenderlove

mrageh:
mind reviewing this pr @tenderlove :)

ioquatix:
Merged.

PR #907, opened 2015-06-30
Update cookie parser to only accept `;` as delimiter

mrageh:
The cookie parse used to accept both `;` and `,` as delimiters. This follows the original specifications in RFC 2109 and RFC 2965, but both of these have been deprecated by RFC 6265 which does not include the possibility of `,` as a delimiter in Cookie: headers. Note that `,` is still not allowed in cookie values (it is explicitly excluded from cookie-octets in RFC 6265).

From now on the cookie parser will only accept `;` as a delimiter to adhere to RFC 6265.

https://github.com/rack/rack/issues/849

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -205,12 +205,11 @@ module Rack
     module_function :parse_cookies
 
     def parse_cookies_header(header)
-      # According to RFC 2109:
-      #   If multiple cookies satisfy the criteria above, they are ordered in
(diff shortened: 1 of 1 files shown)


jvbreen1:
This looks good to me. It's a simple change, and it's well-commented.

mrageh:
This may be ready now cc @tenderlove :smile:

ioquatix:
Merged by hand. Thanks for your contribution.

PR #910, opened 2015-07-02
Introduce a second argument for Request#parse_query

godfat:
I think it's a bad idea to mix types (i.e. String and Hash here),
and Utils.parse_query already accepts a second argument for
passing the delimiter, so why not do the same?

Ref: https://github.com/rack/rack/pull/899

My motivation for patching this is because we have overridden
Request#parse_query as follows, and it broke due to the above
pull request:

    class Request < Rack::Request
      def parse_query(qs)
        Rack::Utils.parse_query(qs, '&')
      end
    end

Given my patch, our application would still break, but at least
we just need to accept the delimiter and pass it to Utils.parse_query,
rather than doing String/Hash handling again.

I know accepting this patch means our application would break again,
but I'll just adopt the change again when a new Rack version is
released.

By the way, the intention for overriding Request#parse_query was
trying to avoid Utils.parse_nested_query because I hate that
behaviour. If this could be an option for Rack::Request, that
would be awesome.

If we're willing to make this change, then I'll try to think about
how to make it an option.

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -188,7 +188,7 @@ module Rack
       if @env["rack.request.query_string"] == query_string
         @env["rack.request.query_hash"]
       else
-        p = parse_query({ :query => query_string, :separator => '&;' })
+        p = parse_query(query_string, '&;')
         @env["rack.request.query_string"] = query_string
         @env["rack.request.query_hash"]   = p
       end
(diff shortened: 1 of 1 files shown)


matthewd:
This is what was happening before cb9a68493b5a17a35c31b2c8cbacd81d5b0e4fae reverted it. 

It's definitely a better API -- which is why we've left it [on master](https://github.com/rack/rack/blob/971209b412e993f47b1956d4711b6ad6faa05689/lib/rack/request.rb#L367).

#899 was the best solution we could find that didn't break Rails (which, like you, is overriding the private `parse_query` method -- but it calls `super` without looking at the argument).

godfat:
Alright... fair enough :( Sorry for not being aware of it.
I probably can't hate Rails more :(

PR #911, opened 2015-07-15
Make static.rb standalone

evanphx:
People sometimes require `rack/static` without `rack`, resulting in constant errors. This fixes that error.


evanphx:
The source of this bug comes from https://github.com/puma/puma/issues/732

PR #913, opened 2015-07-18
Review filename header parsing regexps

matthewd:
This corrects the issues I enumerated in https://github.com/rack/rack/pull/892#issuecomment-112497770, along with a couple of others I spotted along the way.

It's worth noting that we're still not handling [parameter value continuations](https://tools.ietf.org/html/rfc2231#section-3), so our support for RFC 2231 is partial, at best.

Finally, while this brings our current set of regular expressions more in line with the ABNF specified in the updated RFC, I wonder if this is actually far more complication than we need... for our extremely limited purposes, it seems we could get away with a much simpler parser, just by being willing to accept some things the RFC technically disallows.


tenderlove:
> for our extremely limited purposes, it seems we could get away with a much simpler parser, just by being willing to accept some things the RFC technically disallows.

I'm willing to accept that.  As long as we aren't more strict than the RFC, it's probably fine.  I'm going to merge this for now.

PR #915, opened 2015-07-23
Add manually registered servers as optional default server

boazsegev:
When a gem is loaded for a server, it should be assumed that the intention was to use the new server as the default server (or, at the very least, as an optional default server).

This is why the registered handlers registry should be used to chose the default server, falling back on the original puma and webrick servers.

### DIFF
--- a/lib/rack/handler.rb
+++ b/lib/rack/handler.rb
@@ -52,8 +52,9 @@ module Rack
       elsif ENV.include?("RACK_HANDLER")
         self.get(ENV["RACK_HANDLER"])
       else
-        @fall_back ||= ['thin', 'puma', 'webrick', 'lsws', 'cgi', 'fastcgi', 'scgi']
-        pick (@handlers.keys - @fall_back) rescue (pick @fall_back)
+        ordered_defaults = @handlers.keys - ['thin', 'puma', 'webrick', 'lsws', 'cgi', 'fastcgi', 'scgi']
+        ordered_defaults.insert -1, *['thin', 'puma', 'webrick']
+        pick ordered_defaults
       end
     end
 
--- a/lib/rack/handler.rb
+++ b/lib/rack/handler.rb
@@ -52,7 +52,8 @@ module Rack
       elsif ENV.include?("RACK_HANDLER")
         self.get(ENV["RACK_HANDLER"])
       else
-        pick ['thin', 'puma', 'webrick']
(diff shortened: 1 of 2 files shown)


mig-hub:
Your patch seems to fail the CI test.
According to the error log, it happens with CGI handler on JRuby.

Error is:
``` ruby
LoadError: Couldn't find handler for: .
```

You can see the full log [here](https://travis-ci.org/rack/rack/builds/72242028).

boazsegev:
I seems the JRuby build doesn't stop at the webrick handler...

I tried to update the patch, but I can't manage it for some reason... I'll close this push request and open a new one.

boazsegev:
Okay, I managed to commit to this patch - the build was working when I checked - could you have a look?

boazsegev:
The build passes and the pull request is ready to be merged, now the question is if you agree with the premise stated in the pull request:

> When a gem is loaded for a server, it should be assumed that the intention was to use the new server as the default server (or, at the very least, as an optional default server).

> This is why the registered handlers registry should be used to chose the default server, falling back on the original puma and webrick servers.

tenderlove:
Is there any kind of test you can add for this behavior?

boazsegev:
I tried to test the defaults, but I couldn't manage to test the default handler before the RockLobster added itself to the Handler registry.

The test I added shows that the new default method recognizes the added handler and sets it as a default handler, which is the expected behavior for a 'smarter' default method.

boazsegev:
I didn't manage to write tests for the default method for the following reasons:

1. The default value changes depending on the machine's available servers (i.e. if thin/puma are installed) and included gems (any other registered handlers).

2. The test's order is seems to be inconsistent, so that it's impossible to control the expected default value. For instance, during the Ruby 2.2.2 testing the tests ran in reverse, while on JRuby they ran in order of creation... I'm sure I'm missing something there, but on either occasion the other tests affect the expected result for the `default` method, so that the assertions couldn't be pre-determined (some the tests register new handlers, so that the expected `default` value is updated and changed).

Maybe you could help with this... on a clean environment I was expecting:

i.e.: `Rack::Handler.default.name.must_match /webrick|thin|puma/i`

and after a handler's registration, I was expecting `default` to return the registered handler (if a number of new handlers were registered, on of them should have been returned).

i.e.: `Rack::Handler.default.must_equal Rack::Handler::RegisteringMyself`

boazsegev:
Any updates on this?

boazsegev:
I'll be closing this pull request soon, as it isn't going anywhere. I appreciate the time and thoughts that you put into the question of margin this branch.

PR #918, opened 2015-07-28
Fix rack abstract id key option

victormours:
Hi there!

The `:key` option was documented (see https://github.com/rack/rack/blob/master/lib/rack/session/abstract/id.rb#L164) but ignored.

I split this into two commits to make it easier to check that the additional test is failing, but I can squash them into one if you want to keep the git log cleaner (which I assume you do).

### DIFF
--- a/lib/rack/session/abstract/id.rb
+++ b/lib/rack/session/abstract/id.rb
@@ -245,10 +245,10 @@ module Rack
         # metadata into 'rack.session.options'.
 
         def prepare_session(env)
-          session_was               = env[RACK_SESSION]
-          env[RACK_SESSION]         = session_class.new(self, env)
+          session_was               = env[@key]
+          env[@key]         = session_class.new(self, env)
           env[RACK_SESSION_OPTIONS] = @default_options.dup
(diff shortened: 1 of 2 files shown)


nicoolas25:
+1

nicoolas25:
I think there is an mixup here because there is a difference between the cookie's name and the `env`'s key name.

You are not setting the cookie name when you do:

``` ruby
def prepare_session(env)
  session_was = env[@key]
  env[@key] = session_class.new(self, env)
  env[RACK_SESSION_OPTIONS] = @default_options.dup
  env[@key].merge! session_was if session_was
end
```

Your using the cookie's name (`@key`) as the `env`'s key where the session will be contained.

On the other hand, the `key` option seems to change the cookie's name correctly:

``` ruby
def extract_session_id(env)
  request = Rack::Request.new(env)
  sid = request.cookies[@key]
  sid ||= request.params[@key] unless @cookie_only
  sid
end
```

victormours:
Oh, you're totally right, I was confused. Thanks for clearing things up. :-)

Issue #740, opened 2014-10-02
Rack::Static middleware works with and without root url in urls parameter


akonsu:
this is originally posted on stackoverflow: http://stackoverflow.com/questions/26167068/my-rackstatic-middleware-works-with-root-url-in-urls-parameter-and-without
I have this in my `config.ru`:

```
require 'rack'
require './app.rb'

use(Rack::Static,
  :urls => %w(/css /fonts /images /js),
  :root => '../static',
  :index => 'index.html')
end

  run MyApp
```

it correctly serves static files for all four of the above URLs, but in addition, it also serves `index.html` from the root URL as if I had `:urls => %w(/ /css /fonts /images /js)`.

mcantor:
For the sake of clarity, please confirm I have all of this correct:

Your local filesystem looks like this:

```
├── static/
│   ├── css/
│   ├── fonts/
│   ├── images/
│   ├── js/
│   └── index.html
└── app/
    ├── config.ru
    └── app.rb
```

# Expected Behavior

Request  | Response
------------------- | ---------------
`GET / `           | Content of `static/index.html`
`GET /index.html` | `404 Not Found`

# Actual Behavior

Request  | Response
------------------- | ---------------
`GET / `           | ???
`GET /index.html` | Content of `static/index.html`

Is that correct?

mig-hub:
I did not test it yet but it looks like it could because it tests for `route_file || override_file_path`, and `override_file_path` only tests against this regexp `/\/$/`. So it looks like it `can_serve` paths which do not start with the URLs, but end with a forward slash.

Line 99 is not easy to understand for me.

jeremyevans:
This should be fixed in 2.1.0.

PR #919, opened 2015-08-01
Fixes an issue where `Rack::Static` can raise a TypeError

arronmabrey:
Fixes an issue where `Rack::Static` can raise "TypeError: no implicit conversion of nil into String"

This is due a bug that always incorrectly tries to append `@index` to a root path.
This happens even when the "index option" is not given, thus `@index == nil`.

### DIFF
--- a/lib/rack/static.rb
+++ b/lib/rack/static.rb
@@ -95,8 +95,12 @@ module Rack
       @file_server = Rack::File.new(root)
     end
 
+    def add_index_root?(path)
+      @index && path =~ /\/$/
+    end
+
     def overwrite_file_path(path)
-      @urls.kind_of?(Hash) && @urls.key?(path) || @index && path =~ /\/$/
+      @urls.kind_of?(Hash) && @urls.key?(path) || add_index_root?(path)
     end
 
     def route_file(path)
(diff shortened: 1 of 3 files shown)


arronmabrey:
This could also be easily added to the 1.6-stable branch. If you would like, I can create another PR for the as well.

mig-hub:
It seems related to [this issue](https://github.com/rack/rack/issues/740).
I don't know what other people think, but to me the part which defines what can be served could be refactored a bit to make it clearer.

arronmabrey:
@mig-hub Yes I agree, it could use some work for sure. I believe the issue you linked is also related. 

The regex `/\/$/` that's used to determine if it should add `@index` to the path seems to not always be a good default.

I would have been able to fix my own issue, if that regex was easily overridden.

arronmabrey:
This PR at least puts the logic surrounding that regex in a single place. I could easily make it a configurable option, defaulting to what is currently there now. That's if we think adding more surface area to public API is a good idea.

arronmabrey:
@mig-hub do you have the ability to merge PRs or do we need to get a maintainer involved to sign-off on the direction?

mig-hub:
@arronmabrey , no I am not a maintainer. Sorry if I made you think otherwise.

arronmabrey:
@mig-hub oh no you didn't, I was just checking :-)

arronmabrey:
@chneukirchen would be you able to provide any direction on how to proceed from here? I'm willing to do the work :-)

PR #921, opened 2015-08-03
Update full logo with symmetrical icon.

mcantor:
cf. rack/rack.github.com#10

I hope there are no philosophical problems with the fact that I did the update in Adobe Illustrator instead of Inkscape. I kept the old logo as a separate layer in the document for posterity.


chneukirchen:
Perhaps it would be nice to add the XML comment regarding copyright again.

mcantor:
@chneukirchen Good call; done.

PR #922, opened 2015-08-04
Lazily instantiate ERB template object

mcantor:
Implemented @tenderlove's suggestion on @Wardrop's PR #836.

The indentation fix is a little pedantic, but the `gsub()` should only be called once, just like the template instantiation, so it shouldn't be a deal-breaker as far as I can tell.

Please close #836 if/when this is closed.


tenderlove:
(just closing and reopening to get travis to run again)

PR #923, opened 2015-08-05
Fix indent in document

ganmacs:
I think indent is wrong in document. so Fix it.


spastorino:
Add [ci skip] to the commit message

ganmacs:
Sorry, I added `[ci sikp]`.

PR #937, opened 2015-09-03
Remove duplicate assertion from request spec

nmk:


ioquatix:
Merged by hand, thanks for your contribution.

PR #938, opened 2015-09-04
HeaderHash must clear names on clear/replace.

raxoft:
Without this fix, `@names` contains old names after clear/replace call, causing include?/key?/etc. return true even if the key is not there anymore.

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -414,6 +414,12 @@ module Rack
         @names = other.names.dup
       end
 
+      # on clear, we need to clear @names hash
+      def clear
+        super
+        @names = {}
+      end
+
       def each
         super do |k, v|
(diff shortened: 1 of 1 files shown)


jkowens:
Any reason why this hasn't been merged?

lib/rack/utils.rb
```diff
@@ -414,6 +414,12 @@ def initialize_copy(other)
         @names = other.names.dup
       end
 
+      # on clear, we need to clear @names hash
+      def clear
+        super
+        @names = {}
```
mikegee: We can clear the hash without allocating a new one: [`Hash#clear`](http://ruby-doc.org/core-2.2.0/Hash.html#method-i-clear)

ioquatix:
Thanks for your contribution.

PR #943, opened 2015-09-06
Fix bug in parsing of Content-Disposition header where an unquoted name at end-of-line sucked in the trailing newline

ggrossman:
Hi,

I've encountered an edge case in the parsing of the multipart `Content-Disposition` header. [RFC2183](https://tools.ietf.org/html/rfc2183) specifies that the parameter value can be either a token or a quoted string, and Rack [supports this](https://github.com/rack/rack/blob/b2d73960e9ea6b8b15321ef190f13a290d1aedf0/lib/rack/multipart.rb#L20). However, the way the `MULTIPART_CONTENT_DISPOSITION` regex is written, if the quotes are left off and the parameter happens to be at the end of the line, the newline character gets sucked in.

Example: Given this input
```
--AaB03x
Content-Type: text/plain
Content-Disposition: attachment; name=inline

true
--AaB03x--
```
The resulting `params` contains `{"inline\r\n" => "true"}`

With this patch, you get the desired result `params = {"inline" => "true"}`

This PR

* Changes the `MULTIPART_CONTENT_DISPOSITION` regex to more exactly represent what is described in RFC2183: either a token (no tspecial chars), or a quoted string.
* Adds a test for this case

### DIFF
--- a/lib/rack/multipart.rb
+++ b/lib/rack/multipart.rb
@@ -17,7 +17,7 @@ module Rack
     BROKEN_QUOTED = /^#{CONDISP}.*;\sfilename="(.*?)"(?:\s*$|\s*;\s*#{TOKEN}=)/i
     BROKEN_UNQUOTED = /^#{CONDISP}.*;\sfilename=(#{TOKEN})/i
     MULTIPART_CONTENT_TYPE = /Content-Type: (.*)#{EOL}/ni
-    MULTIPART_CONTENT_DISPOSITION = /Content-Disposition:.*\s+name="?([^\";]*)"?/ni
+    MULTIPART_CONTENT_DISPOSITION = /Content-Disposition:.*\s+name=(#{TOKEN}|"[^\";]*")/ni
     MULTIPART_CONTENT_ID = /Content-ID:\s*([^#{EOL}]*)/ni
     # Updated definitions from RFC 2231
     ATTRIBUTE_CHAR = %r{[^ \t\v\n\r)(><@,;:\\"/\[\]?='*%]}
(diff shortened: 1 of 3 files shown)


tenderlove:
Ugh, that's no fun.  Will this also support names with quotes in it?  Like if a parameter name legitimately ends with a quote?  I'm afraid the gsub might remove legitimate quotes.

lib/rack/multipart.rb
```diff
@@ -17,7 +17,7 @@ module Multipart
     BROKEN_QUOTED = /^#{CONDISP}.*;\sfilename="(.*?)"(?:\s*$|\s*;\s*#{TOKEN}=)/i
     BROKEN_UNQUOTED = /^#{CONDISP}.*;\sfilename=(#{TOKEN})/i
     MULTIPART_CONTENT_TYPE = /Content-Type: (.*)#{EOL}/ni
-    MULTIPART_CONTENT_DISPOSITION = /Content-Disposition:.*\s+name="?([^\";]*)"?/ni
+    MULTIPART_CONTENT_DISPOSITION = /Content-Disposition:.*\s+name=(#{TOKEN}|"[^\";]*")/ni
```
grosser: it might be more correct to match separately with: `(?:(#{TOKEN})|"([^\";]*)")`
and then not do a gsub!
```
match = head.match(MULTIPART_CONTENT_DISPOSITION)
match[1] || match[2] || head[MULTIPART_CONTENT_ID, 1]
```

grosser: or do `name.gsub!('\A"(.*)"\Z', "\\1")if name.start_with?('"')` since tokens cannot include a quote

ggrossman: What if we use the `VALUE` regex which seems to capture both the token and quoted-string case, and then use your alternate `gsub!`?

grosser: that would work too ... and it's a nice reuse ... also the quoted string is first, which should be more performant for the 99.99% case ...

ggrossman:
Hi @tenderlove! I tweaked the implementation and it will now support embedded quote chars within a name expressed as a quoted-string.

tenderlove:
Great, thank you!

PR #944, opened 2015-09-08
Convenient cache and content type for response generation

ioquatix:
These three convenient methods make it easy to declaratively specify how a response should be structured. For example:

```
response.do_not_cache!

response.cache!

response.content_type = 'text/html'
```

Open to discussion w.r.t. naming, and behaviour. I monkey patch these in which might be fine for my usage, but I thought it could be useful to a wider audience.

### DIFF
--- a/lib/rack.rb
+++ b/lib/rack.rb
@@ -38,6 +38,7 @@ module Rack
   SERVER_ADDR       = 'SERVER_ADDR'.freeze
   SERVER_PORT       = 'SERVER_PORT'.freeze
   CACHE_CONTROL     = 'Cache-Control'.freeze
+  EXPIRES           = 'Expires'.freeze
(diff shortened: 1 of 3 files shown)


ioquatix:
I have rebased this, is it possible to get some feedback? Thanks :)

ioquatix:
I can rebase this again, but is there any point? @sgrif @matthewd

jeremyevans:
These changes look good to me.  They do not appear to affect backwards compatibility, so I would think we could add them in 2.2 (if that is coming before 3.0).  @ioquatix can you rebase against master?

ioquatix:
Ok :)

ioquatix:
Closing in favour of https://github.com/rack/rack/pull/1555

PR #947, opened 2015-09-13
fix rdoc formatting (remove accidental links)

davidrunger:
In several places, the rdoc comments attempt to indicate hash access, for
example `env['rack.input']`, but rdoc interprets this as a link with
text "env" and href="%27rack.input%27/". Wrapping these snippets of hash
access code in `<tt>` tags escapes the link syntax and also renders the
code snippets in monospace.

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -16,12 +16,12 @@ module Rack
       super(env)
     end
 
-    # shortcut for request.params[key]
+    # shortcut for <tt>request.params[key]</tt>
     def [](key)
       params[key.to_s]
     end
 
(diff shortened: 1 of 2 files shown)


davidrunger:
An example of the current problem can be seen here: http://www.rubydoc.info/github/rack/rack/Rack/Request#update_param-instance_method

The last line of the text under `#update_param(k, v) ⇒ Object` currently reads "env is not touched", but it means to say "env['rack.input'] is not touched". The documentation as it is being generated now is very confusing, because parts of "env" *are* touched - so I think that this is an important update to the doc.

tenderlove:
Thanks!

PR #948, opened 2015-09-15
Use standard header for method override.

tylergannon:
The documentation and specs discuss a header called `X-HTTP-Method-Override` for use as the method override signifier, but in practice it's called `HTTP_X_HTTP_METHOD_OVERRIDE`.  This is not in keeping with standard HTTP header naming convention and it also differs from what others are [using](https://github.com/mgonto/restangular#setmethodoverriders).

This pull request adds the standard header `X-HTTP-Method-Override` into the mix and also adds a deprecation warning for those using `HTTP_X_HTTP_METHOD_OVERRIDE`.

I want to request that we move to the same header name that's used elsewhere in the industry.

If someone can tell me where to go to add information to the correct page in rails guides, I'll happily add this there, as the ability to use a header for Method Override is currently undocumented in rails guides.

thanks!

### DIFF
--- a/lib/rack/method_override.rb
+++ b/lib/rack/method_override.rb
@@ -3,7 +3,8 @@ module Rack
     HTTP_METHODS = %w[GET HEAD PUT POST DELETE OPTIONS PATCH LINK UNLINK]
 
     METHOD_OVERRIDE_PARAM_KEY = "_method".freeze
-    HTTP_METHOD_OVERRIDE_HEADER = "HTTP_X_HTTP_METHOD_OVERRIDE".freeze
+    HTTP_METHOD_OVERRIDE_HEADER_DEPRECATED = "HTTP_X_HTTP_METHOD_OVERRIDE".freeze
+    HTTP_METHOD_OVERRIDE_HEADER = "X-HTTP-Method-Override".freeze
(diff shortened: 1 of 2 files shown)


matthewd:
You seem to be confusing the header name with the env hash key name.

tylergannon:
Can you please explain?  How does the header name differ from what's passed into the `:[]` method on the `env` object?  

To my eye, it looks like the actual [header name](https://github.com/rack/rack/blob/master/lib/rack/method_override.rb#L6) is called `HTTP_X_HTTP_METHOD_OVERRIDE`.  Since the name is set in the above link and then referenced [here](https://github.com/rack/rack/blob/master/lib/rack/method_override.rb#L28) 

To clarify: I have found that if I send a method override in a header called `X-HTTP-Method-Override`, nothing happens.  But if I send a header that's called `HTTP_X_HTTP_METHOD_OVERRIDE`, then I get the correct result.

This is important for JSON requests because rack (AFAIK) hasn't loaded JSON so I can't send `_method` as an argument in the body.

matthewd:
https://github.com/rack/rack/blob/master/SPEC#L54-L65

tylergannon:
Thank you!

PR #949, opened 2015-09-16
Replace \t to 2 whitespaces in HISTORY.md

wildjcrt:
Styling PR.
I replace all \t to 2 whitespaces in HISTORY.md.

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,74 +1,74 @@
 Fri Sep  4 18:34:53 2015  Aaron Patterson <tenderlove@ruby-lang.org>
 
-	* `Rack::Session::Abstract::ID` IS DEPRECATED.  Please switch to
-	`Rack::Session::Abstract::Persisted`.
-	`Rack::Session::Abstract::Persisted` uses a request object rather than
-	the `env` hash.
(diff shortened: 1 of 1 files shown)


tenderlove:
This is the way that vim generates the changelog entries.  Is there a compelling reason to change this since it means I'll have to figure out how to get vim to generate changelog entries with spaces?

arthurnn:
@tenderlove if you wanna change to emacs :trollface: https://github.com/arthurnn/dotfiles/blob/master/.emacs.d/init.d/setup-etc.el#L23-L26

tenderlove:
I don't understand, this ChangeLog style [came from Emacs](http://www.gnu.org/software/emacs/manual/html_node/emacs/Format-of-ChangeLog.html)

arthurnn:
> this ChangeLog style came from Emacs
Thats right. 

Jokes aside. I dont think this PR adds value to the changelog, it is just the style we were following. So :-1:  i would say.

mig-hub:
You guys managed to mix the editor war and the "tab vs spaces" war in a single thread.
It probably is somebody's wet dream.

PR #953, opened 2015-09-25
Avoid re-calculation in Rack::Builder call method 

jackxxu:
in a simple `config.ru` like this one below, 

``` ruby
require 'rack'

class App
  def initialize(); puts "NEW APP INSTANCE CREATED"; end
  def call(env); [200, {}, [""]]; end
end

app = Rack::Builder.new do
  map '/' do
    run App.new
  end
end

run app
```

each time the builder app receives a new request, a new instance of the App class is created. 

```
[2015-09-25 13:14:06] INFO  WEBrick 1.3.1
[2015-09-25 13:14:06] INFO  ruby 2.2.2 (2015-04-13) [x86_64-darwin14]
[2015-09-25 13:14:06] INFO  WEBrick::HTTPServer#start: pid=93874 port=9292
NEW APP INSTANCE CREATED
127.0.0.1 - - [25/Sep/2015 13:14:08] "GET / HTTP/1.1" 200 - 0.0014
NEW APP INSTANCE CREATED
127.0.0.1 - - [25/Sep/2015 13:14:09] "GET / HTTP/1.1" 200 - 0.0003
```

This is not efficient, as the `#to_app` method is invoked to re-calculate the mapped stack for each request. 

This PR simply caches the `#to_app` value in the `call` method.

alternatively, use `Rack::Builder.app` method.

### DIFF
--- a/lib/rack/builder.rb
+++ b/lib/rack/builder.rb
@@ -150,7 +150,7 @@ module Rack
     end
 
     def call(env)
-      to_app.call(env)
+      (@_app ||= to_app).call(env)
     end
 
     private
--- a/test/spec_builder.rb
+++ b/test/spec_builder.rb
@@ -180,6 +180,19 @@ describe Rack::Builder do
     end.must_raise(RuntimeError)
   end
 
+  it "doesn't dupe #to_app when mapping" do
+    app = builder do
+      map '/' do |outer_env|
+        run lambda { |env|  [200, {"Content-Type" => "text/plain"}, [object_id.to_s]] }
+      end
+    end
+
+    builder_app1_id = Rack::MockRequest.new(app).get("/").body.to_s
+    builder_app2_id = Rack::MockRequest.new(app).get("/").body.to_s
+
+    assert_equal builder_app2_id, builder_app1_id
+  end
+
   describe "parse_file" do
     def config_file(name)
       File.join(File.dirname(__FILE__), 'builder', name)


mig-hub:
@jackxxu, this is exactly what the method #to_app is for. In your example you don't use the Rack::Builder class correctly. You should use the #to_app method or use "app" instead of "new".

What you have is only an instance of Rack::Builder, not an app. I know it sounds a bit confusing, especially when you can actually run it, but it makes a difference.

mig

jackxxu:
@mig-hub, thanks for the response. if that is the case, i think that might be a violation of the principle of least surprise (different from how middlewares are normally initialized). after all, https://github.com/rack/rack/blob/master/lib/rack/builder.rb#L7-L15 uses `new` in an example provided in Builder itself.

mig-hub:
Well technically it does not violate the principle of least surprise because it does work.

The example you are referring to would probably be better off using the #app method, you are right.
The second example does, the one below.

I am not blaming you by the way because this is tricky.

That beeing said, forgive me if you already know because I don't know how well you know rack, but one rarely uses `Rack::Builder` like this directly. A `rackup` file is already encapsulated in a `Rack::Builder` app object.

Your example would actually be written like this:

```
class App
  def initialize(); puts "NEW APP INSTANCE CREATED"; end
  def call(env); [200, {}, [""]]; end
end

map '/' do
  run app
end
```

No need to require rack either since it is a `rackup` file.
A `rackup` file is not a standalone ruby file.
It is injected inside a `Rack::Builder` block.
for example by the `rackup` command when you run it.
This is done with this method which receives the content of your `config.ru` file:
https://github.com/rack/rack/blob/master/lib/rack/builder.rb#L48-L51

I personally only used it in the code of other middlewares.

Maybe you already knew and it was only for the sake of the example, I don't know.
But I hope it clarifies things a bit.

Cheers,
mig

jackxxu:
:) although, I am not sure I follow the logic of why it is technically not a violation of least surprise, just because "it does work" even though you mentioned it is confusing. 

i do understand the rack server code, but are there incidents where there may be nested maps, or need to create Builder app instances? 

It will create confusion to a user, no matter how rare it is.

mig-hub:
@jackxxu, I don't want to fall into a "definition" argument because that is just sterile and brings us nowhere. For me a violation of least surprise would mean that it breaks unexpectedly, if you think differently then I won't argue with that.

Now I said it is "confusing", as in it is not easy to grasp immediately. Rack is based on a meta/modular/nested  concept and each time I tried to explain it to another developer, I always struggled a bit. But Rack is still to me a good example of "complex vs complicated", because it is still based on basic concepts.

Now to be proactive, I completely understand the purpose of your pull request but there are 2 problems to it. 

First it does not pass the CI test (that thing with a big red cross below), which means that your code fails the tests with a certain type of OS, or with a certain version of Ruby. Therefore it is not a patch that the committers can merge in its current state.

The second and most important problem, is that you are actually trying to replicate what the `self.app` method is doing. The difference is that the `self.app` allows you to instantiate when you start your server, whereas your version is instantiating on the first request, a bit like a cache.

Now you still can argue that it would be good to make it bulletproof and always cache in any circumstances. I am not necessarily crazy about this because there may be a problem with this for tests or for using a reloader , or maybe it is just me because I want too much control, but if you want to do another version, then it makes more sense to cache the value in `#to_app` itself, not in `#call`. Having caching of the object done when you call this object.

At least if you do a version which passes the CI, you'll have more chances of getting a feedback from a maintainer who will discuss with you the pros and cons of your solution.

Cheers,
mig

raggi:
I have had to squash patches for this so many times. We need to add a doc comment.

This breaks reloader use cases, and it works this way by design. `to_app` is for production use.

ioquatix:
Duplicate of https://github.com/rack/rack/pull/834

raggi:
This just regressed `Rack::Reloader`

This should be rolled back.

See my comments on all the related issues and PRs that stated this.

raggi:
@rafaelfranca ^^^^

rafaelfranca:
Oh! Now I got you comment. Reverted.

matthewd:
Can we add a test to defend the desired behaviour?

rafaelfranca:
Yeah, that is what I'm working right now. Since this is desired behavior this and related PR can be closed.

perlun:
For reference, the added test is here, if I parse this right: https://github.com/rack/rack/commit/ab008307cbb805585449145966989d5274fbe1e4

PR #962, opened 2015-10-22
adding HTTP status codes

tilo:
adding some missing HTTP status codes


perlun:
Looks sane enough to me. Could we get someone to merge this?

gioele:
With the exception of 418, all these status codes are not defined in any RFC and some of them (for example 420) are used by various frameworks with diverging meanings.

talent-io:
Sorry to jump in this 2 years and a half later. 

I was actually looking for 418 error code, and it seems it's actually been removed by https://github.com/rack/rack/pull/933. So I guess this PR can now be closed :).

rafaelfranca:
@gioele is right. We can't add to rack if they are not in the RFCs. Thank you.

PR #963, opened 2015-10-29
[ci skip] Add `#delete_session` to should be overwritten list.

yui-knk:


ioquatix:
Thanks for your contribution.

PR #965, opened 2015-10-31
Remove splat destructuring of Response instances for Ruby 1.9.2

ch1c0t:


ioquatix:
This has been applied to master already. Thanks for your contribution.

PR #966, opened 2015-11-01
Use Float::INFINITY in Rack::URLMap instead of rolling our own

ch1c0t:
`Float::INFINITY` is here since 1.9.2.
`Rack::URLMap::NEGATIVE_INFINITY` seems to be safe to remove since 60c66b5e.


ioquatix:
Thanks for your contribution.

PR #968, opened 2015-11-06
Freeze a few more commonly use strings

mattrobenolt:
This shouldn't impact much, but why not?

I noticed that this was being done as hygiene for lots of the other frequently access cgi vars.

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -139,6 +139,12 @@ module Rack
       HTTP_X_FORWARDED_HOST   = 'HTTP_X_FORWARDED_HOST'.freeze
       HTTP_X_FORWARDED_PORT   = 'HTTP_X_FORWARDED_PORT'.freeze
       HTTP_X_FORWARDED_SSL    = 'HTTP_X_FORWARDED_SSL'.freeze
+      HTTP_X_REQUESTED_WITH   = 'HTTP_X_REQUESTED_WITH'.freeze
+      HTTP_X_FORWARDED_FOR    = 'HTTP_X_FORWARDED_FOR'.freeze
+      HTTP_USER_AGENT         = 'HTTP_USER_AGENT'.freeze
+      HTTP_REFERER            = 'HTTP_REFERER'.freeze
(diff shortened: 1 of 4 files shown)


lib/rack/request.rb
```diff
@@ -151,11 +157,11 @@ def request_method;  get_header(REQUEST_METHOD)                     end
       def query_string;    get_header(QUERY_STRING).to_s                  end
       def content_length;  get_header('CONTENT_LENGTH')                   end
```
mattrobenolt: So I wanted to change this use of `'CONTENT_LENGTH'` and a couple uses of `'CONTENT_TYPE'`, but we have a conundrum.

There's a constant defined `CONTENT_LENGTH = 'Content-Length'.freeze`, notice the NOT ALL CAPS, so switching this to use the constant value is actually incorrect for this case.

raggi: This is not the HTTP header it's an env key that must be supplied by the server per spec.

get_header is a misnomer, added in:

commit ef5546ce355f180c68bd0ce0d326b5e2c2d610ca
Author: Aaron Patterson <aaron.patterson@gmail.com>
Date:   Thu Aug 20 16:28:06 2015 -0700

It probably wants to be fixed to avoid this kind of confusion.

matthewd:
Master requires ruby 2.2, so if anything, we should probably be purging some of the existing constants in favour of inline `.freeze` calls (or plain literals in the case of hash accesses, which freeze automagically).

raggi:
This was my original policy, then we pulled in the constants, then ruby finally fixed these cases.

ioquatix:
@mattrobenolt can you please rebase on master, taking into account we now have `frozen_string_literal: true`

mattrobenolt:
This PR is over 4 years old. I don't use rack or have any vested interest in the Ruby community. This should be trivial for someone to port over to master I'd expect. I don't even know what `frozen_string_literal` means.

😕

ioquatix:
Okay, fair enough, thanks for getting back to me quickly.

PR #969, opened 2015-11-06
Do not reference HTTP_VERSION internally

mattrobenolt:
HTTP_VERSION is supposed to be a client supplied header. This usage
inside Rack is conflating it with SERVER_PROTOCOL, which imo is instead
also conflating it with the client's HTTP version from the request line.

In any of these cases, HTTP_VERSION is set when an existing Version
header doesn't already exist. So it's possible to send a Version header
to conflict with the expected behaviors.

According to the CGI spec
(https://tools.ietf.org/html/draft-robinson-www-interface-00)

>  Environment variables with names beginning with "HTTP_" contain
>    header data read from the client, if the protocol used was HTTP.

This is an anscillary issue with Rack, but will leave that open for
discussion since this behavior already exists.
## Notes

It's worth noting that this incorrect behavior is happening as a result of https://github.com/rack/rack/blob/028438f/lib/rack/handler/cgi.rb#L29 (and in all other handlers). I believe this behavior should be changed and not set to HTTP_VERSION in the first place. See my follow up: #970

Lastly, I'm not super familiar with Rack, but this seems problematic to rely on `SERVER_PROTOCOL` for `Transfer-Encoding` behaviors since what you really want is the client's HTTP version. I've looked into WEBrick only, but their `SERVER_PROTOCOL` does NOT mean the protocol of the request. WEBrick has a separate `request.http_version` as opposed to `request.env_meta['SERVER_PROTOCOL']`. Not sure about the others.

### DIFF
--- a/lib/rack/chunked.rb
+++ b/lib/rack/chunked.rb
@@ -43,7 +43,7 @@ module Rack
     # a version (nor response headers)
     def chunkable_version?(ver)
       case ver
-      when "HTTP/1.0", nil, "HTTP/0.9"
+      when 'HTTP/1.0', nil, 'HTTP/0.9'
         false
       else
         true
@@ -54,7 +54,7 @@ module Rack
       status, headers, body = @app.call(env)
       headers = HeaderHash.new(headers)
 
-      if ! chunkable_version?(env[HTTP_VERSION]) ||
+      if ! chunkable_version?(env[SERVER_PROTOCOL]) ||
          STATUS_WITH_NO_ENTITY_BODY.include?(status) ||
          headers[CONTENT_LENGTH] ||
          headers[TRANSFER_ENCODING]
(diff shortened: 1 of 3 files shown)


jvbreen1:
Looks good. It makes sense to me that the code should not reference the HTTP_VERSION, especially if it results in Rack setting an environment variable that other apps may expect to come from the client.

13121997:
Iam nit understood

ioquatix:
@tenderlove this looks like a good change to me. Can you confirm you are happy to merge?

ioquatix:
I will add entry to changelog if so.

ioquatix:
Rebased and merged by hand.

mattrobenolt:
Happy 4 year anniversary all! 🎉

PR #972, opened 2015-11-10
Pre-initialize reverse enumerator for middleware stack.

marshall-lee:
Hello!

This _clever trick_ eliminates creating a temporary array on every call to `call` :smiley:

### DIFF
--- a/lib/rack/builder.rb
+++ b/lib/rack/builder.rb
@@ -52,6 +52,7 @@ module Rack
 
     def initialize(default_app = nil, &block)
       @use, @map, @run, @warmup = [], nil, default_app, nil
+      @use_reverse = @use.reverse_each
       instance_eval(&block) if block_given?
     end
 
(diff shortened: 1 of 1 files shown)


marshall-lee:
With this patch in basic scenario `call` does not make any allocations other than a chain of middleware instances.

chneukirchen:
Alternatively, #use could just prepend the lambda.  Anyway, you should run #to_app yourself and keep the result for to #call it for every request?

marshall-lee:
@chneukirchen Yeah, it's cool too, but there're two concerns:

1. Lets look at the current implementation of `#use`:

```ruby
  if @map
    mapping, @map = @map, nil
    @use << proc { |app| generate_map app, mapping }
  end
end
```
To rewrite this with `unshift` I should also change the order of these two calls. Here code becomes less readable, IMO — the sense of *stack* fades. As for this patch, it changes the least.

2. As it's a performance-related patch I should say thay `#<<` is faster than `#unshift` by 20-30%. Of course in most cases it does not matter since `Rack::Builder` is being created only once when loading `config.ru` but I'd seen things that invoke `Rack::Builder` dynamically. [Grape](https://github.com/ruby-grape/grape) does it, for instance. Of course grape should be fixed but it's already a case.

marshall-lee:
Alternative: #974

PR #973, opened 2015-11-10
Update example use of Rack::Deflater

mwpastore:
The example if-lambda given in the documentation for Rack::Deflater is dysfunctional. `body` is an array in Rack responses and has been for a while, so checking to see if `body.length > 512` will likely never be true. Additionally, with different encodings taking multiple bytes per character, I think it's more appropriate to check the byte size of the body cf. the string length, so I've also updated the example to reflect that. I've been using this technique for a while and it works well.

### DIFF
--- a/lib/rack/deflater.rb
+++ b/lib/rack/deflater.rb
@@ -22,7 +22,7 @@ module Rack
     # [app] rack app instance
     # [options] hash of deflater options, i.e.
     #           'if' - a lambda enabling / disabling deflation based on returned boolean value
-    #                  e.g use Rack::Deflater, :if => lambda { |env, status, headers, body| body.length > 512 }
+    #                  e.g use Rack::Deflater, :if => lambda { |*, body| body.map(&:bytesize).reduce(0, :+) > 512 }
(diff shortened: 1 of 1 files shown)


lib/rack/deflater.rb
```diff
@@ -22,7 +22,7 @@ class Deflater
     # [app] rack app instance
     # [options] hash of deflater options, i.e.
     #           'if' - a lambda enabling / disabling deflation based on returned boolean value
-    #                  e.g use Rack::Deflater, :if => lambda { |env, status, headers, body| body.length > 512 }
+    #                  e.g use Rack::Deflater, :if => lambda { |*, body| body.map(&:bytesize).reduce(0, :+) > 512 }
```
dentarg: Could be

```ruby
if: ->(*, body) { body.map(&:bytesize).reduce(0, :+) > 512 }
```

which would [please](https://github.com/bbatsov/ruby-style-guide#lambda-multi-line) the [👮](https://github.com/bbatsov/rubocop).

Should be fine as rack now require Ruby 2.2.2.

PR #979, opened 2015-12-02
Rack::File and Rack::Directory to return 400 if path contains null byte

### LINKED ISSUE #971: Path containing null byte returns 500 status code
_From @questionmarc on November 9, 2015 19:31_

Arguably this should instead return the 400 http status code instead, since the path could be considered malformed / invalid. Throwing an error is expensive and should be reserved for truly exceptional situations.

**Steps to reproduce:**
Creating a new rails app
start up the rails server 
run `curl "localhost:3000/index.jsp%00x"` 

**Stacktrace:**

```
[2015-11-09 14:22:59] ERROR ArgumentError: string contains null byte
    /Users/pivotal/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rack-1.6.4/lib/rack/utils.rb:679:in `join'
    /Users/pivotal/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rack-1.6.4/lib/rack/utils.rb:679:in `clean_path_info'
    /Users/pivotal/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/actionpack-4.2.4/lib/action_dispatch/middleware/static.rb:28:in `block in match?'
    /Users/pivotal/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/actionpack-4.2.4/lib/action_dispatch/middleware/static.rb:27:in `map'
    /Users/pivotal/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/actionpack-4.2.4/lib/action_dispatch/middleware/static.rb:27:in `match?'
    /Users/pivotal/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/actionpack-4.2.4/lib/action_dispatch/middleware/static.rb:110:in `call'
    /Users/pivotal/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rack-1.6.4/lib/rack/sendfile.rb:113:in `call'
    /Users/pivotal/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/railties-4.2.4/lib/rails/engine.rb:518:in `call'
    /Users/pivotal/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/railties-4.2.4/lib/rails/application.rb:165:in `call'
    /Users/pivotal/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rack-1.6.4/lib/rack/lock.rb:17:in `call'
    /Users/pivotal/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rack-1.6.4/lib/rack/content_length.rb:15:in `call'
    /Users/pivotal/.rbenv/versions/2.2.2/lib/ruby/gems/2.2.0/gems/rack-1.6.4/lib/rack/handler/webrick.rb:88:in `service'
    /Users/pivotal/.rbenv/versions/2.2.2/lib/ruby/2.2.0/webrick/httpserver.rb:138:in `service'
    /Users/pivotal/.rbenv/versions/2.2.2/lib/ruby/2.2.0/webrick/httpserver.rb:94:in `run'
    /Users/pivotal/.rbenv/versions/2.2.2/lib/ruby/2.2.0/webrick/server.rb:294:in `block in start_thread'
```

_Copied from original issue: rails/rails#22235_

jkowens:
File paths cannot contain null byte characters, so this prevents an `ArgumentError: string contains null byte` from being raised during file path operations (`Rack::Utils::clean_path_info`, etc). Resolves #971.

Regarding backend transcoding [RFC 3986](https://tools.ietf.org/html/rfc3986#section-7.3) states:

>Note, however, that the "%00" percent-encoding (NUL) may require special handling and should be rejected if the application is not expecting to receive raw data within a component.

Any input as to whether this is a good approach(leaving null byte encoded) or if the encoded null byte should be removed or replaced with another character like a space?

### DIFF
--- a/lib/rack/directory.rb
+++ b/lib/rack/directory.rb
@@ -65,12 +65,24 @@ table { width:100%%; }
       script_name = env[SCRIPT_NAME]
       path_info = Utils.unescape_path(env[PATH_INFO])
 
-      if forbidden = check_forbidden(path_info)
+      if bad_request = check_bad_request(path_info)
+        bad_request
+      elsif forbidden = check_forbidden(path_info)
         forbidden
       else
         path = ::File.join(@root, path_info)
         list_path(env, path, path_info, script_name)
       end
+    end
+
+    def check_bad_request(path_info)
+      return if Utils.valid_path?(path_info)
+
+      body = "Bad Request\n"
+      size = body.bytesize
+      return [400, {CONTENT_TYPE => "text/plain",
+        CONTENT_LENGTH => size.to_s,
+        "X-Cascade" => "pass"}, [body]]
     end
 
     def check_forbidden(path_info)
--- a/lib/rack/file.rb
+++ b/lib/rack/file.rb
@@ -38,8 +38,9 @@ module Rack
       end
 
       path_info = Utils.unescape_path request.path_info
-      clean_path_info = Utils.clean_path_info(path_info)
+      return fail(400, "Bad Request") unless Utils.valid_path?(path_info)
 
+      clean_path_info = Utils.clean_path_info(path_info)
       path = ::File.join(@root, clean_path_info)
 
       available = begin
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -609,5 +609,12 @@ module Rack
     end
     module_function :clean_path_info
 
+    NULL_BYTE = "\0"
+
+    def valid_path?(path)
+      path.valid_encoding? && !path.include?(NULL_BYTE)
+    end
+    module_function :valid_path?
+
   end
 end
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -609,7 +609,7 @@ module Rack
     end
     module_function :clean_path_info
 
-    NULL_BYTE = "\0"
+    NULL_BYTE = "\0".freeze
 
     def valid_path?(path)
       path.valid_encoding? && !path.include?(NULL_BYTE)
--- a/test/spec_directory.rb
+++ b/test/spec_directory.rb
@@ -63,6 +63,13 @@ describe Rack::Directory do
     assert_match(res, /passed!/)
   end
 
+  it "serve uri with URL encoded NUL byte (%00) in filenames" do
+    res = Rack::MockRequest.new(Rack::Lint.new(app))
+      .get("/cgi/test%00")
+
+    res.must_be :bad_request?
+  end
+
   it "not allow directory traversal" do
     res = Rack::MockRequest.new(Rack::Lint.new(app)).
       get("/cgi/../test")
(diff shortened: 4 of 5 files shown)


jkowens:
@deepj Thanks for tip about `String#delete`.

deepj:
@jkowens See:

```ruby
require 'benchmark/ips'
require 'uri'

s = 'null%20byte%00'.freeze

Benchmark.ips do |x|
  x.report("String#gsub before unescaping") { ::URI::DEFAULT_PARSER.unescape s.gsub("%00", '') }
  x.report("String#delete after unescaping") { ::URI::DEFAULT_PARSER.unescape(s).delete("\u0000") }
end
```

Results:
```
gsub before unescaping
                        17.952k i/100ms
delete after unescaping
                        16.057k i/100ms
-------------------------------------------------
gsub before unescaping
                        217.278k (± 6.5%) i/s -      1.095M
delete after unescaping
                        179.904k (± 6.1%) i/s -    899.192k
```

jkowens:
That's interesting, I didn't expect gsub before unescaping to be a bit faster.

jkowens:
Instead of removing encoded null byte characters from the request path, I updated the pull request to ignore the encoded null byte when decoding. This will result in a 404 response instead of 500.

deepj:
:+1:

jkowens:
I've gone back to original thinking and decided to remove any encoded null bytes. This seems more consistent with what I've read other places. This applies only to looking for a static file, if the resulting file does not exist it still looks for a route with the requested path.

matthewd:
I think we should probably leave the null byte in the string, but then just ensure we 404 as soon as we notice it (because we have prior knowledge that the presence of a null byte means the path can't be referring to a file on disk). Any variant in which we interpret the path as possibly-valid seems to open the door to security issues.

jkowens:
@matthewd thanks you've helped think more clearly on this issue. I've been approaching it trying to resolve an issue I was seeing as described in #971, but now I realize that issue really needs to be solved in the Rails.

I do think there is a potential issue in `Rack::File` and `Rack:Directory` where it will raise an ArgumentError if PATH_INFO contains a null byte when calling `Rack::Utils#clean_path_info`. As you suggested, I think responding with a 400 or 404 before attempting any file path operations would be best.

jkowens:
@matthewd  if you get a chance, could you take a look at the updates I made to this PR and let me know your thoughts? Thanks!

arthurnn:
I think this make sense. 👍

mickey:
Any updates on this?
We get thousands of exceptions every time a "hacker" tries to find Windows vulnerabilities on our servers.

mickey:
Sorry for the noise, it was fixed on rails master and backported to 4.2.6
https://github.com/rails/rails/pull/23035
https://github.com/rails/rails/pull/23045

PR #981, opened 2015-12-10
Load session for read before accessing keys or values

michaelsauter:
Is there a specific reason why the session is not loaded for `#keys` or `#values`? Looks like it should ...

### DIFF
--- a/lib/rack/session/abstract/id.rb
+++ b/lib/rack/session/abstract/id.rb
@@ -124,10 +124,12 @@ module Rack
         end
 
         def keys
+          load_for_read!
           @data.keys
         end
 
         def values
+          load_for_read!
           @data.values
         end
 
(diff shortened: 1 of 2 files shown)


tarasstruk:
:+1:  Looks fine. These methods are on the same range with `#each`.  If they suppose to be called  only  `if loaded?==true` then they should provide a reasonable exception.

matthewd:
Is there somewhere we can add some tests for this?

michaelsauter:
@matthewd Sorry for the silence on this. I wasn't quite sure where to add tests. It looks like there are none yet for the SessionHash? I made a new test file now (5de59df) which tests the changed behaviour - those tests fail if run without the changes.

matthewd:
Thanks!

PR #985, opened 2015-12-21
Add support for HTTP 451 "Unavailable for Legal Reasons"

ktheory:
Adds support for the newly approved HTTP 451 response code,
“Unavailable for Legal Reasons”.

Adds `Rack::Response#unavailable?` method.

See also:

- [IETF draft specification](https://tools.ietf.org/html/draft-tbray-http-legally-restricted-status-05)
- [Why 451?](https://www.mnot.net/blog/2015/12/18/451)
- [451unavailable.org](http://www.451unavailable.org)

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -551,6 +551,7 @@ module Rack
       428 => 'Precondition Required',
       429 => 'Too Many Requests',
       431 => 'Request Header Fields Too Large',
+      451 => 'Unavailable for Legal Reasons',
       500 => 'Internal Server Error',
       501 => 'Not Implemented',
       502 => 'Bad Gateway',


lib/rack/response.rb
```diff
@@ -128,6 +128,7 @@ def not_found?;           status == 404;                        end
       def method_not_allowed?;  status == 405;                        end
       def precondition_failed?; status == 412;                        end
       def unprocessable?;       status == 422;                        end
+      def unavailable?;         status == 451;                        end
```
ktheory: This could be confusing with `503 Service Unavailable`. I'd be open to renaming this method to `unavailable_for_legal_reasons`, but that's awfully wordy. Figured I'd start simple.

jeremy: We don't have shorthand methods for every status code. Think this is one we'd omit.

colindean: Omitting is probably OK. I imagine use of it would be sufficiently rare that the code could be checked directly.

However, for completion, it might be a good idea to keep it. Maybe call it `legally_unavailable?` or `unavailable_legally?` or `unavailable_by_law?` or `censored?` or `censored_by_law?`

lib/rack/utils.rb
```diff
@@ -551,6 +551,7 @@ def names
       428 => 'Precondition Required',
       429 => 'Too Many Requests',
       431 => 'Request Header Fields Too Large',
+      451 => 'Unavailable For Legal Reasons',
```
colindean: Would it be wise to put a comment on here stating that the code isn't yet present on the IANA list as of the current date, in the unlikely event that someone tries to import from the CSV linked in the comment on [line 503](https://github.com/rack/rack/pull/985/files#diff-7c1a24d5b2fe58a6f925c7cacc6c55e7R503)?

lib/rack/response.rb
ktheory: @jeremy, @colindean: :ok_hand: I removed the method in 040de0d.

lib/rack/utils.rb
ktheory: > Would it be wise to put a comment on here

Seems like a stylistic thing. I'm fine either way, and will defer to @jeremy.

jeremy: Would be appropriate to mention the exception at https://github.com/rack/rack/pull/985#discussion-diff-48199164R501 since the comment says exactly how this list is generated.

lib/rack/response.rb
jeremy: :+1: Please rebase + squash to a single commit as well.

lib/rack/utils.rb
ktheory: @jeremy Ha! Was just writing the comment when I noticed that the [IANA CSV](https://www.iana.org/assignments/http-status-codes/http-status-codes-1.csv) *was updated* to include the new status code. So I updated per the comment on line 503. :tada:

ktheory:
All squashed & rebased & updated per the [IANA CSV of status codes](https://www.iana.org/assignments/http-status-codes/http-status-codes-1.csv). :relieved:

colindean:
:beers:

PR #986, opened 2015-12-21
Response#redirect? micro-optimization

ktheory:
This improves the performance of `Response#redirect?` by ~55% using a set instead of an
array.

See [this gist for benchmarks](https://gist.github.com/ktheory/e62bc43900745d54eb4f).

tl;dr: The new method can run a million redirect checks in 151ms, vs. 338ms for the original method (with a `301` status, the fastest case).

### DIFF
--- a/lib/rack/response.rb
+++ b/lib/rack/response.rb
@@ -24,6 +24,7 @@ module Rack
     alias headers header
 
     CHUNKED = 'chunked'.freeze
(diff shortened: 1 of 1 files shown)


ktheory:
I'm closing this to clean up old pull requests. I'm happy for someone else to advocate for this change; but it seems pretty fiddly & low-impact.

Issue #991, opened 2015-12-24
template in Rack::ShowExceptions works incorrectly when ParameterTypeError is raised. 


rrott:
According to the following doc http://www.rubydoc.info/github/rack/rack/master/Rack/ShowExceptions

> Rack::ShowExceptions catches all exceptions raised from the app it wraps. It shows a useful backtrace with the sourcefile and clickable context, the whole Rack environment and the request data.

This feature does not work correctly and renders exception message in plain text in case of **Rack::QueryParser::ParameterTypeError**.
See attached screenshots:
![screenshot 2015-12-24 18 25 38](https://cloud.githubusercontent.com/assets/746383/11996751/cf9b8e2a-aa72-11e5-8a75-bc0348759df2.png)

It expected to look like:
![screenshot 2015-12-24 18 25 53](https://cloud.githubusercontent.com/assets/746383/11996752/d1c58f0c-aa72-11e5-9011-cef047b6ae18.png)
it is also true for **Rack::QueryParser::PInvalidParameterError** and probably other exceptions.

Issue caused by **TEMPLATE** that tries to GET/POST incorrect url once again just to show "Request information". Because of this a new Rack::QueryParser::ParameterTypeError is raised and templete-less page is rendered.

Affected code is the following:
https://github.com/rack/rack/blob/master/lib/rack/show_exceptions.rb#L288-L319

> <% if req.GET and not req.GET.empty? %>
> 
> ...
> 
> <% req.GET.sort_by { |k, v| k.to_s }.each { |key, val| %>
> 
> ...
> 
> <% if req.POST and not req.POST.empty? %>
> 
> ...
> 
> <% req.POST.sort_by { |k, v| k.to_s }.each { |key, val| %> 

As you can see it does several extra GET and POST calls in this template.

jkowens:
I don't think this is actually an issue since Rack apps should rescue `Rack::Utils::ParameterTypeError` and `Rack::Utils::InvalidParameterError` and then return a 400 status.

jeremyevans:
I agree, this is a bug in Rack::ShowExceptions.  I'll submit a pull request to fix it.

PR #992, opened 2015-12-24
Fixed issue with sending extra POST and GET command from TEMPLATE

rrott:
changed TEMPLATE that tried to not to  GET/POST incorrect url once again just to show "Request information". More information: https://github.com/rack/rack/issues/991

### DIFF
--- a/lib/rack/show_exceptions.rb
+++ b/lib/rack/show_exceptions.rb
@@ -63,6 +63,11 @@ module Rack
     def pretty(env, exception)
       req = Rack::Request.new(env)
 
+      query_params = query_params = {
+        get:  pretty_params(env['QUERY_STRING']),
+        post: pretty_params(env['rack.request.form_vars'])
+      }
+
       # This double assignment is to prevent an "unused variable" warning on
       # Ruby 1.9.3.  Yes, it is dumb, but I don't like Ruby yelling at me.
       path = path = (req.script_name + req.path_info).squeeze("/")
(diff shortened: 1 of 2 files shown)


matthewd:
`req.env` contains something very different from `req.GET` or `req.POST`.

rrott:
My bad, I am sorry.

Now I see that GET and POST update and return RACK_REQUEST_QUERY_HASH and RACK_REQUEST_FORM_HASH respectively. But they have never been shown because of  parse_query that always failed on updating these values.
https://github.com/rack/rack/blob/master/lib/rack/request.rb#L336

I've fixed this to show appropriate data. Could you please have a look?

rrott:
Could you please review it once again?

rrott:
Can you please have a look? 

Note, Travis CI fails on "bundle install" stage, probably because of "This job ran on our new platform for Precise builds. Please read our blog post for more information."

jkowens:
Rack apps should rescue `ParameterTypeError` and `InvalidParameterError` when parsing query parameters and then return a 400 status, which would avoid the issue described.

rrott:
@jkowens Then what do you have [this huge template ](https://github.com/rrott/rack/blob/master/lib/rack/show_exceptions.rb#L110) for?
Should not it be deleted then?

As I see this template as a part of rack and it causes the issue described [here](https://github.com/rack/rack/issues/991)

jkowens:
That template is very useful for displaying traces for any exception other than `ParameterTypeError` and `InvalidParameterError`. Those are the two exceptions Rack expects to be handled, returning a http 400 status (instead of bubbling up to the Rack error handler or being displayed in the browser).

PR #995, opened 2016-01-01
Update license.

gfvcastro:


spastorino:
You're changing an extra line here for no reason

gfvcastro:
@spastorino: You are right. :grimacing:

PR #997, opened 2016-01-15
Fix querystring `foo[]bar=` to be flat key

### LINKED ISSUE #964: Parsing querystring with `foo[]bar` gives unexpected hash
Parsing a querystring of `?foo[]bar` gives a hash of `{"foo"=>[{"bar"=>nil}]}`, which is unexpected and undocumented.

In order to generate that hash from a querystring, I would expect to provide `?foo[][bar]`, which also works.

I'd expect `?foo[]bar` to return `{"foo[]bar" => nil}`. 

For reference, python's `urlparse`:

``` python
>>> parse_qs('foo[]bar', True)
{'foo[]bar': ['’]} 
```

Is this expected behaviour?

hughbien:
Fixes #964

Currently the key `foo[]bar` is parsed as `{"foo"=>[{"bar"=>nil}]}`. This changes it to `{"foo[]bar"=>nil}`. However, this might have been intended behavior (see `test/spec_utils.rb:211` which tests `x[y]z=2`)

I feel like there's a better way to do this than by adding a special case. Maybe by modifying the regex `name =~ %r(\A[\[\]]*([^\[\]]+)\]*)` used in `#normalize_params`?

Any feedback would be appreciated!

### DIFF
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -92,7 +92,7 @@ module Rack
         params[k] ||= []
         raise ParameterTypeError, "expected Array (got #{params[k].class.name}) for param `#{k}'" unless params[k].is_a?(Array)
         params[k] << v
-      elsif after =~ %r(^\[\]\[([^\[\]]+)\]$) || after =~ %r(^\[\](.+)$)
+      elsif after =~ %r(^\[\]\[([^\[\]]+)\]$) || after =~ %r(^\[\](\[.*\])$)
         child_key = $1
         params[k] ||= []
         raise ParameterTypeError, "expected Array (got #{params[k].class.name}) for param `#{k}'" unless params[k].is_a?(Array)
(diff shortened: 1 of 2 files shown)


filipegiusti:
Why do you think it should be a flat key instead of `{"foo"=>[{"bar"=>nil}]}`?

ioquatix:
I reviewed Ruby (`CGI.parse`), PHP (`parse_str`) and Python behaviour, and all of them do different things. I don't believe we should base our decision over some other implementation, and also changing this in a subtle way could break existing systems.

I don't see the value in changing the behaviour here, but I do see a cost, and there is also no standard/specification for what we should do, so unless we can find a standard (e.g. RFC) which tells us our behaviour is wrong, I'm not going to change this.

rafaelfranca:
I agree with that decision.

PR #1001, opened 2016-01-22
support custom trusted proxies

### LINKED ISSUE #735: Allow configuration of trusted proxies
I'm not sure if this has been discussed before, but it seems like it'd be beneficial to allow the configuration of the trusted proxies list used in https://github.com/rack/rack/blob/master/lib/rack/request.rb#L344 (trusted_proxy?).

Right now you have to monkey patch to change this. Rails relies on to get the IP for it's logs, so getting the correct remote IP to show up there requires monkeypatching Rack, which isn't ideal.

sbfaulkner:
here's a PR which closes #735

I propose making `Rack::Request#trusted_proxy?` support custom proxy addresses by allowing them to be specified by adding a `trusted_proxies` configuration option to the `Rack::Utils` module

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -200,6 +200,13 @@ Set to 0 for no limit.
 
 Can also be set via the RACK_MULTIPART_PART_LIMIT environment variable.
 
+=== trusted_proxies
+
+The trusted proxies to be used when determining the IP address of the remote client making a request.
+Client addresses which match the specified proxy addresses are discarded from the X-Forwarded-For addresses.
+
+The value provided may be a string, IPAddr or Regexp object, or an Array including any of those types of objects.
+
 == History
 
 See <https://github.com/rack/rack/blob/master/HISTORY.md>.
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -434,7 +434,7 @@ module Rack
       end
 
       def trusted_proxy?(ip)
-        ip =~ /\A127\.0\.0\.1\Z|\A(10|172\.(1[6-9]|2[0-9]|30|31)|192\.168)\.|\A::1\Z|\Afd[0-9a-f]{2}:.+|\Alocalhost\Z|\Aunix\Z|\Aunix:/i
+        Utils.trusted_proxy?(ip)
       end
 
       private
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -599,5 +599,16 @@ module Rack
     end
     module_function :clean_path_info
 
+    class << self
+      attr_accessor :trusted_proxies
+    end
+
+    # The default regex to match trusted proxy addresses.
+    self.trusted_proxies = /\A127\.0\.0\.1\Z|\A(10|172\.(1[6-9]|2[0-9]|30|31)|192\.168)\.|\A::1\Z|\Afd[0-9a-f]{2}:.+|\Alocalhost\Z|\Aunix\Z|\Aunix:/i
+
+    def self.trusted_proxy?(ip)
+      Array(trusted_proxies).any? { |proxy| proxy === ip }
+    end
+
   end
 end
--- a/test/spec_request.rb
+++ b/test/spec_request.rb
@@ -1283,26 +1283,38 @@ EOF
 
   it "regard local addresses as proxies" do
     req = make_request(Rack::MockRequest.env_for("/"))
-    req.trusted_proxy?('127.0.0.1').must_equal 0
-    req.trusted_proxy?('10.0.0.1').must_equal 0
-    req.trusted_proxy?('172.16.0.1').must_equal 0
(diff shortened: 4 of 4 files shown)


sbfaulkner:
it is unclear to me why the travis build is failing for `ruby-head`
https://travis-ci.org/rack/rack/jobs/104239924

sbfaulkner:
@raggi any thoughts / comments on this ? any chance we get this, or something similar, merged ?

tenderlove:
Being able to configure the trusted proxies seems good, but I'm not sure I like this implementation. Specifically I take issue with mutating a global (the attr accessor on the module).  Since the original issue was for Rails, can we make it configurable via the config object in Rails?  Rails instantiates a subclass of Rack::Request, so we can do whatever we want in that method.

raggi:
Why not put it in the env, like rack.trusted_proxies, then middleware in
rails could set that fetching the canonical value from config as you desire.

sbfaulkner:
@tenderlove 

(I'm totally up for changing how this is configured... I just tried to make it consistent with the other configurable Rack settings)

on the rails side I can partially address it by passing custom_proxies when I build the ActionDispatch::RemoteIp middleware, and using `request.remote_ip` to access the IP... unfortunately the ip address that is logged by rails is the Rack::Request#ip instead (and the ActionDispatch::RemoteIp middleware is later in the chain) so I don't have access to the correctly detected remote ip

although adding yet another middleware to work around this feels a bit unwieldy, how would you feel about an approach doing something like what @raggi is suggesting?

tenderlove:
> on the rails side I can partially address it by passing custom_proxies when I build the ActionDispatch::RemoteIp middleware, and using request.remote_ip to access the IP... unfortunately the ip address that is logged by rails is the Rack::Request#ip instead (and the ActionDispatch::RemoteIp middleware is later in the chain) so I don't have access to the correctly detected remote ip

TBH, it also bothers me that the custom proxies are passed to the middleware code at middleware stack build time (as opposed to runtime).  IMO this should be something the request object calculates rather than a middleware responsibility.

I'm loath to add another key to the env hash, but that's what we have to work with. What do you think about adding a new config option to the Rails config and putting that config object in the env hash?  Rails has a bunch of stuff it reads from the config object and shoves in the env hash today.  It seems like we could just put the config object in there and have other things read from it.

Adding a new key like `rack.trusted_proxies` means we'll have to add "yet another middleware" that sets the key to the custom value.  If we have one key with a config object, Rails can set that at runtime and we don't need more middleware.  It's a bit more work, but WDYT?

Also, if adding the key doesn't work out, I'll be happy to go with the "add a new class method" route.  This is a legit use case that we should address.

raggi:
So it is safe to add new objects to env, and it makes sense in the spec today.

I would drop a `rails.config` into the env hash for this. Similarly I'd memoize `rails.request` and the like that way, if that's still going on.

The global config stuff we have is unfortunate, although I'd say the keyspace stuff was important to have globally applied initially, and our `deployment` vs `production` thing has always meant that many people aren't running the rackup default middleware in all cases (for better and worse).

I could envisage a Rails::Rack middleware that does various things, mapping rails configurations into the env hash early in the stack.

There's nothing against putting "real objects" into the env hash. We've been doing this forever with `rack.input` and such things. It's actually a useful pattern.

Another idea pops to mind that you could have a more general thing like:
```
Rails.application.config.rack.env = { 'rack.trusted_proxies' => [...], 'rails.config' => Rails.application.config }
# then in the middleware
def call(env)
  @app.call(env.merge(Rails.application.config.rack.env))
end
```

The problem we've always had with `Rack::Request` is that it's a slow and somewhat opinionated implementation of a lot of things, and so anything like this that's better off as part of SPEC should really be encoded in `env` per the patterns of today.

sbfaulkner:
@tenderlove am I right that you're suggesting that I/we could "fix" the way that ActionDispatch does things too? one clarification needed... you questioned that the custom proxies were passed a build time, but I'm unclear why this is any different than having them in the Rails config - ie. they'd still be established at load time, rather than as part of the request

do we want to move the remote_ip support from ActionDispatch into rack proper - and simply set it up earlier in the rails middleware stack?

sbfaulkner:
@raggi @tenderlove riffing off of @raggi 's comment... The more general case of a standard Rack::Config middleware that populates env['rack.config'] with the parameters provided at build time appeals to me, but raises the question of build time vs request time, right?

raggi:
Rack shouldn't have globals. A configuration middleware doesn't need to be some kind of class indirection, it just needs to add stuff to env:

```ruby
use ->(env) { env['rack.config'] = (env['rack.config'] || {}).merge( .... )
```

sbfaulkner:
thanks gentlemen... I'll take a first stab at this and ping you when I've updated the PR... cheers

sbfaulkner:
closing for now, and will re-open with a different approach down the line

PR #1004, opened 2016-01-25
Use Mutex instead of Thread.exclusive for reloader

kou:
Because Thread.exclusive is deprecated since Ruby 2.3.0:

  * https://github.com/ruby/ruby/blob/v2_3_0/ChangeLog#L2398-L2400
  * https://github.com/ruby/ruby/blob/v2_3_0/prelude.rb#L11


manveru:
Thanks, that looks like a good change. the travis failure is unrelated so I'll merge this.

PR #1007, opened 2016-01-29
Reduce size of Rack png's logo

deepj:
Used ImageOptim for reducing size.


raggi:
So we reduce the size of the image, but increase the size of the repo by the new file size. What's the purpose of this change?

deepj:
If someone is going to use the Rack logo in PNG so they can use it in reduced version => less data needed to transfer through network or occupy on a disk (except the repository's things as mentioned correctly). If you see it it's pointless, the PR can be closed freely, of course.

PR #1009, opened 2016-01-30
Missing empty multipart params

WojtekKruszewski:
After upgrading to Rails 5.0.0.beta1.1 and Rack 2.0.0.alpha I'm missing empty parameters from params.

I added a failing assertion.

However, this assertion is failing for me also on Rack 1.6.4, so right now I'm totally confused.

### DIFF
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -100,8 +100,6 @@ module Rack
               # Generic multipart cases, not coming from a form
               data = {:type => content_type,
                       :name => name, :tempfile => body, :head => head}
-            elsif !filename && data.empty?
-              return
             end
 
             yield data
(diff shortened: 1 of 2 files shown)


WojtekKruszewski:
The assertion is in fact passing on 1.6.4. I was just running test on that release without bundle exec which made it run against 2.0.0.alpha – d'-oh!

WojtekKruszewski:
I traced it down to: https://github.com/rack/rack/commit/57e030f1db5832c059fde171dca95deec0e6b3af

WojtekKruszewski:
@sigmavirus24 & @burtlo can you take a look?

Could you provide some more details or tests for the problem you were fixing:

> Parsing "files" out of the header was converted to an empty string instead instead of nil when the body was empty.

sjke:
UP! Very important issue

WojtekKruszewski:
Thank you @tenderlove

PR #1011, opened 2016-02-03
Remove json gem

mperham:
The json gem has been included in Ruby stdlib since 1.9.  The gem itself causes difficulties building on Windows so depending on the gem is actually a net negative these days.

### DIFF
--- a/rack.gemspec
+++ b/rack.gemspec
@@ -28,8 +28,6 @@ EOF
   s.homepage        = 'http://rack.github.io/'
   s.required_ruby_version = '>= 2.2.2'
 
-  s.add_dependency 'json'
-
   s.add_development_dependency 'minitest', "~> 5.0"
   s.add_development_dependency 'minitest-sprint'
   s.add_development_dependency 'concurrent-ruby'


tenderlove:
But Rack actually depends on json.  It doesn't seem good to remove a dependency that is actually a dependency.  What, specifically, do we gain by removing it?

mperham:
@tenderlove Sure, and that's why you have to require it.  But it's not a gem dependency.  JSON has been in the stdlib for years and does not need to be declared in gemspec.

As for positives, it means you don't need a compiler and have one less native gem.  I've heard rumors of build difficulties with the gem on Windows that were unnecessary when ruby is distributed pre-built.

tenderlove:
It's a gem, and we depend on it.  I'm really confused.

matthewd:
If it's in stdlib as a gem, rubygems/bundler should be happy to use the existing one without installing a new one, even with the dependency declared... shouldn't it?

tenderlove:
If you just declare the gem as a dependency with no version requirement, then the one that ships with Ruby should be Good Enough™.  I don't see why "difficulties building on Windows" would come in to play when the JSON gem provided with the distribution fulfills an open-ended dependency.

matthewd:
Ah, right... bundler will seek out the latest version available, so as long as that's newer than the pre-installed one, you'll do a build.

tenderlove:
Just because gems are shipped with Ruby doesn't mean that we shouldn't declare them as dependencies.  They are gems, and they are dependencies, so they are gem dependencies, so I do want to declare them.

OTOH, I didn't expect Bundler to install the latest one even though the one on the system would work just fine.  Given that fact, I'm happy to drop the dep.  Is this something we should bother team Bundler with?

raggi:
You know, that kloc we had before was faster and less problematic :(

PR #1014, opened 2016-02-12
Use the first IP address from X-Forwarded-For header for Rack::Request.ip

samlevy:
`Rack::Request.ip` was returning our load balancing server IP address. After reading through the Forwarded For RFC documentation, it looks like the client IP address will be the first in the list.

"In a chain of proxy servers where this is fully utilized, The first 'for'
parameter will disclose the client where the request was first made, followed
by any subsequent proxy identifiers."

https://tools.ietf.org/html/rfc7239#section-5.2

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -278,7 +278,7 @@ module Rack
 
         forwarded_ips = split_ip_addresses(get_header('HTTP_X_FORWARDED_FOR'))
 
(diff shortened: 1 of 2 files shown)


evanleck:
I just ran into this exact problem. I'd love to see this PR merged, but it looks like it's been languishing for a while.

jeremy:
It's selecting the outermost IP that we can trust to not be spoofed.

See https://github.com/rack/rack/issues/1055#issuecomment-212189520 for history.

evanleck:
Right on, thanks for the context!

PR #1016, opened 2016-02-19
Support HTTP Forwarded proto header from RFC 7239

mattbostock:
[RFC 7239][] is an IETF proposed standard for a HTTP `Forwarded` request
header that aims to replace non-standard `X-Forwarded-*` headers.

This commit allows the URI scheme to be detected based on the value of
the `proto` token in that header.

I have prioritised the `Forwarded` header over the non-standard headers
as I think it's preferable to use the standard header when specified.

Multiple values for `proto` are allowed by the standard; the first one
is preferred here as it's the most likely to indicate the protocol used
by the original client.

I've been careful to support mixed case in the header and optional
double quotes for the `proto` value as specified in the RFC.

[RFC 7239]: https://tools.ietf.org/html/rfc7239#section-4

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -134,6 +134,7 @@ module Rack
       # to include the port in a generated URI.
       DEFAULT_PORTS = { 'http' => 80, 'https' => 443, 'coffee' => 80 }
 
+      HTTP_FORWARDED          = 'HTTP_FORWARDED'.freeze
       HTTP_X_FORWARDED_SCHEME = 'HTTP_X_FORWARDED_SCHEME'.freeze
       HTTP_X_FORWARDED_PROTO  = 'HTTP_X_FORWARDED_PROTO'.freeze
       HTTP_X_FORWARDED_HOST   = 'HTTP_X_FORWARDED_HOST'.freeze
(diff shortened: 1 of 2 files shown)


boffbowsh:
This is good, but I think if Rack introduces detection of SSL from this header, it should also give precedence to `for=` values here in place of `X-Forwarded-For` in https://github.com/rack/rack/blob/146c754df01ae768119f2066421b10efd0f0db4f/lib/rack/request.rb#L273. It would be unpredictable behaviour for Rack to support only some parts of this header and not others.

mattbostock:
@boffbowsh That's a good point, thanks @boffbowsh - I'll look at that.

mattbostock:
I've spotted a couple of small improvements I can make - I'll close this will I work on those.

mattbostock:
I've added some tests to `spec_common_logger.rb` and I'm having difficulty getting the logger to recognise the new header.

At the top of that file, I have:

```ruby
app_with_forwarded_for = Rack::Lint.new lambda { |env|
  [200, {"Forwarded" => "for=203.0.113.0"}, []]}
```

In the test, I have:

```ruby
 it "log - RFC 7239 forwarded for header" do
   res = Rack::MockRequest.new(Rack::CommonLogger.new(app_with_forwarded_for)).get("/")

   res.errors.wont_be :empty?
   res.errors.must_match(/203\.0\.113\.0 - /)
end
```

How do I set the new header when mocking a request? I'd be grateful for any help.

I've pushed my code to a temporary branch:
https://github.com/rack/rack/compare/master...alphagov:wip/support_forwarded_proto?expand=1

boffbowsh:
Your `app_with_forwarded_for` is specifying `for=203.0.113.0` as a response header rather than a request header. You want to specify the header as an argument to `get`, similar to https://github.com/rack/rack/compare/master...alphagov:wip/support_forwarded_proto?expand=1#diff-e6241b1590508229ab2a88b63a55c686R1238

mattbostock:
Thanks @boffbowsh, that did it.

I've opened a new PR supporting all relevant tokens from RFC 7239 in https://github.com/rack/rack/pull/1017

PR #1017, opened 2016-02-21
Support RFC 7239: HTTP Forwarded header

mattbostock:
[RFC 7239](https://tools.ietf.org/html/rfc7239) is an IETF proposed standard for a HTTP `Forwarded` request
header that aims to replace non-standard `X-Forwarded-*` headers.

This commit supports all relevant tokens that may be set in the
Forwarded header:
- proto
- host
- for

I have prioritised the Forwarded header over the non-standard headers
as I think it's preferable to use the standard header when specified.

Multiple values for each token are allowed by the standard. When
multiple values are supplied, I've followed these rules:
- for the `proto` token, use the first value specified as it's most
  likely to be the protocol used by the original client
- for the `host` token, use the last value specified following the
  convention already used by Rack for multiple `X-Forwarded-For`
  headers

When both `X-Forwarded-For` and `Forwarded` headers are supplied and the
latter includes `for` values, I've concatenated both sets of values with
the `Forwarded` values coming last.

I've been careful to support mixed case in the header and optional
double quotes encapsulating the token value as specified in the RFC.

### DIFF
--- a/lib/rack/common_logger.rb
+++ b/lib/rack/common_logger.rb
@@ -41,9 +41,10 @@ module Rack
     def log(env, status, header, began_at)
       now = Time.now
       length = extract_content_length(header)
+      forwarded_for = Utils::forwarded_values(env['HTTP_FORWARDED'])[:for].last
 
       msg = FORMAT % [
-        env['HTTP_X_FORWARDED_FOR'] || env["REMOTE_ADDR"] || "-",
+        forwarded_for || env['HTTP_X_FORWARDED_FOR'] || env["REMOTE_ADDR"] || "-",
         env["REMOTE_USER"] || "-",
         now.strftime("%d/%b/%Y:%H:%M:%S %z"),
         env[REQUEST_METHOD],
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -134,6 +134,7 @@ module Rack
       # to include the port in a generated URI.
       DEFAULT_PORTS = { 'http' => 80, 'https' => 443, 'coffee' => 80 }
 
+      HTTP_FORWARDED          = 'HTTP_FORWARDED'.freeze
       HTTP_X_FORWARDED_SCHEME = 'HTTP_X_FORWARDED_SCHEME'.freeze
       HTTP_X_FORWARDED_PROTO  = 'HTTP_X_FORWARDED_PROTO'.freeze
       HTTP_X_FORWARDED_HOST   = 'HTTP_X_FORWARDED_HOST'.freeze
@@ -203,6 +204,10 @@ module Rack
       def scheme
         if get_header(HTTPS) == 'on'
           'https'
+        elsif get_header(HTTP_FORWARDED)
+          # Use first proto field set as it's most likely to be represent the
+          # original client: https://tools.ietf.org/html/rfc7239#section-4
+          get_http_forwarder(:proto)[0]
         elsif get_header(HTTP_X_FORWARDED_SSL) == 'on'
           'https'
         elsif get_header(HTTP_X_FORWARDED_SCHEME)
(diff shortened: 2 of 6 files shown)


mattbostock:
Is this a useful addition? I'd be really grateful for any feedback. :sunglasses:

lib/rack/common_logger.rb
```diff
@@ -41,9 +41,10 @@ def call(env)
     def log(env, status, header, began_at)
       now = Time.now
       length = extract_content_length(header)
+      forwarded_for = Utils::forwarded_values(env['HTTP_FORWARDED'])[:for].last
 
       msg = FORMAT % [
-        env['HTTP_X_FORWARDED_FOR'] || env["REMOTE_ADDR"] || "-",
+        forwarded_for || env['HTTP_X_FORWARDED_FOR'] || env["REMOTE_ADDR"] || "-",
```
jeremy: Seems this is kinda-implementing `Rack::Request#ip`, maybe nicer to reuse that.

lib/rack/common_logger.rb
```diff
@@ -41,9 +41,10 @@ def call(env)
     def log(env, status, header, began_at)
       now = Time.now
       length = extract_content_length(header)
+      forwarded_for = Utils::forwarded_values(env['HTTP_FORWARDED'])[:for].last
```
jeremy: This codebase uses `.` for method calls and `::` for constant scope resolution.

lib/rack/request.rb
```diff
@@ -240,7 +245,9 @@ def xhr?
       end
 
       def host_with_port
-        if forwarded = get_header(HTTP_X_FORWARDED_HOST)
+        if forwarded = get_http_forwarder(:host).last
+          forwarded
+        elsif forwarded = get_header(HTTP_X_FORWARDED_HOST)
           forwarded.split(/,\s?/).last
```
jeremy: Curious that we use the _first_ forward for protocol, the _last_ forward for host, and the _first trusted_ forward for IP.

test/spec_request.rb
```diff
@@ -1188,6 +1217,21 @@ def ip_app
...
+      'HTTP_FORWARDED' => 'for=5.6.7.8'
+    res.body.must_equal '5.6.7.8'
+
+    res = mock.get '/',
+      'HTTP_X_FORWARDED_FOR' => '3.4.5.6',
+      'HTTP_FORWARDED' => 'for=5.6.7.8, for=7.8.9.0'
+    res.body.must_equal '7.8.9.0'
+
+    res = mock.get '/',
+      'REMOTE_ADDR' => '1.2.3.4',
```
jeremy: Test coverage demonstrating that trusted proxy detection behaves with Forwarded, too?

test/spec_utils.rb
```diff
@@ -354,6 +354,43 @@ def initialize(*)
...
+      :for   => [ '3.4.5.6' ],
+      :proto => [ 'http', 'https' ]
+    })
+
+    Rack::Utils.forwarded_values('for=3.4.5.6; proto=http, proto=https; for=1.2.3.4').must_equal({
+      :for   => [ '3.4.5.6', '1.2.3.4' ],
+      :proto => [ 'http', 'https' ]
+    })
+  end
+
```
jeremy: Looks like we're covering https://tools.ietf.org/html/rfc7239#section-7.1 👍 

Can't do much about multiple header testing since that falls on the Rack handler to coalesce.

raggi:
I think this will need to be made configurable now.

The problem here is that proxies that have yet to be updated will not strip these headers, and also will not know to overwrite them.

A failure to protect against outside spoofing can have important security implications. Users must manage the transition with care.

mattbostock:
Sorry for the lack of activity and thanks for the feedback. Hopefully I'll get time to work some more on this in the next couple of weeks.

@raggi: Would you suggest disabling this functionality by default?

raggi:
I would not recommend making major API changes without changing major versions, ideally. This situation is unfortunate and tricky, but I don't think we can safely just break this.

ioquatix:
What's the status of this?

ioquatix:
Can we rebase this and land it in Rack 3.x?

fatkodima:
@ioquatix I will move this forward if @mattbostock don't want (have time) to.
How we should address @raggi 's latest two comments?

ioquatix:
I haven't reviewed this PR for a while, but feel free to rebase it and we can go from there.

jeremyevans:
It seems more likely we will use the updated version of this (#1423) than this, so I'm going to close this now.

Issue #1013, opened 2016-02-11
rackup -O SSLEnable doesn't work for WEBrick


dgbustos:
The WEBrick documentation indicates that enabling SSL should be done with `require 'webrick/https'` and passing `SSLEnable => true` as an option.  `rackup -O SSLEnable` does the second, but not the first.  I think this can be fixed by adding

``` ruby
    if options[:SSLEnable]
      require 'webrick/https'
    end
```

to the `run` method in `Rack::Handler::WEBrick`.

ioquatix:
Fixed.

PR #1018, opened 2016-02-22
fix SSLEnable for Webrick, see https://github.com/rack/rack/issues/1013

gregormelhorn:
see https://github.com/rack/rack/issues/1013


ioquatix:
Thanks for your contribution.

PR #1023, opened 2016-03-01
Changes for frozen string literal compatibility

twalpole:
Changes to allow for use with frozen string literals.


ioquatix:
This has been implemented already. Thanks for your contribution.

PR #1026, opened 2016-03-02
use secure_compare in auth examples

geemus:
I was setting up something for a prototype and found myself following the examples without too much thought, and then stumbled across the secure method and figured better safe than sorry (especially since it is so easy to forget stuff like this).


geemus:
@raggi thanks!

PR #1027, opened 2016-03-02
Puma should be the default webserver for rackup

raggi:
* Thin is based on fully unmaintained code (EM)
 * Puma is significantly faster, on a tuned 16 core VM the difference is 50kqps.


chneukirchen:
+1

chneukirchen:
+1

Issue #1015, opened 2016-02-13
Missing parameters with parse_nested_query


rthbound:
In controller tests, I find that collection objects sent via POST do not make it through to the controller. After stepping through the stack, I've finally identified the source of the information loss. The information is lost here, in [Rack::QueryParser#parse_nested_query](https://github.com/rack/rack/blob/2ac599551d6147184321c42f40a0ab3241499b08/lib/rack/query_parser.rb#L64-L68).  The data resulting from `qs.split(...)` is in a [gist here](https://gist.github.com/rthbound/806a5d2c350ee2ba444f).

I originally opened this issue on the rails project, suspecting rails 5 beta was at fault (see https://github.com/rails/rails/issues/23624).

#951 looks highly similar.

Issue #951, opened 2015-09-21
`parse_nested_query` fails when first member of Array is a Hash


eins78:
form values:

```rb
many_foos[][user][id]=1234&
many_foos[][bar]=baz&
many_foos[][user][id]=5678&
many_foos[][bar]=blub
```

test: 

```rb
Rack::Utils.parse_nested_query("many_foos%5B%5D%5Buser%5D%5Bid%5D=1234&many_foos%5B%5D%5Bbar%5D=baz&many_foos%5B%5D%5Buser%5D%5Bid%5D=5678&many_foos%5B%5D%5Bbar%5D=blub")
=> {"many_foos"=>[{"user"=>{"id"=>"5678"}, "bar"=>"baz"}, {"bar"=>"blub"}]}
```

- result:

    ```rb
    {"many_foos"=>[{"user"=>{"id"=>"5678"}, "bar"=>"baz"}, {"bar"=>"blub"}]}
    ```

- expected:

    ```rb
    {"many_foos"=>[{"bar"=>"baz", "user"=>{"id"=>"1234"}}, {"bar"=>"blub", "user"=>{"id"=>"5678"}}]}
    ```

---

Can be fixed application-side by re-ordering the form fields,
so that no Array starts with a Hash:

form values:

```
many_foos[][bar]=baz&
many_foos[][user][id]=1234&
many_foos[][bar]=blub&
many_foos[][user][id]=5678
```

test:

```rb
Rack::Utils.parse_nested_query("many_foos%5B%5D%5Bbar%5D=baz&many_foos%5B%5D%5Buser%5D%5Bid%5D=1234&many_foos%5B%5D%5Bbar%5D=blub&many_foos%5B%5D%5Buser%5D%5Bid%5D=5678")
```

- result:

    ```rb
    {"many_foos"=>[{"user"=>{"id"=>"5678"}, "bar"=>"baz"}, {"bar"=>"blub"}]}
    ```

prrrnd:
Also having this issue with the following:
```
params = {"form"=>{"things"=>[{"user"=>{"name"=>"a"}, "amount"=>"1"}, {"user"=>{"name"=>"b"}, "amount"=>"2"}]}} 
Rack::Utils.parse_nested_query(Rack::Utils.build_nested_query(params))
# => {"form"=>{"things"=>[{"user"=>{"name"=>"b"}, "amount"=>"1"}, {"amount"=>"2"}]}} 
```

Any updates on this @eins78 ?

[...]

PR #1029, opened 2016-03-05
Fixes 951 fixes 1015

rthbound:
These fixes would close quite a few issues. #951 and #1015 in rack, as well as several in rails.


rthbound:
Fixes https://github.com/rails/rails/issues/23997 && https://github.com/rails/rails/issues/23624

sikachu:
Looks like the test failed on JRuby, but it's not related to the change.

@tenderlove would you mind taking a look at this PR? Maybe restarting the build too?

PR #1031, opened 2016-03-09
Backport rack lock fix

aharpervc:
Brings https://github.com/rack/rack/pull/828 to 1-6-stable. I just cherry-picked the commits. Not sure if there's anything else to do.

PR #1032, opened 2016-03-09
Backport rack lock fix to 1.5

aharpervc:
Same as https://github.com/rack/rack/pull/1031 but for 1.5


ioquatix:
Thanks for your contribution. I'm sorry we did not attend to it in time. We are no longer supporting 1.5 branch.

PR #1033, opened 2016-03-15
Update first-party-only cookie syntax

mastahyeti:
The spec for first-party-only cookies has been updated. The new attribute name is `SameSite` and a few different values are now supported. The new syntax will [supposedly](https://twitter.com/mikewest/status/709676633712676864) be landing in Chrome 51.

[Current spec](https://tools.ietf.org/html/draft-west-first-party-cookies-06)
/cc @mikewest

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -248,13 +248,20 @@ module Rack
           rfc2822(value[:expires].clone.gmtime) if value[:expires]
         secure = "; secure"  if value[:secure]
         httponly = "; HttpOnly" if (value.key?(:httponly) ? value[:httponly] : value[:http_only])
-        first_party = "; First-Party" if value[:first_party]
+        same_site = if value[:same_site]
+          case value[:same_site]
+          when Symbol, String
+            "; SameSite=#{value[:same_site]}"
+          else
+            "; SameSite"
+          end
+        end
         value = value[:value]
       end
       value = [value] unless Array === value
 
       cookie = "#{escape(key)}=#{value.map { |v| escape v }.join('&')}#{domain}" \
-        "#{path}#{max_age}#{expires}#{secure}#{httponly}#{first_party}"
+        "#{path}#{max_age}#{expires}#{secure}#{httponly}#{same_site}"
 
       case header
       when nil, ''
--- a/test/spec_response.rb
+++ b/test/spec_response.rb
@@ -115,16 +115,28 @@ describe Rack::Response do
     response["Set-Cookie"].must_equal "foo=bar"
   end
 
-  it "can set First-Party cookies" do
+  it "can set SameSite cookies with any truthy value" do
     response = Rack::Response.new
-    response.set_cookie "foo", {:value => "bar", :first_party => true}
-    response["Set-Cookie"].must_equal "foo=bar; First-Party"
+    response.set_cookie "foo", {:value => "bar", :same_site => Object.new}
+    response["Set-Cookie"].must_equal "foo=bar; SameSite"
+  end
+
+  it "can set SameSite cookies with string value" do
+    response = Rack::Response.new
+    response.set_cookie "foo", {:value => "bar", :same_site => "Lax"}
+    response["Set-Cookie"].must_equal "foo=bar; SameSite=Lax"
+  end
+
+  it "can set SameSite cookies with symbol value" do
+    response = Rack::Response.new
+    response.set_cookie "foo", {:value => "bar", :same_site => :Strict}
+    response["Set-Cookie"].must_equal "foo=bar; SameSite=Strict"
   end
 
   [ nil, false ].each do |non_truthy|
-    it "omits First-Party attribute given a #{non_truthy.inspect} value" do
+    it "omits SameSite attribute given a #{non_truthy.inspect} value" do
       response = Rack::Response.new
-      response.set_cookie "foo", {:value => "bar", :first_party => non_truthy}
+      response.set_cookie "foo", {:value => "bar", :same_site => non_truthy}
       response["Set-Cookie"].must_equal "foo=bar"
     end
   end


lib/rack/utils.rb
```diff
@@ -248,13 +248,20 @@ def add_cookie_to_header(header, key, value)
           rfc2822(value[:expires].clone.gmtime) if value[:expires]
         secure = "; secure"  if value[:secure]
         httponly = "; HttpOnly" if (value.key?(:httponly) ? value[:httponly] : value[:http_only])
-        first_party = "; First-Party" if value[:first_party]
+        same_site = case (value[:first_party] || value[:same_site]).to_s.downcase
```
mastahyeti: Not sure if it's worth continuing to support the old `:first_party ` option.

jeremy: Let's drop it. It's not in a released Rack yet.

lib/rack/utils.rb
```diff
@@ -248,13 +248,20 @@ def add_cookie_to_header(header, key, value)
...
+        when "lax"
+          "; SameSite=Lax"
+        when "strict"
+          "; SameSite=Strict"
+        when "true"
+          "; SameSite"
```
jeremy: Should we raise ArgumentError in other cases?

jeremy:
Dig the Lax/Strict distinction. Thanks for updating, @mastahyeti!

lib/rack/utils.rb
mastahyeti: There were tests for "falsey" values that would break. I think it's simpler to just ignore any values other than `true` or case-insensitive "lax", or "strict."

jeremy: Ah yes. It came to mind as I read the spec:

> If the "SameSite" attribute has no value, if the value is "Strict", *or if the value is invalid*

Which struck me as a bit puzzling. Whether that's meant to cover implementor mistakes by failing back to a strict default or whether it's meant to act as forward compatibility with new SameSite values.

We could be better off passing through the provided value verbatim. The spec is clear that attribute name and value matching is case insensitive, so passing any variety of lax/strict value would Just Work, and what we see as invalid today may be valid tomorrow, in which case we wouldn't want to be dropping the value entirely, losing strictness.

mastahyeti: Makes sense. I'll change it.

mastahyeti:
I updated it to pass along any String/Symbol values.

lib/rack/utils.rb
```diff
@@ -248,13 +248,18 @@ def add_cookie_to_header(header, key, value)
           rfc2822(value[:expires].clone.gmtime) if value[:expires]
         secure = "; secure"  if value[:secure]
         httponly = "; HttpOnly" if (value.key?(:httponly) ? value[:httponly] : value[:http_only])
-        first_party = "; First-Party" if value[:first_party]
+        same_site = case value[:same_site]
+        when true
+          "; SameSite"
+        when Symbol, String
+          "; SameSite=#{value[:same_site]}"
```
jeremy: Similar strictness-loss issue if we whitelist Symbol/String values. Say someone does `same_site: heuristic_to_decide_same_site?` which has a truthy response. Then they'd lose strictness and never know it's because they didn't provide an allowed value. We could case out non-truthy values to ignore them `else` to the `true` case.

test/spec_response.rb
```diff
@@ -115,16 +115,28 @@
...
+  it "can set SameSite=Lax cookies" do
+    response = Rack::Response.new
+    response.set_cookie "foo", {:value => "bar", :same_site => "Lax"}
+    response["Set-Cookie"].must_equal "foo=bar; SameSite=Lax"
+  end
+
+  it "can set SameSite=Strict cookies" do
+    response = Rack::Response.new
+    response.set_cookie "foo", {:value => "bar", :same_site => "Strict"}
+    response["Set-Cookie"].must_equal "foo=bar; SameSite=Strict"
```
jeremy: These tests imply that it's just these two values that are allowed. Can we write the test in a way that suggests that we don't care about the specific value?

lib/rack/utils.rb
mastahyeti: Hmmm. How about:

```ruby
same_site = if value[:same_site] # anything truthy
  case value[:same_site]
  when Symbol, String
    "; SameSite=#{value[:same_site]}"
  else
    "; SameSite"
  end
end

Comment (jeremy) [inline lib/rack/utils.rb]: Makes sense!

**Comment by mastahyeti** [comment]
I updated the logic again and tested the various possible values.

**Comment by jeremy** [comment]
:+1: Squash commits and :shipit: 

/cc @spastorino @tenderlove

File: test/spec_response.rb [hunk 5]
```diff
@@ -115,16 +115,28 @@
...
+  it "can set SameSite cookies with string value" do
+    response = Rack::Response.new
+    response.set_cookie "foo", {:value => "bar", :same_site => "Lax"}
+    response["Set-Cookie"].must_equal "foo=bar; SameSite=Lax"
+  end
+
+  it "can set SameSite cookies with symbol value" do
+    response = Rack::Response.new
+    response.set_cookie "foo", {:value => "bar", :same_site => :strict}
+    response["Set-Cookie"].must_equal "foo=bar; SameSite=strict"
```
mikewest: Nit: s/=strict/=Strict/

mastahyeti:
The commits are squashed.

tenderlove:
@mastahyeti thanks!

mastahyeti:
Any chance you folks would accept a backport to the 1.6 branch? That's what GitHub is running and it would be nice to have this without monkey patching rack...

tenderlove:
@mastahyeti ya, I'll backport and release. :)

arthurnn:
❤️

tenderlove:
@mastahyeti actually, do you mind backporting and sending a PR? I'm getting conflicts when I try to backport.

mastahyeti:
yeah, no problem.

PR #1034, opened 2016-03-16
When a symbol is passed as a status code it must match a valid http status code.

adambutler:
It is unexpected to get a `500` status code back with no exception raised and no obvious deliberate raising of a `500`.

For instance if one was to provide `:success` as a status symbol by accident instead of `:ok` then a 500 would be raised with no obvious reason.

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -574,7 +574,8 @@ module Rack
 
     def status_code(status)
       if status.is_a?(Symbol)
-        SYMBOL_TO_STATUS_CODE[status] || 500
+        raise ArgumentError, "Unrecognized status_code symbol" unless SYMBOL_TO_STATUS_CODE[status]
+        SYMBOL_TO_STATUS_CODE[status]
       else
         status.to_i
       end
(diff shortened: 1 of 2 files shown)


lib/rack/utils.rb
```diff
@@ -574,7 +574,8 @@ def names
 
     def status_code(status)
       if status.is_a?(Symbol)
-        SYMBOL_TO_STATUS_CODE[status] || 500
+        raise ArgumentError, "Unrecognized status_code symbol" unless SYMBOL_TO_STATUS_CODE[status]
```
gioele: Wouldn't it be better to include the symbol itself?

    "Unrecognized status_code symbol #{status.inspect}"

nateberkopec:
[HTTP status codes are extensible.](https://www.w3.org/Protocols/rfc2616/rfc2616-sec6.html#sec6.1.1)

lib/rack/utils.rb
rafaelfranca: :+1:

rafaelfranca: You could do:

```ruby
SYMBOL_TO_STATUS_CODE.fetch(status) { raise ArgumentError, "Unrecognized status code #{status.inspect}" }
```

oniofchaos:
This seems pretty useful and would help out all over the place. Fixes in other places such as Rails are being held up for a fix here. The fail on CI looks to be weird and unrelated - maybe if latest master was merged in the error would be cleared up?

PR #1035, opened 2016-03-17
Simple refactor Basic::Auth class

davydovanton:


davydovanton:
I have a bundler problem with `ruby-head` and `jruby-head` travis builds :worried:

PR #1038, opened 2016-03-23
strip key name when pares query

wavezhang:
strip key name when pares query


ioquatix:
I'm not sure this behaviour is desirable. I'm going to close this issue, however if you want to address it:

- What situations does this happen in?
- Can you make some related specs.

PR #1042, opened 2016-04-05
Compare host and server name only once per call

gioele:
The host name and the server name are not changed inside `#call(env)`,
so there is no need to compare them every time a mapping is tested.


arthurnn:
👍

PR #1043, opened 2016-04-05
Travis: Allow ruby-head and jruby-head builds to fails

gioele:
Development versions of the various Ruby interpreters may have problems that lead the builds to fail, even though the commits are fine and work without problems with the released versions of the interpreters.

For example, these recent PRs failed because of a temporary misconfiguration of the Travis VMs, unrelated to the PRs: <https://travis-ci.org/rack/rack/jobs/120208553>, <https://travis-ci.org/rack/rack/jobs/120842905>. Other builds failed because of a problem with the compilation of a dependency: <https://travis-ci.org/rack/rack/jobs/120208550>.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,4 +19,6 @@ notifications:
 
 matrix:
   allow_failures:
+    - rvm: ruby-head
+    - rvm: jruby-head
     - rvm: rbx-2


arthurnn:
make sense to me 👍

jeremy:
Allowing jruby-head failures in https://github.com/rack/rack/commit/0c748485a93fcf806741e8afd4dcc10603bbcfcb

Allowing ruby-head unless we get chronic failures. Seeming pretty stable.

PR #1044, opened 2016-04-05
leave modification of rack.multithread from rack::lock

isaiah:
Please correct me if I am wrong, because in the web server's point of view, the modification should at least have the same life span as the mutex, and it's ok to leave it modified until the end of the request.

The situation I am running into is that after `app.call(env)`, the response body is not closed yet, and the mutex is still hold. if the response is handled in a separate thread, if triggers`ThreadError`.

### DIFF
--- a/lib/rack/lock.rb
+++ b/lib/rack/lock.rb
@@ -12,7 +12,7 @@ module Rack
     def call(env)
       @mutex.lock
       begin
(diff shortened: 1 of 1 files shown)


ioquatix:
`Rack::Lock` is a disaster waiting to happen. While I understand the motivation for your change, this is not going to be realistically thread safe.

PR #1045, opened 2016-04-06
Use String interpolation or << instead plus which are faster

shhavel:


lib/rack/auth/digest/params.rb
```diff
@@ -17,7 +17,7 @@ def self.dequote(str) # From WEBrick::HTTPUtils
         end
 
         def self.split_header_value(str)
-          str.scan( /(\w+\=(?:"[^\"]+"|[^,]+))/n ).collect{ |v| v[0] }
+          str.scan(/\w+\=(?:"[^\"]+"|[^,]+)/)
```
gioele: This change looks unrelated to this PR. It also changes the way headers are parsed.

Wrong chunk selected during `git add -p`? :)

shhavel: Ok, sorry. will remove it from PR.

PR #1047, opened 2016-04-08
Backport 1.6 nested query fix to 1.5 stable

natesholland:
This issue was causing us issues in production so I wanted to backport it
to 1.5. Previously if you passed a fixnum into build_nested_query it would
not be put into the query string. This backports a fix to make fixnums show
up in nested queries.


ioquatix:
Thanks for your contribution. We did not attend to your request in time. We are no longer supporting 1.5 branch.

PR #1050, opened 2016-04-16
allow configuration of digest_class in etag.rb.

### LINKED ISSUE #1048: Hard coded MD5 digest breaks rack in FIPS mode
The code below, from etag.rb ( https://github.com/rack/rack/blob/master/lib/rack/etag.rb)  hard codes MD5.  This causes rack to break on any system running in FIPS mode

```
 def digest_body(body)
        parts = []
        digest = nil

        body.each do |part|
          parts << part
          (digest ||= Digest::MD5.new) << part unless part.empty?
        end

        [digest && digest.hexdigest, parts]
      end`
```

woot4moo:
attempts to fix #1048 by enabling configuration of the digest_class parameter.  Parameter was selected to be a default to MD5 so that clients do not have to make any changes in the base case.

### DIFF
--- a/lib/rack/etag.rb
+++ b/lib/rack/etag.rb
@@ -8,17 +8,20 @@ module Rack
   # a sendfile body (body.responds_to :to_path) is given (since such cases
   # should be handled by apache/nginx).
   #
-  # On initialization, you can pass two parameters: a Cache-Control directive
-  # used when Etag is absent and a directive when it is present. The first
-  # defaults to nil, while the second defaults to "max-age=0, private, must-revalidate"
+  # On initialization, you can pass three parameters: a Cache-Control directive
+  # used when Etag is absent, a directive when it is present, and the digest class to be used. The first
+  # defaults to nil, while the second defaults to "max-age=0, private, must-revalidate", and finally the
+  # third defaults to Digest::MD5
   class ETag
+    attr_accessor :digest_class
+
     ETAG_STRING = Rack::ETAG
     DEFAULT_CACHE_CONTROL = "max-age=0, private, must-revalidate".freeze
-
-    def initialize(app, no_cache_control = nil, cache_control = DEFAULT_CACHE_CONTROL)
+    def initialize(app, no_cache_control = nil, cache_control = DEFAULT_CACHE_CONTROL, digest_class = Digest::MD5)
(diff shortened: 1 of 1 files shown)


lib/rack/etag.rb
```diff
@@ -8,17 +8,20 @@ module Rack
...
+  # defaults to nil, while the second defaults to "max-age=0, private, must-revalidate", and finally the
+  # third defaults to Digest::MD5
   class ETag
+    attr_accessor :digest_class
+
     ETAG_STRING = Rack::ETAG
     DEFAULT_CACHE_CONTROL = "max-age=0, private, must-revalidate".freeze
-
-    def initialize(app, no_cache_control = nil, cache_control = DEFAULT_CACHE_CONTROL)
+    def initialize(app, no_cache_control = nil, cache_control = DEFAULT_CACHE_CONTROL, digest_class = Digest::MD5)
```
filipegiusti: There is a lot of parameters, wouldn't it be better to be

```ruby
def initialize(app, no_cache_control = nil, cache_control = DEFAULT_CACHE_CONTROL, digest_class: Digest::MD5)
  # ...
end
```

woot4moo: isn't that what I have? I am on mobile at the moment so formatting may be weird

lib/rack/etag.rb
```diff
@@ -8,17 +8,20 @@ module Rack
...
       @app = app
       @cache_control = cache_control
       @no_cache_control = no_cache_control
+      @digest_class = digest_class.new
```
mikegee: I feel like the "_class" suffix is inaccurate for this instance variable. Maybe we can call it something like "digester"?

ioquatix:
Can you please rebase this on master and use an option as suggested. Then, we can merge! Thanks.

jeremyevans:
MD5 was switched to SHA256 in 7b66d2cdb80a4d6b44fa8c61d92e25fbbda1f152. I don't see a need to make the digest configurable, so I'm going to close this.

ioquatix:
BLAKE2 is supposed to be better, but we couldn't even benchmark it if it's not configurable. Maybe worth considering. In the next release of OpenSSL we will support it.

PR #1051, opened 2016-04-16
Validate the SameSite cookie option

bobjflong:
The [draft spec](https://tools.ietf.org/html/draft-west-first-party-cookies-07) for the SameSite option mentions two configuration
options: Strict & Lax. This commit introduces validation of the
associated same_site attribute.

The main motivation for validating this value is ensuring that awry
option values don't cause unexpected behaviour. As this is a sensitive
security option, I think validation is warranted.

The main drawback of validating the option value is that Rack won't
immediately support new options.

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -251,7 +251,11 @@ module Rack
         same_site = if value[:same_site]
           case value[:same_site]
           when Symbol, String
-            "; SameSite=#{value[:same_site]}"
+            same_site_rule = value[:same_site].to_s
+            unless ["Lax", "Strict"].include?(same_site_rule)
+              raise ArgumentError, "Unrecognized cookie header value for SameSite option: #{same_site_rule}"
+            end
+            "; SameSite=#{same_site_rule}"
           else
             "; SameSite"
           end
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -248,18 +248,17 @@ module Rack
           rfc2822(value[:expires].clone.gmtime) if value[:expires]
         secure = "; secure"  if value[:secure]
         httponly = "; HttpOnly" if (value.key?(:httponly) ? value[:httponly] : value[:http_only])
-        same_site = if value[:same_site]
+        same_site =
           case value[:same_site]
-          when Symbol, String
-            same_site_rule = value[:same_site].to_s
-            unless ["Lax", "Strict"].include?(same_site_rule)
-              raise ArgumentError, "Unrecognized cookie header value for SameSite option: #{same_site_rule}"
-            end
-            "; SameSite=#{same_site_rule}"
+          when false, nil
+            nil
+          when :lax, 'Lax', :Lax
+            '; SameSite=Lax'.freeze
+          when true, :strict, 'Strict', :Strict
+            '; SameSite=Strict'.freeze
           else
-            "; SameSite"
+            raise ArgumentError, "Invalid SameSite value: #{value[:same_site].inspect}"
           end
-        end
         value = value[:value]
       end
       value = [value] unless Array === value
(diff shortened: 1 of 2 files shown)


jeremy:
👍 to validate the attribute value now that it's required by spec, too. This [changed in draft 7](https://tools.ietf.org/rfcdiff?url2=draft-west-first-party-cookies-07.txt). Draft 6 assumed Strict then downgraded to Lax; draft 7 treats the entire cookie as invalid and ignores it unless the attribute value is Strict or Lax.

Looks like this came up @ https://bugs.chromium.org/p/chromium/issues/detail?id=600983

/cc @mastahyeti

jeremy:
Ref #1033

jeremy:
Draft 7 is a little confusing now. It allows SameSite with no attribute value in 3.1 and says that cookies with an invalid SameSite value should be treated as Strict in 3.2, yet 4.1 makes it clear that User Agents must ignore cookies with a SameSite attribute whose value is not Strict or Lax.

3.1
> samesite-av    = "SameSite" / "SameSite=" samesite-value

3.2
> If the "SameSite" attribute's value is "Strict", or if the value is invalid, the cookie will only be sent along with "same-site" requests.

4.1
> 1.  If "cookie-av"'s "attribute-value" is not a case-sensitive match for "Strict" or "Lax", ignore the "cookie-av".

/cc @mikewest

lib/rack/utils.rb
```diff
@@ -251,7 +251,11 @@ def add_cookie_to_header(header, key, value)
...
           when Symbol, String
-            "; SameSite=#{value[:same_site]}"
+            same_site_rule = value[:same_site].to_s
+            unless ["Lax", "Strict"].include?(same_site_rule)
+              raise ArgumentError, "Unrecognized cookie header value for SameSite option: #{same_site_rule}"
+            end
+            "; SameSite=#{same_site_rule}"
           else
             "; SameSite"
           end
```
jeremy: We can dial this back to a stricter whitelist now that the spec requires valid values, e.g.
```ruby
same_site =
  case value[:same_site]
  when false, nil
    nil
  when :lax, 'Lax'
    '; SameSite=Lax'.freeze
  when true, :strict, 'Strict'
    '; SameSite=Strict'.freeze
  else
    raise ArgumentError, "Invalid SameSite value: #{value[:same_site].inspect}"
  end
```

bobjflong: cool 👍 added the uppercase symbols too

test/spec_response.rb
```diff
@@ -115,12 +115,6 @@
...
 
-  it "can set SameSite cookies with any truthy value" do
-    response = Rack::Response.new
-    response.set_cookie "foo", {:value => "bar", :same_site => Object.new}
-    response["Set-Cookie"].must_equal "foo=bar; SameSite"
-  end
-
   it "can set SameSite cookies with string value" do
     response = Rack::Response.new
     response.set_cookie "foo", {:value => "bar", :same_site => "Lax"}
```
jeremy: Let's add specific test coverage for the allowed values, too.

bobjflong: done!

mikewest:
@jeremy: Thanks! Looks like it's time for an -08 draft. :)

The text in 4.1 is what I intended throughout, and matches Chrome's behavior in 51: `SameSite` and `SameSite=Blah` both cause the cookie to be ignored (which gives us some options for extensibility in the future, and feature-detection in the present). I'll adjust the document accordingly in mikewest/internetdrafts#11.

jeremy:
Merged @ https://github.com/rack/rack/commit/f0f828cc1499cf54495e545daecb992a21fef324. Thanks @bobjflong!

reedloden:
Could we get a new `rack` release that includes `SameSite` support? :)

jeremy:
Backported @ https://github.com/rack/rack/commit/6712b864d574b7e09bf365cde3666e2da78b51b4

PR #1058, opened 2016-04-22
Failing test for complex query parsing

vipulnsward:
Ref: https://github.com/rails/rails/issues/23997#issuecomment-213117462

Haven't investigated this. Just copying report from above.

### DIFF
--- a/test/spec_utils.rb
+++ b/test/spec_utils.rb
@@ -316,6 +316,38 @@ describe Rack::Utils do
       message.must_equal "value must be a Hash"
   end
 
+  it 'builds and parses nested complex query correctly' do
+    complex_data = {
+        data: {
+            type: "thing",
+            attributes: {
+                type: "subthing",
+                id: "6f8a6546-5822-4ed5-a5c5-25420dbae498",
+                cards: [
+                    {
+                        default: true,
+                        layers: [
+                            {
+                                type: "title",
+                                attributes: {
+                                    text: "You've Got Mail!"
+                                }
+                            },
+                            {
+                                type: "topic-subject",
+                                attributes: {
+                                    url: "test"
+                                }
+                            }
+                        ]
+                    }
+                ]
+            }
+        }
+    }
+    Rack::Utils.parse_nested_query(Rack::Utils.build_nested_query(complex_data)).must_equal complex_data
+  end
+
   it 'performs the inverse function of #parse_nested_query' do
     [{"foo" => nil, "bar" => ""},
       {"foo" => "bar", "baz" => ""},
--- a/test/spec_utils.rb
+++ b/test/spec_utils.rb
@@ -317,34 +317,7 @@ describe Rack::Utils do
   end
 
   it 'builds and parses nested complex query correctly' do
-    complex_data = {
-        data: {
-            type: "thing",
-            attributes: {
-                type: "subthing",
-                id: "6f8a6546-5822-4ed5-a5c5-25420dbae498",
-                cards: [
-                    {
-                        default: true,
-                        layers: [
-                            {
-                                type: "title",
-                                attributes: {
-                                    text: "You've Got Mail!"
-                                }
-                            },
-                            {
-                                type: "topic-subject",
-                                attributes: {
-                                    url: "test"
-                                }
-                            }
-                        ]
-                    }
-                ]
-            }
-        }
-    }
+    complex_data = { "w" => [{ "x" => [{ "y" => "1" }, { "y" => "2" }] }] }
     Rack::Utils.parse_nested_query(Rack::Utils.build_nested_query(complex_data)).must_equal complex_data
   end
 


gioele:
This test hits three different mismatches:

1. keys are decoded as strings instead of symbols. This looks like the intended behavior.
2. the boolean `true` is decoded as the string `"true"`. This also looks like the intended behavior.
3. arrays nested in hashes are not properly encoded/decode. This one is the real problem.

A minimal test case for problem 3 is

```ruby
complex_data = {
    "array1" => [
        {
            "array2" => [
                { "key" => "A" },
                { "key" => "B" }
            ]
        }
    ]
}
```

```diff
--- expected
+++ actual
@@ -1 +1 @@
-{"array1"=>[{"array2"=>[{"key"=>"A"}, {"key"=>"B"}]}]}
+{"array1"=>[{"array2"=>[{"key"=>"A"}]}, {"array2"=>[{"key"=>"B"}]}]}
```

jrabramson:
The main issue here is that the nesting is casted identically in situations where it isn't, for example, consider the following two datasets: 
```
complex_data1 = {
    "array1" => [
        {
            "array2" => [
                { "key" => "A" },
                { "key" => "B" }
            ]
        }
    ]
}

complex_data2 = {
      "array1" => [
          {
              "array2" => [
                { "key" => "A" }
              ]
          },
          {
              "array2" => [
                { "key" => "B" }
              ]
          }
      ]
    }
```
Both of those are built to look like: ```"array1[][array2][][key]=A&array1[][array2][][key]=B"```

looks identical, thus it falls to the parser to decide if the identical key should be cast as a new item in the array or as a new key.

parse_nested_query previous would parse it as complex_data1 however 7f00781a introduced a change that make it now parse it as complex_data2.

which is right?

vipulnsward:
I will look into this tomorrow.

vipulnsward:
> EDIT: it's too bad the querystring doesn't contain indexes ie: "array1[0][array2][0][key]=A&array1[1][array2][0][key]=B"

We reserve these to build hashes.

vipulnsward:
Is generating a query like `"array1[][array2][][key]=A&;array1[][array2][][key]=B"` (Notice extra ; as delimiter), valid way to segregate these two? 
Or any other segregation ideas? cc @jeremy

bquorning:
>> EDIT: it's too bad the querystring doesn't contain indexes ie: "array1[0][array2][0][key]=A&array1[1][array2][0][key]=B"

> We reserve these to build hashes.

Might using colons for hash keys (inspired by Ruby symbols) help distinguish them from array indexes?
```
"array1[0][:array2][0][:key]=A&array1[1][:array2][0][:key]=B"
```

rubydev:
I have same issue. In my tests missing some parameters:

Test:
<pre>
  post :create, params: { project: valid_attributes }, as: :json, format: :json
</pre>

Input:
<pre>
{
   "name":"Project",
   "gap":1,
   "user_id":1,
   "solar_panel_id":1,
   "wind_exposure":"B",
   "risk_category":"I",
   "roof_type":"comp",
   "clamp_finish":"black",
   "address":"Street 123",
   "address_city":"My City",
   "address_zip":"74601",
   "address_state":"My State",
   "snow_load":0,
   "roof_slope":18,
   "wind_speed":80,
   "layouts":[
      {
         "name":"Layout 1",
         "top_skirt":false,
         "bottom_skirt":true,
         "override_cantilever_length":false,
         "cantilever_length":4,
         "override_row":0,
         "rafter_spacing":24.0,
         "rows":[
            {
               "orientation":"portrait",
               "items":[
                  {
                     "on":true,
                     "space":0
                  },
                  {
                     "on":true,
                     "space":1
                  },
                  {
                     "on":true,
                     "space":1
                  },
                  {
                     "on":true,
                     "space":1
                  }
               ]
            },
            {
               "orientation":"portrait",
               "items":[
                  {
                     "on":true,
                     "space":0
                  },
                  {
                     "on":true,
                     "space":1
                  },
                  {
                     "on":true,
                     "space":1
                  },
                  {
                     "on":true,
                     "space":1
                  }
               ]
            }
         ]
      }
   ]
}
</pre>

Output:
<pre>
{
   "address":"Street 123",
   "address_city":"My City",
   "address_state":"My State",
   "address_zip":"74601",
   "clamp_finish":"black",
   "gap":"1",
   "layouts":[
      {
         "bottom_skirt":"true",
         "cantilever_length":"4",
         "name":"Layout 1",
         "override_cantilever_length":"false",
         "override_row":"0",
         "rafter_spacing":"24.0",
         "rows":[
            {
               "items":[
                  {
                     "on":"true",
                     "space":"0"
                  },
                  {
                     "on":"true",
                     "space":"1"
                  },
                  {
                     "on":"true",
                     "space":"1"
                  },
                  {
                     "on":"true",
                     "space":"1"
                  },
                  {
                     "on":"true",
                     "space":"0"
                  },
                  {
                     "on":"true",
                     "space":"1"
                  },
                  {
                     "on":"true",
                     "space":"1"
                  },
                  {
                     "on":"true",
                     "space":"1"
                  }
               ],
               "orientation":"portrait"
            },
            {
               "orientation":"portrait"
            }
         ],
         "top_skirt":"false"
      }
   ],
   "name":"Project",
   "risk_category":"I",
   "roof_slope":"18",
   "roof_type":"comp",
   "snow_load":"0",
   "solar_panel_id":"1",
   "user_id":"1",
   "wind_exposure":"B",
   "wind_speed":"80"
}
</pre>

Items from second "items" array was moved to first one.

Can you help me please?

Rails version: 5.0.0.1
Ruby version: 2.3.0
Rack version: 2.0.1
Rspec-rails version: 3.5.2

mockdeep:
It looks like this is passing on the latest master.

ioquatix:
@tenderlove can we close this?

jeremyevans:
I think we can close this.  It's supposed to be a failing test for rack's behavior, and it now passes.

In general, you should avoid trying to parse arrays containing arrays or hashes, as behavior is in general ambiguous.  Either use json or use hashes and later convert the hash to arrays (which results in unambiguous behavior).

Issue #945, opened 2015-09-09
Rack::File lint error for HEAD requests that 404 (response body should not be given)


edmorley:
Using Rack v1.6.4 and Ruby v2.1.5p273.

config.ru:
```ruby
run Rack::Static.new(
  @app,
  :urls => [""],
  :index => 'index.html'
)
```

Then after running `rackup`:
```bash
[~/src]$ curl -I "http://localhost:9292/non-existent-file.txt"
HTTP/1.1 500 Internal Server Error
Content-Type: text/html; charset=ISO-8859-1
Content-Length: 39
X-Cascade: pass
Server: WEBrick/1.3.1 (Ruby/2.1.5/2014-11-13)
Date: Wed, 09 Sep 2015 22:05:08 GMT
Connection: close
```

From the Rack logs:
```
::1 - - [09/Sep/2015:23:05:08 +0100] "HEAD /non-existent-file.txt HTTP/1.1" 404 39 0.0007
[2015-09-09 23:05:08] ERROR Rack::Lint::LintError: Response body was given for HEAD request, but should be empty
        /usr/lib/ruby/gems/2.1.0/gems/rack-1.6.4/lib/rack/lint.rb:20:in `assert'
        /usr/lib/ruby/gems/2.1.0/gems/rack-1.6.4/lib/rack/lint.rb:688:in `verify_content_length'
        /usr/lib/ruby/gems/2.1.0/gems/rack-1.6.4/lib/rack/lint.rb:716:in `each'
        /usr/lib/ruby/gems/2.1.0/gems/rack-1.6.4/lib/rack/body_proxy.rb:31:in `each'
        /usr/lib/ruby/gems/2.1.0/gems/rack-1.6.4/lib/rack/handler/webrick.rb:112:in `service'
        /usr/lib/ruby/2.1.0/webrick/httpserver.rb:138:in `service'
        /usr/lib/ruby/2.1.0/webrick/httpserver.rb:94:in `run'
        /usr/lib/ruby/2.1.0/webrick/server.rb:295:in `block in start_thread'
```

Whilst Rack::File correctly handles HEAD requests (ie by returning an empty body) for successfully served files:
https://github.com/rack/rack/blob/c28f271d0c91f45e13bfa8f07bed445ef91f41de/lib/rack/file.rb#L125-L126
```ruby
    def make_body request, path, range
      if request.head?
        []
      else
        ...
```

There is no such handling in `fail()`:
```ruby
    def fail(status, body, headers = {})
[...]

PR #1062, opened 2016-04-24
head requests for missing files are now working (issue #945)

tgrindinger:
`fail` needs to know the request method in order to send the appropriate response, since HEAD responses cannot have a body.

### DIFF
--- a/lib/rack/file.rb
+++ b/lib/rack/file.rb
@@ -27,7 +27,7 @@ module Rack
     def call(env)
       request = Rack::Request.new env
       unless ALLOWED_VERBS.include? request.request_method
-        return fail(405, request.request_method, "Method Not Allowed", {'Allow' => ALLOW_HEADER})
+        return fail_handler(405, request.request_method, "Method Not Allowed", {'Allow' => ALLOW_HEADER})
       end
 
       path_info = Utils.unescape_path request.path_info
@@ -44,7 +44,7 @@ module Rack
       if available
         serving(request, path)
       else
-        fail(404, request.request_method, "File not found: #{path_info}")
+        fail_handler(404, request.request_method, "File not found: #{path_info}")
       end
     end
 
@@ -75,7 +75,7 @@ module Rack
         range = 0..size-1
       elsif ranges.empty?
         # Unsatisfiable. Return error, and file size:
-        response = fail(416, request.request_method, "Byte range unsatisfiable")
+        response = fail_handler(416, request.request_method, "Byte range unsatisfiable")
         response[1]["Content-Range"] = "bytes */#{size}"
         return response
       else
@@ -129,10 +129,12 @@ module Rack
       end
     end
 
-    def fail(status, request_method, body, headers = {})
+    def fail_handler(status, request_method, body, headers = {})
+      request_method == "HEAD" ? [status, headers, []] : fail(status, body, headers)
+    end
+
+    def fail(status, body, headers = {})
       body += "\n"
-      
-      return [status, headers, []] if request_method == "HEAD"
 
       [
         status,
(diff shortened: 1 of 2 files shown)


tgrindinger:
making sure this shows up in issue #945

jeremy:
Changing the `#fail` method signature breaks `Rack::File` subclasses that call `#fail`.

Can we introduce a new method that conditionally handles HEAD requests and otherwise delegates to `#fail`?

Alternatively, we could lean on `Rack::Head` to do this job:
```diff
diff --git a/lib/rack/file.rb b/lib/rack/file.rb
index 5b755f5..552edc9 100644
--- a/lib/rack/file.rb
+++ b/lib/rack/file.rb
@@ -2,6 +2,7 @@ require 'time'
 require 'rack/utils'
 require 'rack/mime'
 require 'rack/request'
+require 'rack/head'
 
 module Rack
   # Rack::File serves files below the +root+ directory given, according to the
@@ -22,9 +23,14 @@ module Rack
       @root = root
       @headers = headers
       @default_mime = default_mime
+      @head = Rack::Head.new(lambda { |env| get env })
     end
 
     def call(env)
+      @head.call env
+    end
+
+    def get(env)
       request = Rack::Request.new env
       unless ALLOWED_VERBS.include? request.request_method
         return fail(405, "Method Not Allowed", {'Allow' => ALLOW_HEADER})
```

tgrindinger:
I did some checking and couldn't find anything outside of `Rack::File` that uses `#fail`, even in other projects. Just out of curiosity, I'd like to see some examples of this.

I think your suggestion of making a wrapper might be the best approach. Some of the code in `#fail` adds some headers for content type and length, which would logically be undefined for "file not found" requests. In that case, I'm not sure if `fail` would... um, fail, but those are headers I wouldn't expect to be present, anyway. I could be wrong about that, though.

jeremy:
I did a GitHub code search and turned up some examples: https://github.com/search?l=ruby&p=1&q=class+%22Rack%3A%3AFile%22+fail&ref=searchresults&type=Code&utf8=✓

e.g. https://github.com/joecorcoran/pannier/blob/5981edd9c13a8752aa0865a72ec7f61ef2b4aff3/lib/pannier/file_handler.rb#L10

tgrindinger:
Interesting, thanks for the info. I'll be sure to use that search pattern in the future.

gioele:
Is it worth complicating the API to preserve compatibility given that the next release will be a 2.0, so API breaks are fine by semver?

jeremy:
@tgrindinger Think we're better off doing the `Rack::Head` wrapper. Introducing a new method does work for our own `#fail` usage, but it leaves subclasses exposed. They'll still wrongly include response bodies for HEAD requests.

@gioele Always worth examining that tradeoff. I think complication v. compatibility is a false dilemma in this case, though. We can preserve compatibility without complicating (augmenting, expanding, changing?) the implementation.

tgrindinger:
@jeremy That's a good point. After some more research into HEAD, it looks like the headers I was concerned about actually SHOULD have been there, so this pull request is not correct. If I were to amend my pull request, it would basically be your code with a new test attached. Should I amend this pull request or do you want to make your own?

jeremy:
Please do 😁

tgrindinger:
Done, but my build is failing on ruby-head. Pretty sure it's unrelated. Looks like maybe there's a bug for if statements with backtick expressions on ruby-head that is being caused by this line:
[`` if `which lighttpd` && $?.success? ``](https://github.com/rack/rack/blob/master/test/helper.rb#L5)
I can fix it in another pull request.

jeremy:
Squashed a893d50ce83e616f06ce24fcac8b902edbee3e31

jkowens:
@tgrindinger @jeremy I believe this fix should be applied to Rack::Directory as well?

jeremy:
Please do PR against other endpoints that should be natively HEAD-aware.
Can survey and hit them all at once.

PR #1064, opened 2016-04-28
split lighttpd condition to fix failing builds on travis

tgrindinger:
ruby-head is failing on this line in [test/helper.rb](https://github.com/rack/rack/blob/master/test/helper.rb#L5):
`` if `which lighttpd` && $?.success? ``
This line works fine in IRB, but if you create a .rb file with just this line:
`` `which lighttpd` && $?.success? ? true : false ``
and run it through command-line ruby, it will give the same error.
`` `which lighttpd` `` will always be truthy, as long as it's run in an environment that has `which`, so it doesn't need to be in the condition, anyway.

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -2,7 +2,8 @@ require 'minitest/autorun'
 
 module Rack
   class TestCase < Minitest::Test
-    if `which lighttpd` && $?.success?
+    `which lighttpd`
+    if $?.success?
       begin
         # Keep this first.
         LIGHTTPD_PID = fork {


jeremy:
The `which lighttpd` is always truthy; it's included inline for a nice concise conditional. Why's it causing a failure. Got a link to an example failure?

jeremy:
I see. The error is that `$?` [is nil at that point](https://travis-ci.org/rack/rack/jobs/126469590#L242)… for some reason. Why?

tgrindinger:
I did some digging into the recent ruby lang repository commits and couldn't find what might be causing it. I know it worked on April 24th, so it must be something that changed in core ruby since then. It appears to specifically be a problem with conditional statements, since
`` `which lighttpd` && $?.success? ``
works just fine, when run through command-line ruby...

jeremy:
Gotcha. Think the code's a bit clearer when separated, too, so went ahead and squashed: f3a108619037d7f1e51230a8437af8f23764eca7

PR #1067, opened 2016-05-04
More consistence for release const

gearnode:
Use same pattern for version and release.
I don't know if made to a File.open is voluntary. So I propose to make the code "more coherent" and "more simple".

### DIFF
--- a/lib/rack.rb
+++ b/lib/rack.rb
@@ -18,11 +18,12 @@ module Rack
     VERSION.join(".")
   end
 
-  RELEASE = "2.0.0.alpha"
+  # The Rack release version number.
+  RELEASE = [2,0,0,'alpha']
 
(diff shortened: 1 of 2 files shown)


gearnode:
@raggi I can't relauch ci, and it's bad surprise the test does't response...

tenderlove:
Adding . to the load path concerns me, this should at least be based of the location of the gemspec file.  Second, I don't want to load lib/rack in the gemspec file.  What is this PR trying to accomplish?  Just removing the regex from the gemspec file?

gearnode:
@tenderlove Sorry the response delay... Yes it's just that... I understand it's not important. 

I would like who is the roadmap for rack 2 ? I would like contribute to rack.
Thanks.

PR #1070, opened 2016-05-11
Use SCRIPT_NAME + PATH_INFO when logging request paths

Erol:
Just noticed that `Rack::CommonLogger` is not using `SCRIPT_NAME` when logging request paths. Not sure if this was intended, but if it's indeed a bug, this PR hopes to address that.


Erol:
Just wanted to follow this up. Is there an update on this?

jeremyevans:
From what I've been reading on the common log format, the logged information for the request line should be: `The request line exactly as it came from the client.`.  This indicates that the SCRIPT_NAME should be included.  I'll send a PR that includes this commit and refactors slightly.

PR #1079, opened 2016-06-09
added the homebrew way to install FCGI in the README file

sophiedeziel:
I could not install the FCGI libraries on my machine because of a 404 error on the package URL (http://www.fastcgi.com/dist/fcgi-2.4.0.tar.gz). Maybe I should try finding a working URL, but it worked fine with Homebrew. 

I found a backup of the old FCGI website but I haven't tested the package below: 

https://github.com/FastCGI-Archives/FastCGI.com/blob/master/original_snapshot/fcgi-2.4.1-SNAP-0910052249.tar.gz


README.rdoc
```diff
@@ -168,9 +168,16 @@ Download and install lighttpd:
 
 Installing the FCGI libraries:
 
+If you use Homebrew on Mac:
```
mikegee: I think Homebrew is available outside of just Mac these days. 🤷‍♂️

sophiedeziel: Oh! 😄

Issue #1077, opened 2016-06-09
PID file deletion race condition when using Puma


sophiedeziel:
I have that same error really well described here:
https://github.com/puma/puma/issues/915

TL;DR : When using Puma, that method of server.rb bellow (Rack) tries to delete the pid file. Every Puma thread executes it and they all see that the file exists. So only the first thread can delete it and every other will raise a `ENOENT No such file tmp/pids/server.pid`

``` ruby
      def write_pid
        ::File.open(options[:pid], ::File::CREAT | ::File::EXCL | ::File::WRONLY ){ |f| f.write("#{Process.pid}") }
        at_exit { ::File.delete(options[:pid]) if ::File.exist?(options[:pid]) }
      rescue Errno::EEXIST
        check_pid!
        retry
      end
```

This is something I'd love to submit a pull request for, but I am not sure of the approach you'd prefer. 

It is clear to me that we could rescue the `Errno::ENOENT`. I'd like to replace the `if` entirely with it but I don't know if you'd like it.

chneukirchen:
The `if` is unnecessary indeed.  Just delete and ignore errors.

steakknife:
This is probably fixed now, yeah?

sophiedeziel:
Yes it is. People will benefit of the fix when a new version is deployed to RubyGems.

PR #1080, opened 2016-06-09
Prevent exception caused by a race condition on multi-threaded servers

sophiedeziel:
I opened an issue earlier today: https://github.com/rack/rack/issues/1077 :

> I have that same error really well described here:
puma/puma#915

> TL;DR : When using Puma, that method of server.rb bellow (Rack) tries to delete the pid file. Every Puma thread executes it and they all see that the file exists. So only the first thread can delete it and every other will raise a ENOENT No such file tmp/pids/server.pid

> ```
  def write_pid
    ::File.open(options[:pid], ::File::CREAT | ::File::EXCL | ::File::WRONLY ){ |f| f.write("#{Process.pid}") }
        at_exit { ::File.delete(options[:pid]) if ::File.exist?(options[:pid]) }
  rescue Errno::EEXIST
    check_pid!
    retry
  end
```

> This is something I'd love to submit a pull request for, but I am not sure of the approach you'd prefer.

> It is clear to me that we could rescue the Errno::ENOENT. I'd like to replace the if entirely with it but I don't know if you'd like it.

### DIFF
--- a/lib/rack/server.rb
+++ b/lib/rack/server.rb
@@ -1,4 +1,5 @@
 require 'optparse'
+require 'fileutils'
 
 
 module Rack
--- a/lib/rack/server.rb
+++ b/lib/rack/server.rb
@@ -359,7 +359,7 @@ module Rack
 
       def write_pid
         ::File.open(options[:pid], ::File::CREAT | ::File::EXCL | ::File::WRONLY ){ |f| f.write("#{Process.pid}") }
-        at_exit { ::File.delete(options[:pid]) rescue Errno::ENOENT }
+        at_exit { ::FileUtils.rm_f(options[:pid]) }
       rescue Errno::EEXIST
         check_pid!
         retry
--- a/lib/rack/server.rb
+++ b/lib/rack/server.rb
@@ -359,7 +359,7 @@ module Rack
 
       def write_pid
         ::File.open(options[:pid], ::File::CREAT | ::File::EXCL | ::File::WRONLY ){ |f| f.write("#{Process.pid}") }
-        at_exit { ::File.delete(options[:pid]) if ::File.exist?(options[:pid]) }
+        at_exit { ::File.delete(options[:pid]) rescue Errno::ENOENT }
       rescue Errno::EEXIST
         check_pid!
         retry

### CODE REVIEW

**Comment by kiramclean** [comment]
+1

File: lib/rack/server.rb [hunk 1]
```diff
@@ -359,7 +359,7 @@ def daemonize_app
 
       def write_pid
         ::File.open(options[:pid], ::File::CREAT | ::File::EXCL | ::File::WRONLY ){ |f| f.write("#{Process.pid}") }
-        at_exit { ::File.delete(options[:pid]) if ::File.exist?(options[:pid]) }
+        at_exit { ::File.delete(options[:pid]) rescue Errno::ENOENT }
```
jeremy: Subtle – need to wrap in `begin … rescue … end` in this case since inline rescues don't take an exception class (this is actually rescuing StandardError and returning Errno::ENOENT). Switching to stdlib `FileUtils#rm_f` would also neatly wrap this up.

sophiedeziel: Thank you for your feedback!

I'll fix that :+1:

rachel-carvalho:
I'm also having this problem and found this PR, is there something I can do to help get it merged? Thanks!

sophiedeziel:
I am available to make any changes required

ericboehs:
Yeah this seems to work for me. Hoping we can get this merged someday.

kiramclean:
Any chance of merging this? I have the same problem and would love to stop getting this exception!

darronschall:
Is there any chance of merging this into 1.6-stable and releasing a 1.6.6 version of the gem?

I see 1.6.5 was released on Nov 10, 2016 around the same time this PR was merged into master, but diff from 1.6.4 to 1.6.5 doesn't contain this fix: https://github.com/rack/rack/compare/ee18520bd9894e68a5d2b26c82835c2e67a43a8b...1886a60e9b96c2a4106fd89eb41dc817696c6369

mhfairbank:
Seconding @darronschall

In the meantime, I've edited the gem source code on my local which seems to work but feels pretty hacky.

PR #1083, opened 2016-06-15
Attempt server.pid deletion without throwing exception

calasyr:
When multiple workers are attempting to delete server.pid at exit, one or more workers can find the file before it is deleted and then try to delete it after its already deleted.  Rescuing the exception that can result eliminates some unnecessary noise.


jeremy:
Thanks @calasyr, but please see #1080 😁

calasyr:
Thanks.  Will check it out.

PR #1084, opened 2016-06-15
Accept multiple require options

syohex:
This PR makes `rackup` accept multiple **require** options. Currently only last **require** library is loaded. Other library are ignored. For example

```
% rackup -r foo -r bar
```

`bar` is loaded but `foo` is not loaded.


jeremyevans:
Recently handled in 723e8fa1f409949c420fd378dbe23a2b9e73fbae.

PR #1085, opened 2016-06-17
Kill Constants

schneems:
Since Rack master is Ruby 2.2+ we get string freezing in hash keys by default. This change was discussed on the list.

We might need to still `.freeze` in a few places where we are calling setter or getter methods for hashes. like `set_header` method in `request.rb`

I can squash and work on it more next week. This took much longer than I would have thought, wanted to get it up on github before the weekend.

### DIFF
--- a/lib/rack.rb
+++ b/lib/rack.rb
@@ -26,7 +26,6 @@ module Rack
   end
 
   # Rack environment variables
-  RACK_SHOWSTATUS_DETAIL              = 'rack.showstatus.detail'.freeze
   RACK_MULTITHREAD                    = 'rack.multithread'.freeze
   RACK_MULTIPROCESS                   = 'rack.multiprocess'.freeze
   RACK_RUNONCE                        = 'rack.run_once'.freeze
--- a/lib/rack.rb
+++ b/lib/rack.rb
@@ -26,7 +26,6 @@ module Rack
   end
 
   # Rack environment variables
-  RACK_HIJACK_IO                      = 'rack.hijack_io'.freeze
   RACK_RECURSIVE_INCLUDE              = 'rack.recursive.include'.freeze
   RACK_MULTIPART_BUFFER_SIZE          = 'rack.multipart.buffer_size'.freeze
   RACK_MULTIPART_TEMPFILE_FACTORY     = 'rack.multipart.tempfile_factory'.freeze
--- a/lib/rack.rb
+++ b/lib/rack.rb
@@ -25,7 +25,6 @@ module Rack
     RELEASE
   end
 
-  HTTP_HOST         = 'HTTP_HOST'.freeze
   HTTP_VERSION      = 'HTTP_VERSION'.freeze
   HTTPS             = 'HTTPS'.freeze
   PATH_INFO         = 'PATH_INFO'.freeze
--- a/lib/rack.rb
+++ b/lib/rack.rb
@@ -26,7 +26,6 @@ module Rack
   end
 
   # Rack environment variables
-  RACK_HIJACK                         = 'rack.hijack'.freeze
   RACK_IS_HIJACK                      = 'rack.hijack?'.freeze
   RACK_HIJACK_IO                      = 'rack.hijack_io'.freeze
   RACK_RECURSIVE_INCLUDE              = 'rack.recursive.include'.freeze
--- a/lib/rack.rb
+++ b/lib/rack.rb
@@ -25,7 +25,6 @@ module Rack
     RELEASE
   end
 
-  REQUEST_PATH      = 'REQUEST_PATH'.freeze
   SCRIPT_NAME       = 'SCRIPT_NAME'.freeze
   QUERY_STRING      = 'QUERY_STRING'.freeze
   SERVER_PROTOCOL   = 'SERVER_PROTOCOL'.freeze
--- a/lib/rack.rb
+++ b/lib/rack.rb
@@ -26,7 +26,6 @@ module Rack
   end
 
   # Rack environment variables
-  RACK_REQUEST_QUERY_STRING           = 'rack.request.query_string'.freeze
   RACK_METHODOVERRIDE_ORIGINAL_METHOD = 'rack.methodoverride.original_method'.freeze
   RACK_SESSION_UNPACKED_COOKIE_DATA   = 'rack.session.unpacked_cookie_data'.freeze
 
--- a/lib/rack.rb
+++ b/lib/rack.rb
@@ -25,7 +25,6 @@ module Rack
     RELEASE
   end
 
-  SERVER_PORT       = 'SERVER_PORT'.freeze
   CONTENT_LENGTH    = 'Content-Length'.freeze
   CONTENT_TYPE      = 'Content-Type'.freeze
   SET_COOKIE        = 'Set-Cookie'.freeze
--- a/lib/rack.rb
+++ b/lib/rack.rb
@@ -25,7 +25,6 @@ module Rack
     RELEASE
   end
 
-  REQUEST_METHOD    = 'REQUEST_METHOD'.freeze
   REQUEST_PATH      = 'REQUEST_PATH'.freeze
   SCRIPT_NAME       = 'SCRIPT_NAME'.freeze
   QUERY_STRING      = 'QUERY_STRING'.freeze
--- a/lib/rack.rb
+++ b/lib/rack.rb
@@ -25,7 +25,7 @@ module Rack
     RELEASE
   end
 
-  CONTENT_LENGTH    = 'Content-Length'.freeze
+
   CONTENT_TYPE      = 'Content-Type'.freeze
   SET_COOKIE        = 'Set-Cookie'.freeze
   TRANSFER_ENCODING = 'Transfer-Encoding'.freeze
--- a/lib/rack.rb
+++ b/lib/rack.rb
@@ -25,7 +25,6 @@ module Rack
     RELEASE
   end
 
-  SERVER_ADDR       = 'SERVER_ADDR'.freeze
   SERVER_PORT       = 'SERVER_PORT'.freeze
   CACHE_CONTROL     = 'Cache-Control'.freeze
   CONTENT_LENGTH    = 'Content-Length'.freeze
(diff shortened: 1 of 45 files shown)


lib/rack/recursive.rb
```diff
@@ -14,11 +14,11 @@ def initialize(url, env={})
       @url = URI(url)
       @env = env
 
-      @env[PATH_INFO] =       @url.path
-      @env[QUERY_STRING] =    @url.query  if @url.query
-      @env[HTTP_HOST] =       @url.host   if @url.host
+      @env['PATH_INFO'] =       @url.path
+      @env['QUERY_STRING'] =    @url.query  if @url.query
+      @env['HTTP_HOST'] =       @url.host   if @url.host
       @env["HTTP_PORT"] =     @url.port   if @url.port
```
stoivo: Do you know if there is a difference between  single and double quotes?

Maybe change this one too, just to stay consistent.

schneems: updated and fixed the formatting

lib/rack.rb
```diff
@@ -25,67 +25,6 @@ def self.release
...
-  RACK_REQUEST_FORM_INPUT             = 'rack.request.form_input'.freeze
-  RACK_REQUEST_FORM_HASH              = 'rack.request.form_hash'.freeze
-  RACK_REQUEST_FORM_VARS              = 'rack.request.form_vars'.freeze
-  RACK_REQUEST_COOKIE_HASH            = 'rack.request.cookie_hash'.freeze
-  RACK_REQUEST_COOKIE_STRING          = 'rack.request.cookie_string'.freeze
-  RACK_REQUEST_QUERY_HASH             = 'rack.request.query_hash'.freeze
-  RACK_REQUEST_QUERY_STRING           = 'rack.request.query_string'.freeze
-  RACK_METHODOVERRIDE_ORIGINAL_METHOD = 'rack.methodoverride.original_method'.freeze
-  RACK_SESSION_UNPACKED_COOKIE_DATA   = 'rack.session.unpacked_cookie_data'.freeze
-
```
jeremy: Many of these constants are used by other libs, middleware, and apps. A quick GitHub search for for PATH_INFO, for example:

- https://github.com/tk-hamaguchi/rack_request_builder/blob/a7b634413162b02a238f7482a6760b8580aa3c25/lib/rack_request_builder/rack.rb#L1-L24
- https://github.com/rack-app/rack-app/blob/3347f38bd4767dd3be3484aac8a6a1378448a9b3/lib/rack/app/request_configurator.rb#L12-L16
- https://github.com/ioquatix/utopia/blob/9e016656353bd2884a11e7122ed6d305c62988b2/spec/utopia/session_spec.ru#L4-L20
- https://github.com/literalice/rack-pushstate-static/blob/c69ef8ae388785bdfcc1544b39465b94e99e902e/lib/rack/pushstate_static.rb#L26-L45

lib/rack/etag.rb
```diff
@@ -12,7 +12,7 @@ module Rack
   # used when Etag is absent and a directive when it is present. The first
   # defaults to nil, while the second defaults to "max-age=0, private, must-revalidate"
   class ETag
-    ETAG_STRING = Rack::ETAG
+    ETAG_STRING = 'ETag'
```
jeremy: No longer using this constant, also swapped for string literals below

schneems:
I updated to add the constants back in as some people will still be using them. There were also some codepaths not caught in testing that still had some constants.

If we're doing this for perf we should also either add the special comment to files to freeze strings, or go through and find which strings are being used when passed to methods instead of as hash arguments and freeze them.

fidelisrafael:
I just wondering why exactly changing this can be a good thing...
(we know that) theres a lot of middlewares  that can be depending on these constants....I'ts not easier to `freeze` constants values in assignment?

schneems:
It's faster, check out my blog www.schneems.com i've written a lot about performance, I wrote about this type of optimization in a speeding up rails post i think. The TLDR is that String#freeze is faster than a constant since the constant requires a lookup on every iteration. You can prove it with a trivial benchmark ips script. Give it a shot, and let me know if you've got questions.

Also, i should really finish this up.

fidelisrafael:
I'm wondering, this will be true for users running ruby with immutable strings (`#frozen_string_literal` or `--enable-frozen-string-literal`), right? I mean, in this scenario new strings objects will not be allocated, since freezing will always use the same reference for literal strings (like symbol)

For users running ruby without turning immutable strings activated this will allocate much more objects[1] per request or I'm missing something?

Maybe its a good idea appending `# frozen_string_literal: true` to changed files to force immutability in environments where this is not enabled.

[[1]](https://github.com/rack/rack/commit/dc53a8c26dc55d21240233b3d83d36efdef6e924)

ioquatix:
This would break tons of code. I know I use `Rack::*` constants quite a bit. It also helps avoid typos, so it's not just about performance.

fidelisrafael:
@ioquatix The constants are not removed from code, for what I saw in the last time, so people can still use rack constants

oniofchaos:
With https://github.com/rack/rack/pull/1250 merged I think this PR makes a lot more sense to wrap up as all of these files should now have frozen strings by default.

jeremyevans:
I think it is a good idea to keep the constants to not break external software, but change all constant references to literal strings for performance to avoid constant lookups at runtime.  Any interest in updating this PR or submitting a new PR for that, @schneems ?  I promise to review it quickly and merge it if there are no issues.

ioquatix:
I have a couple of issues with this:

- Having a single constant is generally good in programming rather than scattering strings all over the  place. If there is a performance issue in Ruby, maybe that should be addressed (e.g. caching constants).
- We can't remove existing constants as they are part of the public interface.
- If we are going to do this, should we also follow the conventions for the latest HTTP specification which uses a lower case normal form for headers?

ioquatix:
I discussed this performance issue with @mame. It turns out we do have optimisations in place for constant lookup.

> Writing a literal directly is faster than cache reading

```
$ ruby --dump=i -e 'A'
== disasm: #<ISeq:<main>@-e:1 (1,0)-(1,1)> (catch: FALSE)
0000 opt_getinlinecache                     9, <is:0>                 (   1)[Li]
0003 putobject                              true
0005 getconstant                            :A
0007 opt_setinlinecache                     <is:0>
0009 leave

$ ruby --enable-frozen-string-literal --dump=i -e '"foo"'
== disasm: #<ISeq:<main>@-e:1 (1,0)-(1,5)> (catch: FALSE)
0000 putobject                              "foo"                     (   1)[Li]
0002 leave
```

We looked at performance:

```
$ time ./miniruby -e 'A = "foo"; def forty_two; 42; end; 10000000.times { { A => forty_two } }'
real	0m1.557s
user	0m1.538s
sys	0m0.017s
```

vs

```
$ time ./miniruby -e 'A = "foo"; def forty_two; 42; end; 10000000.times { { "foo" => forty_two } }'
real	0m0.934s
user	0m0.910s
sys	0m0.020s
```

In this micro-benchmark it's about 30% faster.

That being said, I'd personally like to see if this impacts real web application performance before making a decision, because I think from a code POV, it's a bit worse.

headius:
Constant access doesn't need to be any slower than literal access, but it may require a (de)optimizing runtime that can treat constants as constant with some optimization guard to back off when that's not the case. It's not really my place to argue whether it's appropriate to optimize based solely on the capabilities of CRuby.

Here's an example on JRuby, using all the right tricks to reduce the cost of code around the constant/literal access. You can see it's basically identical performance.

JRuby:
```
[] ~/projects/jruby $ jruby -Xfixnum.cache=false -Xcompile.invokedynamic -e 'A = "foo"; def forty_two; 42; end; 10.times { t = Time.now; x = nil; i = 0; while i < 100000000; i+=1; x ||= { "foo" => forty_two }; end; p Time.now - t }'
0.15387
0.250353
0.119222
0.119235
0.11909099999999999
0.119112
0.11880199999999999
0.11924699999999999
0.118906
0.119182

[] ~/projects/jruby $ jruby -Xfixnum.cache=false -Xcompile.invokedynamic -e 'A = "foo"; def forty_two; 42; end; 10.times { t = Time.now; x = nil; i = 0; while i < 100000000; i+=1; x ||= { A => forty_two }; end; p Time.now - t }'
0.150609
0.249679
0.11995499999999999
0.119977
0.11980299999999999
0.118951
0.11916199999999999
0.118879
0.118672
0.119158
```

TruffleRuby runs slightly faster 0.09s range) but the result is the same: an immediate literal (e.g. a frozen string) has no performance difference compared to a constant.

tenderlove:
> Looking at it another way, if constant lookups were cached and as fast as string literals (definitely possible), would we do this?

No, definitely not.

> To me, there is value in using constants, which is not just about performance.

Yes.  Not only are the constants part of Rack's API, we use these "magic strings" in multiple places.  It's good practice to put those types of things in constants so we don't have to go hunting for strings if we change them.

> That being said, I'd personally like to see if this impacts real web application performance before making a decision, because I think from a code POV, it's a bit worse.

I agree with this.  I'm not super keen on this change just because it wins in micro benchmarks.  It loses in code hygiene IMO.  If constant lookups were as fast as frozen string literals, I wouldn't make this change so I'm reticent to accept it (not to mention Charlie's point about this being an optimization for an implementation detail of one specific Ruby implementation).  Maybe we can speed up constant lookups and then more people win?

If we could show this has measurable impact in an app (doesn't need to be Rails), then I'd be more happy to merge this (until we can make constant lookup faster of course, and then I'd revert)

schneems:
Hello 👋 . Author here. I'm closing this PR. We can stop debating.

For some context, I also introduced the string and CONSTANTIZING freezing PR. One of the main concerns was that this would litter the code and that we should update it to remove constants eventually. I got the change merged based on that understanding.

I made this PR as part of my prior agreement. I think that if we were really serious about wanting to remove the contstants to begin with, then what we should have done is use `private_constant` to make sure that no one else uses these constants. We didn't do that, so deprecating these in a reasonable way would be awful. Simply removing them would be a breaking change.

I actually like having these constants for use in other places. Due to this being a breaking change and multiple people chiming in saying we should keep the constants, I agree. 

Thanks all for the comments.

PR #1088, opened 2016-06-20
Minor query parser refactorings

bquorning:
This PR cleans up query_parser ~3~ 2 places:
1. In `#parse_nested_query` there is a guard clause for `qs` being nil or empty, so there should be no need for the `qs || ''` conditional. I applied the same guard clause to `#parse_query`.
2. The logic determining the regex for splitting the query string (`d ? (COMMON_SEP[d] || /[#{d}] */n) : DEFAULT_SEP`) looks a bit too complex for using directly as a method argument. I extracted it to a local variable `delimiter`, and also changed it to use `#fetch` with a block, instead of using `||` for fallback.
3. ~Speed up extracting `first_key` from `child_key`: Using `gsub` and splitting over the inserted whitespaces, only to use the first element isn’t particularly fast. It’s much faster to fetch the first match using a regex:~ 

Please let me know if I should split into multiple PRs.

### DIFF
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -30,11 +30,12 @@ module Rack
     # cookies by changing the characters used in the second
     # parameter (which defaults to '&;').
     def parse_query(qs, d = nil, &unescaper)
+      return {} if qs.nil? || qs.empty?
       unescaper ||= method(:unescape)
 
       params = make_params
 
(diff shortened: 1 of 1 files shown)


jeremyevans:
This would make `parse_query` not handle nil or empty queries correctly, returning a hash instead of an instance of the appropriate params class. That wasn't an issue when this PR was submitted, but it is now.

I don't think the logic here is complicated enough to require refactoring, so I'm going to close this.

PR #1089, opened 2016-06-23
Try harder when deciding whether to add a new array element

### LINKED ISSUE #1086: Incorrect parsing of nested query
Trying to upgrade an app to Rails 5.0.0.rc1, I noticed a bug in Rack’s query parser. I wrote two test cases for spec_utils.rb:

```rb
# Query to hash - fails
Rack::Utils.parse_nested_query("x[][y][][z]=1&x[][y][][w]=2").
  must_equal "x" => [{ "y" => [{ "z" => "1", "w" => "2" }] }]

# Hash to query - passes
Rack::Utils.build_nested_query("x" => [{ "y" => [{ "z" => "1", "w" => "2" }] }]).
  must_equal "x[][y][][z]=1&x[][y][][w]=2"
```

It looks like both these tests would’ve passed before #1029, so there’s a starting point.

Should I open an issue on [rails/rails](https://github.com/rails/rails) too? I personally feel this might be a blocker for their v5.0.0 release.

cc/ @rthbound, @tenderlove (related to https://github.com/rails/rails/issues/23997#issuecomment-224726220 ?)

matthewd:
Only move to a new entry if the end key is taken; checking only the top-level key is insufficient.

This implementation is pretty unfortunate, but is the option that minimizes changes to `normalize_params`. I'm tempted to try for a more invasive (and correspondingly less tacked-on) alternative, though.

Fixes #1086, https://github.com/rails/rails/issues/23997

cc @rthbound @bquorning @bmedenwald

### DIFF
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -102,8 +102,7 @@ module Rack
         child_key = $1
         params[k] ||= []
         raise ParameterTypeError, "expected Array (got #{params[k].class.name}) for param `#{k}'" unless params[k].is_a?(Array)
-        first_key = child_key.gsub(/[\[\]]/, ' ').split.first
-        if params_hash_type?(params[k].last) && !params[k].last.key?(first_key)
+        if params_hash_type?(params[k].last) && !params_hash_has_key?(params[k].last, child_key)
           normalize_params(params[k].last, child_key, v, depth - 1)
         else
           params[k] << normalize_params(make_params, child_key, v, depth - 1)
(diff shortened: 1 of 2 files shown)


bmedenwald:
For what it's worth, this fix does successfully address all the use cases I reported.

lib/rack/query_parser.rb
```diff
@@ -135,6 +134,18 @@ def params_hash_type?(obj)
       obj.kind_of?(@params_class)
     end
 
+    def params_hash_has_key?(hash, key)
+      return false if key =~ /\[\]/
```
kbrock: I'm probably missing something, but it would seem since you are splitting on `[]`, that the guard clause would ensure that `[]` exists in the query?

```ruby
return false if key !~ /\[\]/
```

Then again, I probably just missed the nuance here.

Yay make rack better!

Issue #461, opened 2012-11-26
SPEC does not require `to_hash` on `rack.session` interface


pmahoney:
I encountered this in Rack 1.3.6, but it seems to be also present on master.

The SPEC claims `rack.session` must implement:
- `store(key, value) (aliased as []=)`
- `fetch(key, default = nil) (aliased as [])`
- `delete(key)`
- `clear`

But [session/abstract/id.rb:238](https://github.com/rack/rack/blob/master/lib/rack/session/abstract/id.rb#L238) expects to be able to `merge!` a `rack.session` into a `SessionHash`.  I was able to work around this by having my `rack.session` also implement `to_hash`.

Either I am misunderstanding something, or SPEC should be updated to also require `to_hash`.

ioquatix:
This needs to be reviewed.

jeremyevans:
I agree with @raggi that `rack.session` needs to implement `to_hash`.  Effectively this is required by rack's current session support, so it should be added to SPEC/Lint.  I'll submit a pull request for this.

PR #1090, opened 2016-06-26
fix(session): add a proper #fetch method to Sessionhash

raszi:
Fixes rails/rails#25489

### DIFF
--- a/lib/rack/session/abstract/id.rb
+++ b/lib/rack/session/abstract/id.rb
@@ -54,7 +54,10 @@ module Rack
           load_for_read!
           @data[key.to_s]
         end
-        alias :fetch :[]
+
+        def fetch(key, default=nil)
+          has_key?(key) ? self[key] : default
+        end
 
         def has_key?(key)
           load_for_read!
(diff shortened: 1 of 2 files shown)


raszi:
This PR is also relates to rack/rack#461

raszi:
The failing tests seem to be unrelated to this PR.

tenderlove:
Can you remove the changes in this patch that aren't related to the feature? Specifically the string quoting changes.

raszi:
Well, I was following the Boy Scout Rule, but it is your call.

raszi:
@tenderlove what do you think, is it okay now or should I rebase it?

tenderlove:
I like the idea of a proper fetch method, but I'd like it better if the behavior was more similar to the fetch method on Hash. Can you make it behave the same way as Hash#fetch ?

--
Aaron Patterson
http://tenderlovemaking.com/
I'm on an iPhone so I apologize for top posting.

raszi:
@tenderlove I am waiting for your answer on this.

lerencao:
I think what @tenderlove  means is that `raise` when no key found and no default supplied, and yes also a block argument form.

raszi:
The upstream took a different approach.

PR #1091, opened 2016-06-27
use freeze string constants

quake:


jeremyevans:
Now that frozen string literals are used, there is not much reason to move things to constants (it can actually make code slower).  So I'm going to close this.

PR #1099, opened 2016-07-21
Fix typo in README

vandrijevik:
Alumni is the plural form of alumnus, so there should not be an s at the end.


vandrijevik:
This has been sitting untouched for over 2 months now, so I’m closing it.

PR #1103, opened 2016-07-27
Add deprecation warning for Rack::Utils.bytesize

glaucocustodio:


glaucocustodio:
The test which failed is over `ruby-head` (2.4.0dev). Maybe we should move it to the "Allowed Failures" section.

jeremy:
Could you add a test verifying the the method works as intended? That'll prevent regression as well (due to someone removing it from the public API again).

glaucocustodio:
For sure, I am going to add @jeremy.

glaucocustodio:
Please, take a look now @jeremy.

jeremy:
Wonder whether we should wait until 3.0 until removal (instead of 2.1). Make it easy for Rack libraries to support both Rack 1.x and 2.x without having to go 2.x only.

glaucocustodio:
Until 3.0 seems better indeed. I will change so.

dmitry:
#1141 merged

jeremyevans:
Considering the method was removed close to 5 years ago and has never been in a Rack 2 release, I don't see the point in adding it back in deprecated form at this point, so I'm going to close this.

PR #1105, opened 2016-08-19
Use RFC 6265 date format for expires time.

### LINKED ISSUE #1100: Cookie expires format is invalid (take 2)
It appears that the date format used by the `expires` attribute the `Set-cookie` header is incorrect. This was broken by this commit: 5e0a9413a80e9b0aa8a74bfc806fb20f2ea0bf3c

The comment in the commit is incorrect. It implies that the date format should be as defined in RFC 2822, however I think this is a misreading of section [RFC 2616 section 3.3.1](https://tools.ietf.org/html/rfc2616#section-3.3.1) which includes its own specification of the date format. In particular the difference is that the timezone must be specified as "GMT".

RFC2616 has been superseded by rfc7231 which has the same definition in [section 7.1.1.1](https://tools.ietf.org/html/rfc7231#section-7.1.1.1).

In practice this is probably not causing a lot of problems because most cookie parsers are robust to the error. However occasionally there are problems.

This was first reported in #527.

stereoscott:
Many browsers are forgiving when it comes to interpreting the `expires` value set in a cookie. 

With that said, Rack has been using a date format for its `expires` time that is not consistent with the rest of the web. This PR changes one tiny thing... the timezone format is "GMT" and not "-0000". This is consistent with `Time#httpdate` and `CGI::Cookie` formatting found elsewhere in Ruby.

Exploring the expiration times for various popular sites around the web illustrate this format:

```
curl -I http://www.wikipedia.com | grep Set-Cookie
Set-Cookie: WMF-Last-Access=19-Aug-2016;Path=/;HttpOnly;secure;Expires=Tue, 20 Sep 2016 12:00:00 GMT

curl -I http://www.google.com | grep Set-Cookie
Set-Cookie: xxx; expires=Sat, 18-Feb-2017 18:33:43 GMT; path=/; domain=.google.com; HttpOnly
```

This closes #1100. See the discussion there for more details.

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -221,31 +221,8 @@ module Rack
         domain  = "; domain=#{value[:domain]}"   if value[:domain]
         path    = "; path=#{value[:path]}"       if value[:path]
         max_age = "; max-age=#{value[:max_age]}" if value[:max_age]
-        # There is an RFC mess in the area of date formatting for Cookies. Not
-        # only are there contradicting RFCs and examples within RFC text, but
-        # there are also numerous conflicting names of fields and partially
-        # cross-applicable specifications.
-        #
-        # These are best described in RFC 2616 3.3.1. This RFC text also
-        # specifies that RFC 822 as updated by RFC 1123 is preferred. That is a
-        # fixed length format with space-date delimited fields.
-        #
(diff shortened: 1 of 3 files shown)


rafaelfranca:
Fixed #1144

PR #1107, opened 2016-08-28
Websocket support specification proposal

boazsegev:
After [a somewhat lengthly discussion](https://github.com/boazsegev/iodine/issues/6) including the Thin, Puma and Iodine server implementors, as well as some input from the Rails core team, we have come up with the following websocket specification proposal for Rack.

Please review this proposal and incorporate it into the Rack specification.

---
## The Websocket Callback Object

Web socket connection upgrade and handling is performed using a Websocket Callback Object.

The Websocket Callback Object should be a class (or an instance of such class) who's instances implement any of the following callbacks:
- `on_open()` WILL be called once the upgrade had completed.
- `on_message(data)` WILL be called when incoming Websocket data is received. `data` will be a String with an encoding of UTF-8 for text messages and `binary` for non-text messages (as specified by the Websocket Protocol).
  
   The client must assume that the `data` String will be a recyclable buffer and that it's content will be corrupted the moment the `on_message` callback returns.
- `on_ready()` MAY be called when the state of the out-going socket buffer changes from full back to read (data can be sent to the socket). If `has_pending?` returns `true`, the `on_ready` callback MUST be called once the buffer state changes.
- `on_shutdown()` MAY be called during the server's graceful shutdown process, _before_ the connection is closed and in addition to the `on_close` function (which is called _after_ the connection is closed.
- `on_close()` WILL be called _after_ the connection was closed for whatever reason (socket errors, parsing errors, timeouts, client disconnection, `close` being called, etc').
- `on_open`, `on_ready`, `on_shutdown` and `on_close` shouldn't expect any arguments (`arity == 0`).

The following method names are reserved for the network implementation: `write`, `close` and `has_pending?`.

The server **must** extend the Websocket Callback Object's class using `extend`, so that the  Websocket Callback Object inherits the following methods:
- `write(data)` will attempt to send the data through the websocket connection. `data` **must** be a String. If `data` is UTF-8 encoded, the data will be sent as text. If `data` is binary encoded it will be sent as non-text (as specified by the Websocket Protocol).
  
  `write` has the same delivery promise as `Socket#write` (a successful `write` does **not** mean any of the data actually reached the other side).
  
  `write` shall return `true` on success and `false` if the websocket is closed.
- `close` closes the connection once all the data in the outgoing queue was sent. If `close` is called while there is still data to be sent, `close` will only take effect once the data was sent.
  
  `close` shall always return `nil`.
- `has_pending?` queries the state of the server's buffer for the specific connection (i.e., if the server has any data it is waiting to send through the socket).
  
  `has_pending?`, shall return `true` if the server has data waiting to be written to the socket and the server will call the `on_ready` callback once the buffer is empty and the socket is writable. Otherwise (i.e., if the server doesn't support the `on_ready` callback), `has_pending?` shall return `false`.
  
  To clarify: implementing `has_pending?` is semi-optional, meaning that a server may choose to always return `false`, no matter the actual state of the socket's buffer.

The following keywords (both as method names and instance variable names) are reserved for the internal server implementation: `_server_ws` and `conn_id`.
- The `_server_ws` object is private and shouldn't be accessed by the client.
- The `conn_id` object may be used as a connection ID for any functionality not specified herein.

Connection `ping` / `pong`, timeouts and network considerations should be implemented by the server. It is **recommended** (but not required) that the server send `ping`s to prevent connection timeouts and detect network failure.

Server settings **may** (not required) be provided to allow for customization and adaptation for different network environments or websocket extensions. It is **recommended** that any settings be available as command line arguments and **not** incorporated into the application's logic.
## Upgrading
- **Server**: When an upgrade request is received, the server will set the `env['upgrade.websocket?']` flag to `true`, indicating that: 1. this specific request is upgradable; and 2. this server supports specification.
- **Client**: When a client decides to upgrade a request, they will place a Websocket Callback Object (either a class or an instance) in the `env['upgrade.websocket']` Hash key.
- **Server**: The server will review the `env` Hash _before_ sending the response. If the `env['upgrade.websocket']` was set, the server will perform the upgrade, sending the correct response status and any headers present in the response as well as performing any required housekeeping, such as closing the response body, if exists. The response status provided by the response object shall be ignored and the correct response status shall be set by the server.
- **Server**: Once the upgrade had completed, The server will add the required websocket/network functions to the callback handler or it's class (as aforementioned). If the callback handler is a Class object, the server will create a new instance of that class.
- **Server**: The server will call the `on_open` callback.
  
  No other callbacks shall be called until the `on_open` callback had returned.
  
  Web socket messages shall be handled by the `on_message` callback in the same order in which they arrive and they will **not** be executed concurrently for the same connection.
  
  The `on_close` callback will **not** be called while `on_message` or `on_open` callbacks are running.
  
  The `on_ready` callback might be called concurrently with the `on_message` callback, allowing data to be sent even while other data is being processed. Multi-threading considerations may apply.

### DIFF
--- a/SPEC
+++ b/SPEC
@@ -190,8 +190,8 @@ The Websocket Callback Object should be a class (or an instance of such class) w
 The following method names are reserved for the network implementation: +write+, +close+ and +has_pending?+.
 The server **must** extend the Websocket Callback Object's class using +extend+, so that the  Websocket Callback Object inherits the following methods:
 * +write(data)+ will attempt to send the data through the websocket connection. +data+ **must** be a String. If +data+ is UTF-8 encoded, the data will be sent as text. If +data+ is binary encoded it will be sent as non-text (as specified by the Websocket Protocol).
-  +write+ only promises to schedule an attempt at sending the data. A successful +write+ does **not** mean any of the data actually reached the other side.
-  +write+ shall return +true+ if the data was scheduled to be sent and +false+ if the websocket is closed. +true+ does not mean that any of the data was actually sent.
+  +write+ has the same delivery promise as +Socket#write+ (a successful +write+ does **not** mean any of the data actually reached the other side).
+  +write+ shall return +true+ on success and +false+ if the websocket is closed.
 * +close+ closes the connection once all the data in the outgoing queue was sent. If +close+ is called while there is still data to be sent, +close+ will only take effect once the data was sent.
   +close+ shall always return +nil+.
 * +has_pending?+ queries the state of the server's buffer for the specific connection (i.e., if the server has any data it is waiting to send through the socket).


matthewd:
Setting aside the details of the proposed change, I think it would be more appropriate for servers and applications to implement this as a draft rack extension before it gets merged into the spec proper. The API under discussion is arguably more complex in implementation than the entire current spec; to expect it to be right the first time seems unrealistic, and to force all of rack to declare incompatible version bumps while smoothing it out seems disruptive.

boazsegev:
@matthewd ,

Thank you for your response.

#### Publishing the specification as a draft

Publishing the specification as a draft sounds like a very good idea.

I'm sure that when we implement and use something, a lot more comes to light (such as the Rails team's suggestion to add the semi-optional feature of `has_pending?`).

For example, there was a big debate on whether we should add a `each`-`defer` functionality that will allow users to use the server's queuing / locking mechanism to serialize concurrent connection tasks. (I was probably too intense about the issue or the debate would have been smaller).

Also, some things were left out (such as `close!` or `closed?`).

I'm sure people might have update suggestions after a while... How do you propose we publish the specification as a draft?

#### a "Rack Extension"

I'm not sure what you might mean by "Rack Extension"... I assume you are referring to the same status conferred to the `hijack` API, it being optional to implement...? If so, that's what we had in mind as well.

#### Complexity

As a side note regarding the proposed specification's implementation:

I believe it's simpler then you think, since there's already a lot of code that allows for easy implementation.

It's more about plugging existing things into the server (removing a lot of duplicated functionality) rather then the way things are organized at the moment (the code being attached to the application layer).

We aren't really reinventing the wheel.

boazsegev:
P.S.

Where / How did you publish the `hijack` API during it's time as a draft?

matthewd:
I guess I was thinking of some sort of little-e extension.. perhaps just by publishing it somewhere and having suitably-adventurous people implement it, either with an understanding that it could change out from under them, or using a less official env key (e.g. `draft.websockets?`), or both. Basically, to me this feels like something Rack could adopt after it was already a de facto standard: it doesn't redefine any existing rack semantics, and is purely additive.

In fact, lack of gained efficiencies aside, I think this could be implemented entirely by a hijacking middleware. Maybe that would make for a useful reference implementation, and allow freer adoption on the application side?

Have you (collectively, cc @evanphx) considered whether this even belongs in Rack-proper? It could equally stand alone as a separate specification, which just specifies how to bootstrap over a Rack request. Perhaps with a matching "if you want websockets.." link from Rack, say. (NB, I'm not saying this is what I think should happen... it's just a thought.)

---

As for complexity... I didn't mean difficulty. There is a lot more API here than "call `call` and pass it a hash".

I'm not sure how hijack was handled. That was still a relatively simple API, though: no-one's going to change their mind about exactly what methods should exist if the only method under discussion is "give me a raw socket, and promise you won't touch it again" -- that's something that servers either can or cannot support, without much room for negotiation or adjustment.

matthewd:
> the Rails team's suggestion to add the semi-optional feature of `has_pending?`

Well, as you know, that's not actually what the suggestion was. But yes, that is part of what I'm looking at when I predict further evolution: the current description of write behaviour seems much too vague for me, even ignoring my deep philosophical objections to the use of `extend`.

boazsegev:
> In fact, lack of gained efficiencies aside, I think this could be implemented entirely by a hijacking middleware. Maybe that would make for a useful reference implementation, and allow freer adoption on the application side?

It's not a bad idea to write a "fallback" middleware... however, I doubt everyone will use the same `hijack` solution during a transitional period.

The gained efficiencies are a big part of the sales pitch. Without them, why not keep the current code? (I can think of numerous reasons, but I doubt anyone would listen)

Also, this proposal is designed to allow direct access to the websocket layer, no middleware necessary.

> As for complexity... I didn't mean difficulty. There is a lot more API here than "call call and pass it a hash".

Putting the `env` hash complexity aside, you are right.

There can definitely be further discussion about what methods could be added.

However, the current specification (putting aside the buffer access feature and the proposed reserved names) follows the javascript specification for callbacks and adds `write` and `close`.

It's pretty much "essentials only".

I know I, for one, would love to add a bit more. But growth is as easy as deprecation is hard.

> the current description of write behaviour seems much too vague for me

I'm assuming you're referring to the question of data being written to the socket or not.

There's an issue many developers ignore when dealing with TCP/IP connections, and it's the fact that connections have more then the two states of open and closed.

This is why I thought it's important to say that having data "written" to the socket doesn't mean it actually reaches the other side.

If you're referring to the String encoding details, I guess they are very specific to the two flavors of websocket messages (text vs. binary messages)... but...

 I don't find the description vague or unclear, so I'm not sure what to do about this. If you have better phrasing suggestions (considering I'm not a native english speaker), I would appreciate your input.

> perhaps just by publishing it somewhere and having suitably-adventurous people implement it, either with an understanding that it could change out from under them, or using a less official env key (e.g. draft.websockets?), or both.

[Done](https://bowild.wordpress.com/2016/07/31/the-dark-side-of-the-rack). Didn't produce meaningful results or interest from other server implementations.

So far, [Iodine](https://github.com/boazsegev/iodine) is the only server implementing the proposed specification (and a few added features).

> It could equally stand alone as a separate specification, which just specifies how to bootstrap over a Rack request. Perhaps with a matching "if you want websockets.." link from Rack, say. (NB, I'm not saying this is what I think should happen... it's just a thought.)

I'm not against a modular approach to the Rack specification, but unless it's under the same roof and has the same integration level as `hijack`, it will be seen as an outsider, which will translate to less servers implementing the specification (or testing it, while it's still in draft stages).

If we take this modular approach, maybe the main specification should only deal with the core (required) `env` details and both `hijack` and `upgrade.websockets?` can be placed in separate "extension" files in the Rack repository, maybe adding extension version support rather then a global specification version.

> even ignoring my deep philosophical objections to the use of `extend`

Yes, please ignore this.

These objections are all valid, I'm sure, but at the end of the day pragmatic considerations should dictate the question of best approach.

There's a reason EventMachine chose this approach and there's a reason why I rewrote Iodine server to adopt `extend`.

In addition, if we ignore the prejudice against `extend`, there's no real reason not to use it.

Everyone knows ActiveSupport monkey patches the core Ruby library when they include the gem. People will as easily expect websocket functionality from their websocket callback object. The use of `extend` won't surprise anyone.

evanphx:
Being an optional protocol, it really is little e extension. The main reason to have it be referenced by Rack is so that implementors know what implementors don't accidentally introduce incompatible but similar apis. There really isn't a huge worry about that given the number of possible implementors, so if it's agreed to upon by most of the implementors, it's effectively a done deal.

The main reason for the discussion is making sure that this is the correct API. Both the implementors and the users have their viewpoints and so we need to get a majority consensus on things. 

On the topic of `write`, I think that what we should say is that `write` has the same delivery promise as `Socket#write`. That's expected by users and isn't generally a problem to implement and explain.

boazsegev:
I updated `write` as suggested by @evanphx, I hope it's better now (we can also get rid of the info in the parenthesis, if it's an overkill). Thanks, Evan.

boazsegev:
Do you guys think it would be more constructive if we placed the proposed extension in a separate file? i.e. `SPEC-Websocket-draft.md`?

Perhaps this would allow PRs from the community and an easier way to discuss the proposal?

matthewd:
My concern is that it doesn't specify whether `write` blocks.

That's a problem for me, because if it does block, I can't use it (https://github.com/rails/rails/pull/26646).

If servers can optionally hard-code `has_pending?` to return false, does that mean they're obliged to maintain an infinite buffer for themselves, or does it mean they're permitted to make `write` block if their buffer is full?

boazsegev:
@matthewd ,

I hope to convince you that this issue should remain in the purview of the server implementation and shouldn't be controlled by the specification.

I'm not sure what @evanphx or @macournoyer think, but it seems to me that the differences between servers allow us to choose the best server for the specific application.

As for the [issue you pointed out](https://github.com/rails/rails/pull/26646), I think this demonstrates why the implementation preference shouldn't leak into the specification.

Sometimes, having threads "block" and the system "lag" is the preferred result (as opposed to high memory usage). Having this "lag" might help load balancing when more than a single process has access to the same listening socket or when using multiple machines.

Other times the lag might be unbearable and might leak into the user experience.

In my opinion (just my 2¢), different servers and setups have different strengths and weaknesses, so that diversity allows us to find the tools that fits our needs.

On a practical note, I feel that an infinite buffer could be impractical for many app developers, since I believe it would be too easy to introduce security vulnerabilities that could crash the machine through memory starvation (vs. DoS against the specific app / service).

Based on my beliefs and priorities, Iodine introduces a limited buffer that is shared among all sockets (~4Mb, maybe could be bigger). Once the buffer is full, sockets are flushed in rotation until a buffer "packet" (~16Kb) becomes available.

I'm quite certain that @evanphx feels different than I and that Puma is based on different priorities than Iodine.

These differences, IMHO are good and allow people to choose the "right tool for the job".

matthewd:
I agree an infinite buffer is problematic. But if the spec allows `write` to block without warning, that seems to introduce some pretty strong architectural constraints on the application side -- if the application uses a small(ish) worker pool, the problem in the linked PR is going to arise.

Yes, that "just" means some applications won't be compatible with some servers... but I don't see much point to having a unified API if servers and applications are each divided into groups, physically API-compatible but semantically mismatched.

---

But I'll skip forward a step: I'm currently just asking you to amend the specification to state what behaviour the server is obliged to provide, and what behaviour the application is permitted to expect, from the `write` method with regard to buffering and/or blocking. If the answer to those is  "implementation-defined", it's still helpful to make that clear, instead of leaving it as an apparent oversight. In the latter case, there's a risk someone will test the behaviour of one server, then assume that holds for others.

Explicitly stating what's not promised allows a potential client author to evaluate that, and decide whether they're happy to call such a method. I'm pretty sure I know what my decision would/will be, but I'm sure other potential users will have different priorities.

boazsegev:
@matthewd , @evanphx , what do you think about adding the following to `write`:

>     A server **should** document whether `write` will block or return immediately. It is recommended that servers implement some type of internal (and possibly limited) buffer, allowing `write` to return immediately when resources allow.

I'm placing [a temporary draft of the extension](https://github.com/boazsegev/iodine/blob/master/SPEC-Websocket-Draft.md) in the Iodine repo, I hope this will make future discussions and changes easier, since this allows us to use PRs to discuss specific changes instead of placing everything in a global thread.

For now I added this paragraph, but please feel free to PR any changes you think are a good fit.

evanphx:
Indicating if `#write` blocks is good and should be done. I think that @matthewd's point that a blocking `#write` is pretty problematic because it locks up the thread is quite important. Given all that, I suspect we need to put in the spec that `#write` can not block, otherwise the behavior will just be too difference between servers and create non-portable apps.

boazsegev:
@evanphx , my only concerns are these:

* Do we really want to excludes a blocking IO design?
* Are we excluding a limited buffer design (where `write` blocks when the internal buffer is full)?

Locking up the thread might be a serious issue, but I'm not sure that's always the case. Maybe I would be more positive towards these restrictions if I knew their full effect (especially when the GIL comes into play)...

...but I can't help thinking that "locking" the thread is mostly an issue when the serve's design has a flaw that is exposed when a thread "blocks".

I can see where this might be an issue where the server's design uses an event loop, when it risks a dead-lock or when the design handles race conditions for the same connection in a way that might cause a chain reaction (all threads are "waiting" or event rescheduling loops recursively)... but I'm not sure how thread blocking is an issue itself. If a server had a thread per IO (ugly in my opinion, but hey, who am I to judge), is blocking really an issue?

I'm not sure how the application waiting on `write` would require a different design, since I'm not sure the server's semantics will effect the application itself.

---

Since, so far, I'm in the minority on this one, do you guys want to put in a PR with the change for [the draft](https://github.com/boazsegev/iodine/blob/master/SPEC-Websocket-Draft.md)?

We can invite everyone to vote on it and merge it into the draft.

It's really not my draft, we're all part of it and I think voting after focused and short discussions in separate detailed PRs will be the most effective way to get input and discuss specific changes such as this one.

tinco:
@evanphx what would be the alternative to blocking? would the `write` throw an exception, or would an exception method (or `on_close`?) on the function be called, or would the data be silently dropped?

I agree that it would be bad if `write` could block.

jeremy:
Blocking/unbuffered writes can be appropriate, too, e.g. to exert backpressure. If we look at prior art for flow control in other WebSockets/streams systems, we could leapfrog these early design concerns, or at least decide what's deferring for later. This is well-trod territory.

Buffered IO is pretty common. Return immediately (appearing "non-blocking") until the buffer's exhausted, then return an error or… block until buffer's available. Leaving these decisions up to the server implementor (barring further survey of existing streams/flow control work) seems a good default.

In any case, thoroughly agreed that the server implementor should be crystal clear about flow control behavior, in every sense. Having the protocol suggest—much less require—that we have infinite concurrency available does nobody any favors!

boazsegev:
I'm not sure how the votes stand and what changes we're moving towards.

If I understand correctly, the current suggestions seems to be:

1. Either, following what I understand to be @jeremy's line of thought:

    > A server **should** document whether `write` will block or return immediately. It is **recommended** that servers implement a buffered IO, allowing `write` to return immediately when resources allow and block (or, possibly, disconnect) when the buffer is full.

    - Vote for this option using 👍 

2. OR, following what I understand to be @matthewd's line of thought:

    > A server **must** implement buffered IO, allowing `write` to return immediately when resources allow and block (or, possibly, disconnect) when the IO buffer is full. A server **should** document `write`'s behavior.

    - Vote for this option using ❤️ 

If you want to suggest a different approach, go for it - especially since I'm not sure if I understand everyone correctly.

I hope as many people as possible vote (@matthewd , @evanphx , @tinco, @jeremy , @macournoyer, @dmitry).

If no one suggests something new then I guess the vote could finalize [this part of the discussion](https://github.com/boazsegev/iodine/blame/c5623cd45d6750f1fb14924a824be14adaa6c4d0/SPEC-Websocket-Draft.md#L39).

jeremy:
My suggestion is to do more research & review before making a lasting decision on a spec. 😊

boazsegev:
According to the only two votes, I'm updating [the temporary draft](https://github.com/boazsegev/iodine/blob/master/SPEC-Websocket-Draft.md) to read (for now):

> A server should document whether write will block or return immediately. It is recommended that servers implement a buffered IO, allowing write to return immediately when resources allow and block (or, possibly, disconnect) when the buffer is full.

In accordance with @jeremy's comment, I've been attempting to research the issue, but I must admit that I couldn't find much information about this at the moment (tons about non-blocking `read`, not much about buffered IO for `write`).

If anyone comes up with interesting data that will help decide the issue (one way or the other), please post a link ( @matthewd , wasn't there a blog post you guys were writing about the blocking `write` issue? ).

tinco:
Are you following [RFC2119](https://www.ietf.org/rfc/rfc2119.txt) in this spec? If so then maybe you could capitalize your SHOULD's and RECOMMEND's to emphasize their standardized meanings. I'll back a SHOULD at least for implementing write buffer. I can imagine one reason for not implementing it would be to give the application developer control over the exact mechanism, which for a general purpose application server should not be the default, so that might be a reason for not making it MUST, though you could make the exception explicit with a "MUST .. unless application explicitly disables it".

boazsegev:
I didn't know about [RFC2219](https://www.ietf.org/rfc/rfc2119.txt), I just used the language I read so often in the different specifications - thanks 👍

As for capital letters: every time someone uses CAPITALS, I feel as if they are shouting, or just didn't get to discover the **bold** thing... even my C macros are usually lowercase (people sometimes get upset with me about this)...

...But I guess you're right, I should probably go over the text and capitalize the terms.

brodock:
are there any work being done in the regards of getting websocket "great again" in ruby?

boazsegev:
Iodine is working properly, as far as I know, using the new specification (although I just got a report demonstrating that it's `each` extension breaks on Linux although it works flawlessly on OS X).

I do wonder how things are progressing for everyone else, but it seems progress is somewhat stunted.

Lately I was exposed to a ["Websocket Shootout"](https://hashrocket.com/blog/posts/websocket-shootout)...

The blog post demonstrated the poor situation in Ruby and proved that some of the issues that the new Rack specification aims at helping resolve are indeed very acute... but I have doubts that the community gives Websockets much of a priority.

I have no idea how much progress was actually made.

Anyone?

boazsegev:
@brodock  - both I and @tinco  posted updates in the current home for the proposal, at [iodine (issue # 10)](https://github.com/boazsegev/iodine/issues/10).

boazsegev:
Any updates? Are we moving forward with this?

Please  [post a comment here](https://github.com/boazsegev/iodine/issues/10) with your status regarding this proposal.

boazsegev:
I'm closing this discussion due to inactivity.

If anyone is promoting a solution that allows separation of concerns between websocket applications and websocket servers,  please let me know.

If you're a server implementor and you're implementing the proposed specification, please feel free to re-open the discussion.

Issue #1111, opened 2016-09-13
Add custom middleware to rack app, remove #reverse requirement


HoneyryderChuck:
Coming from [here](https://github.com/rails/rails/issues/26479), where I describe my experiments trying to reuse a rails app middleware in other rack apps.

In theory it should work, as rack defines (or should define) the expectations on the middleware chain and the app. However, in practice it breaks. Although the Rails middleware is (it seems) not designed to be used outside of the rails app scope, part of the limitations are enforced by rack. 

First, I don't have a way to build an app and pass a middleware (I would have to insert it one by one). In case of such "customized" middlewares like rails, that would be nice. Something like:

```
# in a config.ru from some rails app, after initializing the app....
# ... 
api = Rack::Builder.new do
  use Rack.application.config.middleware
  run App::API
end

This would allow me to set it up. 

The other problem is the way the Rack::Builder handles the middleware chain, as currently it requires it to be an Array, instead of an Enumerable (which the Rails middleware already is). This main culprit is [this call to #reverse](https://github.com/rack/rack/blob/master/lib/rack/builder.rb#L147) and `#<<` but this could be done differently(?). 
```

PR #1112, opened 2016-09-13
Middleware Stack

HoneyryderChuck:
This proposes a basic spec for the middleware chain/stack, which is built within the Rack::Builder class (ivar `@use`). 

A class implementing the middleware stack should implement both `#use`and `#to_app`,where:
- `#use` receives a middleware class, arguments and block;
- `#to_app` receives a rack app and returns the app chained with the middlewares. 

This is an attempt to support #1111 , which proposes to replace the main rack builder app middleware stack with the rails application middleware object, removing the need to support 2 different middleware stacks.

### DIFF
--- a/example/lobster_stack.ru
+++ b/example/lobster_stack.ru
@@ -1,0 +1,9 @@
+require 'rack/lobster'
+
+stack = Rack::MiddlewareStack.new
+stack.use Rack::Chunked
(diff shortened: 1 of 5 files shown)


jeremyevans:
I think middleware stacks are something best handled by frameworks using rack as opposed to rack itself.  I don't think think we want to add this feature to rack.

PR #1114, opened 2016-09-20
Fix handling of parsing empty nested queries

jeremyevans:
If the query string is empty, instead of using a plain hash, we
still need to use params.to_params_hash, as params.to_params_hash
may return a different object, such a hash with a default proc.

This fixes request.params in cases where indifferent query
params are used, and r.GET is empty but r.POST is not.  Before
this commit, this would result in request.params ending up with
a non-indifferent hash, when it should have used an indifferent
hash.

The cause of this issue can originally be traced to an optimization
in c4596b3e3d821d19b25c2d48d619097a3ad39ec0, which was before I
added the code to allow user-configurable query params in
7e7a3890449b5cf5b86929c79373506e5f1909fb.  Now that
user-configurable query params are supported, the optimization is
invalid and needs to be removed.

### DIFF
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -64,13 +64,14 @@ module Rack
     # ParameterTypeError is raised. Users are encouraged to return a 400 in this
     # case.
     def parse_nested_query(qs, d = nil)
-      return {} if qs.nil? || qs.empty?
       params = make_params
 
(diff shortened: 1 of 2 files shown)


jeremyevans:
The Travis failures are only on ruby-head, and appear to relate to the changes in proxy handling in ruby-head, they do not appear to be due to the changes in this commit.

PR #1117, opened 2016-10-08
Fix warnings and usage of $VERBOSE

### LINKED ISSUE #1116: global variable `$verbose' not initialized
When running with `--warnings`, the following shows up in `lib/rack/request.rb`

```
^_^ > irb
2.3.0 :001 > $verbose
 => nil 
2.3.0 :002 > $VERBOSE
 => false 
2.3.0 :003 > defined? $verbose
 => nil 
2.3.0 :004 > defined? $VERBOSE
 => "global-variable" 
```

That's because it should probably be $VERBOSE, not $verbose.

ioquatix:
The correct global variable is called $VERBOSE, not $verbose. Fixed #1116

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -422,7 +422,7 @@ module Rack
 
       # shortcut for <tt>request.params[key]</tt>
       def [](key)
-        if $verbose
+        if $VERBOSE
           warn("Request#[] is deprecated and will be removed in a future version of Rack. Please use request.params[] instead")
         end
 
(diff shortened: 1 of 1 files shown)


matthewd:
cc @sgrif

ioquatix:
Do you think it's possible to do a small point release to include this? It's a pretty noisy problem when running rspec with --warnings

rafaelfranca:
Sure. I can't release rack, but @tenderlove and @spastorino can.

axfcampos:
I would also like a small point release for the same reason @ioquatix mentioned 🙏, running rspec with warnings turned on.

PR #1122, opened 2016-10-19
accept multiple libraries to require at start method

khotta:
accept multiple libraries to require at start method

### DIFF
--- a/lib/rack/server.rb
+++ b/lib/rack/server.rb
@@ -264,7 +264,13 @@ module Rack
       end
 
       if library = options[:require]
-        require library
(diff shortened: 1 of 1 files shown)


mikegee:
@khotta is this still important?

by the way, here's a simpler way to write that code.

```ruby
Array(library).each do |library|
  require library
end
```

jeremyevans:
Accepting multiple `-r` options is how `ruby` and many similar programs work.  This change doesn't get us all the way there.  We also need to fix the options handling so this support is usable from the command line.  I'll work on a PR for that.

PR #1125, opened 2016-10-24
Improve and test SessionHash#fetch

yannvanhalewyn:
### Summary

Implement `Rack::Session::Abstract::SessionHash#fetch` to accept a default argument or a block. 
### Reason

We had some exceptions in production where we'd expect to be working with an object that works like a hash. When calling `session.fetch("foo", "bar")`, we'd get the error:

```
ArgumentError: wrong number of arguments (given 2, expected 1)
File "/local/ruby23/lib/ruby/gems/2.3.0/gems/rack-1.6.4/lib/rack/session/abstract/id.rb" line 59 in []
```

Which was pretty cryptic at first, because we weren't using `[]`. That's when we realised it was aliased.

### DIFF
--- a/test/spec_session_abstract_session_hash.rb
+++ b/test/spec_session_abstract_session_hash.rb
@@ -37,5 +37,9 @@ describe Rack::Session::Abstract::SessionHash do
     it "works with a block" do
       assert_equal :default, hash.fetch(:unkown) { :default }
     end
+
+    it "it raises when fetching unknown keys without defaults" do
+      lambda { hash.fetch(:unknown) }.must_raise KeyError
+    end
   end
 end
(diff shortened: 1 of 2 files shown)


test/spec_session_abstract_session_hash.rb
```diff
@@ -25,4 +25,17 @@ def session_exists?(req)
     assert_equal [:bar, :qux], hash.values
   end
 
+  describe "#fetch" do
+    it "returns value for a matching key" do
```
rafaelfranca: It is missing a test where you don't have a default block or value and the value is unknow

yannvanhalewyn:
Updated!

rafaelfranca:
Thank you.

PR #1127, opened 2016-11-20
Fix Puma + Rack + Rails cannot restart itself

### LINKED ISSUE #1159: Allow starting the server when pid is the same as current process
Puma restarts by replacing itself via `exec`, so they new process has the same pid as the old process, that trips up https://github.com/rack/rack/blob/master/lib/rack/server.rb#L379-L393 which then decides that the server must still be running since the pid is healthy and refuses to boot.

Would you accept a PR that adds a `:dead if old_pid == Process.pid` check ?

see https://github.com/puma/puma/issues/1060

steakknife:
This PR fixes the rack part, puma is almost done with their part.

More detail https://github.com/puma/puma/issues/1117

It's not just Puma, but how Rails configures Rack AND Process.pid doesn't change on restart with Puma servers when it detects `tmp/restart.txt` changes.

Fixes #1159

### DIFF
--- a/lib/rack/server.rb
+++ b/lib/rack/server.rb
@@ -381,6 +381,7 @@ module Rack
 
         pid = ::File.read(options[:pid]).to_i
         return :dead if pid == 0
+        return :this_process if pid == Process.pid
 
         Process.kill(0, pid)
         :running
--- a/test/spec_server.rb
+++ b/test/spec_server.rb
@@ -4,6 +4,7 @@ require 'rack/server'
 require 'tempfile'
 require 'socket'
 require 'open-uri'
+require 'rbconfig'
 
 module Minitest::Spec::DSL
   alias :should :it
@@ -16,6 +17,10 @@ describe Rack::Server do
 
   def app
     lambda { |env| [200, {'Content-Type' => 'text/plain'}, ['success']] }
+  end
+
+  def ruby_exe
+    File.join(RbConfig::CONFIG["bindir"], RbConfig::CONFIG["RUBY_INSTALL_NAME"] + RbConfig::CONFIG["EXEEXT"])
   end
 
   def with_stderr
@@ -141,7 +146,10 @@ describe Rack::Server do
   it "check pid file presence and running process" do
     pidfile = Tempfile.open('pidfile') { |f| f.write($$); break f }.path
     server = Rack::Server.new(:pid => pidfile)
-    server.send(:pidfile_process_status).must_equal :running
+    server.send(:pidfile_process_status).must_equal :this_process
+
+    script = "print Rack::Server.new(:pid => \"#{pidfile}\").send(:pidfile_process_status).inspect"
+    `'#{ruby_exe}' -I./lib -rrack/server -e '#{script}'`.must_equal ":running"
   end
 
   it "check pid file presence and dead process" do


rafaelfranca:
Could you elaborate how the changes you made here are going to fix the linked issue?

steakknife:
When puma is started via `rails s` and restarted via `touch tmp/restart.txt`, the Rails-configured rack pidfile isn't deleted (which is correct) but attempts to recreate it (it always exists). Rack should ignore pidfiles containing `Process.pid` because deleting the pidfile between restarts would introduce a race-condition. It's safer to write and delete pidfile only once each during the entire pid lifecycle.

rafaelfranca:
This make sense to me. WDYT @tenderlove @matthewd?

matthewd:
Looks like we'll need a new test for an actual running process, but otherwise seems fine.

(You can `fork` to get a known-other running process.)

steakknife:
@matthewd `fork` is unimplemented on JRuby and Windows, but a Ruby one-liner does the trick though.

test/spec_server.rb
```diff
@@ -141,7 +141,8 @@ def with_stderr
   it "check pid file presence and running process" do
     pidfile = Tempfile.open('pidfile') { |f| f.write($$); break f }.path
     server = Rack::Server.new(:pid => pidfile)
-    server.send(:pidfile_process_status).must_equal :running
+    server.send(:pidfile_process_status).must_equal :this_process
+    `ruby -rrack/server -e 'print Rack::Server.new(:pid => "#{pidfile}").send(:pidfile_process_status).inspect'`.must_equal ":running"
```
matthewd: We should use the running ruby, not whatever happens to be on the path.

And similarly, better if we can avoid relying on what library [versions] it can find.

Along these lines, we could instead run a `-e` that'll ~ just sit, then do the pid check in the current process.

(I'd personally be fine with a test that just skipped where fork was unavailable, but as long as the complexity doesn't grow much, this is good too.)

steakknife: I dug around stdlib and special variables for an actual C argv[0]. Do you have any idea where it's at?

steakknife: Nm, found it. https://stackoverflow.com/questions/2814077/how-do-i-find-the-ruby-interpreter

steakknife: @matthewd Ready for re-review ( 678790d )

steakknife:
Re-review, por favor.

cc: @tenderlove @matthewd @rafaelfranca

lib/rack/server.rb
```diff
@@ -381,6 +381,7 @@ def pidfile_process_status
 
         pid = ::File.read(options[:pid]).to_i
         return :dead if pid == 0
+        return :this_process if pid == Process.pid
```
grosser: maybe `:self` ?

test/spec_server.rb
```diff
@@ -141,7 +146,10 @@ def with_stderr
   it "check pid file presence and running process" do
     pidfile = Tempfile.open('pidfile') { |f| f.write($$); break f }.path
     server = Rack::Server.new(:pid => pidfile)
-    server.send(:pidfile_process_status).must_equal :running
+    server.send(:pidfile_process_status).must_equal :this_process
+
+    script = "print Rack::Server.new(:pid => \"#{pidfile}\").send(:pidfile_process_status).inspect"
+    `'#{ruby_exe}' -I./lib -rrack/server -e '#{script}'`.must_equal ":running"
```
grosser: maybe simpler to stub the `Process.pid` ?

lib/rack/server.rb
steakknife: It seems like possible bikeshedding, but `:this_process` is semantically-clear whereas `:self` conveys less information, more ambiguously.

test/spec_server.rb
steakknife: It was intended as an integration test to make sure it works end-to-end. Stubbing would turn this check into a less-powerful unit test. If someone wants to refactor this later and make improvements, it's fine.

steakknife:
Closing this since it's never getting merged.

grosser:
@matthewd ok if we just test this unit-style with a stubbed pid ... would love to get this issue solved ... can maybe leave the `how to perfectly test` discussion to another PR ... a pretty obvious bug with an obvious solution ... so I'd love to get this done :(

PR #1131, opened 2016-12-05
Escape URL to render HTML

yous:
DIR_FILE have `<a href='%s'>`, so single quotes in URL should be escaped.

### DIFF
--- a/lib/rack/directory.rb
+++ b/lib/rack/directory.rb
@@ -44,15 +44,14 @@ table { width:100%%; }
     class DirectoryBody < Struct.new(:root, :path, :files)
       def each
         show_path = Rack::Utils.escape_html(path.sub(/^#{root}/, ''))
-        listings = files.map{|f| DIR_FILE % DIR_FILE_escape(*f) } * "\n"
+        listings = files.map{|f| DIR_FILE % DIR_FILE_escape(f) } * "\n"
         page = DIR_PAGE % [ show_path, show_path, listings ]
         page.each_line{|l| yield l }
       end
 
       private
-      # Assumes url is already escaped.
-      def DIR_FILE_escape url, *html
-        [url, *html.map { |e| Utils.escape_html(e) }]
+      def DIR_FILE_escape htmls
+        htmls.map { |e| Utils.escape_html(e) }
       end
     end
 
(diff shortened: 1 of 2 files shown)


yous:
You can simply reproduce this problem.

``` sh
mkdir test
mkdir test/foo\'bar
touch test/foo\'bar/omg\'omg.txt
```

`config.ru`:

``` ruby
run Rack::Directory.new('test')
```

``` sh
rackup config.ru
```

When I click the link to `foo'bar`, it actually goes to `/foo`.

screenshot 2016-12-07 10 09 28

tonytonyjan:
It's caused by the changed behavior of `Rack::Utils#escape_path`:

```ruby
# rack 1.6.5
Rack::Utils.escape_path(?') # => "%27"
```

```ruby
# rack master
Rack::Utils.escape_path(?') # => "'"
```

Why not patch `#escape_path` instead, I wonder.

yous:
That change was came from the patch for #265. Seeing https://github.com/rack/rack/pull/265#issuecomment-4022291, `'` is a valid character for path.

And actually this patch is needed only because `DIR_FILE` uses `<a href='%s'>`. If we change this to `<a href="%s">`, then we don't need `gsub` as `::URI::DEFAULT_PARSER.escape` replaces `"` to `%22`. Maybe changing `DIR_FILE` is more proper solution. How do you think?

tonytonyjan:
I found it's been URI escaped [here](https://github.com/yous/rack/blob/334d5a97572c4ff61b112357f0a9d23ccc59a06d/lib/rack/directory.rb#L117), but not yet HTML escaped.

Since the `#DIR_FILE_escape` will return contents (url, basename, size, type, mtime) which are going to be used to render HTML [here](https://github.com/yous/rack/blob/334d5a97572c4ff61b112357f0a9d23ccc59a06d/lib/rack/directory.rb#L45), they should be all HTML escaped already.

In my opinion, this way can be more decent:

```ruby
def each
  # ...
  listings = files.map{|f| DIR_FILE % DIR_FILE_escape(f) }*"\n"
  # ...
end
# ...
def DIR_FILE_escape(html)
  html.map { |e| Utils.escape_html(e) }
end
```

yous:
Yes, I think that's more proper. I just updated.

yous:
Any update on this? I can reproduce this with rack 2.0.7.

jeremyevans:
This change looks good to me.  CI failures look unrelated.  I'll push the changes as a new PR and if CI passes I'll merge it.

jeremyevans:
Merged in #1524

yous:
Thanks!

PR #1134, opened 2016-12-07
Add built-in coders in `Rack::Session::Cookie::Coder` using decorator pattern.

tonytonyjan:
I saw there has been some built-in coder, such as `Base64::Marshal`, `Base64::JSON`, `Base64::ZipJSON`. However, since they are all built on top of `Rack::Cookie::Base64`, it makes them not flexible, reusable and unable to be used without base64.

With this patch, we can build a complex coder by wrapping multiple coders, for example, to encode with Marshal and Base64:

```ruby
coder = Coder::Base64.new(Coder::Marshal.new, strict: true)
use Session::Cookie, coder: coder
```

Encode with JSON and Zip:

```ruby
coder = Coder::Base64.new(Coder::Marshal.new)
use Session::Cookie, coder: coder
```

# Advice Wanted

1. `Rack::Session::Cookie::Coder` can be used stand-alone, and seems not part of `Rack::Session::Cookie`'s responsibility. I am considering whether I should extract those code to another module, maybe `Rack::Util`?

2. I've also written `Encrypt` and `HMAC` coders, It's very useful when we want a signed, encrypted secure cookie like Rails did, for example:

    ```ruby
    coder = Coder::Base64.new(Coder::Marshal.new)
    coder = Coder::HMAC.new(coder, secret: 'secret', digest: 'SHA1')
    coder = Coder::Encrypt.new(coder, secret: 'secret', cipher: 'AES-256-CBC')
    ```

   However, those coders are not included in this patch. `Rack::Session::Cookie` will automatically sign if the `:secret` option being set. I want to refine this portion in `#find_session` (explicitly, `#unpacked_cookie_data`), and `#write_session`, making this handled by coder. Am I on the right direction?

3. Wrapping coders is a bit tedious for me, I am thinking about DSL design:

      ```ruby
      coder = Coder.build do
        use :marshal
        use :base64
        use :encrypt, secret: 'secret', cipher: 'AES-256-CBC'
        use :hmac, secret: 'secret', digest: 'SHA1'
      end
      ```

   Or it can be method chaining:

      ```ruby
      Coder.marshal.base64.encrypt(options).hmac(options)
      ```

   Any idea?

I am not a native speaker, hope my English is readable enough. :)

### DIFF
--- a/lib/rack.rb
+++ b/lib/rack.rb
@@ -144,4 +144,16 @@ module Rack
     autoload :Pool, "rack/session/pool"
     autoload :Memcache, "rack/session/memcache"
   end
+
+  module Coders
+    autoload :Base64, 'rack/coders/base64'
+    autoload :Cipher, 'rack/coders/cipher'
+    autoload :Coder, 'rack/coders/coder'
+    autoload :HMAC, 'rack/coders/hmac'
+    autoload :Identity, 'rack/coders/identity'
+    autoload :JSON, 'rack/coders/json'
+    autoload :Marshal, 'rack/coders/marshal'
+    autoload :Rescue, 'rack/coders/rescue'
+    autoload :Zip, 'rack/coders/zip'
+  end
 end
--- a/lib/rack/coders/base64.rb
+++ b/lib/rack/coders/base64.rb
@@ -1,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'rack/coders/coder'
+module Rack
+  module Coders
+    class Base64 < Coder # :nodoc:
+      def initialize(coder = nil, strict: true)
+        super(coder)
+        @template_str = strict ? 'm0' : 'm'
+      end
+
+      def encode(obj)
+        [coder.encode(obj)].pack(@template_str)
+      end
+
+      def decode(str)
+        coder.decode(str.unpack(@template_str).first)
+      end
+    end
+  end
+end
--- a/lib/rack/coders/cipher.rb
+++ b/lib/rack/coders/cipher.rb
@@ -1,0 +1,45 @@
+# frozen_string_literal: true
+
+require 'rack/coders/coder'
+require 'openssl'
+
+module Rack
+  module Coders
+    # Encrypt the data into this format:
+    #
+    #     "#{encrypted_data}--#{initial_vector}"
+    #
+    class Cipher < Coder
+      def initialize(coder = nil, secret:, old_secret: nil, cipher: 'AES-256-CBC')
+        super(coder)
+        @secret = secret
+        @old_secret = old_secret
+        @cipher = OpenSSL::Cipher.new(cipher)
+      end
+
+      def encode(obj)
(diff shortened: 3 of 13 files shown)


tonytonyjan:
And I've released a gem for this issue: https://github.com/tonytonyjan/coder_decorator just in case if this pull request won't be accepted.

rafaelfranca:
Thank you for the patch. I like the idea of composing the coders using decorators. It seems you also take care of backward compatibility here.

A think that even `Coder` is not responsibility of `Rack::Session::Cookie` they are made to be used with cookie, so I'd not move it out of the constant.

> I want to refine this portion in #find_session (explicitly, #unpacked_cookie_data), and #write_session, making this handled by coder. Am I on the right direction?

Yeah, it make sense to leave the coder to deal with it. But it may break backward compatibility. Since rack master is rack 3 now, I think it is fine.

> Wrapping coders is a bit tedious for me, I am thinking about DSL design:

You usually only build the coder once so I think a DSL may be overkill here.

@tenderlove opinions about this?

Kingdonie:
Man f off

tenderlove:
I agree the DSL is a little overkill, but I do like this idea.  I also agree about moving the constant outside, but I'm not sure where we should put it.

+1 on the overall idea, but we need somewhere to put the coder class.

tonytonyjan:
Sorry for the delayed response. Thanks for approving of this idea, I am considering to put them under `Rack::Coders`, for example:

```ruby
module Rack::Coders
  class Coder
    def encode(obj); raise NotImplementedError; end
    def decode(obj); raise NotImplementedError; end
  end

  class Marshal < Coder
  end

  class Base64 < Coder
  end
end
```

It can be a better place in my opinion.

tonytonyjan:
Just updated this pull request with 2 commits. The first one is to add some useful coders under `Rack::Coders`:

* `Rack::Coders::Base64`
* `Rack::Coders::Cipher`
* `Rack::Coders::Coder`
* `Rack::Coders::HMAC`
* `Rack::Coders::Identity`
* `Rack::Coders::JSON`
* `Rack::Coders::Marshal`
* `Rack::Coders::Rescue`
* `Rack::Coders::Zip`

Another commit is to apply those coders to `Rack::Session::Cookie` without breaking compatibility.

As to using coders to deal with signature and encryption, I wonder if we need to remove `:secret`, `:old_secrets` and `:let_coder_handle_secure_encoding` from options. Actually, it can still be implemented with backward compatibility, but I don't know if it is worth.

ioquatix:
I understand the implementation, but what's the use-case? How many users decided to change the encoding of cookies?

ioquatix:
I am going to close this as out of scope. That doesn't mean it's not useful. I checked your gem. Great work. However, we already provide the interface for users to swap in their own preference for coder. I don't think a full array of coders belongs in the rack gem, we provide some basic ones by default, which is sufficient for most use cases.

PR #1136, opened 2016-12-08
Add middleware type - middleware can also be proc.

cavinkwon:
By default, the Rack middleware must be a class. Proc-style middleware provides simplicity when someone wants to learn Rack middleware. So I added some code.

Someone can write:
```ruby
require 'rack'

upcase_middleware = proc do |app,env|
  st, hd, body = app.call(env)
  [st, hd, body.map(&:upcase)]
end

app = Rack::Builder.app do
  use upcase_middleware
  run lambda { |env| [200, {"Content-Type" => "text/plain"}, ['Hi Boss']] }
end

Rack::Handler::WEBrick.run app
```

### DIFF
--- a/lib/rack/builder.rb
+++ b/lib/rack/builder.rb
@@ -83,7 +83,13 @@ module Rack
         mapping, @map = @map, nil
         @use << proc { |app| generate_map app, mapping }
       end
-      @use << proc { |app| middleware.new(app, *args, &block) }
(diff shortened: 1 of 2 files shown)


rafaelfranca:
Thank you for the pull request but I prefer to the API to be as simpler as possible and adding support to a different kind of proc (that accept the application as argument) will just make the API more complex.

PR #1144, opened 2017-02-19
Use .httpdate for expires date formatting

### LINKED ISSUE #1100: Cookie expires format is invalid (take 2)

nanaya:
Specified in [RFC 7231, section 7.1.1.2: Date](https://tools.ietf.org/html/rfc7231#section-7.1.1.2) (which obsoletes RFC 2616 though no change in date format).

Different take on #1105 and should fix #1100.

Quite sure will break rails tests ([example](/rails/rails/blob/e51572dce00c5f011f019397984ed8e82d94fc84/actionpack/test/dispatch/cookies_test.rb#L323), and various other places).

I also noticed that at least `Rack::Utils.rfc2822` and `Rack::Utils.rfc2109` aren't used anywhere but as they might be used by external libraries/apps, I didn't touch it.

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -221,31 +221,7 @@ module Rack
         domain  = "; domain=#{value[:domain]}"   if value[:domain]
         path    = "; path=#{value[:path]}"       if value[:path]
         max_age = "; max-age=#{value[:max_age]}" if value[:max_age]
-        # There is an RFC mess in the area of date formatting for Cookies. Not
-        # only are there contradicting RFCs and examples within RFC text, but
-        # there are also numerous conflicting names of fields and partially
-        # cross-applicable specifications.
-        #
-        # These are best described in RFC 2616 3.3.1. This RFC text also
-        # specifies that RFC 822 as updated by RFC 1123 is preferred. That is a
-        # fixed length format with space-date delimited fields.
-        #
-        # See also RFC 1123 section 5.2.14.
-        #
-        # RFC 6265 also specifies "sane-cookie-date" as RFC 1123 date, defined
-        # in RFC 2616 3.3.1. RFC 6265 also gives examples that clearly denote
-        # the space delimited format. These formats are compliant with RFC 2822.
-        #
-        # For reference, all involved RFCs are:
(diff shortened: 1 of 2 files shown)


rafaelfranca:
Yeah, this seems to break a lot of things. @matthewd @tenderlove WDYT?

tenderlove:
Seems like a good fix for the bug.  @rafaelfranca what type of stuff is breaking?  If it's breaking too much stuff, maybe we should put this in a Y release (`X.Y.Z`).

rafaelfranca:
I mean, there are a lot of tests that check the entire string and it is expecting that format. Maybe there are also some code that expect that format. But agree we should put it in a X format.

tenderlove:
> But agree we should put it in a X format.

You mean a major release, or this string format?

rafaelfranca:
Sorry, I edited the comment later. I mean Y release.

tenderlove:
@rafaelfranca cool.  Do you want to merge this and do the branch dance, or shall I?

rafaelfranca:
I can do it. I'm already with the hands dirty

tenderlove:
@rafaelfranca lol, thank you!

tremby:
There still hasn't been a release with this fix in it. It looks like master hasn't been merged into the 2.x or 1.x line in over four years!

Issue #1010, opened 2016-02-03
Middleware should be frozen by default


ioquatix:
This is a high level conceptual issue, but it seems to me that after being constructed, middleware should be frozen to avoid potential issues with handling concurrent requests.

Perhaps this could be some kind of option in the `config.ru` file to avoid backwards compatibility issues. However, a default to freeze middleware seems like the only sane option based on my experience.

ioquatix:
just to touch on this again, it could be as simple as calling `freeze` on the top level app with the expectation that it will propagate the call to descendant @app instances. It would be nice if this was easy to do in the context of a config.ru file.

ioquatix:
I've thought about this issue some more.

In `Rack::Builder#to_app` it would be easy to call `freeze` on the top level app.

Provided that apps in the chain implement freeze something like:

```ruby
class MyMiddleware
    def freeze
        @app.freeze
        super
    end
...
```

It's possible to even invoke this in warmup:

```ruby
warmup do |app|
	# Freeze all the middleware so that mutation bugs are detected.
	app.freeze
end
```

But each middleware needs to implement it correctly, e.g. `URLMap` needs to freeze each one of it's url specific apps.

ioquatix:
So, in the mean time, I've put together https://github.com/ioquatix/rack-freeze which can be used to retro-fit this policy on existing middleware. It's an acceptable solution, but I think that middleware should generally adopt this policy.

ioquatix:
I'm going to close this issue as I believe the merged PR is suitable. However, if you prefer to impose this policy to your entire application by default, `rack-freeze` is still a good option and I'll continue to use it and maintain it.

PR #1149, opened 2017-03-14
Add Builder#freeze_app to freeze application and all middleware instances

jeremyevans:
Freezing the application and middleware instances can prevent
thread-safety issues at runtime.  This addresses the issues
discussed in #1010.

Unlike the solution proposed by #1010, this is backwards compatible
as it is opt-in, the application and middleware instances are
only frozen if you choose to use freeze_app.

### DIFF
--- a/lib/rack/builder.rb
+++ b/lib/rack/builder.rb
@@ -51,7 +51,7 @@ module Rack
     end
 
     def initialize(default_app = nil, &block)
-      @use, @map, @run, @warmup = [], nil, default_app, nil
+      @use, @map, @run, @warmup, @freeze_app = [], nil, default_app, nil, false
       instance_eval(&block) if block_given?
     end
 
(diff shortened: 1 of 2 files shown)


PikachuEXE:
What should I do if plan to use this method instead of [`rack-freeze`] in a rails app?

[`rack-freeze`]: https://github.com/ioquatix/rack-freeze

jeremyevans:
@PikachuEXE Probably something better asked on a Rails forum.  I'm not sure if Rails uses Rack::Builder for building the middleware chain, I believe it uses something custom.   A code search for "Rack::Builder" doesn't bring up much: https://github.com/rails/rails/search?utf8=%E2%9C%93&q=%22Rack%3A%3ABuilder%22&type=

ioquatix:
I've been thinking about this issue today, in the context of https://www.youtube.com/watch?v=fVIVY7INWWQ and I think that opt-in is as good as broken by default. I'm sure it's not easy to make major changes, but perhaps for the next major release, it should be the default. If people want to opt out, they should explicitly add

```
def freeze
end
```

to their middleware. I know it's a tough call since it's going to break things but in a sense they are already broken, we are just exposing the madness.

PR #1153, opened 2017-04-09
Avoid overwriting response body when it's already overridden

iamvery:
It appears that the intention of #570 was to provide a seam on
Rack::File where the response body may be provided by an overriding
class (see their markdown example). However e0ac32 overwrites the value
of response[2] (the body) after this is set.

This is the simplest possible fix, ignore the step of overwriting if a
value is already set. However it's worth noting that the design of the
entire method would probably be served well to improve. At this point, I
don't want to take that on.

Also of note, there may very well be a latent bug in the previous
implementation. Specifically, when response_body is overridden, it
affects the calculation of filesize, therefore if the method is
overridden the filesize of that overridden response_body would be
returned which is probably not the correct size for the response read
from the path. If that is in fact a defect, this PR may very well
address it to, though it is unverified.

### DIFF
--- a/lib/rack/files.rb
+++ b/lib/rack/files.rb
@@ -99,7 +99,7 @@ module Rack
       response[2] = [response_body] unless response_body.nil?
 
       response[1][CONTENT_LENGTH] = size.to_s
-      response[2] = make_body request, path, range
+      response[2] = make_body request, path, range unless response[2]
       response
     end
 
--- a/test/spec_files.rb
+++ b/test/spec_files.rb
@@ -261,6 +261,7 @@ describe Rack::Files do
     res  = Rack::MockRequest.new(file).get("/cgi/test")
 
     res.must_be :ok?
+    res.body.must_equal "hello world"
   end
 
 end


ioquatix:
Can you rebase on latest master?

iamvery:
>Can you rebase on latest master?

Yes, done. However, this is a very old PR, and I don't recall the full intention at this point. It seems well-meaning, so let me know if you're unable to get to the root of the issue from the original comment.

Thanks for following up!

jeremyevans:
The current code definitely has an issue, because the `response[2] = [response_body] unless response_body.nil?` is now pointless.  

We may be better off removing the `response_body` feature, which was added in #570.  The feature does not handle range requests correctly.  It was added with no specs for the behavior. The only spec that uses `response_body` was added in response to the removal of `Rack::Utils.bytesize` in b2e8b0fab8b49030a342074227820f3f6cc0d644.  We could remove it, and then add a `method_added` hook so that if someone else is using the feature, we can warn them (maybe even raise an exception).

iamvery:
Thanks for the insightful reply, @jeremyevans. Obviously this PR is a bit long in the tooth, but I'll keep a task around for myself to circle back one day when I get some free time (unless it has been addressed before then). Feel free to take over (or delegate) if it's important to complete in the meantime.

ioquatix:
Closing in favour of https://github.com/rack/rack/pull/1559. Thanks for your effort!

PR #1156, opened 2017-04-17
Dynamic cookie key for session ID

### LINKED ISSUE #1155: Dynamic cookie key for session stores
The abstract session store implementation ([lib/rack/session/abstract/id.rb](https://github.com/rack/rack/blob/master/lib/rack/session/abstract/id.rb)) uses an instance variable, initialized at startup, to determine the name of the cookie it will use to fetch/set a session's id.

I recently ran into a scenario where it would be convenient to have the option to make this cookie name dynamic. For example, a user could visit my website at `www.example.com` and have one session, then they could visit `www.example.com/admin` and have a completely different one. I could also see instances where the session could change dynamically based on cookies, headers, IP address, or any other information the webserver has.

Since these cases are very rare, and may not be best practice, I wanted to get some feedback on this feature request. If there's any desire for such functionality, I'll happily submit a PR.

I'll post one example of how this feature request could be addressed. Please let me know if this is something desirable in the rack mainline.

j1mmie:
Allows custom session stores that extend the abstract session store to
dynamically select which cookie key will be used for fetching/setting
the session's id. This allows the user to maintain multiple, different sessions.
The web application can determine which session to be used for the
current request based on any number of conditions (URL, headers, cookies, IP,
etc).

Fixes #1155

### DIFF
--- a/lib/rack/session/abstract/id.rb
+++ b/lib/rack/session/abstract/id.rb
@@ -284,8 +284,9 @@ module Rack
         # Extract session id from request object.
 
         def extract_session_id(request)
-          sid = request.cookies[@key]
-          sid ||= request.params[@key] unless @cookie_only
+          cookie_key = key
+          sid = request.cookies[cookie_key]
+          sid ||= request.params[cookie_key] unless @cookie_only
           sid
         end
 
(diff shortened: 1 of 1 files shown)


j1mmie:
Closing - branch name and commit message referenced the wrong issue!

PR #1157, opened 2017-04-17
Dynamic cookie key for session ID

### LINKED ISSUE #1155: Dynamic cookie key for session stores

j1mmie:
Allows custom session stores that extend the abstract session store to
dynamically select which cookie key will be used for fetching/setting
the session's id. This allows the user to maintain multiple, different sessions.
The web application can determine which session to be used for the
current request based on any number of conditions (URL, headers, cookies, IP,
etc).

Fixes #1155

No tests are written for this change yet. Please let me know if this
functionality is desired in rack, or similar functionality with a different
implementation, and I'll happily add tests and seek review. Thanks!

### DIFF
--- a/lib/rack/session/abstract/id.rb
+++ b/lib/rack/session/abstract/id.rb
@@ -235,6 +235,10 @@ module Rack
           [status, headers, body]
         end
 
+        def cookie_key(request)
+          @key
+        end
+
         private
 
         def make_request(env)
@@ -284,8 +288,9 @@ module Rack
         # Extract session id from request object.
 
         def extract_session_id(request)
-          sid = request.cookies[@key]
-          sid ||= request.params[@key] unless @cookie_only
+          key = cookie_key(request)
+          sid = request.cookies[key]
+          sid ||= request.params[key] unless @cookie_only
           sid
         end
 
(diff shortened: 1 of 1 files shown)


rafaelfranca:
I like the idea. @tenderlove @eileencodes thoughts?

tenderlove:
I'm fine with this. Needs tests and docs though. The method might get
deleted without them!

rafaelfranca:
@jeremy what you are describing is basically what is possible today https://github.com/rack/rack/issues/1155#issuecomment-301169716.

I also think it is better, but the current API is not good in my opinion. It know too much about the rack implementation.

jeremy:
Using "well-known" Rack env keys is the normal, public way of interacting with Rack components. It's not like peeking underneath a public interface at private implementation details—it _is_ the public interface. So, given the Rack API style, providing runtime config in this way seems as natural as we can hope.

j1mmie:
Oops didn't mean to close this. Reopened.

@jeremy @rafaelfranca Can either of you confirm that @hiteshsharma's suggestion works in the latest version of Rack? I haven't had a chance to test, but in my version this did not work.

I agree with @jeremy's suggestion, the cookie name should be decided *very* early on, possibly at the Rack level.

jeremyevans:
As `env['rack.session.options']` already exists and is supported, I plan to close this unless another core team member objects.

PR #1160, opened 2017-04-21
Don't fall back to REMOTE_ADDR unless X_FORWARDED_FOR is empty

KaptajnKold:
The previous, faulty behaviour was to remove all trusted ip addresses
from X_FORWARDED_FOR and returning the last one remaining, falling back
to the REMOTE_ADDR if none remain.

A problem arises if all the address in X_FORWARDED_FOR are trusted.
They can't all be proxies. One of them (the first one) is the ip that
made the request. The correct behaviour is to return the first one, but
the previous implementation would remove them all, and fall back to
REMOTE_ADDR.

This patch removes trusted ips only from the tail of X_FORWARDED_FOR,
and returns the last address in the resulting list, ensuring that we
don't fall back to REMOTE_ADDR if X_FORWARDED_FOR is non-empty.

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -261,7 +261,9 @@ module Rack
 
         forwarded_ips = split_ip_addresses(get_header('HTTP_X_FORWARDED_FOR'))
 
-        return reject_trusted_ip_addresses(forwarded_ips).last || get_header("REMOTE_ADDR")
+        return get_header("REMOTE_ADDR") if forwarded_ips.empty?
+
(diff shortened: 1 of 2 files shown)


timon:
Hi @KaptajnKold 
I'm experiencing the same issue, thanks for the fix!

SamSaffron:
Almost the same PR was also submitted by me 

https://github.com/rack/rack/pull/1224

Such a shame this is in the pipeline for so long

mikegee:
@KaptajnKold does #1224 resolve the issue for you too?

KaptajnKold:
@mikegee yes, it does.

PR #1162, opened 2017-04-24
Added (Unused) 306 HTTP Status Code

snood1205:
Per RFC 7231 Section 6.4.6, I added Status Code 306. While it is unused, it is a reserved status code and unlike the other status codes that are not in this document, this was explicitly noted as unused and not unassigned making it a possible safeguard against people trying to use it for custom status codes (much like nginx does with 499 and 599).

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -509,6 +509,7 @@ module Rack
       303 => 'See Other',
       304 => 'Not Modified',
       305 => 'Use Proxy',
+      306 => '(Unused)',
       307 => 'Temporary Redirect',
       308 => 'Permanent Redirect',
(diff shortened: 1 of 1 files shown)


matthewd:
Hi!

I don't see what we gain by adding this. Could you expand on why you think we should?

snood1205:
@matthewd the thought was that because it was denoted differently in the RFC from all of those that are unassigned, and that 306 is the only (unused) status code, that it might be useful to have there.

ioquatix:
Checked RFC, it's valid, and it's marked as literally "(Unused)" in exactly the same way all the other status codes are labeled, so for consistency and the other reasons stated above, merged.

PR #1165, opened 2017-05-11
Allow specifying array index in #parse_query

odedniv:
Once it's been established that a certain param is an array, allow referencing parts of the array with an index. Some of the new tests show outputs that would otherwise be impossible, the most prominent is an array of different types and an array of objects with different keys.

Note that this means that the first element of an array must still start with a `[]`, since `[0]` is ambiguous for being an object with `"0"` key.

Also removed the seemingly unnecessary check `after =~ %r(^\[\]\[([^\[\]]+)\]$)`, maybe someone could shed some light on that check - this is obviously not the main feature in this PR.

### DIFF
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -12,6 +12,11 @@ module Rack
     # sequence.
     class InvalidParameterError < ArgumentError; end
 
+    # MissingParameterError is the error that is raised when incoming structural
+    # parameters (parsed by parse_nested_query) contain missing elements, such
+    # as skipping an array index.
+    class MissingParameterError < IndexError; end
+
     def self.make_default(key_space_limit, param_depth_limit)
       new Params, key_space_limit, param_depth_limit
     end
@@ -98,7 +103,11 @@ module Rack
         params[k] ||= []
         raise ParameterTypeError, "expected Array (got #{params[k].class.name}) for param `#{k}'" unless params[k].is_a?(Array)
         params[k] << v
-      elsif after =~ %r(^\[\]\[([^\[\]]+)\]$) || after =~ %r(^\[\](.+)$)
+      elsif params[k].is_a?(Array) && after =~ %r(^\[(\d+)\]$)
+        index = $1.to_i
+        raise MissingParameterError, "skipped Array index (got #{index} while current length is #{params[k].length}) for param `#{k}'" if index > params[k].length
+        params[k][index] = v
+      elsif after =~ %r(^\[\](.+)$)
         child_key = $1
         params[k] ||= []
         raise ParameterTypeError, "expected Array (got #{params[k].class.name}) for param `#{k}'" unless params[k].is_a?(Array)
@@ -107,6 +116,12 @@ module Rack
         else
           params[k] << normalize_params(make_params, child_key, v, depth - 1)
         end
+      elsif params[k].is_a?(Array) && after =~ %r(^\[(\d)+\](.+)$)
+        index = $1.to_i
+        raise MissingParameterError, "skipped Array index (got #{index} while current length is #{params[k].length}) for param `#{k}'" if index > params[k].length
+        child_key = $2
+        params[k][index] ||= make_params
+        normalize_params(params[k][index], child_key, v, depth - 1)
       else
         params[k] ||= make_params
         raise ParameterTypeError, "expected Hash (got #{params[k].class.name}) for param `#{k}'" unless params_hash_type?(params[k])
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -13,6 +13,7 @@ module Rack
   module Utils
     ParameterTypeError = QueryParser::ParameterTypeError
     InvalidParameterError = QueryParser::InvalidParameterError
+    MissingParameterError = QueryParser::MissingParameterError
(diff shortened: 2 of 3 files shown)


tenderlove:
Hi,

I like the idea of specifying array indices, but the syntax seems unintuitive.  What if we started with `1` rather than `0`?  0 makes it seem like the first element was actually the `-1`th element.

tenderlove:
Another problem is that Ruby arrays aren't sparse, so someone could do `x[][x]=x&x[9999999][x]=x` and gobble up someone's memory very easily.

odedniv:
@tenderlove fixed the leak, not sure I understand what you said about starting with `1` rather than `0`.

The reason you have `0` is to modify the already existing element in the array, so `x[][a]=1&x[0][b]=2` is parsed as `{ "x": [{ "a": 1, "b": 2 }] }` rather than `{ "x": [{ "a": 1 }, "b": 2 }] }` which before this fix had to be a wild guess according to existing keys in the last array element.

tenderlove:
> The reason you have 0 is to modify the already existing element in the array, so x[][a]=1&x[0][b]=2 is parsed as { "x": [{ "a": 1, "b": 2 }] } rather than { "x": [{ "a": 1 }, "b": 2 }] } which before this fix had to be a wild guess according to existing keys in the last array element.

ahhh, gotcha.  I'm sorry, I totally misread what you were proposing.  Ya, I think this makes sense.

> fixed the leak

Just to confirm my understanding, this fix requires that all parameters are filled in.  IOW, you can't skip parameter indices.

odedniv:
> ahhh, gotcha. I'm sorry, I totally misread what you were proposing. Ya, I think this makes sense.

Before pulling this I suggest taking a look at the tests I added to see how far the rabbit hole goes (what new cases are meant to be supported), just to make sure it all makes sense. Sorry I didn't write examples in the PR description, figured the tests are the best place to view that.

> Just to confirm my understanding, this fix requires that all parameters are filled in. IOW, you can't skip parameter indices.

That is correct, skipping an array index would cause an exception similar to having an invalid argument. It seemed to fit the exception's documentation so I reused it, should I create a different class? (I don't see how this should ever be differentiated though)

Also, negative indexes (which is what I thought you meant at first) are not a problem (I used `>` so I thought they might), since the regex forces only digits. We could add the option for negative indexes with `\[(\-?\d+)\]` (that would work like ruby arrays do) but I think that's a little too confusing.

lib/rack/query_parser.rb
```diff
@@ -98,7 +98,11 @@ def normalize_params(params, name, v, depth)
         params[k] ||= []
         raise ParameterTypeError, "expected Array (got #{params[k].class.name}) for param `#{k}'" unless params[k].is_a?(Array)
         params[k] << v
-      elsif after =~ %r(^\[\]\[([^\[\]]+)\]$) || after =~ %r(^\[\](.+)$)
+      elsif params[k].is_a?(Array) && after =~ %r(^\[(\d+)\]$)
+        index = $1.to_i
+        raise InvalidParameterError, "skipped Array index (got #{index} while current length is #{params[k].length}) for param `#{k}'" if index > params[k].length
```
eileencodes: I think you should make a new error class. It's close but it doesn't contain "invalid" parameters, rather it's missing parameters (ie you're asking for index 5 when index 5 doesn't exist).

odedniv: Done.

At first I inherited from `ArgumentError` which seemed to make sense but then I figured maybe `IndexError` makes more?

There was also a problem with using `ArgumentError`, it's too broadly caught. Shouldn't it be caught at the right times? (the current tests imply catching in `#unescape` and when matching `name` in `#normalize_params`).

odedniv:
Done.

At first I inherited from `ArgumentError` which seemed to make sense but then I figured maybe `IndexError` makes more?

There was also a problem with using `ArgumentError`, it's too broadly caught. Shouldn't it be caught at the right times? (the current tests imply catching in `#unescape` and when matching `name` in `#normalize_params`).

odedniv:
Wow Jeremy, you are right. This is really not necessary and I could implement my API with a very simple format (not JSON though, that would look real bad).

I believe this PR does no harm, it's completely backward compatible and could help someone some day, but I don't mind if you don't pull it.

I'll tell you the story just to let you in on the fun:

I have a restful API. Some resources can include other resources when fetched, but I don't want them to *always* be included because some specific case needs them, so I implemented an inclusion mechanism with an `include` query parameter.

Now this inclusion can be complex, like so: get user with its comments and posts, and bring the picture for each post. The include parameter should *mean* something like that: `/api/users?include=[:comments, { posts: :picture }]` (the query parameter in Ruby for simplicity). In this case for example this PR is needed.

I'm going to change that parameter to a simple comma+colon separated values, which would make it much simpler on the eye (and on my tests!) like so: `/api/users?include=comments,posts:picture`

Thanks for your advice!

jeremyevans:
As @jeremy stated:

> Introducing array indexing is asking for massive confusion + ambiguity with Hash indexing, though.

For that reason alone I am against this.  If you want an array of hashes, I think it's better to use the hash syntax and convert the hash to an array after Rack parses it, as that is unambiguous.  Unless there are objections from the core team, I plan on closing this.

PR #1169, opened 2017-05-16
Fix mistake in encoding change

### LINKED ISSUE #1168: 1.6.7 release mock endpoint broken on ruby 1.9.3
Hi,
[This commit](https://github.com/rack/rack/commit/c7ee52f6ba4684ce917b97c1ff4c7c523760a9d2), which appears to have been released in version 1.6.7, invokes the ``String#b`` method which is not present in Ruby 1.9.3 (https://ruby-doc.org/core-1.9.3/String.html).

It looks like the code attempts to protect against this exact failure by [checking](https://github.com/rack/rack/commit/c7ee52f6ba4684ce917b97c1ff4c7c523760a9d2#diff-4fc9bc1f7d91630f4f9f47fc6663f3f7R146) for the ``:encoding`` method, but Ruby 1.9.3 does indeed [respond to this method](https://ruby-doc.org/core-1.9.3/String.html#method-i-encoding), and so it proceeds to fail on the ``.b`` invocation.  (Ruby 1.8 has neither the ``:encoding`` or ``:b`` methods.)

```
Failures:

  1) Sinatra API should serve status endpoint
     Failure/Error: get '/status'
     
     NoMethodError:
       undefined method `b' for "GET":String
     # /Users/derek/.rvm/gems/ruby-1.9.3-p374/gems/rack-1.6.7/lib/rack/mock.rb:148:in `env_with_encoding'
     # /Users/derek/.rvm/gems/ruby-1.9.3-p374/gems/rack-1.6.7/lib/rack/mock.rb:94:in `env_for'
     # /Users/derek/.rvm/gems/ruby-1.9.3-p374/gems/rack-test-0.6.3/lib/rack/test.rb:237:in `env_for'
     # /Users/derek/.rvm/gems/ruby-1.9.3-p374/gems/rack-test-0.6.3/lib/rack/test.rb:57:in `get'
     # ./spec/api_spec.rb:31:in `block (2 levels) in <top (required)>'
```

eileencodes:
1) I wasn't even testing my change on the branches I made changes on
because I accidentally used `unless` instead of `if`. This test should
only run IF encoding them to binary is supported. :flushed:

2) 1.9.3 does respond to `:encoding` but it doesn't respond to `:b` and
since it's `:b` we're calling on these env vars it's best to ask if it
responds to that method, not to `:encoding`.

Fixes #1168

### DIFF
--- a/lib/rack/mock.rb
+++ b/lib/rack/mock.rb
@@ -143,7 +143,7 @@ module Rack
       env
     end
 
-    if "<3".respond_to? :encoding
+    if "<3".respond_to? :b
       def self.env_with_encoding(env, opts, uri)
         env[REQUEST_METHOD] = (opts[:method] ? opts[:method].to_s.upcase : "GET").b
         env["SERVER_NAME"] = (uri.host || "example.org").b
--- a/test/spec_mock.rb
+++ b/test/spec_mock.rb
@@ -212,7 +212,7 @@ describe Rack::MockRequest do
     called.should.equal true
   end
 
-  unless "<3".respond_to? :encoding
+  if "<3".respond_to? :b
     should "defaults encoding to ASCII 8BIT" do
       req = Rack::MockRequest.env_for("/foo")
 


eileencodes:
cc/ @tenderlove @rafaelfranca

rafaelfranca:
The build is still broken on 1.8.7. Not sure if related with your change though

eileencodes:
1.8.7 tests are failing because of an earlier backport Aaron and I did. Will merge this and then fix the other one (https://github.com/rack/rack/compare/2e6313c292ca...c0598cd42b24) before releasing.

Issue #1171, opened 2017-05-19
when is line 4 in test/spec_cgi.rb executed?


radekosmulski:
**newb question - please delete if this is inappropriate / is wasting your time**

I think I have everything installed that I need, helper.rb starts lighttpd no problem and assigns pid to LIGHTTPD_PID, but how can `if defined? LIGHTTPD_PID` ever evaluate to true (line 4 in test/spec_cgi.rb)? Do the specs need to be run with some special options / via some other means than via rake / rspec from command line? 

BTW really enjoying reading the gem, is there anything a newb could do to help around here?

Thx for your time and sorry to bother

tenderlove:
Hi @radekosmulski this is a great question.  I don't think the `if defined? LIGHTTPD_PID` can ever evaluate to true, so this seems to be a bug.

> BTW really enjoying reading the gem, is there anything a newb could do to help around here?

Can you fix this bug?  I think you just need to change it to `Rack::TestCase::LIGHTTPD_PID`.  If you need any help with this, please let me know!

> Thx for your time and sorry to bother

No problem.  Thank you for filing this ticket!

radekosmulski:
Done :)

https://github.com/rack/rack/pull/1172

ioquatix:
Merged. Thanks for your effort!

PR #1172, opened 2017-05-25
Fix cgi specs

radekosmulski:
Correctly rerefernce LIGHTTPD_PID from spec_cgi.rb and spec_fastcgi.rb
allowing specs to run.

See https://github.com/rack/rack/issues/1171


ioquatix:
Thanks for your contribution to Rack.

PR #1173, opened 2017-06-04
Use slice! to extract session cookie digest

mjc-gh:
For more details on this change see the discussions below:

- https://github.com/rack/rack/commit/734a00c5f4bb46e9a5e6e2677d89a2f285dcc185
- https://gist.github.com/mikeycgto/0339162c97ffd4257d674c0bfb71e37f

### DIFF
--- a/lib/rack/session/cookie.rb
+++ b/lib/rack/session/cookie.rb
@@ -107,6 +107,11 @@ module Rack
         @secrets = options.values_at(:secret, :old_secret).compact
         @hmac = options.fetch(:hmac, OpenSSL::Digest::SHA1)
 
+        # Multiply the :digest_length: by 2 because this value is the length of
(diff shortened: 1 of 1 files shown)


mikeycgto:
This comes from my in progress work for adding AES GCM session cookies to Rack and Sinatra. See issue sinatra/sinatra/issues/1300 for more details.

mikeycgto:
Did a little more research into the history of this. Seems the original commit which moved away from a simple split was commit ae360063. This appears to be done to allow for any arbitrary "coder" which many permit double dashes in it's output. While no coder included in rack by default would allow such output, it's probably best to theoretically support.

jeremyevans:
I don't think this is useful enough by itself, so I'm going to close this.  We may consider changing the format for session cookies in the future (with the ability to upgrade older cookies transparently).

PR #1175, opened 2017-06-05
Use "strict encoding" for Base64 encoded session cookies

mjc-gh:
The prior implementation would include new lines characters. Subsequently, these characters would then be URI encoded when the headers are written (as "%0A"). Not including new lines via "strict encoding" will slightly reduce the size for long session values.

In order to handle existing sessions encoded with new lines, Base64#decode will handle ArgumentError exceptions and try to decode using non-strict encoding. A couple of small specs have also been added
for this case.

### DIFF
--- a/lib/rack/session/cookie.rb
+++ b/lib/rack/session/cookie.rb
@@ -49,7 +49,7 @@ module Rack
       # Encode session cookies as Base64
       class Base64
         def encode(str)
-          [str].pack('m')
+          [str].pack('m0')
         end
 
         def decode(str)
--- a/test/spec_session_cookie.rb
+++ b/test/spec_session_cookie.rb
@@ -74,26 +74,32 @@ describe Rack::Session::Cookie do
     it 'uses base64 to encode' do
       coder = Rack::Session::Cookie::Base64.new
       str   = 'fuuuuu'
-      coder.encode(str).must_equal [str].pack('m')
+      coder.encode(str).must_equal [str].pack('m0')
     end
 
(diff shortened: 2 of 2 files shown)


lib/rack/session/cookie.rb
```diff
@@ -48,11 +48,17 @@ module Session
...
+        end
+
         def encode(str)
-          [str].pack('m')
+          Base64.strict_encode(str)
         end
 
         def decode(str)
+          str.unpack('m0').first
+        rescue ArgumentError
```
matthewd: Is there an advantage to this? AFAICT `m` unpacks the output of `m0` just fine.

mikeycgto: At least on Ruby 2.4.1, trying to `unpack('m0')` with a string that was encoded with `pack('m')` is raising an `ArgumentError` for me:

```
$ ruby -e "p ['A' * 256].pack('m').unpack('m0').first"
-e:1:in `unpack': invalid base64 (ArgumentError)
	from -e:1:in `<main>'
```

matthewd: Right. But what do we gain by changing this method at all?

mikeycgto: Ah, I see what you mean. Good call as this method can be totally left alone then!

I can amend my commit to not touch this method. Thanks for the feedback!

ioquatix:
I am manually rebasing and merging this.

ioquatix:
https://github.com/rack/rack/pull/1460

PR #1177, opened 2017-06-09
Rack Authenticated and Encrypted session cookies

mjc-gh:
This PR adds AES GCM support to `Rack::Session::Cookie` and is able to upgrade legacy HMAC sessions to the new scheme (when configured correctly).

See sinatra/sinatra/issues/1300 for some context on this feature.


test/spec_encryptor.rb
```diff
@@ -0,0 +1,51 @@
...
+    cmsg = Rack::Encryptor.encrypt_message('hello world', @secret)
+    pmsg = Rack::Encryptor.decrypt_message(cmsg, @secret)
+
+    pmsg.must_equal 'hello world'
+  end
+
+  it 'encryptor and decryptor handles overly long keys' do
+    new_secret = "#{@secret}abcdef123456"
+
+    # These methos should truncate the long key (so OpenSSL raise exceptions)
```
olleolleolle: Typo: methos

olleolleolle: Q: does the last bit of this sentence mean "so as to avoid OpenSSL raising exceptions"?

test/spec_session_cookie.rb
```diff
@@ -148,7 +161,15 @@ def response_for(options={})
     Rack::Session::Cookie.new(incrementor)
     @warnings.first.must_match(/no secret/i)
     @warnings.clear
-    Rack::Session::Cookie.new(incrementor, :secret => 'abc')
+    Rack::Session::Cookie.new(incrementor, :secret => @secret)
+    @warnings.must_be :empty?
+  end
+
+  it 'warns if secret is to short' do
```
olleolleolle: Typo: to short => too short

test/spec_session_cookie.rb
```diff
@@ -262,69 +283,98 @@ def decode(str); @calls << :decode; str; end
...
     response = response_for(:app => app, :cookie => tampered_with_cookie)
     response.body.must_equal '{"counter"=>1}'
   end
 
+  it 'ignores tampered with legacy hmac cookie' do
+    legacy_session = Rack::Session::Cookie::Base64::Marshal.new.encode({ 'counter' => 1, 'session_id' => 'abcdef' })
+    legacy_secret  = 'test legacy secret'
+    legacy_digest  = OpenSSL::HMAC.hexdigest(OpenSSL::Digest::SHA1.new, legacy_secret, legacy_session).reverse
+
+    legacy_cookie = "rack.session=#{legacy_session}--#{legacy_digest}; path=/; HttpOnly"
```
olleolleolle: Q: Below, the Rack::Utils.escape was used on the legacy_session - not here. Does this matter?

test/spec_encryptor.rb
mikeycgto: Thanks!

test/spec_session_cookie.rb
mikeycgto: We actually don't need `escape` when an encoder with strict base64 encoding is used.

tonytonyjan:
This makes me think of a pull request sent by me long time ago.
ref: https://github.com/rack/rack/pull/1134

jeremyevans:
I think it is a good idea for Rack to support encrypted session cookies.  I prefer this approach over #1134, mostly because for things related to cryptography, it's better to have as few knobs as possible.

Changes I think we should make:

* `Assume keys are hex strings` is definitely something that we shouldn't do
* Encoded format would be better if it was `base64(iv + auth_tag + cipher_text)`, as `iv` and `auth tag` size are known, avoiding the need for delimeters and using only single `base64` call
* Webrick spec change appears unrelated

What do other core team members think?

mjc-gh:
Thanks for the feedback @jeremyevans! I agree with the changes you've suggested and can simplify the encoding process and rely on fixed length string slices in order to remove the need for a delimiter.

I think at this point though, if we're looking to introduce encrypted sessions directly to Rack, we should look into using something like [miscreant.rb](https://github.com/miscreant/miscreant.rb ) which provides for "nonce-reuse misuse-resistance" modes of AES. When using GCM mode, the security of the encryption completely fails if a nonce is reused. The nonce for GCM comes from the Initialization Vector, which is generated for us by OpenSSL and is 96-bits in size. While the probability of a randomly generated IV being reused is low, it is still theoretically possible. Thus, I believe it's best to introduce something like AES-SIV which miscreant provides for at this point in time. Additionally, [miscreant's API](https://github.com/miscreant/miscreant.rb#api) is extremely simple to use which goes well with your "few knobs as possible" comment.

As an aside, [rbnacl](https://github.com/RubyCrypto/rbnacl) which is being opted for instead of OpenSSL in some projects has similar security considerations: https://github.com/RubyCrypto/rbnacl/wiki/Secret-Key-Encryption#important-usage-notes

I'm hoping to make similar changes to Rails soon and finish up work with regards to Key Rotation within Rails. I'm happy to make the necessary changes to use miscreant. Thoughts?

jeremyevans:
Unfortunately, while I agree miscreant's approach is safer than AES-GCM, I'm not sure we want to add a dependency to rack, especially if this session support is designed to be the default for cookie sessions.  Adding it as a gem dependency would pull it in for people not using session cookie support, and adding it as an implicit dependency would break people upgrading from older versions of rack.

FWIW, when I designed Roda's cookie session support, I ultimately chose AES-CTR+HMAC-SHA256 instead of AES-GCM.  This way decryption is not even attempted unless HMAC verification succeeds.    To avoid the nonce-reuse issue, per-cookie secrets are used by default, with the cipher secret being an HMAC of random data stored in the cookie with the master secret.  This approach is slower than AES-GCM, though probably not enough to make a significant difference in the average application. On the plus side, it doesn't require external libraries.

mjc-gh:
Definitely understand not wanting to add a dependency, especially to a gem like Rack. Perhaps we should rework this PR then to use either AES-CTR or AES-CBC and use a MAC like HMAC-256 for authentication purposes? I'd like to take a deeper look at how Roda's encrypted cookie sessions work but it sounds like a similar approach could be used here.

I am more than happy to do this reimplementation in this PR as well as resolve all the conflicts as well.

Rails versions between 3.2 and 5.1 used AES-CBC prior to the updates to the `ActiveSupport::MessageEncryptor` and how sessions cookies are in handled in 5.2.

mjc-gh:
@jeremyevans got this PR updated to the latest.

I am going to now work on rewriting the encryption aspects to use your approach from Roda (AES-CTR with SHA-256 HMAC for authentication). I also really like your padding approach to help limit session length information leakage. I'm hoping to just rewrite the `Rack::Encryptor` module to achieve this.

Currently, this PR uses `Marshal` for serialization aspects. Do you think we should switch that over to JSON and limit what goes into a session? Do you also think it is worth introducing the "version" byte? This could be very useful in the future.

jeremyevans:
I choose to use JSON instead of Marshal in Roda because I think it is a more secure approach in the face of compromise (Marshal makes RCE much easier), and I didn't have to worry about backwards compatibility.  Unfortunately, if we plan to make this the default, we have to worry about backwards compatibility, and symbols and many other classes do not round trip through JSON. Supporting JSON as an option and recommending the use of the option is a possibility.  If you use a bitmap similar to Roda's approach, you can store the Marshal/JSON selection in the bitmap.

I think the version byte approach is useful to allow for gradual updates.  Using this approach allowed for adding the per-cookie cipher secrets approach in a backwards compatible way to Roda (those were not part of the original design).

mjc-gh:
Agreed about the use of JSON with regards to security and backwards compatibility. We can definitely support it as an option, explain the security reasoning, and make it easy to enable. Definitely like the use of the bitmapping and masking in order to store these sorts of settings as well as the version number.

Thanks for all the feedback! Should have this reworked in a few days time when I get some time to work on it.

ioquatix:
Do you think we can get this merged in the next week? Because it would be great to get 2.2 out the door soon, and this is the last major piece which I've included in the milestone. It seems like it's almost done, but I guess there are some minor outstanding issues?

.travis.yml
```diff
@@ -0,0 +1,38 @@
+language: ruby
```
ioquatix: Why is a travis config part of this PR?

mjc-gh: I will remove this with a rebase.

mjc-gh:
@ioquatix I should be able to rewrite the `Encryptor` module this weekend and get this ready for merging.

I also need to review the tests again, but we need to ensure migrating from the current session cookie setup is seamlessly handled through this new encrypted session cookie approach.

lib/rack/session/cookie.rb
```diff
@@ -1,34 +1,66 @@
...
+    # * :serialize_json
+    #     Use JSON for message serialization instead of Marshal. This can be
+    #     viewed as a security ehancement.
+    # * :gzip_over
+    #     For message data over this many bytes, compress it with the deflate
+    #     algorithm.
+    #
+    # Refer to Rack::Encryptor for more details on these options.
+    #
+    # Prior to version TODO, the session hash was stored as base64 encoded
```
mjc-gh: Need to add the version this middleware will change.

mjc-gh:
Made some major updates to this PR. It now uses the same encryption and HMAC verification approach that Roda uses.

There's still a few outstanding things to consider:

- Finish some of the documentation around these changes.
- Ensure sufficient coverage for legacy cookie "migrations". 
- Consider adding the padding feature to prevent session leaks. This is something that Roda supports that could be added to the `Rack::Encryptor` class.
- How are secrets passed to this middleware? Are they raw binary strings or are they encoded as Base64 or Hex?
- Remove the spec for weak keys. I've left the below spec failing to flag this:
  ```
  1) Error:
  Rack::Session::Cookie#test_0002_warns if secret is too short:
  ArgumentError: invalid secret: 16, must be >=64
    /home/mcoyne/Projects/rack/lib/rack/encryptor.rb:48:in `initialize'
    /home/mcoyne/Projects/rack/lib/rack/session/cookie.rb:162:in `new'
    /home/mcoyne/Projects/rack/lib/rack/session/cookie.rb:162:in `block in initialize'
    /home/mcoyne/Projects/rack/lib/rack/session/cookie.rb:161:in `map'
    /home/mcoyne/Projects/rack/lib/rack/session/cookie.rb:161:in `initialize'
    /home/mcoyne/Projects/rack/test/spec_session_cookie.rb:168:in `new'
    /home/mcoyne/Projects/rack/test/spec_session_cookie.rb:168:in `block (2 levels) in <top (required)>'
  ```

Note that I've made comments on this PR pointing out where I need to make more updates within the actual code and docs.

.travis.yml
mjc-gh: This is now removed.

lib/rack/session/cookie.rb
```diff
@@ -102,11 +134,43 @@ def encode(str); str; end
...
+          ::Marshal.load(str) if str
+        end
+      end
+
+      attr_reader :coder, :encryptors
 
       def initialize(app, options = {})
-        @secrets = options.values_at(:secret, :old_secret).compact
-        @hmac = options.fetch(:hmac, OpenSSL::Digest::SHA1)
+        # TODO handle decoding the secrets... are they base64? hex? raw binary
```
mjc-gh: We need to consider how we want to handle the `:secrets` option. If this option is encoded as base64 or hex, we need to decode. If these encoded values are not handled correctly, key material can be mishandled.

mjc-gh:
I've also renamed the `:secret` option to `:secrets`. This was done for two reasons:

1. Any existing `:secret` value in likely not compatible with this new encrypted approach. The secret keys now need to be 64 bytes in length as required by `Rack::Encryptor`.
2. The option is plural because it allows for an Array of Strings. When an Array is used, key rotation is attempted and older secrets can migrated out over time.

lib/rack/session/cookie.rb
jeremyevans: I think we should document that secrets are expected to be raw.  If users want to use base64 or hex, they should decode before passing the option to the middleware.

lib/rack/session/cookie.rb
```diff
@@ -135,14 +203,36 @@ def extract_session_id(request)
...
+          cookie_data = request.cookies[@key]
+          session_data = nil
+
+          # Try to decrypt the session data with our encryptors
+          encryptors.each do |encryptor|
+            begin
+              session_data = encryptor.decrypt(cookie_data) if cookie_data
+              break
+            rescue Rack::Encryptor::Error
+              # TODO warn? should there be some sort of external signal?
```
jeremyevans: Logging to `request.env[rack.errors]` is probably the best way to do this, I think.

lib/rack/session/cookie.rb
```diff
@@ -182,22 +268,34 @@ def delete_session(req, session_id, options)
...
-        OpenSSL::HMAC.hexdigest(@hmac.new, secret, data)
+      def legacy_generate_hmac(data)
+        OpenSSL::HMAC.hexdigest(@legacy_hmac.new, @legacy_hmac_secret, data)
       end
 
-      def secure?(options)
-        @secrets.size >= 1 ||
-        (options[:coder] && options[:let_coder_handle_secure_encoding])
+      def encode_session_data(session)
+        if encryptors.empty? then coder.encode(session)
```
jeremyevans: Probably split this into two lines.

lib/rack/encryptor.rb
```diff
@@ -0,0 +1,166 @@
...
+    # * :serialize_json
+    #     Use JSON for message serialization instead of Marshal. This can be
+    #     viewed as a security ehancement.
+    # * :gzip_over
+    #     For message data over this many bytes, compress it with the deflate
+    #     algorithm.
+    # * :purpose
+    #     Limit messages to a specific purpose. This can be viewed as a
+    #     security enhancement to prevent message reuse from different contexts
+    #     if keys are reused.
```
jeremyevans: Roda uses the cookie key for this.  Do you think there is a reason to use a separate option?  If so, do you think using the cookie key is a good default?

ioquatix:
I am sick so the earliest time I can review it will be next week. However, it would be great to ensure all tests are passing.

lib/rack/encryptor.rb
```diff
@@ -0,0 +1,166 @@
...
+    # will be used for the encryption cipher key. The remainder will be used
+    # for an HMAC key.
+    #
+    # Options may include:
+    # * :serialize_json
+    #     Use JSON for message serialization instead of Marshal. This can be
+    #     viewed as a security ehancement.
+    # * :gzip_over
+    #     For message data over this many bytes, compress it with the deflate
+    #     algorithm.
```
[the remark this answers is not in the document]
mjc-gh: Sounds good and agreed.

I also wonder if we should be concerned about how the first two bytes of a Marshaled Ruby hash is always `"\x04\b"`? Knowing this is the same in every message can be useful to an attacker.

lib/rack/session/cookie.rb
mjc-gh: Sounds good, I can make changes as needed.

I think the documentation just needs to emphasis the importance of how to handle the secret key(s). Often, these values are coming from environment variable or configuration files are very likely to be encoded somehow.

My other thought on this was we can have `:secrets` handle raw keys and add `:base64_secrets` and `:hex_secrets` which imply a decoding step before passing the keys off to the Encryptor.

lib/rack/encryptor.rb
mjc-gh: This feature is similar to an option offered by `ActiveSupport::MessageEncryptor`: https://api.rubyonrails.org/v5.2.3/classes/ActiveSupport/MessageEncryptor.html#class-ActiveSupport::MessageEncryptor-label-Confining+messages+to+a+specific+purpose

My hope was keep `Rack::Encryptor` as general purpose as possible, hence adding this option. It's used in the same way Roda uses the cookie key name as part of the authenticated message. Generally, I think this is good idea as it does add some effective security. However, one should not be reusing keys in the first place, but you never know!

lib/rack/encryptor.rb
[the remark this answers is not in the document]
jeremyevans: Unfortunately, I don't think there is anything we can do about that.  A similar issue exists for JSON, too.

lib/rack/encryptor.rb
jeremyevans: I'm fine keeping the option.  However, unless you have an objection to it, I would default the value to the cookie key.

mjc-gh:
@jeremyevans do you think it is worth adding the padding option to `Rack::Encryptor`? I do believe this adds some security value overall.

jeremyevans:
I think adding padding is a worthwhile addition.

lib/rack/encryptor.rb
mjc-gh: I now realized I didn't wire up using the `:purpose` option from session/cookie.rb. Will add that and will use the cookie name as the value.

jeremyevans:
I've reviewed the recent commits and they look good to me.

mjc-gh:
No sure why 2 checks have failed. The specs that failed have to do with tampered encrypted messages (the specs alter the MAC and expect verification to fail). Is there a way to rerun these via Github actions?

mjc-gh:
Realized the issue! Was setting the last byte of the MAC to `0x00`. In some cases, when the MAC actually had this value for the last byte, the resulting code was not modified and not "tampered" with. As a result, the specs failed. I modified these "tampered" specs to instead use an XOR with `1` on the last byte to ensure some bits are always changed.

Feel like we got lucky with CI failing as these specs were originally unstable 😬

mjc-gh:
Just updated this to deal with the merge conflicts.

ioquatix:
Okay great why don't we try to get some eyes on this with an aim to merge it.

mjc-gh:
Sounds good to me! Happy to make any changes as needed, just ping me 😄

mjc-gh:
Hey all! I caught a really bad bug in my implementation where I was using the incorrect key instead of the pre-session/message secret key (caught it due to an unused var).

Very happy I found this before it went any further but I think all the more reason to get a bunch of eyes on this before we merge it.

mjc-gh:
Was thinking about this more last night and I tried to write a test case to guard against this problem: https://github.com/mjc-gh/rack/commit/bb2c520529d3b7a831aebdb2572ad657147f2d49

The test is intentionally doing decryption using what is expected to be the wrong key and thus will get back random data. Not entirely sure the best way to test for this. If anyone has more ideas or test cases to guard against this, happy to add them to this spec.

jeremyevans:
If you want to test that your decrypt implementation doesn't use the raw key in the message, extract a method for setting the decryption cipher key, and in the test, override the method to check that the method is not called with the raw key.

zzak:
@mjc-gh Could anyone explain to me what the blockers are left on this? Is it just getting someone from rack-core to approve it?

ioquatix:
@mjc-gh thanks so much for your continued effort to bring this to completion. It's amazing that this PR has been active since 2017.

mjc-gh:
@ioquatix very happy to help where I can! I also agree this needs significant review by the Rack team and others.

I've added another spec as per Jeremy's suggestion. I temporarily reverted my fix locally and both of these two specs failed as expected. It can be difficult to write useful and valuable tests for this stuff, so let me know if anyone has more test cases or additional coverage they would like to see.

What's left is:

- [ ] We need to update this code comment here to specify the right versions before merging and/or releasing this: https://github.com/mjc-gh/rack/blob/2ab4cbddf2bb646979cdebd3adfdf9e6b4f33186/lib/rack/session/cookie.rb#L36
- [ ] I'd also like to add some additional security documentation around this feature. In particular, I think it is worth explaining how to properly handle keys as most users will likely be using some type of text encoding of their keys (base64) that will need to be decoded before passing it in as a `:secrets` (https://github.com/mjc-gh/rack/blob/2ab4cbddf2bb646979cdebd3adfdf9e6b4f33186/lib/rack/session/cookie.rb#L52) This document could also cover key rotations.

Also let me know if I should do a full rebase of this PR as I just did a merge. Thanks!

zzak:
@ioquatix Thanks for adding that context about the current state of this PR, and setting reasonable expectations. Tbh, I wasn't asking because I wanted someone from rack-core to magically appear and rubber stamp this but because it's been around for so long with many comments and changes -- I simply didn't want to read all of that.

So I really appreciate the tl;dr and happy to help where I can.

ioquatix:
> Also let me know if I should do a full rebase of this PR as I just did a merge. Thanks!

If it's easier, you can do squash and then rebase on master.

mjc-gh:
Reset the base to the upstream branch and squashed this down to 1 commit. Pushed up a branch on my fork that has all the history preserved if we ever need a reference. Want to just review this again with fresh eyes tomorrow and then add some docs if possible.

ioquatix:
@jeremyevans What do you think about merging this as it is now, and then reworking it if needed after extracting it to `rack-session` gem?

ioquatix:
Maybe we should also consider the features proposed in https://github.com/rack/rack/pull/1134 after extracted to a separate gem.

jeremyevans:
> @jeremyevans What do you think about merging this as it is now, and then reworking it if needed after extracting it to `rack-session` gem?

I'm fine with that approach. I haven't had time to review the current state of this PR, but I should be able to do so before we release a rack-session gem.  We should keep the session code in the rack repository until after we release a rack-session gem, to make it easier for users who track the main branch and are using the session code.

ioquatix:
I am going to merge this by hand into `rack-session`.

ioquatix:
It's done and in https://github.com/rack/rack-session now.

PR #1178, opened 2017-06-14
Unfreeze default session options

rinmu:
I would like to suggest to unfreeze default session options.

Freezing session options causes an error when combining certain libraries.
For example, session options are modified in warden gem.
https://github.com/hassox/warden/blob/v1.2.7/lib/warden/proxy.rb#L175

### DIFF
--- a/lib/rack/session/abstract/id.rb
+++ b/lib/rack/session/abstract/id.rb
@@ -210,7 +210,7 @@ module Rack
           :sidbits =>       128,
           :cookie_only =>   true,
           :secure_random => ::SecureRandom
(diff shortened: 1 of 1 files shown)


matthewd:
That line in warden is attempting to change the options for the current individual session, not modify a constant that will affect all sessions forever. If it's getting a reference to this original hash, the only thing preventing it creating Actual Problems *is* the fact it's frozen.

See https://github.com/hassox/warden/issues/147#issuecomment-306349101 for more information.

rinmu:
I understand. Thank you.

Issue #1075, opened 2016-05-24
Large file upload consumes 100% cpu for a very long time


olegantonyan:
Rack 2.0.0.rc1 with Rails 5.0.0.rc1 and Puma 3.4.0

The problem is that there is a loop in `Rack::Multipart::Parser.parse` (https://github.com/rack/rack/blob/master/lib/rack/multipart/parser.rb#L70). On each iteration it reads at most `Rack::Multipart::Parser::BUFSIZE` bytes which is 16384. But for the 200Mb file this means 12000 iterations and this is very slow.

My dirty hack:

``` ruby
# config/initializers/rack.rb
Rack::Multipart::Parser.const_set('BUFSIZE', 10_000_000)
```

Maybe there should be a config option to override default `Rack::Multipart::Parser::BUFSIZE`. Or maybe a better solution?

miroslavLalev:
Since no one has commented, you can pass header 'rack.multipart.buffer_size' with your response in order to specify the size of buffer instead of doing such workarounds.

olegantonyan:
Thank you @miroslavLalev !

vizo:
@miroslavLalev Can you please explain how to do it? Where to put it?

disastrous-charly:
@vizo I didn't succeed to send the header, I just did the dirty hack of @olegantonyan and it's finally working fine.

olegantonyan:
Actually, i didn't work for me either, and I couldn't find why in rack's source code. I've just implemented clunked upload, so this option no longer required.

mahnunchik:
Any news with this issue?

disastrous-charly:
@mahnunchik The "dirty" (but very useful) hack of @olegantonyan is working great. I just have a warning when starting the server.

Chunked upload is a good alternative (easy to do with https://github.com/tors/jquery-fileupload-rails ) but if you provide a public API, users will also have to implement the chunked upload. (lost my hairs trying to do it with cURL)

So that's why personnaly I go with the dirty hack.

darfux:
You can create a middleware and just insert it before Rack, such as:
```ruby
class MultipartBufferSetter

  def initialize(app)
    @app = app
  end

  def call(env)
    env.merge!(Rack::RACK_MULTIPART_BUFFER_SIZE => 100*1024*1024)
    @app.call(env)
  end
  
end
```

and then:
```ruby
config.middleware.insert_before "Rack::Runtime", "MultipartBufferSetter"
```

vizo:
[...]

PR #1179, opened 2017-06-14
Large file upload performance fix

tompng:
fixes issue #1075 

The problem is that:
method `handle_mime_body` tries to check `@buf =~ rx` every time `@buf` is changed,
and `@buf =~ rx` is slow because it tries to seach `rx` from the beginning of `@buf` every time.

I changed `@buf =~ rx` to `@buf.index rx, appropriate_offset` to fix this.

### DIFF
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -166,6 +166,7 @@ module Rack
 
       def initialize(boundary, tempfile, bufsize, query_parser)
         @buf            = String.new
+        @buf_rx_start   = 0
 
         @query_parser   = query_parser
         @params         = query_parser.make_params
@@ -174,6 +175,7 @@ module Rack
         @bufsize        = bufsize
 
         @rx = /(?:#{EOL})?#{Regexp.quote(@boundary)}(#{EOL}|--)/n
+        @rx_max_size = EOL.size + @boundary.size + [EOL.size, '--'.size].max
         @full_boundary = @boundary
         @end_boundary = @boundary + '--'
         @state = :FAST_FORWARD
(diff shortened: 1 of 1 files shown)


tompng:
Benchmarks

```
# before
1MB: 0.018405 sec
2MB: 0.045491 sec
4MB: 0.132708 sec
8MB: 0.464206 sec
16MB: 1.57744 sec
32MB: 5.183627 sec
64MB: 26.803011 sec
execution expired (HTTPClient::ReceiveTimeoutError)

# after
1MB: 0.022998 sec
2MB: 0.01944 sec
4MB: 0.043587 sec  # x3 faster
8MB: 0.086238 sec
16MB: 0.131328 sec # x12 faster
32MB: 0.234914 sec
64MB: 0.491324 sec # x54 faster
128MB: 1.159996 sec
```


Benchmark codes

Gemfile
```ruby
source 'https://rubygems.org'
ruby '2.4.0'
gem 'sinatra'
gem 'httpclient'
# gem 'rack', git: 'https://github.com/tompng/rack'
```

server
```sh
bundle install
be ruby -e "require 'sinatra';post('/upload'){params[:file][:tempfile].size.to_s}"
```

benchmark
```ruby
require 'httpclient'
def test size
  file = Tempfile.new
  file.write 'a'*size
  file.rewind
  option = { 'content-type' => "multipart/form-data, boundary=#{rand}" }
  t = Time.now
  res = HTTPClient.post('http://localhost:4567/upload', { file: file }, option).body
  raise if res.to_i != size
  Time.now - t
end
[1,2,4,8,16,32,64,128].each do |mb|
  time = test mb*1000*1000
  puts "#{mb}MB: #{time} sec"
end
```

PR #1181, opened 2017-06-21
Add falcon as an option by default.

ioquatix:
* Server code is available here: https://github.com/socketry/falcon

### DIFF
--- a/lib/rack/handler.rb
+++ b/lib/rack/handler.rb
@@ -52,7 +52,7 @@ module Rack
       elsif ENV.include?("RACK_HANDLER")
         self.get(ENV["RACK_HANDLER"])
       else
-        pick ['puma', 'thin', 'webrick']
+        pick ['puma', 'thin', 'falcon', 'webrick']
       end
     end
 


rafaelfranca:
Thank you for the pull request but I think it is too soon to add it to the default list.

ioquatix:
Why not make it possible for web servers to register themselves somehow?

rafaelfranca:
That sounds good to me

parndt:
@rafaelfranca is it still too soon? I'm using Falcon for things in production already and it's a great option. Thanks 😄

ioquatix:
I would like to implement a PR to allow servers to dynamically add themselves. We could merge this in the meantime, but I think it makes sense for it to be dynamic based on what gems the user has loaded.

simi:
:thinking: What changed your opinion @rafaelfranca?

rafaelfranca:
More than one year with falcon being out there and people using it.

ioquatix:
Thanks for merging.

Just FYI, I've added falcon tests using rack 1.x and master branch: https://travis-ci.com/socketry/falcon/builds/98681821 (check the Gemfile used).

e.g. https://travis-ci.com/socketry/falcon/jobs/173157848

It will be interesting to track changes to ensure compatibility.

Is there any discussion around what Rack 3 might look like?

I know there was at least some discussion about removing the need for rewindability.

PR #1182, opened 2017-06-21
Updates for frozen string literal compatibility.

pat:
To be clear, @twalpole did all the hard work previously in #1023, so I've doubled up on their efforts (taking a bit of a different approach, mostly to use `dup` given that seems to be the preference). I've not gone and added all the pragma comments to the files though.

Given that all the testing dependencies seem to be frozen-string-literal friendly, you could take the approach of enabling the flag for MRI 2.4 onwards in the Travis configuration to avoid regressions:

```yml
before_script:
- if (ruby -e "exit RUBY_VERSION.to_f >= 2.4"); then export RUBYOPT="--enable-frozen-string-literal"; fi; echo $RUBYOPT
```

Or adding the pragma comments will do the trick as well.

### DIFF
--- a/lib/rack/auth/digest/params.rb
+++ b/lib/rack/auth/digest/params.rb
@@ -38,12 +38,12 @@ module Rack
 
         def to_s
           map do |k, v|
-            "#{k}=" << (UNQUOTED.include?(k) ? v.to_s : quote(v))
+            "#{k}=#{(UNQUOTED.include?(k) ? v.to_s : quote(v))}"
           end.join(', ')
         end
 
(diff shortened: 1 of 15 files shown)


schneems:
Is there a game plan for adding the pragma to rack? https://github.com/rack/rack/pull/1203

pat:
@schneems I've generally left that up to the maintainers preference, but I'm happy to put the work in and get it in all the files (in this PR or a separate one).

twalpole:
@pat Without commitment that it will actually get merged, I'd suggest you wait on putting in the work - been there done that (as you pointed out originally)

PR #1184, opened 2017-06-27
sort select_best_encoding candidates by available_encodings order

wjordan:
This PR makes a small change to [`Rack::Utils#select_best_encoding`](https://github.com/wjordan/rack/blob/03850b61b38b32fc0258d4698fad8a97df82f105/lib/rack/utils.rb#L174-L202) that allows extra server-side control over the encoding returned in the case of a tie (multiple encodings at the same qvalue).

The reason this extra control is necessary is because the `Accept-Encoding` header field (ref. [rfc7231#5.3.4](https://tools.ietf.org/html/rfc7231#section-5.3.4)) uses HTTP content negotiation which contains a bit of ambiguity regarding tie-breakers at the same qvalue.

My use-case is the Chrome browser currently sending the following `Accept-Encoding` header for most requests over an HTTPS connection:

    Accept-Encoding: gzip, deflate, br

Because `br` (the [official](https://www.iana.org/assignments/http-parameters/http-parameters.xml#content-coding) identifier for [Brotli Compressed Data Format](https://tools.ietf.org/html/rfc7932)) has a better compression ratio than `gzip` or `deflate`, I'd like to tell Rack to prefer sending `br`-encoded content above the others.

The current implementation of `#select_best_encoding` interprets these encoding preferences in sorted-order, e.g.: "I prefer `gzip` most, next `deflate`, and `br` least", so `gzip` is returned.

This PR interprets these encoding preferences equally, e.g.: "I prefer `gzip`, `deflate`, and `br` equally, so send me whatever the server decides is best". The tie-breaker would be determined by the order of the encodings provided to `#select_best_encoding`, so `#select_best_encoding(%w(br gzip identity), request.accept_encoding)` for the above request would return `br` instead of `gzip`, as desired.

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -185,7 +185,9 @@ module Rack
           mem + list
         }
 
-      encoding_candidates = expanded_accept_encoding.sort_by { |_, q| -q }.map { |m, _| m }
+      encoding_candidates = expanded_accept_encoding.sort_by do |m, q|
+        [-q, available_encodings.index(m) || available_encodings.length]
+      end.map { |m, _| m }
 
       unless encoding_candidates.include?("identity")
         encoding_candidates.push("identity")
--- a/test/spec_utils.rb
+++ b/test/spec_utils.rb
@@ -435,6 +435,7 @@ describe Rack::Utils do
 
     helper.call(%w(compress gzip identity), [["compress", 1.0], ["gzip", 1.0]]).must_equal "compress"
     helper.call(%w(compress gzip identity), [["compress", 0.5], ["gzip", 1.0]]).must_equal "gzip"
+    helper.call(%w(compress gzip identity), [["gzip", 1.0], ["compress", 1.0]]).must_equal "compress"
 
     helper.call(%w(foo bar identity), []).must_equal "identity"
     helper.call(%w(foo bar identity), [["*", 1.0]]).must_equal "foo"


HoneyryderChuck:
you shouldn't need this IMO, you just need to pass the available encodings in order of preference to the first argument (see its usage in the deflater middleware). I'm assuming you're br-encoding them in a middleware of yours, as rack doesn't handle brotli by default.

wjordan:
@HoneyryderChuck refer to the following test I've included in this PR:

    helper.call(%w(compress gzip identity), [["gzip", 1.0], ["compress", 1.0]]).must_equal "compress"

If what you're saying is true, this test should pass in the current implementation, but it does not.

HoneyryderChuck:
@wjordan , got it finally. I guess I'm not sure what to say here. As you stated, the RFC doesn't cover this case, and all other documentation found leans toward "undefined behaviour". 

However, I'd keep it as it is. The order in which the client sends the encodings reflects preference (or browser defaults... ). Google, which champions brotli, also responds with a gzipped response in such a scenario, as with all of the other examples I've found. Can you provide me with URLs which match your desired behaviour?

AFAIK you can always skip the deflater middleware (most applications encode at the load balancer / on deploy time) and insert your own which handles brotli only. It might not be what you're looking after though.

wjordan:
> Can you provide me with URLs which match your desired behaviour?

facebook.com, for one:

    $ curl -sSL facebook.com -H "Accept-Encoding: gzip, deflate, br" -I | grep Content-Encoding
    Content-Encoding: br

HoneyryderChuck:
Thx for the links, I think I got it now. I've also reasoned about why the google main page isn't brotli'ing, and that might be because when it comes to dynamic content, the compression rate is beaten by the compress time(?). 

So I think that your proposal makes sense.

jeremyevans:
This change seems reasonable to me.  Can you please rebase against master and add documentation?

wjordan:
> Can you please rebase against master and add documentation?

What kind of documentation do you have in mind for a small change like this? The only thing [`CONTRIBUTING`](https://github.com/rack/rack/blob/master/CONTRIBUTING.md#write-documentation) suggests is to "Document any external behavior in the [README](https://github.com/rack/rack/blob/master/README.rdoc)" but that seems like overkill for this tiny detail.

jeremyevans:
I was thinking adding a method rdoc comment (one doesn't currently exist).  Unfortunately, rack doesn't document all existing methods, but we should  start moving in that direction.

ioquatix:
I will merge this by hand. Thanks for your effort.

PR #1187, opened 2017-07-03
Add multi mapping support via header

yoshuki:
nignx sets "HTTP_X_ACCEL_MAPPING" from multi mappings to comma splited string.
like,
```
proxy_set_header X-Accel-Mapping /var/www1/=/files1/;
proxy_set_header X-Accel-Mapping /var/www2/=/files2/;
```
to
```
'/var/www1/=/files1/, /var/www2/=/files2/'
```

### DIFF
--- a/lib/rack/sendfile.rb
+++ b/lib/rack/sendfile.rb
@@ -150,8 +150,12 @@ module Rack
       if mapping = @mappings.find { |internal,_| internal =~ path }
         path.sub(*mapping)
       elsif mapping = env['HTTP_X_ACCEL_MAPPING']
-        internal, external = mapping.split('=', 2).map(&:strip)
-        path.sub(/^#{internal}/i, external)
+        mapping.split(',').map(&:strip).each do |m|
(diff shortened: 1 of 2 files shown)


firien:
I do not believe this has made it into any release yet - any plans to merge into `2-0-stable` branch?

https://github.com/rack/rack/blob/2.0.5/lib/rack/sendfile.rb#L153
https://github.com/rack/rack/blob/1.6.10/lib/rack/sendfile.rb#L155

pabuisson:
I second @firien question. My team and I have been struggling with this for a while, and I just found this PR. Is there something preventing this from being released at the moment? And is there any way we can move things forward?

nbibler:
Seemingly this duplicates #853, although it uses commas instead of semicolons for the separator. I guess the other PR should be closed?

pabuisson:
Those two PRs are really similar indeed. I also think PR #853 could then be closed since the current PR we're commenting on has already been merged into master.

aidewoode:
I'm wondering why don't release it?

PR #1189, opened 2017-07-07
Stop replacing the environment in Rack::Lock

lugray:
While the flow of environment information is generally downstream only, test frameworks sometimes rely on being able to pull information added to the environment after `.call`.

This can be seen
 - In [rack-test](https://github.com/rack-test/rack-test/blob/fcfcd172fdfc0b5d9ca4cd27f8a5c8f68fbbecec/lib/rack/mock_session.rb#L32)
 - In [Rails](https://github.com/rails/rails/blob/98f8f81879825a009b0c751b3178fcb687e3033a/actionpack/lib/action_dispatch/testing/integration.rb#L271) where the controller instance is pulled from the environment, which fails causing Rails/rails#29608 when `Rack::Lock` is being used.

~~In order to accomplish this, I had to change [a test](https://github.com/rack/rack/pull/1189/files#diff-6482bbd415bc6ca058c28b364a19aa7dR151), but this line was introduced in #828, (which introduced the problem) and seemed to mainly be concerned with the other tests being added.  Thoughts @dkubb ?~~

### DIFF
--- a/lib/rack/lock.rb
+++ b/lib/rack/lock.rb
@@ -11,12 +11,21 @@ module Rack
 
     def call(env)
       @mutex.lock
+      @env = env
+      @old_rack_multithread = env[RACK_MULTITHREAD]
       begin
-        response = @app.call(env.merge(RACK_MULTITHREAD => false))
-        returned = response << BodyProxy.new(response.pop) { @mutex.unlock }
+        response = @app.call(env.merge!(RACK_MULTITHREAD => false))
+        returned = response << BodyProxy.new(response.pop) { unlock }
       ensure
-        @mutex.unlock unless returned
+        unlock unless returned
       end
+    end
+
+    private
+
+    def unlock
+      @mutex.unlock
+      @env[RACK_MULTITHREAD] = @old_rack_multithread
     end
   end
 end
--- a/test/spec_lock.rb
+++ b/test/spec_lock.rb
@@ -147,7 +147,8 @@ describe Rack::Lock do
     }, false)
     env = Rack::MockRequest.env_for("/")
     env['rack.multithread'].must_equal true
-    app.call(env)
+    _, _, body = app.call(env)
+    body.close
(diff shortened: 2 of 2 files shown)


matthewd:
The line in the test was new, but [the behaviour it is asserting was not](https://github.com/rack/rack/pull/828/files#diff-c06815bc484e53b5575852fc37c8b27bL23), and claiming the request is being processed in a single-threaded environment when it's not would be Bad, so we'll want to retain that.

As for the actual change.. I'm not sure how I feel about it. A strict reading of the spec would say that Rails is in the wrong: the spec makes no promise that the original `env` will pass through middleware (indeed, the spec barely acknowledges the existence of middleware: as far as it's concerned, they're basically two independent rack applications), and therefore one may not rely on such a behaviour. But in practice, almost every middleware does pass along the original. Notably, that's true even of [`Rack::Cascade`](https://github.com/rack/rack/blob/0362a54dba92626582d42f3343c209b7cdb7e713/lib/rack/cascade.rb#L33), which was my first thought of one which had a strong excuse not to.

I somewhat lean toward changing Rack::Lock to retain the original, thus conforming to the prevailing convention and avoiding inflicting any unnecessary surprises.. but still suggesting that Rails should try to avoid depending on that behaviour: a 3rd party middleware may be less accomodating.

lugray:
> the behaviour it is asserting was not

True enough.  I was in too much of a rush to get this open on Friday after spending the better part of the day tracking this down 😆.  I've pushed up a new version that maintains the behaviour, provided that the body is closed (which *is* part of the rack spec).

I agree that rails is relying on un-spec'd behaviour, but it is quite widely followed, and it's only relying on it for testing conveniences, not for production use.

lugray:
@matthewd, you've been extremely helpful on my [rails/rails](https://github.com/rails/rails) PRs, but can I bug you to take another look at this?

matthewd:
👍🏻

PR #1194, opened 2017-07-21
Add Hanami to the list of frameworks that use Rack.

oana-sipos:


oana-sipos:
Thank you @rafaelfranca

PR #1196, opened 2017-08-10
Fix typos

AlexKVal:


lib/rack/request.rb
```diff
@@ -98,15 +98,15 @@ def initialize_copy(other)
 
     module Helpers
       # The set of form-data media-types. Requests that do not indicate
-      # one of the media types presents in this list will not be eligible
+      # one of the media types presented in this list will not be eligible
```
matthewd: I think "present" would be closer to the intended meaning

AlexKVal:
> I think "present" would be closer to the intended meaning

Sure. Thank you for helping 🍒

matthewd:
Thanks!

PR #1197, opened 2017-08-14
set minimum version of concurrent-ruby

pravi:
concurrent/utility/native_integer (used in spec_webrick.rb) was introduced only in 1.0.3


matthewd:
Nice catch! I've solved this from a slightly different angle in #1198.

PR #1199, opened 2017-08-18
Don't swallow backtrace when reraising query parser error

jeremyevans:


ioquatix:
@Jeremy I assume `Exception#cause` is now sufficient and this PR is redundant?

jeremyevans:
Ruby only started printing causes for exceptions in 2.5.  So this would definitely be helpful in Ruby 2.2-2.4, which Rack still supports.

ioquatix:
While pragmatic, I don’t think we should do this? I prefer we just depend on `#cause`. That being said, I’m not bothered by this. @tenderlove?

ioquatix:
Sorry got wrong button on mobile phone...

matthewd:
@ioquatix agree.

In practice I don't think the Ruby exception-printer behaviour is a big factor here, FWIW: this is going to be called from inside a request handler, and all request-handling servers are going to have a top level rescue around the request that prevents this exception ending up on an actual terminal. 

Given that `cause` exists, and has for some time now, I'd rather not introduce a sometimes-confusing lie ("how did this line raise an `InvalidParameterError`, it clearly says `raise ArgumentError`?!")

jeremyevans:
In most cases where I've seen exception handling that prints backtraces, only the backtrace of the current exception is printed, it does not recursively print exceptions.  For the best example of this, see Ruby's stdlib logger: https://github.com/ruby/logger/blob/master/lib/logger/formatter.rb#L30.  Unicorn uses this feature of stdlib logger to log exceptions: https://bogomips.org/unicorn.git/tree/lib/unicorn/http_server.rb#n556

If you don't do this, what happens is the backtrace logged doesn't show you where the underlying exception actually occurs.  This significantly complicates debugging.

The reason for the code in question is to use a rack-specific exception class instead of a generic argument error.  The place where the exception class wrapping occurs is almost never going to be of interest to the programming debugging the error.

matthewd:
Both valid points. I might argue the former is a shortcoming of pre-`cause` exception handling implementation(s) that should be updated, but that doesn't help us right now. 🤷🏻‍♂️

ioquatix:
Would another solution be to fix logger?

ioquatix:
I also agree, it's super confusing when an exception of one type is transformed into another.

jeremyevans:
This issue is not solely in logger.  There are plenty of places where exceptions are handled and backtraces are printed without checking `cause`.  It would say if you find code that prints a backtrace for an exception, it is more likely it doesn't do it recursively for `cause`.

ioquatix:
There is only one main reason why I wouldn't merge this.

Because Ruby itself does log cause correctly:

```
koyoko% ruby
begin
raise "Boom1"
rescue
raise "Boom2"
end
Traceback (most recent call last):
-:2:in `<main>': Boom1 (RuntimeError)
	1: from -:1:in `<main>'
-:4:in `rescue in <main>': Boom2 (RuntimeError)
```

And also, my gem `console` logs it correctly too.

By putting the backtrace twice, you penalise logging systems that do the right thing..

My vote would be to fix `logger` to print cause recursively.

That being said, if you still feel strongly about this, you can merge it. But I don't personally feel strongly one way or the other, just the above case sticks out to me.

jeremyevans:
With the current state of things, I do feel strongly we should merge this, so I'll do so.  We can revisit this in the future if more exception handling respects `cause` and someone actually complains about this.

ioquatix:
Okay great.

PR #1200, opened 2017-08-18
Use buffer string when parsing multipart requests

janko-m:
Passing a buffer string as a second argument to `IO#read` will make the Rack input IO, instead of creating a new string object on each read, load each chunk into the same string instance (overwriting the previous content). This greatly reduces string allocations.


janko-m:
When testing with a 5MB request body, this patch reduced memory allocation from 18.4MB to 14.4MB.

PR #1201, opened 2017-08-18
Don't use #eof? when parsing multipart

janko-m:
On chunked requests (`Transfer-Encoding: chunked`) content length won't be present, so `#eof?` will be called on the Rack input directly. However, the Rack input doesn't have to respond to `#eof?`, it's not part of the Rack specification. For example, `Rack::Lint::InputWrapper` and `Unicorn::StreamInput` don't respond to `#eof?`. This means that in development and on Unicorn, multipart parsing will fail for chunked requests.

This change refactors the multipart parser so that it doesn't call `#eof?`.

### DIFF
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -39,8 +39,6 @@ module Rack
           str
         end
 
-        def eof?; @content_length == @cursor; end
-
         def rewind
           @io.rewind
         end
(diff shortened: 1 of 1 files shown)


ioquatix:
This is a good idea and should be merged!

stanhu:
@janko-m We just ran into private method errors where `eof?` in `Unicorn::StreamInput`. Would you mind tagging a new release with this?

dirkjonker:
We are experiencing a an issue with uploading documents that is very difficult to reproduce but randomly fails because rack is calling `eof?` as described in [this issue on Passenger](https://github.com/phusion/passenger/issues/2150).

@tenderlove would you consider tagging a new release for this? Technically, currently the latest rack release (2.0.6) does not conform to its own specifications because of this.

urkle:
I also was just bit by this bug.  Any timeline of pushing a 2.0.7 rack version that includes this fix?

PR #1202, opened 2017-08-24
Faster Utils.clean_path_info

schneems:
The purpose of the `File.join` is to deal with redundant slashes for example

```
File.join("foo/", "bar")
# => "foo/bar"
```

In this case we are guaranteed that there will be not slashes in our data because we are first splitting by separators:

```
parts = path_info.split PATH_SEPS
```

We can speed up this call then by using the `Array#join` method which does less work:

```
require 'benchmark/ips'

SOURCE_ARRAY = ["foo", "bar"]

Benchmark.ips do |x|
  x.report("Array#join") do |times|
    i = 0
    while i < times
      SOURCE_ARRAY.join(::File::SEPARATOR)
      i +=1
    end
  end
  x.report("File#join") do |times|
    i = 0
    while i < times
      ::File.join(SOURCE_ARRAY)
      i +=1
    end
  end
  x.compare!
end
```

You can see this method of building strings is roughly 2.55 times faster than the current method

```
          Array#join   111.966k i/100ms
           File#join    62.000k i/100ms
          Array#join      2.075M (±12.1%) i/s -     10.189M in   5.022957s
           File#join    813.613k (±11.5%) i/s -      4.030M in   5.052667s

Comparison:
          Array#join:  2075017.5 i/s
           File#join:   813613.1 i/s - 2.55x  slower
```

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -580,9 +580,9 @@ module Rack
         part == '..' ? clean.pop : clean << part
       end
 
-      clean.unshift '/' if parts.empty? || parts.first.empty?
-
-      ::File.join clean
+      clean_path = clean.join(::File::SEPARATOR)
+      clean_path.prepend("/") if parts.empty? || parts.first.empty?
+      clean_path
     end
 
     NULL_BYTE = "\0"


matthewd:
Assuming this method is hot enough to care about, a couple of extra quick-glance observations:

* In the common case of a single path_sep, the split should be measurably faster with a string instead of a regexp
* I suspect this often has nothing to actually clean... could it be worth an early no-op return unless the original path matches `/#{PATH_SEPS}\.*#{PATH_SEPS}/`?

schneems:
If we could early return than we could skip two array allocations, and that would be much faster. I'll throw that on here in a bit. I would want to use `Regexp#match?` so i'll need to add a backwards compat shim if it's not somewhere in here already. My only worry would be security. 

We could simply check for any presence of a dot `/\./` and we might have higher false positives for things like assets `foo.jpg` but I would be less worried about obscure edge cases.

For splitting via string instead of regex, we're using both the path separator and an alternate path separator. You can't split on two strings.

matthewd:
> so i'll need to add a backwards compat shim

I think we have precedent somewhere that the compatibility would be in the form of a conditional around this method, defining two different versions based on whether the thing we need (`#match?`, in this case) is available.

> We could simply check for any presence of a dot

The algorithm's not that complicated... the only conditions inside the loop body are for an empty string, exactly `'.'`, and exactly `'..'`.

> For splitting via string instead of regex, we're using both the path separator and an alternate path separator. You can't split on two strings.

Right, but those only differ on Windows. On my laptop, and most server systems:

```
[::File::SEPARATOR, ::File::ALT_SEPARATOR].compact
=> ["/"]
```

So it could be a regex when they differ, and a string when they don't.

ioquatix:
@schneems can you please rebase on master.

schneems:
Rebased, I didn't end up implementing any of the early return optimizations as discussed earlier though. I would probably split off into a separate PR at this point.

Tests are failing. Not sure the status of CI on master.

jeremyevans:
Looks like CI on master was not working on November 15, 2019.  Can you please rebase on current master?  Assuming CI passes we can merge this.

schneems:
Rebased again.

schneems:
Thanks all!

Issue #738, opened 2014-10-02
Utils::HeaderHash is EXTREMELY expensive


schneems:
Hypothetically if it was safe to remove HeaderHash usage right now, we would see

```
rack/lib x 7318
rack/lib x 5678
```

22.4 % fewer objects per request, and here's the kicker:

Speed with Header Hash

```
Calculating -------------------------------------
                 ips        51 i/100ms
-------------------------------------------------
                 ips      524.8 (±9.3%) i/s -       2601 in   5.005893s
```

Speed Without header Hash

```
Calculating -------------------------------------
                 ips        62 i/100ms
-------------------------------------------------
                 ips      649.2 (±8.8%) i/s -       3224 in   5.005930s
```

We get a `(649.2 - 524.8 )/524.8  * 100 # => 23` % faster response only by removing HeaderHash. This is against a whole Rails app, not just Rack benchmarks in isolation. I don't know about you, but that sounds like a lot.
## Reality

I know we can't remove `HeaderHash` wholesale, but we should do an audit of the places it's used, maybe it's not necessarry everywhere. 

Perhaps we can cross reference documentation and for all classes that explicitly mention returning a header hash with this behavior (preserves original keys, and but allows for access with varying case), if documentation doesn't exist re-write to not use HeaderHash. Is that crazy? This util class is new to me, so I would appreciate any background info you might have.

With no changes or fixes, wholesale removing HeaderHash behavior from Rack causes 5 failures out of 2020 assertions in the test suite.

wjordan:
@schneems do you have a working repo or other steps to replicate your benchmarking tests?

I did a bit of exploration and it looks possible to reduce the number of excessive per-request object allocations caused by HeaderHash by memoizing the results of `downcase` methods into a class variable, e.g.: f17d66dab152da4e5eb5ad427c88b66af57442b9

If the cost of using HeaderHash is mainly due to string object allocations caused by `downcase`, then a change like this could be worth the slight extra cost of extra memory usage, and an internal performance fix could be more easily backported without breaking API changes.

Using this technique I see a slight reduction in total object allocations (from 1173 to 1131 objects profiling a simple 'hello world' Rails 4.2 app with [derailed](https://github.com/schneems/derailed_benchmarks)), but negligible `ips` improvement w/Ruby 2.2- perhaps this is no longer an issue with the updated Ruby GC? Or do some Rails apps use HeaderHash much more extensively than the default `rails new` stack I tested on?

raggi:
[...]

PR #1204, opened 2017-08-25
Allow Disabling Header Hash

schneems:
Using a HeaderHash in place of a hash is extremely expensive. The hash seeks to normalize the case of keys. For any apps using correctly cased keys they should be allowed to bypass this behavior.

Using an a super simple Rack app, I was able to see a 22.77% performance increase by disabling this behavior.

## Benchmark Reproduction

Make an app with a `Gemfile`

```ruby
# Gemfile
source 'https://rubygems.org'

gem 'rack', path: '~/documents/projects/rack'
gem 'rack-test'
gem 'derailed'
```

And a `perf.rake`

```ruby
# perf.rake
CONST_RESPONSE = [
    200,
    {
      'Content-Type'  => 'text/html',
    },
    ["hello world".freeze]
  ]

namespace :perf do
  task :rack_load do
    DERAILED_APP = lambda { |env| CONST_RESPONSE }
  end
end
```

Then hit the app with the flag on and off

```term
export TEST_COUNT=100_000

echo "============ CURRENT =================="
for run in {1..20}
do
  echo "=="
  bundle exec derailed exec perf:test
done

echo "=============================="
echo "=============================="
echo "=============================="
echo "=============================="
echo "=============================="
echo "=============================="


echo "============ PATCH =================="

for run in {1..20}
do
  echo "=="
  RACK_DISABLE_HEADER_HASH=1 bundle exec derailed exec perf:test
done
```

I then graphed the results

## Why it's faster

There are interpreter level hash optimizations such as automatically [freezing key strings](https://github.com/ruby/ruby/commit/58f800a278b8b70463f4afdbb23a918d8ab441ff), and [hash lookups](https://github.com/ruby/ruby/blob/066e9a8b4ac36907f8e7915b5f94fdc7ca851de0/vm_insnhelper.c#L3575-L3577). When you subclass hash many of these optimizations cannot be applied.

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -410,7 +410,11 @@ module Rack
     # header when set.
     class HeaderHash < Hash
       def self.new(hash={})
-        HeaderHash === hash ? hash : super(hash)
+        if ENV["RACK_DISABLE_HEADER_HASH"]
+          return hash
+        else
+          HeaderHash === hash ? hash : super(hash)
+        end
       end
 
       def initialize(hash={})


schneems:
I should add, I ran CodeTriage on this patch and it runs fine out of the box with Puma and Rails 5.

matthewd:
This just seems like way too much of a fragile hack to warrant inclusion in the library, in any form, to me. To do it with an environment variable is, just... um... architecturally impure.

To me, it's the sort of thing best approached by leaving the application to monkey-patch the method: that has the right "this is super fragile, and if/when it breaks you get to keep all the pieces" feel.

schneems:
Sure, FWIW I don't intend this to go in as is. It's more of a "hey this tiny seeming thing is adding huge overhead, let's all get together and brainstorm a future where it doesn't have to exist".

ioquatix:
Unfortunately headers can not be stored in a hash. Best case, an array of key-value pairs.

schneems:
> Unfortunately headers can not be stored in a hash

There is an object entitled `HeaderHash` which this issue is filed about. 

I'm fine for this PR to be closed if you're cleaning up issues as it's not actively being worked on.  Not sure if that's what you're doing or not.

ioquatix:
Would it be safe to write `HeaderHash = Hash` and be done with it?

schneems:
Related to #738 #1549

PR #1208, opened 2017-10-16
Add file-based session mechanism

graywolf:
This pull request adds new file-based session mechanism. Only session id is stored in cookie (with hmac protection same way as Session::Cookie) and actual session is stored on disk using PStore.

Code is heavily inspired by how Session::Cookie works. Same with tests.

Please consider for merging, feedback is welcomed :)

### DIFF
--- a/lib/rack.rb
+++ b/lib/rack.rb
@@ -143,5 +143,6 @@ module Rack
     autoload :Cookie, "rack/session/cookie"
     autoload :Pool, "rack/session/pool"
     autoload :Memcache, "rack/session/memcache"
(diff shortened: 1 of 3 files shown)


jeremyevans:
This looks like something you could ship as an external gem, and I don't think the need for this common enough to warrant inclusion in Rack.  If we did decide to include it, I think we would want to require that a secret be provided.  It also would result in conflicts with https://github.com/dayflower/rack-session-file (though that appears to be dead). 

I'm going to close this now. If another committer thinks we should accept it, we can reopen.

PR #1213, opened 2017-11-01
 Enclose IPv6 address in X-Forwarded-Host in brackets

mwpastore:
Prevent Rack::Request#host from stripping off the last hextet of an IPv6 address contained in X-Forwarded-Host (and returned by Rack::Request#host_with_port) by enclosing the address in brackets.

IPv6 addresses in the HTTP_HOST, SERVER_NAME, and SERVER_ADDR CGI variables will (should) always be enclosed in brackets.

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -1,5 +1,6 @@
 require 'rack/utils'
 require 'rack/media_type'
+require 'resolv'
 
 module Rack
   # Rack::Request provides a convenient interface to a Rack
@@ -224,7 +225,12 @@ module Rack
 
       def host_with_port
         if forwarded = get_header(HTTP_X_FORWARDED_HOST)
-          forwarded.split(/,\s?/).last
+          host = forwarded.split(/,\s?/).last
+
+          # If the reverse proxy sends an IPv6 address without brackets,
+          # prevent the last hextet from being stripped off by host() by
+          # enclosing the address in brackets.
+          host =~ Resolv::IPv6::Regex ? "[#{host}]" : host
         else
           get_header(HTTP_HOST) || "#{get_header(SERVER_NAME) || get_header(SERVER_ADDR)}:#{get_header(SERVER_PORT)}"
         end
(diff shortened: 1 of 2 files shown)


matthewd:
> IPv6 addresses in the HTTP_HOST, SERVER_NAME, and SERVER_ADDR CGI variables will (should) always be enclosed in brackets.

Indeed. And X-Forwarded-Host is a forwarded value for HTTP_HOST.

This sounds rather more like a hack around an upstream misconfiguration than a necessary correction to me. Could you give some more detail on how you came to encounter such a value?

mwpastore:
Yes, and the best reverse proxies I've seen don't even use the X-Forwarded-Host header, they just pass the Host value along. RFC 7239 says that IPv6 addresses in this header _should_ be enclosed in square brackets. 

I haven't seen it it in the wild in person, just in forum posts and the like, so this may be a case of overly-defensive programming. I agree that if the reverse proxy is sending an invalid value, it should be fixed there. It's just such a subtle, silent issue—and the X- headers seem to be especially prone to errors and misconfiguration—that a guard here makes sense to me. 

Take it or leave it!

HoneyryderChuck:
Take a look at [this](https://bugs.ruby-lang.org/issues/14016), there's quite a lot of sketchy IPv6 support in ruby's core libs.

jeremyevans:
RFC 7239 Section 7.4:

```
   Note that IPv6 addresses may not be quoted in
   X-Forwarded-For and may not be enclosed by square brackets, but they
   are quoted and enclosed in square brackets in "Forwarded".

       X-Forwarded-For: 192.0.2.43, 2001:db8:cafe::17

   becomes:

       Forwarded: for=192.0.2.43, for="[2001:db8:cafe::17]"
```

So I think we should handle this for `X-Forwarded-*`.  but not `Forwarded` (when support is added for `Forwarded` #1423).

Could you please rebase against master?

ioquatix:
This makes perfect sense to me, thanks for the specs. I'll merge it.

ioquatix:
I reviewed the PR and it looks like we recently merged https://github.com/rack/rack/pull/1538 which prefered the non-square-brackets representation. However, this has not been released yet. So we could adjust it.

Based on the defintion of `URI#host`, I think we should probably prefer the square brackets representation... but I'm not sure what is canonical. I need to do some research.

PR #1221, opened 2017-12-28
More efficient multipart? method

sjshuck:
Do not continue to traverse params instance variable after determining multipart is true.


jeremy:
Thanks @sjshuck !

PR #1223, opened 2018-01-05
Allow subclasses to override ShowExceptions::TEMPLATE

jkowens:
This issue came up in the Sinatra project. Sinatra extends `Rack::ShowExceptions` to implement a custom template, but the only way to do that right now with this code:

https://github.com/sinatra/sinatra/blob/master/lib/sinatra/show_exceptions.rb#L364

```
Rack::ShowExceptions.send :remove_const, "TEMPLATE"
Rack::ShowExceptions.const_set "TEMPLATE", Sinatra::ShowExceptions::TEMPLATE
```

This was added after #922, but this causes problems when Sinatra is mounted along with other Rack apps.

See: sinatra/sinatra/issues/1376

### DIFF
--- a/lib/rack/show_exceptions.rb
+++ b/lib/rack/show_exceptions.rb
@@ -97,7 +97,7 @@ module Rack
     end
 
     def template
-      self.class::TEMPLATE
+      TEMPLATE
     end
 
     def h(obj)                  # :nodoc:
--- a/lib/rack/show_exceptions.rb
+++ b/lib/rack/show_exceptions.rb
@@ -93,7 +93,11 @@ module Rack
         end
       }.compact
 
-      TEMPLATE.result(binding)
+      template.result(binding)
+    end
+
+    def template
+      self.class::TEMPLATE
     end
 
     def h(obj)                  # :nodoc:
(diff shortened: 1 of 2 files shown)


namusyaka:
Hello, rack maintainers.

As Rack's opinion, is it assumed that the framework that conforms to Rack inherits this exception class and uses it as a use case? If so, you should provide a standard way, not just this way.
Unfortunately, due to #922, we simply lost the technique to inherit and extend this class for defining custom template.
I am pleased to receive your opinion.

lib/rack/show_exceptions.rb
```diff
@@ -93,7 +93,11 @@ def pretty(env, exception)
         end
       }.compact
 
-      TEMPLATE.result(binding)
+      template.result(binding)
+    end
+
+    def template
+      self.class::TEMPLATE
```
jkowens: Referencing `TEMPLATE` in this way allows subclasses to override the `TEMPLATE` constant. Subclasses also have the choice of overriding the `#template` method instead.

rafaelfranca: Thanks. It is better to use just `TEMPLATE` here and only allow subclasses to override `#template` instead of a constant.

jkowens: Ok good call. I've updated the PR, thanks!

rafaelfranca:
hello @namusyaka. After this PR do you still think it is not possible to inherit and extend this class to define custom templates?

namusyaka:
Nope, current change looks fine to me. Thank you! @rafaelfranca @jkowens

PR #1224, opened 2018-01-05
Preserve forwarded IP address for trusted proxy chains

SamSaffron:
Sometimes proxies make requests to Rack applications, for example
HAProxy health checks and so on.

Previously the forwarded IP implementation ate up these IP addresses,
making it hard to tell in Rack applications who made the request.

@rafaelfranca @matthewd  thoughts on this? 

Rack is eating up my IP addresses so its totally mucking with my rate limiter. 

My actual problem is that I have a "global" rate limiter in the app and for whatever crazy reason AWS ELB loves making enormous amounts of requests for its health checks, I noticed just for meta.discourse.org we had 5 different load balancers checking every cycle. They all have unique private IP addresses though, but the traffic goes ELB -> NGINX -> Unicorn. So it needs to get the ELB address out as remote IP which is a private IP.

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -261,7 +261,7 @@ module Rack
 
         forwarded_ips = split_ip_addresses(get_header('HTTP_X_FORWARDED_FOR'))
 
-        return reject_trusted_ip_addresses(forwarded_ips).last || get_header("REMOTE_ADDR")
+        return reject_trusted_ip_addresses(forwarded_ips).last || forwarded_ips.first || get_header("REMOTE_ADDR")
       end
 
       # The media type (type/subtype) portion of the CONTENT_TYPE header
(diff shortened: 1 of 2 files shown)


SamSaffron:
one sec ... let me update this ... need to fix something

matthewd:
Yeah, this sounds right to me (see also https://github.com/rails/rails/pull/14600, which I've still never revisited :pensive:)

SamSaffron:
thanks @matthewd , better ping @indirect here since he was responsible for the first implementation here.

SamSaffron:
failures seem unrelated, travis config here is a bit ill

timeout in one case, broken on ruby head in another due to rack

SamSaffron:
anything? anyone?

SamSaffron:
Almost the same PR is here: https://github.com/rack/rack/pull/1160

PR #1233, opened 2018-02-12
Documenting Rack::Utils::HeaderHash keys restrictions [ci skip]

esparta:
Currently the class does not accept nil as keys on their operations

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -409,6 +409,9 @@ module Rack
 
     # A case-insensitive Hash that preserves the original case of a
     # header when set.
+    # It does not support nil as key on any operation: instantiation,
+    # access, assigment, key checking (#include?, #has_key?, #member?)
+    # neither merging/replacing with another hash
     class HeaderHash < Hash
       def self.new(hash={})
         HeaderHash === hash ? hash : super(hash)
--- a/test/spec_utils.rb
+++ b/test/spec_utils.rb
@@ -685,6 +685,50 @@ describe Rack::Utils::HeaderHash do
     h['foo'].must_be_nil
     h.wont_include 'foo'
   end
+
+  it "nil keys are not supported on instantiation" do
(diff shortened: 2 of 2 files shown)


tenderlove:
PR looks good, but I'm not sure if we explicitly don't support `nil`, or we just don't know what to do to support it.  I think if we were to explicitly not support `nil`, it should raise an `ArgumentError` rather than just getting an exception from `downcase` called on `nil`.

rafaelfranca:
By the purpose of the hash I'd say we should not support it. Unless you think `nil` has a header name make sense.

tenderlove:
> By the purpose of the hash I'd say we should not support it. Unless you think nil has a header name make sense.

I really don't think `nil` as a header name makes sense. @esparta can you change these to raise an `ArgumentError`?

esparta:
I guess I can do it, but I don't know if we should.

Adding a begin block +re-raise or an additional if for a case it shouldn't happen frequently (or is something wrong with the call chain), would be costly. I don't know if originally was intended to let it fail.

The methods on `HeaderHash` (`#initialize`, `#merge`, etc) are called by `#initialize` or `#call` within these classes: `Chunked`, `CommonLogger`, `ConditionalLogger`, `ContentType`, `ContentLength`, `Deflater`, `Link`, `ShowStatus`. The one concern most is `Response#initialize`.

I got this because some internal/in-house middleware was misbehaving and leads me to that error, that took me a some time to figure out what was going on, so that's why I mean to at least document it on comments and tests.

So, should I proceed?

jeremyevans:
I don't think we should add tests for this.  It doesn't make sense to call out `nil` as not being supported, anything not a string will probably result in the same error (e.g. `Rack::Utils::HeaderHash.new(1=>2)`).  The documentation for HeaderHash states: `A case-insensitive Hash that preserves the original case of a header when set.`  I think this implies that keys should be strings.  At most we should just add a note to explicitly state that.

PR #1236, opened 2018-02-23
Correct the FCGI installation instructions

### LINKED ISSUE #1102: [Replace or Remove] http://www.fastcgi.com is unavailable on README.rdoc
http://www.fastcgi.com is no longer available, so the **FCGI installation example**(from line 160) which on **README.rdoc** should be updated or removed.  I found [FastCGI-Archives/FastCGI.com](https://github.com/FastCGI-Archives/FastCGI.com) on [Wikipedia](https://en.wikipedia.org/wiki/FastCGI) and created a **Docker image** on [Docker Hub](https://hub.docker.com/r/ysksn/dockerfile_fcgi/)(see its [Dockerfile](https://hub.docker.com/r/ysksn/dockerfile_fcgi/~/dockerfile/)).

Could I replace

> Installing the FCGI libraries:
> 
> ```
> curl -O http://www.fastcgi.com/dist/fcgi-2.4.0.tar.gz
> tar xzvf fcgi-2.4.0.tar.gz
> cd fcgi-2.4.0
> ./configure --prefix=/usr/local
> make
> sudo make install
> cd ..
> ```

with

Installing the FCGI libraries:

```
curl -O -k -L https://github.com/FastCGI-Archives/FastCGI.com/raw/master/original_snapshot/fcgi-2.4.1-SNAP-0910052249.tar.gz
tar xvfz fcgi-2.4.1-SNAP-0910052249.tar.gz
cd fcgi-2.4.1-SNAP-0910052249
./configure --prefix=/usr/local
mv libfcgi/fcgio.cpp /tmp/original
echo "#include <stdio.h>" > libfcgi/fcgio.cpp
cat /tmp/original >> libfcgi/fcgio.cpp
make
sudo make install
```

or just remove it?

What
1. `mv libfcgi/fcgio.cpp /tmp/original`
2. `echo "#include <stdio.h>" > libfcgi/fcgio.cpp`
3. `cat /tmp/original >> libfcgi/fcgio.cpp`

are doing is just inserting header file `#include <stdio.h>` top of of `fcgio.cpp` which is somehow missing. I know it's weird though.

Edouard-chin:
Fix the readme:

- I'm taking over #1102 since it has been open for a long time now (Thanks @yskn for opening the issue!)
- http://www.fastcgi.com is no longer available, this replaces the instruction to download the tar from a github archive
- Closes #1102

cc/ @rafaelfranca

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -170,9 +170,9 @@ Download and install lighttpd:
 
 Installing the FCGI libraries:
 
-    curl -O http://www.fastcgi.com/dist/fcgi-2.4.0.tar.gz
-    tar xzvf fcgi-2.4.0.tar.gz
-    cd fcgi-2.4.0
+    curl -O -k -L https://github.com/FastCGI-Archives/FastCGI.com/raw/master/original_snapshot/fcgi-2.4.1-SNAP-0910052249.tar.gz
+    tar xvfz fcgi-2.4.1-SNAP-0910052249.tar.gz
+    cd fcgi-2.4.1-SNAP-0910052249
     ./configure --prefix=/usr/local
     make
     sudo make install


Edouard-chin:
Looks like builds are broken on ruby-head since at least revision [62380](https://bugs.ruby-lang.org/projects/ruby-trunk/repository/revisions/62380) (issues with bundler 1.16.1)

Not sure why build are failing on jruby though

matthewd:
> `BodyProxy#each` should return an enumerator when no block is passed

Should it?
 https://github.com/rack/rack/blob/d1363a66ab21738d53bf13cece3652ada0f01976/SPEC#L245-L246

Edouard-chin:
Hmm indeed, didn't read the SPEC. So maybe it's [ActionDispatch::Response::RackBody](https://github.com/rails/rails/blob/20c91119903f70eb19aed33fe78417789dbf070f/actionpack/lib/action_dispatch/http/response.rb#L124) that needs to be fixed.

I opened the PR because I saw an inconsistency between both

rafaelfranca:
It is open for interpretations. If `each` return a enumerator, it still respond to each and only yield string values.

But most of the body objects don't return an enumerator, except for Rails' `ActionDispatch::Response::Buffer`.

Edouard-chin:
Doing a spring clean of my open PR, I have removed the code to `BodyProxy#each` since it's open for interpretation and we fixed the code in our own app at some point instead.

This PR now just corrects the installation instruction which is currently broken

PR #1238, opened 2018-02-26
Attempt to fix the ruby head build

eileencodes:
Debugging builds with this PR

---

Always install bundler so it gets updated. This will fix the ruby-head
build that's failing on master and 1-6-stable with

```
bundler: failed to load command: rake
(/home/travis/build/rack/rack/vendor/bundle/ruby/2.6.0/bin/rake)
LoadError: cannot load such file -- bundler/dep_proxy
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ addons:
 
 before_install:
   - gem env version | grep '^\(2\|1.\(8\|9\|[0-9][0-9]\)\)' || gem update --system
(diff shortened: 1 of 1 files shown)


rafaelfranca:
Ah, I have a branch that fix it https://github.com/rack/rack/compare/fix-travis. It is missing only clean up the commit

eileencodes:
@rafaelfranca cool I'll close this one then

PR #1247, opened 2018-04-11
Update URLMap Spec

### LINKED ISSUE #990: Rails App with map and X-Forwarded-Host fails
# Steps to reproduce:

1. Create a rails app: `rails new mapbug`
2. Use a config.ru with rails inside a map block (sample below)
3. Send a request with X-Forwarded-Host (example curl request below)

# What happens:

the app 404s without ever reaching rails (or at least without logging?)

# What I expect to happen:

the app returns the webpage

# Sample config.ru:

```
require ::File.expand_path('../config/environment',  __FILE__)

map "/" do
  run Rails.application
end
```

# Curl examples

This curl request works:

`curl -s http://localhost:3000 | head`

But this one will throw a 404 and rails will not log a request:

`curl --header "X-Forwarded-Host: localhost" -s http://localhost:3000`

I'm not 100% certain that rack is at fault here, but since Rails isn't logging the request that's what I'm guessing? If you can tell me how to debug this setup to see if the problem isn't WEBrick or Rails I'd be happy to do so, this is a very mystifying bug to me.

Jamie0:
This PR modifies URLMap's functionality to fix #990.

Currently, when using URLMap, if you request a HTTP_HOST that isn't technically valid, instead of defaulting to whatever the first route is without a specified host, it fails, returning a 404. This is mainly apparent on reverse proxies (like mod_proxy) that add the HTTP_X_FORWARDED_HOST header. 

I can't say I'm wholly sure of every URLMap use case, but it passes all existing tests, plus two extra assertions. There aren't any existing tests for regex-based routes, so I'm not sure how it will work on those.

### DIFF
--- a/lib/rack/urlmap.rb
+++ b/lib/rack/urlmap.rb
@@ -20,9 +20,11 @@ module Rack
     end
 
     def remap(map)
+      @known_hosts = []
       @mapping = map.map { |location, app|
         if location =~ %r{\Ahttps?://(.*?)(/.*)}
           host, location = $1, $2
+          @known_hosts << host
         else
           host = nil
         end
@@ -50,10 +52,13 @@ module Rack
       is_same_server = casecmp?(http_host, server_name) ||
                        casecmp?(http_host, "#{server_name}:#{server_port}")
 
+      is_host_known = @known_hosts.include? http_host
+
       @mapping.each do |host, location, match, app|
         unless casecmp?(http_host, host) \
             || casecmp?(server_name, host) \
-            || (!host && is_same_server)
+            || (!host && is_same_server) \
+            || (!host && !is_host_known) # If we don't have a matching host, default to the first without a specified host
           next
         end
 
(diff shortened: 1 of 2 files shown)


lib/rack/urlmap.rb
```diff
@@ -75,7 +80,6 @@ def call(env)
       env[SCRIPT_NAME] = script_name
     end
 
-    private
```
rafaelfranca: Any reason why `casecmp?` was made public? It appears to be used only in this class.

Jamie0: Ah, typo. 

Originally moved the contents of `@mapping.each` out to its own private method, but this would have added some unnecessary complexity.

lib/rack/urlmap.rb
```diff
@@ -20,9 +20,11 @@ def initialize(map = {})
     end
 
     def remap(map)
+      @known_hosts = []
```
tenderlove: Should we make this a `Set`?  I'm not sure how many known hosts a typical app is going to deal with, so maybe an array is fine.

Jamie0: A `Set` isn't a bad shout at all.

PR #1252, opened 2018-04-16
Support unicode byte order mark in config.ru

### LINKED ISSUE #1021: Strip BOM from the beginning of the rackup file
Some editors insert a [BOM (Byte Order Mark)](https://en.wikipedia.org/wiki/Byte_order_mark) at the beginning of each file, which causes issues that are [difficult to debug](http://stackoverflow.com/q/35635029/157328).

To reproduce, first generate a `config.ru` file with this script:

``` ruby
File.open('config.ru', 'w') do |f|
  f.write "\xEF\xBB\xBF" # UTF8 BOM
  f.puts "require 'date'"
  f.puts "run -> (env) { [200, {}, ['OK']] }"
end
```

Then run it with `rackup`:

```
rackup config.ru
```

This gives:

```
config.ru:1:in `block in <main>': undefined method `require' for #<Rack::Builder:0x007ffe5b885368> (NoMethodError)
```

And you'll be wondering why the heck is `"require"` not defined, when it's actually `"\xEF\xBB\xBFrequire"`.

Ruby seems to handle it well, the problem doesn't manifest with `ruby config.ru`.

mikegee:
fixes #1021

### DIFF
--- a/lib/rack/builder.rb
+++ b/lib/rack/builder.rb
@@ -31,10 +31,13 @@ module Rack
   # You can use +map+ to construct a Rack::URLMap in a convenient way.
 
   class Builder
+    UTF_8_BOM = '\xef\xbb\xbf'
+
     def self.parse_file(config, opts = Server::Options.new)
       options = {}
       if config =~ /\.ru$/
         cfgfile = ::File.read(config)
+        cfgfile.slice!(/\A#{UTF_8_BOM}/) if cfgfile.encoding == Encoding::UTF_8
         if cfgfile[/^#\\(.*)/] && opts
           options = opts.parse! $1.split(/\s+/)
         end
--- a/test/builder/bom.ru
+++ b/test/builder/bom.ru
@@ -1,0 +1,1 @@
+﻿run -> (env) { [200, {'Content-Type' => 'text/plain'}, ['OK']] }
--- a/test/spec_builder.rb
+++ b/test/spec_builder.rb
@@ -199,7 +199,7 @@ describe Rack::Builder do
       end)
       o = Object.new
       def o.call(env)
-        @a = 1 if env['PATH_INFO'] == '/b'; 
+        @a = 1 if env['PATH_INFO'] == '/b';
         [200, {}, []]
       end
       run o
@@ -257,6 +257,11 @@ describe Rack::Builder do
     it "sets __LINE__ correctly" do
       app, _ = Rack::Builder.parse_file config_file('line.ru')
       Rack::MockRequest.new(app).get("/").body.to_s.must_equal '1'
+    end
+
+    it "strips leading unicode byte order mark when present" do
+      app, _ = Rack::Builder.parse_file config_file('bom.ru')
+      Rack::MockRequest.new(app).get("/").body.to_s.must_equal 'OK'
     end
   end
 


lib/rack/builder.rb
```diff
@@ -34,7 +34,7 @@ class Builder
     def self.parse_file(config, opts = Server::Options.new)
       options = {}
       if config =~ /\.ru$/
-        cfgfile = ::File.read(config)
+        cfgfile = ::File.read(config, mode: 'r:bom|utf-8')
```
jeremy: Does this introduce a unstated requirement that `config.ru` is UTF-8?

mikegee: Seems that way, doesn’t it? I’ll concoct a utf-16 config.ru or perhaps another encoding and see what happens. 

Thanks for giving this PR a look!

mikegee: This change does not introduce a new requirement. Instead it makes a previous requirement explicit.

When not told the encoding of the file, `::File.read` assumes the content is UTF-8 and returns a string encoded as UTF-8 even when file is encoded as UTF-16. This fails because when the string is `eval`ed, the first null byte is interpreted as EOF, which leads to a `SyntaxError`.

I'm not sure it's possible to infer the encoding of the `config.ru` file by looking at it, in general. I believe it is preferable to raise a bit earlier with a more reasonable error, as my proposed change would.

tested with Ruby 2.5.1

manveru: Sorry to disappoint, but `::File.read` doesn't assume UTF-8 unless your `LC_CTYPE` environment variable or `Encoding.default_external` say so. So I don't think that this is backwards compatible.

mikegee: Thanks for the clarification on that point, @manveru.

Unfortunately, changing my `Encoding.default_external` to match the content of my `config.ru` only leads to different problems.

First, the simple call to `File.read` fails with "ArgumentError: ASCII incompatible encoding needs binmode".

Second, `File.read` with `mode: 'rb'`, doesn't assume the content matches my default external encoding and provides `ASCII-8BIT` instead.

Finally, if I explicitly declare the content type for `File.read` with `mode: 'rb:utf-16le'`, it fails to successfully concatenate with the literal strings in the next method.
`Encoding::CompatibilityError: incompatible character encodings: UTF-8 and UTF-16LE`

Additional pointers on these issues are welcome.

manveru: I don't think we should support non-ASCII compatible encodings for source code. There is no way to set `LC_*` to UTF-16.
Also try to run the following file:

    # Encoding: UTF-16LE
    p 'hi'.encoding

And see that Ruby fails with `UTF-16LE is not ASCII compatible (ArgumentError)`.

mikegee: Thanks again, @manveru. I think I've got it now. 

I was able to create a CP1252 `config.ru` with some accented characters and it worked as expected. Assuming `config.ru` is UTF-8 would be a breaking change.

----

Regarding the issue this PR is meant to solve, how do you feel about a solution like:

- if the string encoding is UTF-8 after reading the file,
- strip the BOM with `String` methods or a regexp

?

mikegee:
@jeremy I addressed your feedback. Would you have a moment to check this out again?

lortza:
Thanks for doing this! This problem has definitely sucked hours of my life. 🙌 🎉

PR #1256, opened 2018-04-16
Enable Style/FrozenStringLiteralComment in rubocop

yhirano55:
## Summary

* Enabled `Style/FrozenStringLiteralComment` in rubocop, and executed auto-correct.
* Removed duplicated magic comments in `Rack::RewindableInput`
* Enabled `Layout/LeadingCommentSpace` in rubocop, and executed


.rubocop.yml
```diff
@@ -4,6 +4,24 @@ AllCops:
   Exclude:
     - '**/vendor/**/*'
 
+Style/FrozenStringLiteralComment:
+  Enabled: true
+  EnforcedStyle: always
+  Exclude:
+    - 'example/**/*'
+    - 'test/builder/**/*'
```
rafaelfranca: Why is this rule disabled on tests?

yhirano55: @rafaelfranca Thanks for the comment. You're right. It doesn't need to set `Exclude`. So I've removed it and added `frozen_string_literal` comments to all files.

PR #1261, opened 2018-04-17
Update LICENSE

yhirano55:
## Summary

* Renamed COPYING to MIT-LICENSE. Because currently MIT-LICESE is used in almost gems.
* Updated README, and added license section instead of copyright section.
* Updated author url.
* Updated copyright's end of the term be ~"present"~ 2018.
* Updated Django license ([link](https://github.com/django/django))

### DIFF
--- a/MIT-LICENSE
+++ b/MIT-LICENSE
@@ -1,6 +1,6 @@
 The MIT License (MIT)
 
-Copyright (C) 2007-present Christian Neukirchen <http://chneukirchen.org/infopage.html>
+Copyright (C) 2007-2018 Christian Neukirchen <http://chneukirchen.org/infopage.html>
 
 Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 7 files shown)


MIT-LICENSE
```diff
@@ -1,4 +1,6 @@
-Copyright (c) 2007-2016 Christian Neukirchen <purl.org/net/chneukirchen>
+The MIT License (MIT)
+
+Copyright (C) 2007-present Christian Neukirchen <http://chneukirchen.org/infopage.html>
```
rafaelfranca: Unfortunately we need to be explicit about the year. So it should be 2007-2018

yhirano55: Noted with thanks. I fixed it. be698cb

jeremy:
Loving your close & careful attention to improving the Rack project. Thank you, @yhirano55!! ❤️

PR #1271, opened 2018-04-26
Add trailer headers to Rack

eileencodes:
Trailer headers allow the sender to add additional information to a
chunked message.

An example message would look like this (taken from mozilla docs:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Trailer)

```
HTTP/1.1 200 OK
Content-Type: text/plain
Transfer-Encoding: chunked
Trailer: Expires

7\r\n
Mozilla\r\n
9\r\n
Developer\r\n
7\r\n
Network\r\n
0\r\n
Expires: Wed, 21 Oct 2015 07:28:00 GMT\r\n
\r\n
```

This could be useful for sending an expiration but other trailer headers
as well like server timing.

While this works with cURL while implementing this we found that Chrome
doesn't support trailer headers (it seems like it might be supported for
HTTP/2 but I had trouble proving that). Once chrome does support trailer
headers we could implement server-timing and use the network toolbar
to get performance metrics from chunked messages. This post uses a
server-timing header, but if we can verify that Chrome support trailers
we could use this header in the trailer instead of a direct header sent.
See https://ma.ttias.be/server-timings-chrome-devtools/

Co-authored-by: Aaron Patterson <aaron.patterson@gmail.com>

cc/ @tenderlove

### DIFF
--- a/lib/rack/chunked.rb
+++ b/lib/rack/chunked.rb
@@ -12,7 +12,7 @@ module Rack
     # A body wrapper that emits chunked responses
     class Body
       TERM = "\r\n"
-      TAIL = "0#{TERM}#{TERM}"
+      TAIL = "0#{TERM}"
 
       include Rack::Utils
 
@@ -20,7 +20,7 @@ module Rack
         @body = body
       end
 
-      def each
+      def each(&block)
         term = TERM
         @body.each do |chunk|
           size = chunk.bytesize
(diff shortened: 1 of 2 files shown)


tenderlove:
It would be really nice to send the server timing header as a trailer (so that we can time anything happening when calculating the response), but it seems like Chrome doesn't support it.  The [spec mentions parsing strategies for trailers](https://www.w3.org/TR/server-timing/#-dfn-server-timing-header-parsing-algorithm-dfn-), but I guess that doesn't indicate whether or not browsers will support trailers at all (though [MDN seems to indicate all browsers support trailers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Trailer)).

@surma Do you know if Chrome support this in the future?

ioquatix:
I didn't know about this, but it should be trivial to implement in Falcon, I will add support.

tenderlove:
Looks like Firefox [may support the Server-Timing trailer](https://bugzilla.mozilla.org/show_bug.cgi?id=1413999)

olleolleolle:
Here's a closed-as-fixed [Chromium issue on "HTTP/2 trailers"](https://bugs.chromium.org/p/chromium/issues/detail?id=422958).

PR #1272, opened 2018-04-28
Add WebSocket and SSE options

ohler55:
This suggested addition is the result of a collaboration between @boazsegev of Iodine fame and myself. Both Agoo and Iodine will support these suggested additions. At this point I don't think any other servers are supporting WebSockets and SSE so this would set the standard for others to follow.

### DIFF
--- a/SPEC
+++ b/SPEC
@@ -119,16 +119,16 @@ Servers that choose to support WebSockets and SSE connections should follow thes
                          :websocket for a WebSocket upgrade, or :sse for an SSE upgrade.
 <tt>rack.upgrade</tt>:: Is used to pass a handler instance or class back to the server for WebSocket and SSE connections.
                      The handler MAY implement any or all of the following callbacks:
-                      on_open(client)      # called when the connection is complete with a client object.
-                      on_message(message)  # called when a WebSocket message is received by server. <tt>message</tt> will be UTF-8 encoded for text messages and Binary encoded for binary messages.
-                      on_shutdown          # may be called before a connection is closed due to server shutdown.
-                      on_close             # called when the connection is closed
-                      on_drained(client)   # may be called when the number of pending writes drops to zero.
+                      on_open(client)              # called when the connection is complete with a client object.
+                      on_message(client, message)  # called when a WebSocket message is received by server. <tt>message</tt> will be UTF-8 encoded for text messages and Binary encoded for binary messages.
+                      on_shutdown                  # may be called before a connection is closed due to server shutdown.
+                      on_close                     # called when the connection is closed
+                      on_drained(client)           # may be called when the number of pending writes drops to zero.
                      The <tt>client</tt> is used for writing and checking status of the upgraded connect. It has these methods.
-                      write(message)       # writes to the WebSocket or SSE connection
-                      close()              # forces a close of the WebSocket or SSE connection
-                      open?()              # returns true if the connection is open, false otherwise
-                      pending()            # returns the number of pending writes or -1 if the connection is closed
+                      write(message)               # writes to the WebSocket or SSE connection
+                      close()                      # forces a close of the WebSocket or SSE connection
+                      open?()                      # returns true if the connection is open, false otherwise
+                      pending()                    # returns the number of pending writes or -1 if the connection is closed
                      The <tt>env['rack.upgrade']</tt> option should only be set it the environment options has a non-nil value for the <tt>rack.upgrade?</tt> option,
                      If the response status is 300 or higher, the server MUST ignore the <tt>rack.upgrade</tt> value (send the response without performing an upgrade).
 
--- a/SPEC
+++ b/SPEC
@@ -121,8 +121,8 @@ Servers that choose to support WebSockets and SSE connections should follow thes
                      The handler MAY implement any or all of the following callbacks:
                       on_open(client)              # called when the connection is complete with a client object.
                       on_message(client, message)  # called when a WebSocket message is received by server. <tt>message</tt> will be UTF-8 encoded for text messages and Binary encoded for binary messages.
-                      on_shutdown                  # may be called before a connection is closed due to server shutdown.
-                      on_close                     # called when the connection is closed
+                      on_shutdown(client)          # may be called before a connection is closed due to server shutdown.
+                      on_close(client)             # called after the connection is closed.
                       on_drained(client)           # may be called when the number of pending writes drops to zero.
                      The <tt>client</tt> is used for writing and checking status of the upgraded connect. It has these methods.
                       write(message)               # writes to the WebSocket or SSE connection
--- a/SPEC
+++ b/SPEC
@@ -119,17 +119,17 @@ Servers that choose to support WebSockets and SSE connections should follow thes
                          :websocket for a WebSocket upgrade, or :sse for an SSE upgrade.
 <tt>rack.upgrade</tt>:: Is used to pass a handler instance or class back to the server for WebSocket and SSE connections.
                      The handler MAY implement any or all of the following callbacks:
-                      on_open()                        # called when the connection is complete
-                      on_message(message)              # called when a WebSocket message is received by server. <tt>message</tt> will be UTF-8 encoded for text messages and Binary encoded for binary messages.
-                      on_shutdown()                    # may be called before a connection is closed due to server shutdown.
-                      on_close()                       # called when the connection is closed
-                      on_drained                       # may be called when the number of pending writes drops to zero.
-                     The object will be extended by the server to include:
-                      write(message)  # writes to the WebSocket or SSE connection
-                      close()         # forces a close of the WebSocket or SSE connection
-                      open?()         # returns true if the connection is open, false otherwise
-                      pending()       # returns the number of pending writes or -1 if the connection is closed
-                     This option should only be set it the environment options has a non-nil value for the <tt>rack.upgrade?</tt> option,
+                      on_open(client)      # called when the connection is complete with a client object.
+                      on_message(message)  # called when a WebSocket message is received by server. <tt>message</tt> will be UTF-8 encoded for text messages and Binary encoded for binary messages.
+                      on_shutdown          # may be called before a connection is closed due to server shutdown.
+                      on_close             # called when the connection is closed
+                      on_drained(client)   # may be called when the number of pending writes drops to zero.
+                     The <tt>client</tt> is used for writing and checking status of the upgraded connect. It has these methods.
+                      write(message)       # writes to the WebSocket or SSE connection
+                      close()              # forces a close of the WebSocket or SSE connection
+                      open?()              # returns true if the connection is open, false otherwise
+                      pending()            # returns the number of pending writes or -1 if the connection is closed
+                     The <tt>env['rack.upgrade']</tt> option should only be set it the environment options has a non-nil value for the <tt>rack.upgrade?</tt> option,
                      If the response status is 300 or higher, the server MUST ignore the <tt>rack.upgrade</tt> value (send the response without performing an upgrade).
 
 The server or the application can store their own data in the
--- a/SPEC
+++ b/SPEC
@@ -113,6 +113,25 @@ be implemented by the server.
                         fatal(message, &block)
 <tt>rack.multipart.buffer_size</tt>:: An Integer hint to the multipart parser as to what chunk size to use for reads and writes.
 <tt>rack.multipart.tempfile_factory</tt>:: An object responding to #call with two arguments, the filename and content_type given for the multipart form field, and returning an IO-like object that responds to #<< and optionally #rewind. This factory will be used to instantiate the tempfile for each multipart form file upload field, rather than the default class of Tempfile.
+
+Servers that choose to support WebSockets and SSE connections should follow these additional environment specifications:
+<tt>rack.upgrade?</tt>:: Is nil or missing if no connection upgrade is possible,
+                         :websocket for a WebSocket upgrade, or :sse for an SSE upgrade.
+<tt>rack.upgrade</tt>:: Is used to pass a handler instance or class back to the server for WebSocket and SSE connections.
+                     The handler MAY implement any or all of the following callbacks:
+                      on_open()                        # called when the connection is complete
+                      on_message(message)              # called when a WebSocket message is received by server. <tt>message</tt> will be UTF-8 encoded for text messages and Binary encoded for binary messages.
+                      on_shutdown()                    # may be called before a connection is closed due to server shutdown.
+                      on_close()                       # called when the connection is closed
+                      on_drained                       # may be called when the number of pending writes drops to zero.
+                     The object will be extended by the server to include:
+                      write(message)  # writes to the WebSocket or SSE connection
+                      close()         # forces a close of the WebSocket or SSE connection
+                      open?()         # returns true if the connection is open, false otherwise
+                      pending()       # returns the number of pending writes or -1 if the connection is closed
+                     This option should only be set it the environment options has a non-nil value for the <tt>rack.upgrade?</tt> option,
+                     If the response status is 300 or higher, the server MUST ignore the <tt>rack.upgrade</tt> value (send the response without performing an upgrade).
+
 The server or the application can store their own data in the
 environment, too.  The keys must contain at least one dot,
 and should be prefixed uniquely.  The prefix <tt>rack.</tt>
--- a/SPEC
+++ b/SPEC
@@ -130,6 +130,7 @@ The <tt>client</tt> is used for writing and checking status of the upgraded conn
                       close()                      # forces a close of the WebSocket or SSE connection
                       open?()                      # returns true if the connection is open, false otherwise
                       pending()                    # returns the number of pending writes or -1 if the connection is closed
+                      env()                        # returns the `env` Hash from the inital call to `#call()`. Note some elements may no longer be relevant.
 
 The <tt>env['rack.upgrade']</tt> option should only be set it the environment options has a non-nil value for the <tt>rack.upgrade?</tt> option,
 If the response status is 300 or higher, the server MUST ignore the <tt>rack.upgrade</tt> value (send the response without performing an upgrade).
--- a/SPEC
+++ b/SPEC
@@ -117,7 +117,7 @@ be implemented by the server.
 Servers that choose to support WebSockets and SSE connections should follow these additional environment specifications:
 <tt>rack.upgrade?</tt>:: Is nil or missing if no connection upgrade is possible,
                          :websocket for a WebSocket upgrade, or :sse for an SSE upgrade.
-<tt>rack.upgrade</tt>:: Is used to pass a handler instance or class back to the server for WebSocket and SSE connections.
+<tt>rack.upgrade</tt>:: Is used to pass a handler object back to the server for WebSocket and SSE connections.
                      The handler MAY implement any or all of the following callbacks:
                       on_open(client)              # called when the connection is complete with a client object.
                       on_message(client, message)  # called when a WebSocket message is received by server. <tt>message</tt> will be UTF-8 encoded for text messages and Binary encoded for binary messages.
--- a/SPEC
+++ b/SPEC
@@ -115,22 +115,24 @@ be implemented by the server.
 <tt>rack.multipart.tempfile_factory</tt>:: An object responding to #call with two arguments, the filename and content_type given for the multipart form field, and returning an IO-like object that responds to #<< and optionally #rewind. This factory will be used to instantiate the tempfile for each multipart form file upload field, rather than the default class of Tempfile.
 
 Servers that choose to support WebSockets and SSE connections should follow these additional environment specifications:
-<tt>rack.upgrade?</tt>:: Is nil or missing if no connection upgrade is possible,
-                         :websocket for a WebSocket upgrade, or :sse for an SSE upgrade.
-<tt>rack.upgrade</tt>:: Is used to pass a handler object back to the server for WebSocket and SSE connections.
-                     The handler MAY implement any or all of the following callbacks:
+<tt>rack.upgrade?</tt>:: Is nil or missing if no connection upgrade is possible, :websocket for a WebSocket upgrade, or :sse for an SSE upgrade.
+<tt>rack.upgrade</tt>:: Is used to pass a handler object back to the server for WebSocket and SSE connections. The handler MAY implement any or all of the following callbacks.
                       on_open(client)              # called when the connection is complete with a client object.
-                      on_message(client, message)  # called when a WebSocket message is received by server. <tt>message</tt> will be UTF-8 encoded for text messages and Binary encoded for binary messages.
+                      on_message(client, message)  # called when a WebSocket message is received by server.
+                                                   # <tt>message</tt> will be UTF-8 encoded for text messages
+                                                   # and Binary encoded for binary messages.
                       on_shutdown(client)          # may be called before a connection is closed due to server shutdown.
                       on_close(client)             # called after the connection is closed.
                       on_drained(client)           # may be called when the number of pending writes drops to zero.
-                     The <tt>client</tt> is used for writing and checking status of the upgraded connect. It has these methods.
+
+The <tt>client</tt> is used for writing and checking status of the upgraded connect. It has these methods.
                       write(message)               # writes to the WebSocket or SSE connection
                       close()                      # forces a close of the WebSocket or SSE connection
                       open?()                      # returns true if the connection is open, false otherwise
                       pending()                    # returns the number of pending writes or -1 if the connection is closed
-                     The <tt>env['rack.upgrade']</tt> option should only be set it the environment options has a non-nil value for the <tt>rack.upgrade?</tt> option,
-                     If the response status is 300 or higher, the server MUST ignore the <tt>rack.upgrade</tt> value (send the response without performing an upgrade).
+
+The <tt>env['rack.upgrade']</tt> option should only be set it the environment options has a non-nil value for the <tt>rack.upgrade?</tt> option,
+If the response status is 300 or higher, the server MUST ignore the <tt>rack.upgrade</tt> value (send the response without performing an upgrade).
 
 The server or the application can store their own data in the
 environment, too.  The keys must contain at least one dot,


ioquatix:
Interesting.

I already support completely transparent asynchronous web sockets in [falcon](https://github.com/socketry/falcon). It uses `rack.hijack` just fine. An example is given here: https://github.com/socketry/async-websocket/blob/master/examples/chat/config.ru

Of course with this model supporting SSE is also trivial.

Regarding the event driven model which is being exposed via `rack.upgrade`: I don't think exposing an event driven API is a great way to make systems asynchronous. It's kind of limiting.

Regarding the overall design - careful thought needs to be given to how HTTP/2 fits into this picture. Ideally, SSE, WebSockets, and other "interactive" streams, transit via reading from the request body and writing to the response body. That's the model which I feel works best with HTTP/2, where each request/response is encapsulated into a full duplex stream.

ioquatix:
Bare in mind that `async-websocket` doesn't support web sockets over http/2 yet, this is something I'll work on in the coming months.

This is also one of the challenges of implementing `rack.hijack` for HTTP/2. You need to be careful of what it means, exactly, to hijack a connection (or stream in the case of HTTP/2).

boazsegev:
I think the HTTP/2 layer is something that's handled only by the server. It shouldn't change the `rack.upgrade` semantics since it isn't one of the protocols / behaviors that should be handled by the application.

I think the idea is great because it decouples the application from any network logic.

While `hijack` solutions remove the server, changing the relationship from `network<=>server<=>application` to `network<=>application`, this approach keeps the application away from the network, keeping the initial (and desired) relationship intact.

ohler55:
I'd like to point out that Rack is event based as are the WebSocket and SSE API/specs. The Rack hijack offers a streaming back door that is not addressed by the spec. This PR attempts to formalize an approach to using WebSocket and SSE that is consistent with the use pattern common in modern web applications with JavaScript front ends. It is not an attempt to get rid of the hijack option. There is certainly reason to continue support for hijacking but that should not be a reason to reject this PR.

boazsegev:
@ioquatix , you're totally right, I can't implement that code... nobody can unless they implement `Async::Task` and add that as a Rack dependency.

As for streaming solutions: SSE **is** a streaming solution.

If you refer to video streaming or binary data requirements, WebSockets are perfect for that.

If you meant old-school HTTP streaming... yes, that's true. I'm not even sure that I should, but I'm aware that I don't. In fact, I'm pretty sure [the Rack model wouldn't be ideal for streaming](http://blog.plataformatec.com.br/2012/06/why-your-web-framework-should-not-adopt-rack-api/) anyway.

ioquatix:
> I'd like to point out that Rack is event based

[I'm not convinced by this](https://github.com/rack/rack/search?utf8=✓&q=on_&type=)

If anything, rack is based on composable middleware. There is no event handling.

> The Rack hijack offers a streaming back door that is not addressed by the spec

Sorry, but this is complete dump trucks: https://github.com/rack/rack/blob/154ac5255323021ef74039e11a0a7376af889eaa/SPEC#L87-L96

> This PR attempts to formalize an approach to using WebSocket and SSE that is consistent with the use pattern common in modern web applications with JavaScript front ends

This is the guts of the matter. Why is this necessary? It's clearly already possible to do it - so we have to ask the question, do we need to formalise this in rack, or is there something simpler that we could formalise which allows more possibilities. We should accept or reject this PR based on the merits of what it adds to Rack, as a specification. I think there are better, more generic, options.

ioquatix:
> @ioquatix , you're totally right, I can't implement that code... nobody can unless they implement Async::Task and add that as a Rack dependency.

I don't think I proposed that anywhere. I just said, with the current Rack spec, you can do it, and it scales well. We don't need to change the rack spec to have a great implementation of web sockets.

> As for streaming solutions: SSE is a streaming solution.

It only works in one direction and it only works in a very specific context. It's not general response streaming, as in `body.each {|chunk| ... peer.write(chunk) ...}`

> If you refer to video streaming or binary data requirements, WebSockets are perfect for that.

> If you meant old-school HTTP streaming... yes, that's true. I'm not even sure that I should, but I'm aware that I don't. In fact, I'm pretty sure the Rack model wouldn't be ideal for streaming anyway.

The Rack model is great for streaming responses. That's the whole point of the response body responding to `each`.

ioquatix:
By the way, I'm not trying to shit all over your PR, although it probably feels like it. Congratulations for making something, and implementing it and proposing it. Basically, I just don't see the logic of it though. I think there are better things in Rack that need to be addressed, and I think the direction of HTTP/2 confirms this.

ohler55:
Of course Rack is event based. The trigger for all Rack code is the `#call(env)` callback where `env` is the encapsulation of the event that has occurred. That event being an HTTP request. After that you are correct, the handling is based on a middleware model.

boazsegev:
I'm paraphrasing slightly to focus on your question, forgive me and correct me if I misunderstood:

> Why is this necessary? ... is there something **simpler** that we could formalise which allows more possibilities.

I believe this is necessary because `hijack` is complex.

There's a difference between simple and easy. `hijack` is easy - but not simple.

It might simple for Rack and maybe even simple in the server, but it's a super complex solution that requires applications to add IO handling code.

The `hijack` pattern is an anti-DRY and anti-SOLID pattern that breaks away from object oriented approaches and requires applications to become familiar with the networking layers.

Sure, we might need to keep `hijack` around because some things are just impossible without it, but that doesn't make the best solution for everything. 

On the other hand, the callback approach in this PR is **simple**, but it's clearly not **easy**.

Servers will have to put in some work (less work than applications need to put in for `hijack`, but still).

However, the solution is simple and follows both DRY and SOLID principles.

The callback approach isn't a generic solution, it's just good for things that lend themselves to using callbacks, but this is the advantage of this approach.

IMHO, This is a perfect fit for everything callback related and it's extendible (if WebSockets 2.0 come around, nothing needs to be changed on the Rack/Application side, just the server).

boazsegev:
@ioquatix , I was looking through the falcon repo and you seem to have done great work there.

I understand your concerns and your wish to have a better solution. It seems to me that you yourself put in a lot of time and effort into improving the sated of affairs and making things easier and better for everyone.

I appreciate your approach and I find myself slightly surprised by the challenges you raise against this PR.

It seems to me that you're using similar semantics with different method names (i.e., your WebSocket `next_message` seems to block the current thread, allowing other threads to handle other server events until a new message becomes available... or maybe I misunderstood).

Sure, my implementation is evented and your implementation is multi-threaded, but both lend themselves to the same approach - application side "callbacks".

Your implementation could just as easily remain threaded and hide the `next_event` loop away so it calls the correct callback.

I'm not sure I understand your reservations.

ioquatix:
> I believe this is necessary because hijack is complex.

Those are all really good points and I appreciate your thinking.

To be honest, in falcon, implementing your proposed API wouldn't be more than about 100 lines of code, perhaps less. So, it's not complicated to implement, because falcon already has a well defined and good concurrency model. That's the context in which I'm thinking about your PR, by the way.

Whether the complexity is in the server, or the application code, is an interesting point and one that I can agree with to a certain extent. I don't think there is a strong argument either way.

> @ioquatix , I was looking through the falcon repo and you seem to have done great work there.

Thanks, I can tell we are going to get along well now.

> I understand your concerns and your wish to have a better solution. It seems to me that you yourself put in a lot of time and effort into improving the sated of affairs and making things easier and better for everyone.

Yes, but I still haven't made a 1.0 release yet. It's coming soon hopefully.

> I appreciate your approach and I find myself slightly surprised by the challenges you raise against this PR.

Fair enough.

> It seems to me that you're using similar semantics with different method names (i.e., your WebSocket next_message seems to block the current thread, allowing other threads to handle other server events until a new message becomes available... or maybe I misunderstood).

That's almost right. Async doesn't use threads, it uses fibers which are cooperatively scheduled. They have much less overhead than threads.

> Sure, my implementation is evented and your implementation is multi-threaded, but both lend themselves to the same approach - application side "callbacks".

Async is event driven. When a fiber performs an operation that would block, it yields back to the reactor which resumes once the operation can continue.

I deliberately try to avoid callbacks because in my experience they lead to callback hell. The difference is that with callbacks you need to use variables to track state, and each time the callback is invoked you have to process that state to figure out what to do next. With fibers, you naturally resume where you were with your stack intact. You can implement complicated state machines with ease.

> Your implementation could just as easily remain threaded and hide the next_event loop away so it calls the correct callback.

Yes, this is feasible.

> I'm not sure I understand your reservations.

I like this PR from the point of view that it tries to provide some generic interface for web sockets and server sent events. Making a generic interface for that is both simple and difficult. I admire the you've done it and implemented it.

My main reservation is that your proposed API is incomplete. The rack specification as it stands, is really simple and allows for a lot of flexibility. What you've proposed is an extension which fundamentally encodes a model for concurrency into Rack, something that it hasn't had except in a very precise location (response body `#each`).

To look at it another way, `falcon` passes the rack linter, yet it implements real-time streaming requests and responses. So, within the current confines of Rack, we can enjoy these advanced features.

Your proposed API exposes an incomplete model for concurrency and the word that comes to my mind is dangerous. What happens if you stick a `RestClient.get` into `on_message` - does it stall the entire server? How do you communicate between web sockets (in your PR I assume you'd need a shared backend like redis)? Whether or not you agree with it, the node model of single process, multiple connections is one that works very well - you can have an array of web socket connections, send messages to all of them, etc. This model can work well with Rack as shown by `falcon` (e.g. https://github.com/socketry/async-websocket/tree/master/examples/chat).

Whatever way you cut it, Ruby has a pretty frustrating model for concurrency right now (and IO too, but it's slowly getting better). People are working on it, myself, others (Eric, https://bugs.ruby-lang.org/issues/13618), ko1 (guilds) and I have no doubt that good things are coming to this space.

But, Rack, right now, has to be a specification that works with Ruby as it is today. If `Thribers` become part of Ruby (I personally hope not), the way servers should implement web-sockets should be totally different. It might be possible to make a MVP, like this PR, but I think it's loading Rack up with too much complexity.

Personally, what I think we need to focus on is making `rack.input` better suited for reading from HTTP/2 streams, and response `body` better suited for generating streaming responses. Finding the right model here should allow for SSE, WebSockets, and whatever other things

boazsegev:
@ioquatix interesting, thanks for taking the time to explain.

I love the fiber approach, as it hides away the event reactor and make things more comfortable to manage.

> My main reservation is that your proposed API is incomplete.  

Yes, you are right that there are concerns related to the proposed approach. It doesn't deal with IPC or client<=>client communications. It also allows developers to violate multi-threading best practices.

However, this proposal seems pretty balanced. It doesn't enforce a huge change on existing servers and the "missing" pieces are easily complemented by existing solutions (Redis being just one of them).

Personally, I implement pub/sub in iodine to "fill the gap". However, I'm not sure this should be part of the specification since other developers might wish to "fill the gap" using a different approach, such as limiting the server to a single process and using an array.

> What you've proposed is an extension which fundamentally encodes a model for concurrency into Rack...

I'm not sure I understand this part.

The server can be a single threaded server, a thread-per connection server, a fiber based server and practically any type of concurrency model can be implemented.

I'm not sure I see where the PR requires a specific concurrency model.

> What happens if you stick a `RestClient.get` into `on_message` - does it stall the entire server?

That really depends on the server, I guess.

What happens if I call `sleep(10)` while handling an HTTP request (except getting fired, that is)?

It's probably the same answer.

Iodine, for example, supports multi-threading and cluster mode (pretty much the same model as Puma). Agoo (I think) supports multi-threading.

For both of these servers I would recommend avoiding blocking calls within the thread, but they both offer some level of protection before experiencing DoS.

I don't use fibers so I'm not sure how they would react to this scenario. Hopefully they will react better.

Either way, I doubt the question of having fibers replace threads is part of the discussion here.

> Whether or not you agree with it, the node model of single process, multiple connections is one that works very well...

I agree with it, but I don't see why that matters.

Even the node model needs to be adjusted when scaling horizontally.

All models have properties that define their strengths and weaknesses.

IMHO, this PR proposes a mechanism that's independent of model used by the server, making it flexible enough for everyone to implement.

ioquatix:
> However, this proposal seems pretty balanced. It doesn't enforce a huge change on existing servers and the "missing" pieces are easily complemented by existing solutions (Redis being just one of them).

How do servers support your proposal? By pulling in `websocket-driver`? It's a large surface area, and it does get pretty tricky. I wouldn't call it balanced - it's heavily biased towards implementing WebSockets.

> I'm not sure I understand this part.
> The server can be a single threaded server, a thread-per connection server, a fiber based server and practically any type of concurrency model can be implemented.
> I'm not sure I see where the PR requires a specific concurrency model.

Invoking a Rack app is for the most part trivial. It's a function call that returns a result. No real model for concurrency is needed for this basic definition. You scale up by executing the function on different CPU cores, but fundamentally you can't change it to a non-linear event driven callback model or some other asynchronous model (`async` provides transparent inversion of control back to the reactor, but it doesn't change or require changes in flow control).

This proposal embeds non-linear flow control into the Rack spec. What I mean is, it's impossible to implement the given proposal without some kind concurrency. I'm not saying that any particular model for concurrency is being encoded, but just that by your approach a model for currency is now necessary.

This has a huge down-stream effect, since all code that depends on Rack now has to be aware of and capable of asynchronous execution. For example, how would you change `rack-test` to support this PR?

I think this equally applies to `rack.hijack`, and I also don't like that approach, but it's pretty much set in stone now. For example, how do you implement `rack.hijack` with HTTP/2? It's probably not possible in the general sense.

> What happens if I call sleep(10) while handling an HTTP request (except getting fired, that is)?

In `async` provided you call `task.sleep(10)` the fiber defers for 10 seconds. It doesn't block the server. In `puma` it would block the worker thread.

Have you tried using ActiveRecord in your `on_message` callback? How did that scale for you? My experience is that the ActiveRecord `ConnectionPool` design is very poor for highly concurrent workloads. These are very tricky issues to get right (although you can do it as shown here: https://github.com/socketry/async-postgres and it does scale up pretty well).

> [regarding node] I agree with it, but I don't see why that matters.
> Even the node model needs to be adjusted when scaling horizontally.

Because with your proposed API, implementing a basic Node style server like this is not possible. Right from the get go you need additional machinery to do pub/sub or other kinds of communication. Even the situation with streaming responses is not improved without additional work. It's a very specific proposal designed for a very specific kind of scalability. It's far too specific IMHO.

What we need is a proposal that better aligns with HTTP/2 streams since it's clear to me that it should be the future of HTTP and Rack. It should have a clear model for concurrency that fits with the existing multiprocess/multithread/worker implementations - i.e. reading from `rack.input` might block the request and writing to the response body might block the request due to buffering. On top of that, as already demonstrated, you can implement highly scalable asynchronous servers.

Streaming request and response bodies in `falcon` directly improve the latency of existing apps with no changes. Here is an implementation of `rack.input` which streams (and buffers) the input: https://github.com/socketry/falcon/blob/master/lib/falcon/adapters/input.rb

But this PR requires significant changes to existing apps for any kind of benefit. Not only that, but it only supports a very specific kind of scalability with an under-specified concurrency model (i.e. what happens if you block in `on_message`).

Let me finish with the following question: Do even think there is a future for WebSockets? https://datatracker.ietf.org/doc/draft-hirano-httpbis-websocket-over-http2/ hasn't been touched since 2014. There is an interesting write up here: https://daniel.haxx.se/blog/2016/06/15/no-WebSockets-over-http2/ - WebSockets are something which has never been a good fit for the request/response paradigm and that's something which fundamentally underpins Rack.

SPEC
```diff
@@ -113,6 +113,25 @@ be implemented by the server.
...
+<tt>rack.upgrade?</tt>:: Is nil or missing if no connection upgrade is possible,
+                         :websocket for a WebSocket upgrade, or :sse for an SSE upgrade.
+<tt>rack.upgrade</tt>:: Is used to pass a handler instance or class back to the server for WebSocket and SSE connections.
+                     The handler MAY implement any or all of the following callbacks:
+                      on_open()                        # called when the connection is complete
+                      on_message(message)              # called when a WebSocket message is received by server. <tt>message</tt> will be UTF-8 encoded for text messages and Binary encoded for binary messages.
+                      on_shutdown()                    # may be called before a connection is closed due to server shutdown.
+                      on_close()                       # called when the connection is closed
+                      on_drained                       # may be called when the number of pending writes drops to zero.
+                     The object will be extended by the server to include:
```
tenderlove: This will break method caches.  I'd prefer a solution that doesn't lean on runtime method extension.  Maybe something like this:

```ruby
class MyWSEObject
  def initialize io
    @io = io
  end

  def on_open
    @io.write("neat")
  end
end

app = lambda do |env|
  env['rack.upgrade'] = MyWSEObject.new(env['rack.wse.object'])
end
```

ioquatix:
```
class MyWSEObject
  def on_open(io)
    io.write("neat")
  end
end

app = lambda do |env|
  env['rack.upgrade'] = MyWSEObject.new
end
```

would be simpler

tenderlove: That would be fine too. 😄

ioquatix: So, the next question, is it a good idea to write into the request `env` to essentially send a response? To me, that's confusing. Perhaps at best it could be something like

```ruby
if upgrade = env['rack.upgrade']
  return upgrade.call(MyWSEObject.new)
end

return [400, {}, []]
```

evanphx:
Hi all! Websockets and rack again, let's do this!

Up front, last time this came up, I prototyped it in puma (https://github.com/puma/puma/pull/1054). It was an experiment we didn't merge in, but it totally works and we might revisit it.

The API of on_* methods for events is totally fine, no issue there.

Extending the object to decorate it with a write method is pretty ugly, from both an implementation statement as well as a performance one. Passing an object that implements #write to the on_* methods to write back is less code and performs better.

If frameworks want to take that object and use extend to make it's methods available on the handler, that's totally fine and up to the framework. Rack should not do that on it's own, it should provide a lower requirement.

SamSaffron:
You should be able to opt to the "type" of upgrade you are after

I can already see 2 use cases that this protocol does not help at all and will be stuff on hijack. 

1. https://github.com/discourse/discourse/blob/master/lib/hijack.rb 

With the current example use of:

https://github.com/discourse/discourse/blob/master/app/controllers/user_avatars_controller.rb#L71-L78

2. Messagebus also controls transport via: https://github.com/SamSaffron/message_bus/blob/master/lib/message_bus/client.rb and there is no clean protocol here for a "chunked encoding" http req short of SSE (which happens to fit sort of except for demanding `"text/event-stream"` as the content type

I would recommend

```
class Raw
  def type
      :raw  # or :sse or :socket
   end

  def on_open(io)
      # custom header support
      io.write_headers({  }) 
     
      # transparent chunked encoding support
      io.write_chunk("a chunk") 
  end
end

 env['rack.raw'] = Raw.new 
```

I am not super happy about `rack.upgrade` as a name cause it very tied to web sockets kind of terminology

SPEC
tenderlove: TBH it doesn't really matter to me.  Unfortunately though, `env` is really the only way that middleware can communicate.  So writing to `env` is the only way that one middleware could indicate to the next "I'm responding by writing to a websocket".  If the intention is that any "callable" (middleware / app) that sets the WSE object must not call anything else in the middleware chain, then that needs to be explicitly stated in SPEC.

ioquatix:
Rather than `env['rack.raw'] = Raw.new`, why not just `return [200, {}, Raw.new]` and have servers do different things depending on `#type`.

SamSaffron:
Another thing that makes me somewhat uneasy here is that it still suffers from the very ugly trait that deep down the middleware stack someone returns rubbish to all the rest of the pieces that then gets thrown in bin.

So... to mitigate this very bad problem I would recommend:

```
  # only gets called AFTER everything walked down the middleware stack. 
  def on_open(io, env, response)
      status, headers, result = response

      # custom header support
      io.write_headers({  }) 
     
      # transparent chunked encoding support
      io.write_chunk("a chunk") 
  end
```

Overall I am leaning toward not even mucking with rack long term for a lot of this stuff and moving to fiber servers that can pause and resume more cleanly, multiplexing tons of connections cleanly. A big blocker there though is the MRI does not let you ship fibers between threads now which is not idea. 

That said this works ... so ... yay ?https://github.com/SamSaffron/performance/blob/master/fiber_server/server.ru

SPEC
boazsegev: I understand the need to maintain the method caches, especially as we move towards JIT compilation.

As far as iodine is concerned, I am totally flexible on all points as long as the negative impact on performance isn't too high.

On the other hand, please consider a number of good reasons to prefer the `extend` approach over the `on_open(io)`:

1. It's likely that the method caches will break only the first time a callback object is returned.

   Assuming the same object class is used for all callback objects, the `extend` operations becomes a no-op, keeping the method caches intact (since the class was already extended).

2. Using an `io` object adds an obvious requirement for the `io` object to be stored somewhere by the receiver (otherwise, `write` could never get called).

   This makes the `on_open(io)` method a requirement rather than optional and it adds a required use-pattern (`def on_open(io); @io=io; end`).

   IMHO, although not always the case, this is a strong indication of a design flaw (requiring the user to complete the design).

3. For C extensions, there's double the amount of objects per connection. For Ruby implementations it's an increase of 150% in object count (I implemented in C, so my Ruby estimates are theoretical assuming 3 objects: raw IO, wrapper IO and callback objects).

   This puts an extra burden on the memory management system and hastens the (inevitable?) memory fragmentation issue that many Ruby servers suffer from.

4. As a side note, this might make future extensions harder, since future gem authors wouldn't be sure how to access the `io` object from within a connection.

I noticed that @evanphx 's experience is the opposite of mine.

While I found reduced performance and higher complexity when using an `io` object, @evanphx reports improved performance when using this approach.

SamSaffron:
Also... one extra big question here is ... do we want to bloat all of our Rack servers out there with all this extra logic when simply adding a single dependency to middlware that implements this today is already feasible on top of hijack? 

should we not start with `gem install magic_generic_web_socket_support_middleware` prior to pushing on implementers to all implement the same thing?

SPEC
boazsegev: As for the `.call`, I find it confusing.

Let's assume I want to upgrade to SSE (or WebSockets) as well as set a cookie.

When using `call` (besides the extra Proc object), it's unclear to me when the `call` is actually performed. Is the upgrade immediate? Are the headers in the final response are sent?

Passing an object is both cheaper (no Proc object) and, I feel, simpler.

As for `[200, {}, MyObject]`, I think this breaks existing servers, so I'm pretty sure it wouldn't work as smoothly.

ioquatix:
> A big blocker there though is the MRI does not let you ship fibers between threads now which is not ideal.

Fibers are actually bound to the thread that creates them by definition.

Some designs use a more generic coroutine structure (e.g. green threads) which you can move between threads. If you are interested in this, you might find https://bugs.ruby-lang.org/issues/13618 interesting.

If you are interested in Fiber based servers, check out falcon. https://github.com/socketry/falcon/blob/master/examples/beer/config.ru and https://github.com/socketry/ which is a complete stack of asynchronous components.

> Also... one extra big question here is ... do we want to bloat all of our Rack servers out there with all this extra logic when simply adding a single dependency to middlware that implements this today is already feasible on top of hijack?

Yes, I strongly agree with this, and the answer IMHO is no.

SPEC
ioquatix: > As for [200, {}, MyObject], I think this breaks existing servers, so I'm pretty sure it wouldn't work as smoothly.

As long as MyObject responds to `#each` as a fallback, it won't break existing servers.

SamSaffron:
yes, I am across Eric's work there and hope we get something in MRI, I also think Koichi is open to having a protocol for allowing to move Fibers between threads, and I want to see proper green threads back. 

Also seen Falcon, I agree that a fiber based server is very very appealing for a bunch of workloads it simplifies so much of the mess we have now with pure threaded servers and slow requests, especially cause you can walk down the middleware stack properly in the right time.

tenderlove:
> should we not start with gem install magic_generic_web_socket_support_middleware prior to pushing on implementers to all implement the same thing?

TBH I'm OK with it if it's an opt-in.  This proposal seems simple enough that someone could implement a "hijack" based approach using a middleware.  The upside of formalizing it is that if webservers want to implement a high performance version of `magic_generic_web_socket_support_middleware` can.  I think of it as similar to `X-Sendfile`: yes we have a middleware that can do it, but webservers can implement an accelerated version since we have the standard.

SPEC
boazsegev: > > For C extensions, there's double the amount of objects per connection.

> Why is this?

@tenderlove , I assume you know that the C server doesn't need an IO object, it uses a simple integer to keep track of IO.

This means that a C extension just needs to keep the callback object itself and add a secret variable to that object with the socket number.

This secret variable isn't possible in Ruby, but C extensions are allowed to create them. Since that variable is a Number, it's very small (it doesn't create an Object in the ObjectSpace realm and it's immutable).

This means a single object in C (the callback object). If we add the IO object, it's two objects - a 200% increase in object count.

In Ruby, we will have the callback handler, the raw IO object (both pre-existing) and an IO wrapper, a 150% increase in object count.

ohler55:
@SamSaffron in regard to on_open taking an io, env, and repsonse. By the time `on_open` is called the connection has already been established. Since it is either WebSocket or SSE a response and env are no longer relevant. The intent was to allow middleware to decide if a connection shoulc be established based on the return status in the `#call(env)` method that asked for the upgrade.

ioquatix:
I only think `on_open` should take an IO. It should be the responsibility of the instance to store the IO IMHO. This way you can call `@io.write` at any time, e.g. from a timer.

tenderlove:
> I only think on_open should take an IO. It should be the responsibility of the instance to store the IO IMHO. This way you can call @io.write at any time, e.g. from a timer.

I'd rather they all take the IO object.  You can save to an ivar if you want.

ioquatix:
> I'd rather they all take the IO object.

What's the reasoning behind that? Is it more efficient?

tenderlove:
To address @SamSaffron's "junk data" comment, I totally agree.  Maybe we could use headers to indicate to the webserver that the body object is "special" (it knows websockets):

```ruby
class WSBody
  def each
    yield "Webserver said it does WS, but lied"
  end

  def on_open io
  end

  # ... more on_* methods
end

# App

app = lambda do |env|
  if env['rack.supports_ws?']
    [ 200, { "X-Websocket" => "on" }, WSBody.new ]
  else
    [ 404, {}, [ "This was supposed to be a WS response" ] ]
  end
end

# Webserver

status, headers, body = app.call("rack.supports_ws?" => true)
if headers["X-Websocket"]
  # do websocket stuff with the body
  body.on_open io
else
  # do the normal `each` thing
end
```

I think it would eliminate questions about what to do WRT middleware (this type of API would insist that the thing responding to Websockets *not* forward to another middleware)

ohler55:
@SamSaffron code duplication on the server implementation is really a matter left up to the server author. I don't think that is relevant to the proposed spec addition, which is proposed to be optional.

@ioquatix using an io object or the extended handler itself is pretty much the same in regard to some external writer loop. The example code on Agoo demonstrates that. Thats not to say that using an io object might be preferred fro other reasons though.

boazsegev:
@SamSaffron 

> My issue though is that encourages a whole bunch of code duplication

I think code duplication is one of the concerns this PR will solve.

Rright now we have tons of applications each implementing their own WebSocket protocol (there's only one standard protocol, so historical variants aren't necessary).

On top of that, all these application have their own implementation of IO handling logic (their own reactor, `nio4r` or some other approach).

Consolidating the dependency in the server actually minimizes code bloat as most of the code is already there. Sure, adding the WebSocket parser might be new code, but the network and IO handling logic is already there.

tenderlove:
> What's the reasoning behind that? Is it more efficient?

It means that you can implement a stateless object.  I prefer to avoid maintaining state when possible.  If the methods take the IO object as a parameter, I am allowed to implement a stateless object, but it is *my* choice.  Insisting that only `on_open` takes the IO object forces me to maintain state (I no longer have a choice).

SamSaffron:
> Since it is either WebSocket or SSE a response and env are no longer relevant. 

I think chunked encoding responses are just as good as a candidate, they are more widely supported than SSE anyway. In fact I can see almost zero reason to support SSE cause it is pretty much a novelty protocol. https://caniuse.com/#feat=eventsource compared to https://caniuse.com/#feat=xhr2 which is supported on IE/edge. 

env is very relevant for chunked encoding you may want to add headers deep in your middleware for CORS and various things like that.

SPEC
tenderlove: > This means a single object in C (the callback object). If we add the IO object, it's two objects - a 200% increase in object count.

What % increase is that in an actual app?

ohler55:
@tenderlove in you code example you seem to be assuming WebSocket and SSE calls are triggered by an event from the browser. Maybe I misread the code. My apologies if I did. Anyway, I think a common use case will be to push event from Ruby to the browser. Do foresee a more request response pattern initiated from the browser using WebSockets instead of using HTTP?

tenderlove:
> @tenderlove in you code example you seem to be assuming WebSocket and SSE calls are triggered by an event from the browser. Maybe I misread the code. My apologies if I did. Anyway, I think a common use case will be to push event from Ruby to the browser. Do foresee a more request response pattern initiated from the browser using WebSockets instead of using HTTP?

Derp, yes, sorry. It's been a long day. 😞

ohler55:
@tenderlove regarding stateless objects, I think there is some merit to what you say. The principles of the API remains the same but adding the io object does allow some additional options. It certainly doesn't hurt.

boazsegev:
@SamSaffron 

> I think chunked encoding responses are just as good as a candidate

I'm not sure if the PR is supposed to be a `hijack` replacement.

It's definitely a step towards retiring `hijack`, but not necessarily a complete replacement.

The idea is that response that passes through the middleware is actually sent by the server.

i.e., notice the header in the response (place in `config.ru` and run with `iodine`):

```ruby
# Place in config.ru
RESPONSE = [200, { 'Content-Type' => 'text/html',
          'Content-Length' => '12' }, [ 'Hello World!' ] ]
# a Callback class
class MyCallbacks
  def initialize env
     @name = env["PATH_INFO"][1..-1]
     @name = "unknown" if(@name.length == 0)
  end
  def on_open
    subscribe :chat
    publish :chat, "#{@name} joined the chat."
  end
  def on_message data
    publish :chat, "#{@name}: #{data}"
  end
  def on_close
    publish :chat, "#{@name} left the chat."
  end
end
# The actual Rack application
APP = Proc.new do |env|
  if(env['rack.upgrade?'] == :websocket)
    env['rack.upgrade'] = MyCallbacks.new(env)
    [200, { "X-Header" => "This is sent" }, []]
  else
    RESPONSE
  end
end
# The Rack DSL used to run the application
run APP
```

Yes, I suppose long polling can be added to this list, using chunked encoding, but I this means that the headers are sent immediately, since this is part of the design's premise.

SamSaffron:
> Sure, adding the WebSocket parser might be new code, but the network and IO handling logic is already there.

This is my point though... What are @evanphx @FooBarWidget and Eric going to do here? They are pretty much the people who need to be convinced this is good.

- Add a hard dependency to "standard websocket parser gem" (I give it zero % chance this will happen for Unicorn, but don't know where Evan and Hongli stand) 

- Add "websocket_encoder.rb" file and "websocket_upgrader.rb" and "magic_rack_protocol.rb" they hand curate and is duplicated 3 times between the 3

- Add a soft dependency we need to add to our Gemfile to "activate" 

I don't see any clear cut good answer here.

SamSaffron:
This `env['rack.upgrade?'] == :websocket` is a big concern for me protocol wise, if `/some/route` is meant to be SSE, and `/some/ws` is meant to be web sockets, how is the server going to decide this upfront. Not allowing the router any say here is a problem imo. 

`if env['rack.can_upgrade?`]` seems far more flexible.

ohler55:
@SamSaffron I'm surprised we are considering rejecting the PR because it might be hard for some of the current server gems to implement. I would have thought Rack was intended to make life easier for developers/users and not for the server implementers.

boazsegev:
> if /some/route is meant to be SSE, and /some/ws is meant to be web sockets, how is the server going to decide this upfront. 

The server doesn't.

The server lets the application know what the user asked for (if the user asks for WebSockets or SSE).

The application is free to refuse the connection if the route doesn't match.

The application is also free to implement both solutions on the same route.

ohler55:
@SamSaffron `env[`rack.upgrade`] can be tested for `nil` or for s specific value. Hiding the fact that the upgrade is either WebSocket or SSE is limiting. Providing more information only help make the decision easier for more specialized applications.

boazsegev:
@tenderlove , @SamSaffron 

I'm not sure I understand what you mean by "junk data".

As I understand the PR, the `env` object performs the roundtrip through the middleware and afterwards the server tests for the callback object. At what point is there a risk for junk data?

SamSaffron:
My personal goal here would be adoption, and if we are introducing a protocol here that can not easily be adopted (or will be rejected by the big 3) then chances of faye,message_bus and family all moving to it are very low. 

Also it places unfair pressure. "Look passenger did it why don't you?"

That said, since this can be implemented today in middleware it is not the end of the world if the big 3 reject it. 

> The server lets the application know what the user asked for (if the user asks for WebSockets or SSE).

I follow so you are looking at `Upgrade: websocket` header, then the server knows it can implement it so it sets `['rack.upgrade?']` to websocket. But ... SSE ship `Accept: text/event-stream` and chunked encoding ships nothing. and in all cases they could potentially use this protocol. The terminology is very websocket centric.

julik:
I don't know if this comment is helpful to the discussion at large, so pardon me if this feels as an intervention. There are legitimate cases where hijack is used because it offers access to the socket. For example, we use it to call `sendfile` multiple times in sequence to server very large binary responses, and it works well.

The "downgrade" in convenience that `hijack` brings is well compensated by the interesting things you can do to it. Imagine with this proposal one of the use cases would be _not_ using SSE (and not using WebSockets) but instead stuffing large binary responses into the socket from the filesystem (as we do in https://github.com/WeTransfer/fast_send). Would such an "upgrade wrapper" facility help that use case in any capacity? For example, would it give me a way to "yield back control" from a sending loop - which I ideally would have synchronous or in a Fiber - back to the webserver?

In practical terms, using Hijack was a bit tricky in the first place because it was not clear _which exact subset_ of an `IO` the given socket-ish object supports, and I feel this might be the case for this proposal too. Personally I would love to see that IO specified somewhat tighter ("here are the methods that _must_ be available, here are the methods that _may_ be available if the server decides to"). I don't know if having something so tightly scoped to be a WebSockets support library is the answer to "hijack API is obtuse" TBH.

SamSaffron:
> In practical terms, using Hijack was a bit tricky in the first place because it was not clear which exact subset of an IO the given socket-ish object supports, and I feel this might be the case for this proposal too. 

Yes absolutely, we don't even know which type of exception this object can raise and the list goes on. I would love to see this thing properly documented. 

Also another missing practical item is that you have no way of setting a callback to happen after the server finished dealing with the hijack which is a huge hole. eg: `env['hijacked] = lambda{|r| ...}` 

> I don't know if having something so tightly scoped to be a WebSockets support library is the answer to "hijack API is obtuse" TBH.

I agree with this. I would like to see hijack better specified.

SPEC
boazsegev: > What % increase is that in an actual app?

That really depends on the app's use-case.

A micro services app running a WebSocket API will get hit harder than an HTTP app with the occasional WebSocket upload manager.

I think the issue is more pronounced because we are talking about long term objects.

Short term object count doesn't effect memory fragmentation as much as the fact that there's a long-term object blocking a memory "block" / "arena".

The more long-term objects are created (and eventually destroyed), the faster the fragmentation.

Having said that, even a 200% increase in object count is probably better than the solutions we currently employ when implementing WebSockets using `hijack`.

I won't make an issue out of it. I'll do what everyone thinks best, but I know in very practical terms (having coded both implementations in C) that `extend` is cheaper both in memory and in the number of method tree traversals required.

ohler55:
@SamSaffron regarding `env['rack.upgrade?']`, the server has already looked at the headers and determined the requested upgrade. There is no need for the application to look at the headers to determine the upgrade type. It was provided as a convenience so the application did not need to look at the various headers that were used to determine the upgrade requested. It also leaves the API open for future additions if some new event based connection protocol is supported.

boazsegev:
@SamSaffron 

> I follow so you are looking at Upgrade: websocket header, then the server knows it can implement it so it sets ['rack.upgrade?'] to websocket. But ... SSE ship Accept: text/event-stream and chunked encoding ships nothing. and in all cases they could potentially use this protocol. The terminology is very websocket centric.

Yes, the idea is that the `env['rack.upgrade?']` value is set according to the headers client send (i.e., `Accept: text/event-stream` or `Upgrade: websocket`, for now).

If there was a way to detect XHR, it would probably be easy enough to manage... but there isn't. This might remain in the `hijack` realm for now.

As @ohler55 [mentioned](https://github.com/rack/rack/pull/1272#issuecomment-385836295), this allows the application to ignore the HTTP headers and distance itself from the HTTP protocol.

As a side-note:

I loved XHR when I didn't have WebSockets. I hardly ever use SSE or XHR for push notifications (I do use XHR for REST calls). However, I can see HTTP/2 bringing a strong movement towards SSE, so it might be in my future.

boazsegev:
@tenderlove , @ohler55 , I feel that the static object support is a very strong argument in favor of an `io` object (although I find the name `io` misleading, maybe `client` more befitting).

I feel emotional resistance against this added complexity, but I can understand the logical benefits.

However, retaining a copy of the `env` object is a major memory concern as a lot of data has to be kept "alive" and moves into long-term memory storage... even though this might shrink the number of cases where static objects could be used.

This will make `on_message(client, data)`, `on_close(client)`, etc' less enticing, but it will work.

ohler55:
@boazsegev I know we discussed having an `io` object or what ever the label is. I don't have a strong opinion able the label. I don't see the need for passing around the `env` object though. I'm in agreement with you. The data in the `env` is stale by the time it is used in any of the callbacks. It is only relevant on the initial invocation of the `#call(env)` method.

boazsegev:
@ohler55 , yes, you were the one who suggested the `io` object during our discussion and I was the one with the strong feelings against the idea.

It's an implementation perspective. The extra object is somewhat of a headache and increases performance costs...

...but I'd rather have something everyone agrees to implement than have nothing.

Besides, if we all write amazing servers, I'll retire iodine in favor of whichever server is the best, so the extra implementation and maintenance cost is avoidable as far as I'm concerned ;-)

SPEC
```diff
@@ -119,17 +119,17 @@ Servers that choose to support WebSockets and SSE connections should follow thes
...
-                     The object will be extended by the server to include:
-                      write(message)  # writes to the WebSocket or SSE connection
-                      close()         # forces a close of the WebSocket or SSE connection
-                      open?()         # returns true if the connection is open, false otherwise
-                      pending()       # returns the number of pending writes or -1 if the connection is closed
-                     This option should only be set it the environment options has a non-nil value for the <tt>rack.upgrade?</tt> option,
+                      on_open(client)      # called when the connection is complete with a client object.
+                      on_message(message)  # called when a WebSocket message is received by server. <tt>message</tt> will be UTF-8 encoded for text messages and Binary encoded for binary messages.
+                      on_shutdown          # may be called before a connection is closed due to server shutdown.
+                      on_close             # called when the connection is closed
```
boazsegev: I think it's only fair that the API informs the receiver about the identity of the closed client (i.e., for `collection.delete client` type of situations)?

so `on_close(client)` as well.

boazsegev: Oh, sorry, I misunderstood... I think this update assumes the client object is stored by the receiver.

ohler55: The call to `on_close` indicates the connection has been closed so the client is no longer viable or at least any method called on it would raise an exception as not being connected so it doesn't seem useful to include it as an argument to the close callback.

ohler55: Did miss adding the client to the `on_message`. Fixed that.

ohler55: On second though. I hadn't fully bend my mind around what the change would mean. With a stateless approach as @tenderlove suggested a single upgrade handler could be used. It might then want to store an array of clients thereby avoiding creating a two new objects for every connection. Maybe thats what you were thinking of before my hasty response.

I'll update the spec with all calls taking the client.

SPEC
```diff
@@ -113,6 +113,25 @@ be implemented by the server.
                         fatal(message, &block)
 <tt>rack.multipart.buffer_size</tt>:: An Integer hint to the multipart parser as to what chunk size to use for reads and writes.
 <tt>rack.multipart.tempfile_factory</tt>:: An object responding to #call with two arguments, the filename and content_type given for the multipart form field, and returning an IO-like object that responds to #<< and optionally #rewind. This factory will be used to instantiate the tempfile for each multipart form file upload field, rather than the default class of Tempfile.
+
+Servers that choose to support WebSockets and SSE connections should follow these additional environment specifications:
+<tt>rack.upgrade?</tt>:: Is nil or missing if no connection upgrade is possible,
+                         :websocket for a WebSocket upgrade, or :sse for an SSE upgrade.
+<tt>rack.upgrade</tt>:: Is used to pass a handler instance or class back to the server for WebSocket and SSE connections.
```
boazsegev: Maybe, now that we support static handlers, this line should be altered?

Perhaps:

> <tt>rack.upgrade</tt>:: Is used to pass a handler object back to the server for WebSocket and SSE connections.

ohler55: Sure, minor change.

ioquatix:
I've been mucking around with a sample implementation of this proposal.

I can implement it entirely as middleware. So, in my mind, it could be something independent of Rack. I didn't implement it exactly because I just reused what I've already done with `async-websocket`.

# Specific feedback regarding the proposal.

With reference to https://developer.mozilla.org/en-US/docs/Web/API/WebSocket

## `rack.upgrade?`

It's okay. It's trivial enough to implement.

It's not clear if the user should return anything. In my case, I use hijack, so no return is necessary (or desirable).

## `on_open`

Makes sense. Do you mean the connection at the socket level or the connection at the web-socket level?

## `on_message(client, message)`

Makes sense.

## `on_shutdown(client)`

Not exactly sure where this fits in. It's similar to `on_close`.

## `on_close`

Do you mean when the websocket connection has been closed by the remote end only (and gracefully?) i.e. when ready state of web socket is changed to close? What about EOF?

## `on_drained(client)`

This seems very asynchronous model dependent. I don't know how to implement this. My model is completely asynchronous so when you call `write` it won't return until the write is complete, but it's non-blocking. Remove this or make it completely optional (as in, it should still work if unimplemented).

## `write(message)`

Pretty crappy name IMHO, because it directly clashes with the WebSocket::Driver implementation's use of `#write` (https://github.com/faye/websocket-driver-ruby#usage). If you have to wrap the underlying driver, it's another object allocation. Bleh. Rename it `#send` or `#text`?

It would be pretty damn convenient if `#send` would convert it to JSON first. Perhaps if `Accept: application/json` is set, it can be done automatically. We could also support other formats. So, user doesn't need to worry about underlying transport to send complex objects (e.g. hash, array, and so on).

## `close()`

If user calls this, what is the sequence of events afterwards? `on_close`? `on_shutdown`? Nothing?

## `open?()`

In distributed systems, it's hard to know if anything is truely open. But you can usually tell with certainty if something is closed. Rename to `closed?` - it's typical Ruby `IO#closed?` `Socket#closed?` and so on.

## `pending()`

Again, this is very async model specific. Remove?

I still stand by my original assessment. But, I think it's an interesting idea. This proposal implements something which can be done entirely in middleware with minimal overhead if your server has a half decent concurrency model.. I don't see a clear benefit to making Rack more complex in this regard and encoding a concurrency model into the `SPEC`. However, I can see value in making some kind of standard layer. I just don't feel like it belongs in the server to implement it, middleware would make more sense.

I still believe streaming requests/response bodies to be the right way forward, but I'm still learning about how this all fits together with HTTP/2 and whether it's possible to use XMLHTTPRequest in a similar way to WebSockets in the browser. It's simpler to set up, multiplexes better (when running on HTTP/2), and has a simpler model - basically read chunk from stream, write chunk to stream, and it works with existing request/response model without any real changes.

Anyway that's my 2c. Feel free to play around with the implementation :)

ioquatix:
An addendum, a concurrent `rack.hijack` IO should be possible for most servers with very minimal surface area (it's what `falcon` does). It would still require some kind of `rack.upgrade` approach since to make a worker truely async you'd need to expose part of the concurrency model, as in "Your rack.upgrade object will be run in it's own thread with blocking IO" as the base requirement, but with more advanced servers able to provide better guarantees (e.g. rack.multithread could be false, you can assume an `async` like model if things are running concurrently). Again, it starts to get pretty hairy, in terms of how you define that model, and whether that should really be in Rack, but it's probably a more generic approach and it would work with HTTP/2 as well (in general, rack.hijack with an IO like API should map directly to HTTP/2 streams, and it's the next thing I want to try in `falcon`).. It wouldn't even preclude you from implementing web sockets on top of it with whatever you currently use and a default implementation as a middleware which really does just make a thread per request is entirely possible.

ohler55:
I appreciate you taking the PR seriously enough to devote time to exploring implementation options. Thank you. Nice summary as well. I'd like to point out that all the method are optional. If the handler does not respond to the method then it is not called.

`rack.upgrade` :heavy_check_mark: 

`on_open` indicated the connection has beenn upgraded to WebSocket or SSE and is ready for sending messages.

`on_message` :heavy_check_mark: 

`on_shutdown` I agree, it does not seem necessary.

`on_close` Yes, called when the socket is closed. Ideally gracefully but that assumes the client did a graceful close which may not be the case. Basically the socket is no longer write able and nothing else will be received from it. Since both WebSocket and SSE are message based the notion of EOF doesn't really come into play as it is a streaming concept. Best case is EOF is the same as closed or maybe another way to look at it is if the server is reading a message and gets an EOF error then it assumes the socket is closed and tells the application that the connection is classed with the `on_close` callback. That is the way Agoo handles it anyway.

`on_drained` indicates no more messages are waiting to be sent. Depending on the implementation this could indicate the previous message was sent and another can be sent or the backlog of messages has dropped to zero. It is meant as a way for the application to meter itself and detecting slow consumers. In my opinion it is nice to have to avoid slow consumer attack vectors.

`write` could certainly be renamed but `send` would conflict in the base Ruby Object `#send` so something different. `#write` is the second best choice. You seem to have agreed since you used it in Faye which of course is where the conflict lies. How about `push`?

`close` when called initiates a graceful close of the connection. Once closed the `on_close` callback should be called.

`open?` flipping the call to `closed?` makes sense. Agreed.

`pending` is async specific. It could be removed and everything would still be fine. It was added to address the slow client issues and allow and application to address slow client attacks. How about return some value that indicates it is not supported?

Thanks again for the well thought out response.

ohler55:
I like the idea of using `rack.multithread` as an indicator to the application as to what it can expect in regard to the async model especially since it is already part of the spec and seems to be a perfect fit.

ioquatix:
> write could certainly be renamed but send would conflict in the base Ruby Object #send so something different. #write is the second best choice. You seem to have agreed since you used it in Faye which of course is where the conflict lies. How about push?

Don't get me started with `send` and Ruby. I had no end of problems with it several years ago, when I first started playing with Ruby's IO. Fortunately, now `__send__` exist in Ruby. I sometimes wonder if we are writing Python :p

To me, send/recv are socket level, but they resonate for higher level network IO too.

read/write make me think of buffered IO.

As this is a message based protocol, and you chose `on_message`, why not then `send_message`?

> pending and on_drained  ...

I understand where you are coming from.

In my client code, the following happens:

```ruby
connection.send_message(....) -> @driver.text -> @socket.write(protocol level data) -> Socket#write_nonblock -> okay, :wait_writable or EPIPE
connection.send_message(....) -> @driver.text -> @socket.write(protocol level data) -> Socket#write_nonblock -> okay, :wait_writable or EPIPE
connection.send_message(....) -> @driver.text -> @socket.write(protocol level data) -> Socket#write_nonblock -> okay, :wait_writable or EPIPE
```

You literally can't send messages faster than they would be written to the socket, because you'll yield on `:wait_writable` and `select(io, writable)`. Once the data is written to OS socket, how do you know `pending` or not? I mean, eventually that buffer will fill up. But do we get receipt of messages in WebSockets? I don't know enough about it, I only used `websocket-driver` and left the rest up to that gem. Can we get pending acknowledged count from that implementation?

ioquatix:
In an implementation where messages are buffered some how, where do you propose to handle EPIPE? Or should it never occur and `on_close` should be invoked? or do we need `on_error`? To me, implementing something like `on_error` to handle EPIPE?

If your concurrency model is more event driven, I can imagine you would buffer the message and actual writing might occur on a different call stack.

ioquatix:
In my implementation, when you have a loop like `while event = @driver.next_event`, if you receive EOF, event is nil. It's not an exception. Then, it would call `on_close`. However, sometimes it might be caused by ECONNRESET. This is really an error condition. should it be `on_close(error = nil)`?

but if you call send_message, and remote end is gone, you make get EPIPE, Errno::ECONNRESET, or on macOS even EPROTOTYPE (https://bugs.ruby-lang.org/issues/14713). It's not graceful. Sometimes yo can get E

boazsegev:
@ioquatix, it is my humble understanding that this specification allows us to abstract away the network layer as much as we can.

This allows all persistent connections to share the same API regardless of their type (WebSockets, SSE, raw TCP/IP, etc').

It's my understanding that EOF and other network details should be handled by the server. They aren't application related events.

The reverse is also true. JSON is as application concern, it doesn't fit every persistent connection out there (i.e., binary transmissions in WebSocket data).

It's my recommendation that the server require a String object and let the application deal with making sure the data formatted as a String.

Network details could be inferred by the server (UTF-8 for text encoded messages and otherwise the server should probably assume binary data, which some connection objects might not support).


### `on_shutdown`

> `on_shutdown` I agree, it does not seem necessary.

I humbly disagree.

Maybe I'm assuming but I think applications can benefit from the knowledge that the client is closing because the server is shutting down.

There are three possible disconnection causes:

1. The connection is deemed unnecessary (this could be a proxy timeout due to inactivity).
1. The server shutting down (the client might be mid transmission).
1. Network errors (which we can do nothing about).

True, `on_shutdown` only called when the server is shutting down (not for every connection), but it's a good way to prevent communication errors and keep the disconnection polite - especially when using server clusters or load balancing.

This could be use by load-balanced micro-services as well to indicate a node is going away (and perhaps send a list of alternate nodes in a gossip based system).

### `on_drained` / pending()

@ioquatix , since you're using a blocking model, where the method returns only after `write` was complete, `pending` can easily return `0` - no pending `write`, indicates that `on_drained` isn't about to be called (maybe ever).

Do note that this blocking design might be risky, if - for example - a DataBase connection was checked out of a connection or the `write` method was called within a critical section (i.e., looping over all clients).


### `write`

> `write` - Pretty crappy name IMHO...

Actually, just because other implementations use this amazing name (which is also the name of an underlying system call), doesn't mean that the specification should choose a poor name...

...in fact, it just supports the choice as a good one - developers will have a good idea as to what happens when they call this method.

IMHO `send` conflicts with the Ruby object `send` which might prevent an error from being detected (i.e., if `send` is called on another object, it will not return a "missing method `write`, but rather a long gibberish message).

`push` seems (IMHO) less intuitive and more protocol specific (SSE). It also implies objects rather than data.

IMHO, `write` describes exactly what's happening. It might behave slightly differently from an IO `write` (no partial `write` allowed), but it performs the same function.

### `open?`

> open? flipping the call to closed? makes sense. Agreed.

Actually I think this `closed?` will make application code harder to maintain. It's too easy to mixup `close` for `closed?` and vice verses.

Using `open?` is safer, even if the server's information regarding the connection is never fail proof.

### `on_close(err)` / `on_error(err)`

I really think network / connection errors aren't an application concern, they are a server concern.

However, if we are to report an error, I would wonder if it would make sense to place the error information in a different callback, preventing the need for the obvious `if(err)` block of code in `on_close`.

boazsegev:
@ioquatix 

> In my implementation, when you have a loop like while event = @driver.next_event, if you receive EOF, event is nil. It's not an exception. Then, it would call on_close. However, sometimes it might be caused by ECONNRESET. This is really an error condition. should it be on_close(error = nil)?

`ECONNRESET` is a perfect example for an error that the application really shouldn't receive. Why should the application have any information about network errors? What would it do with it?

IMHO, this is a Server related concern and it should be handled by the server.

ohler55:
I have less of an issue passing some error back to the application with `on_close`. The application can't do anything about it but it isn't harmful either.

@boazsegev you make some good points I was a little too willing to compromise on.

I'm not sure where we go from here. I assume we need a 'member' to step in again. How about it @tenderlove , how should we proceed on this PR?

boazsegev:
@ioquatix 

> In my client code, the following happens... 
>
> You literally can't send messages faster than they would be written to the socket...
>... Once the data is written to OS socket, how do you know pending or not?

I think the question isn't about data pending **after** performing `write` to the socket. It's about data pending **before** `write` is performed (if applicable).

In your  (fiber-per-connection) model `write` always completes. But consider the node.js model (which iodine emulates in this regard):

* An application calls `connection.write`.
* The server places the data in the connection's queue and adds a `flush_connection` event to the event loop.
* `connection.write` returns.

    Note that _**no network actions were performed**_. At this point `pending()` would probably return `1` (or perhaps a higher value).

* The application code ends and the server handles the "flush" event, flushing the connection's buffer using the system call `write` (or `socket.write`).
* Once all the data was written to the socket (which might require a number of events) the connection's `on_drained` event is scheduled.

   This doesn't guarantee that the data actually reached the browser/client (a network error might have occurred, the OS might be waiting on a timeout).

   This only guarantees that the server won't call `write` (or `sock.write`) until more data is scheduled to be sent - the data is out of the server's control.

ohler55:
Agoo uses a similar approach.

boazsegev:
**`write` - exception vs. `true`/`false`?**

Reading through the thread I noticed how hard it might be to predict a `write` failing.

This means that user code must use `begin`/`rescue` blocks to address network issues.

Since a failing `write` is a predictable and possible result in the normal flow, wouldn't it be better if the server handled any exceptions and simply returned `true` if the data was (or scheduled to be) written to the socket and `false` otherwise?

I think this will improve performance and also prevent the obvious nesting of exception handling code (on the server and the app).

Thoughts?

ioquatix:
I think the next steps are:

- To finish several implementations of the proposed spec, this will expose any obvious design issues with the spec.
- To implement several practical applications, and test interop - if we can't make useful applications that work on different servers (i.e. if you start relying on iodines pub/sub which can't be implemented in puma), whats the point?
- To prepare a finalised spec which takes into account the above and then have a formal vote, probably only involving actual stake-holders: Ruby server implementors and Rack contributors.

ioquatix:
> There are currently two implementations of the originally proposed spec, Agoo and Iodine. I'd think that if proof that the spec is workable.

I think we need to see a working implementation in at least Puma & Passenger as a baseline since that is what is very common for Ruby development and production.

ohler55:
I believe the suggested bar was several practical applications. There are already sample applications in the Agoo and Iodine repositories.

ioquatix:
> To implement several practical applications

Something more practical than "Hello World". For example, a shared real-time Todo list :p

I think we should be involving some database activity in the callback, probably using ActiveRecord or Redis.

If you've already got several sample applications, once we finalise a draft SPEC and we all implement it, we should be able to all run those apps on our respective servers, right? That's the whole point of having a shared SPEC.

matthewd:
I still don't see a statement about whether `write` can block.

I'm also unconvinced that pending/drained is an effective solution for non-blocking writes.

I remain of the opinion that this should remain a draft / small-e extension while real-world implementations adopt it and fully exercise it.

Easy-mode for convincing me would be a PR to Rails adapting Action Cable to this API -- which would also allow like-for-like benchmarking.

boazsegev:
@matthewd ,

Although I understand the suggestion (adding a Rails PR), I find that quite unfair.

I can program a server in C and I can wrap in a gem for Ruby MRI... but I know nothing about ActionCable, it's inner workings or the Client<=>Application protocol used above the WebSocket layer.

The learning curve and time spent on a Rails patch would only be viable if the PR is conceptually acceptable and we all intend to move forward with the proposal and implement it.

I really doubt if pushing a PR to Rails is something that should happen before we agree that the PR is conceptually acceptable (i.e., contingent on benchmarks, or whatever).

As a side note, I also don't think this should enter be a Rails 5.x release. This PR has the potential to retire the Rails IO handling layer (the one based on `nio4r`) and reduce the existing codebase significantly - at least where WebSockets and SSE (and perhaps long polling could be achieved by [Javascript's polyfills](https://github.com/Yaffle/EventSource), retiring `hijack` altogether).

SPEC
```diff
@@ -113,6 +113,25 @@ be implemented by the server.
...
+<tt>rack.upgrade</tt>:: Is used to pass a handler object back to the server for WebSocket and SSE connections.
+                     The handler MAY implement any or all of the following callbacks:
+                      on_open(client)              # called when the connection is complete with a client object.
+                      on_message(client, message)  # called when a WebSocket message is received by server. <tt>message</tt> will be UTF-8 encoded for text messages and Binary encoded for binary messages.
+                      on_shutdown(client)          # may be called before a connection is closed due to server shutdown.
+                      on_close(client)             # called after the connection is closed.
+                      on_drained(client)           # may be called when the number of pending writes drops to zero.
+                     The <tt>client</tt> is used for writing and checking status of the upgraded connect. It has these methods.
+                      write(message)               # writes to the WebSocket or SSE connection
+                      close()                      # forces a close of the WebSocket or SSE connection
```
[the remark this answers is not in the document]
ioquatix: If you want to have that, implement it as `close_write` which is what Ruby calls `shutdown(SHUT_WR)`. `close` should guarantee after returning that the underlying socket is closed. Otherwise, you are in for a world of pain.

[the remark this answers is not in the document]
ohler55: Agoo treats a cal to close as a request just like a write and places the close on the queue.

ohler55:
@matthewd would a statement leaving the blocking behavior up the the server be enough or maybe a method to ask if blocking would block other threads?

SPEC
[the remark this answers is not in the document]
boazsegev: @ohler55, so does iodine. I place a "close" marker on the packet queue, so `close` is always performed after all scheduled data is sent.


@ioquatix , I think your missing the point o f abstracting away the network and the protocol.

My suggestion was about clarifying this little bit, not changing what both iodine and agoo already implement.

We aren't authoring a network layer. We are authoring an abstracted application side API.

The reasonable exception is that `write` is perfumed before `close`. i.e., if my code is:

```ruby
write "hello"
close
```

The reasonable expectation is that "hello" is actually written.

There's no `force_close` or `close_write` in the specification because the application shouldn't be concerned with these things. If the application doesn't want data written, it can avoid writing it.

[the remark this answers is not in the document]
ioquatix: You can make close call `flush`, or you can flush after every write. But you make close call flush, you better be careful about `EPIPE`.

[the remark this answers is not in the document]
boazsegev: @ioquatix 

I think these are implementation details that each server could work out according to it's architecture.

The falcon implementation, for example, seems to block `write` so it always completes before `close` (or so I understand) while iodine's implementation places a flag and adds an event to the event queue.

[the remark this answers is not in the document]
ioquatix: Sure as long as there are no pratical differences to the user, it would be reasonable. However, if you are buffering events, you might end up with the user having the wrong expectations.

```
10.times do
    connection.send_message(...)
end

# got here, were all events sent? it depends on implementation. Can I call flush to confirm they were sent?
```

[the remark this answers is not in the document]
ohler55: I'm not sure what you expect `flush` to do. In most uses flush blocks until all outstanding writes have completed. Is that the intent, block until completed? The PR proposes the `on_drained` callback for that purpose or a `pending` count of zero. Are you proposing a `wait_for_all_writes_to_complete` method called `flush` in addition to the existing ones?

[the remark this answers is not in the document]
ioquatix: For a high level protocol like this, calling flush after each write would make sense to me.

It provides the user with a strong expectation, that after calling write, the data has been sent, and pending a network failure, would arrive, or else the write fails right then, with, say, `EPIPE`. Otherwise you'll just end up with a spaghetti state machine trying to handle all these conditions.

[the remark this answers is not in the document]
boazsegev: @ioquatix ,

I hope I don't seem too blunt or crude. I very much appreciate the interest and willingness to polish the specification and make it both as clear and as practical as can be.

However, please consider the model to be totally separate from the network - there is no network. There's only this API.

We can change the API if we need features, but we don't expose network bits or logic because part of our job is to abstract these things away - so there is no network, there is no protocol (as much as possible).

In this sense, `flush` doesn't exist. It's a network / server detail that the client never sees, abstracted away by the server.

The closest an application can come to ask about these things is to ask about all the `pending` events that haven't yet completed. This allows an application to know if the `on_drain` callback is somewhere in their future.

The `pending` query doesn't to expose the network, it exposes the progress of existing API calls. This provides important information about the possibility of a slow client (or a slow clearing "queue") allowing an application to stop serving a resource hungry "client".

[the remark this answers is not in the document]
ohler55: If I understand your response you are answering yes to the question of whether or not a `flush` method should be added. Then in any application you wrote you would block until the write completes instead of making use of the `on_drained` callback. That is your choice of course.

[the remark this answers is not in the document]
ioquatix: > However, please consider the model to be totally separate from the network - there is no network. There's only this API.

Fair enough.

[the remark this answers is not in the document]
ioquatix: > If I understand your response you are answering yes to the question of whether or not a flush method should be added. Then in any application you wrote you would block until the write completes instead of making use of the on_drained callback. That is your choice of course.

I really find the inverted flow control of callback style programming horrible. So, I prefer `#flush` over an `#on_drained` callback. Callbacks = state machine spaghetti = hard to maintain/buggy code. It's just my personal opinion, FYI.

boazsegev:
I have an initial implementation of the updated draft, placed in [iodine's 0.6.0 version branch](https://github.com/boazsegev/iodine/tree/0.6-stripped) (installable from source).

There's some features being re-written (such as pub/sub extensions, including Redis integration), but the core WebSocket/SSE and pub/sub API will work.

I think this would be enough to author a test application using a monkey-patched Rails...

@matthewd , any final changes / requests / tips before we start woking on a "like-for-like benchmarking" example app? How do you want to benchmark this? Should we use the [websocket-shootout approach](https://github.com/hashrocket/websocket-shootout/tree/master/ruby/action-cable-server)?

ohler55:
The Agoo develop branch is also compatible with the spec as described in the current state of this PR. Should make a versioned release later this week once the named pub-sub feature is completed.

ohler55:
I cleaned up the formatting of the SPEC around the upgrade section. An HTML version is at [http://www.ohler.com/agoo/rack/file.SPEC.html](http://www.ohler.com/agoo/rack/file.SPEC.html).

boazsegev:
@evanphx , how is the Puma adaptation going?

So far, the API iodine implements for the `client` object looks [like this](https://www.rubydoc.info/gems/iodine/0.6.0/Iodine/Connection).

It includes a number of methods not listed in the specification besides the Pub/Sub extensions (the `timeout`, `timeout=` and `protocol` method).

I also added the `env` accessor so I could avoid making objects just to keep the `env` alive. This is something I need for an upcoming plezi.io release and I'm not sure if it should be adopted (as it adds longevity to all the objects in the `env` Hash, which is bound to increase memory fragmentation).

Anyway, I hope to see this move forward. Iodine has been serving clients like this for the last couple of year or so (albeit using `extent` rather than a client object) and so far the feedback is positive and the performance is promising.

boazsegev:
@ioquatix , this was my thought as well, which is why I wrote it here. Perhaps these added methods (`timeout`, `timeout=` and `protocol`) should be part of the specification.

On the other hand, I don't have to have the `env` variable, I just as easily could have create a callback object (using a class instance) instead of static module... it was more of a design preference that I thought I should mention.

ohler55:
The entire pub-sub pattern can be implemented by keeping an Array of connections. A publish would then call write on each one. The reason pub-sub was not included in the PR was that it added another new use pattern to the spec that would take even more effort for servers to comply with.

Like @boazsegev and Iodine Agoo includes a pub-sub implementation that is compatible to a large degree with Iodine. In then end we wanted something basic that could be used with a minimum of changes in Rack and the way developers use it.

The current PR focuses just on Push without also introducing pub-sub. Maybe a future PR can propose the addition of pub-sub. One step at a time.

ioquatix:
> The entire pub-sub pattern can be implemented by keeping an Array of connections. A publish would then call write on each one. The reason pub-sub was not included in the PR was that it added another new use pattern to the spec that would take even more effort for servers to comply with.

This wouldn't work with falcon in forked mode, since it forks one process per CPU core and they all handle connections independently.

Even in threaded mode, some kind of synchronisation is required. I believe the same problem would affect puma in cluster mode, for example.

My opinion is that this spec needs to be feature complete, as in, it's possible within the confines of the spec to implement actual useful websocket applications. Otherwise, what's the point? If every server supports the core spec but provides their own incompatible models on top of that, this isn't going to work out very well.

ohler55:
The PR, as it stands, is feature complete for WebSocket and SSE. Sure, extending the PR to include pub-sub would be a problem for servers that don't share state but that has nothing to do with WebSocket and SSE support proposed here.

@ioquatix, Slamming @boazsegev example code that is meant to demonstrate not only the proposed Rack additions but also other features of Iodine and then claiming that example invalidates this PR is ludicrous. The PR offers a clean and simple way for even new developers to take advantage of WebSockets and SSE. Some of the comments such as those from @tenderlove were constructive and genuinely aimed toward enhancing the current Rack spec with the addition of WebSockets and SSE. You have fought tooth and nail against it from the start with an extreme passion to block any variation on the PR. I don't understand why you feel so threatened by this PR. Can you explain why you have such strong feelings about offering Rack users an simple to use approach to WebSocket and SSE?

matthewd:
Okay, this conversation clearly needs to go back on ice for a bit.

I don't agree that pub/sub is within scope, but if we can't do each other the courtesy of assuming genuine opinions and positive/constructive intentions, then we're not going to get anywhere useful.

tenderlove:
TBH I'm kind of worried about committing to an API without a real world app using it.  The reason is because I want to make sure we don't forget anything we need to implement actual apps.  If we could get Action Cable running on top of a server that used this API, then I'd be satisfied to merge.

ioquatix:
There has been some great discussion here. Thanks everyone for their effort and passion. I am going to close this issue, because I don't think we reached a consensus.

That being said, further discussion is required, backed by concrete implementations. Specifically, I think more emphasis should be placed on HTTP/2 (and HTTP/3) and union of the semantic model when HTTP/1 is also included in the mix.

We welcome proposals for a better stream model for Rack 3.0 and we are currently considering how to make `rack.hijack` a first class citizen: https://github.com/rack/rack/issues/1550

PR #1276, opened 2018-05-09
Add support for rfc6598 public shared networks

btalbot:
Allow Request#trusted_proxy? to handle 100.64.0.0/10 network addresses as other
special networks and be ignored as the 'correct' remote ip.
https://tools.ietf.org/html/rfc6598

Kubernetes installations using CNI networking (overlay networks) are often using the 100.64/10 network for pod addresses. This causes rack apps running on those clusters to to use the REMOTE_ADDR of an upstream pod (possibly a proxy or ingress) instead of evaluating the X-Forwarded-For headers as expected. This patch expands the addresses handled by Request#ip() to include the 100.64/10 public shared network block.

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -420,7 +420,7 @@ module Rack
       end
 
       def trusted_proxy?(ip)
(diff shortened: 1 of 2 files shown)


matthewd:
Where Service Providers "use this Shared Address Space to number the interfaces that connect CGN devices to Customer Premises Equipment (CPE)", that crosses a trust boundary.

Consequently, I think we need to leave it up to individual applications to choose to trust these IPs, based on their specific knowledge of their surrounding networks.

jeremyevans:
#1290 implemented the ability to customize the trusted proxy filtering, so I think this can be closed.

PR #1281, opened 2018-07-06
Take out `Utils.build_nested_query` to the separate gem

AlexWayfer:
Let's unify this code between gems, because `URI.encode_www_form` from `stdlib` can't deal with nested objects.

Also, `Utils.build_query` looks unused.


rafaelfranca:
Thank you but the code duplication that this removed don't justify adding a new dependency.

PR #1282, opened 2018-07-18
Fix gemspec to not include test files from gem

ChrisBr:
as these shouldn't be shipped.

This causes some troubles with generating RPM packages for openSUSE distribution as rpmlint is complaining.


mikegee:
👍  The `gem test` command has been removed from rubygems and testing shipped gems has fallen out of favor in the Ruby community.

PR #1283, opened 2018-07-29
remove .freeze calls

ahorek:
after https://github.com/rack/rack/pull/1256 there's no need to freeze strings explicitly.


kamipo:
Rack would drop Ruby 2.2 support?
`frozen_string_literal` is only work after Ruby 2.3.

kamipo:
I think most `.freeze` is not for perf but is for prohibiting mutation, since it is achieved by reusing the values assigned to constants.

ahorek:
Ok, I agree. @tenderlove please revert this change

kamipo:
@mikegee
> `frozen_string_literal` is only work after Ruby 2.3.

matthewd:
Given that 2.2 is EOS, this seems fine to me personally... it doesn't break compatibility with the unsupported version: it just means that if you're running on an unsupported version _and_ then you attempt an unsupported mutation, the result could propagate into a shared string instead of raising.

mikegee:
Right. I figured there would be another PR to drop Ruby 2.2 support before the next release anyway.

Issue #1263, opened 2018-04-17
Retire rack.github.io site


antonpaisov:
When a person visits https://github.com/rack/rack, the description says "a modular Ruby webserver interface http://rack.github.io/".
However, the rack.github.io site is seriously out of date with the latest post in news dating "February 7th, 2013" and the latest release marked as 1.4.1.
This is very confusing.
Given that https://github.com/rack/rack.github.com is basically a static HTML page with hardcoded content (meaning that neither releases nor other content fetched from outside) and it seems that no one is interested in updating and maintaining it, I suggest it's better to remove the repo and the site altogether and keep all the related info in the main repo/documentation.

If, however, someone would be willing to update/maintain rack.github.io, that person would need to make a new site anyways.

Opinions?

yhirano55:
I'm going to update rack.github.com in the near future. I feel it's not good to maintain with hardcoding, so I'm going to propose that we adopt static generator like Jekyll (like https://github.com/rails/homepage)

antonpaisov:
@yhirano55 cool! When I saw the site I also thought about making a completely new version with something like Jekyll :)
One of my concerns is that there's no CHANGELOG (or I haven't found one) at the repo. This fact made me also think that if the basic info is not supported in the repo it will be twice as hard to keep the separate site up to date. Hope that I'm wrong with this assumption :)

chneukirchen:
Reminder that the rack homepage existed before Github was in use. :)

I think it makes sense to kill it and just redirect to this repo and fill the README with the appropriate info.

There is a HISTORY.md file, but it's not maintained anymore it seeems.

antonpaisov:
IMO having README and CHANGELOG is most important and is actually enough.
I think if there will be a separate site (which I still think is not needed anymore) - it would basically be a page that combines README and CHANGELOG, or what else could there be?

On a side note - is there a quick way to create and fill RELEASES.md for rack?

yhirano55:
> I think it makes sense to kill it and just redirect to this repo and fill the README with the appropriate info.

+100

yhirano55:
@chneukirchen This repo's Wiki is quite old, too. Do you have the plan delete it?

antonpaisov:
@yhirano55 yeah, I also think that would make sense.

In general, I think the approach of getting to the MVP of up to date information with `readme.md`, `releases.md` and `contributors.md` is the best here.

So now someone with an admin access to the repo just needs to do that :)

ioquatix:
[...]

Issue #1245, opened 2018-04-04
Document maintainership


leahneukirchen:
I just update README.rdoc and I think the contents are quite outdated.

The Core Team should list the actual current committers, resp. people on rack-core;
also we should extend the Gem metadata to a relevant set of authors.

fxn:
Don't know off the top of my head... maybe ask in the Contributors room?

rafaelfranca:
@yhirano55 is doing some good work related to documentation on Rails, so maybe he may want to help us here?

yhirano55:
It's my pleasure. I do my best.

Drenmi:
> I'll make a pass through it, but it would be nice if we had more documentation focused folks on the team.

@tenderlove: I've been keen to help since your Request & Response talk at RDRC, but haven't had the time. Work arrangements have changed though, and I'd be happy to help out in any capacity.

tenderlove:
@Drenmi yes please!  TBH, I would like to eliminate `HISTORY` in favor of using commit history.  The thing I think would be most useful for end users is a `NEWS` file that contains more high level information.  I don't think we need to tell everyone about every single commit, but we definitely need to communicate new features / bug fixes / security fixes.

We could definitely use help digesting the commits in to something easily consumable by people using the library.  What do you think?

Drenmi:
@tenderlove I was thinking the same thing. I’m flying back from Ruby Conference Taiwan today. Was hoping to see you here, but I’ll drop you an e-mail or a ping when I’m home and have looked into it.

jeremyevans:
I updated the README to add Samuel and me to the core team.  I'm not sure whether all other people listed as core team members are still active.  If not, they should be moved to the Alumni section.  Looking at the commit log, it seems like the following people may no longer be active, using month of last commit:

* James Tucker (March 2016)
* Josh Peek (June 2011)
* Jose Valim (March 2012)
* Michael Fellinger (May 2016)
* Konstantin Haase (February 2015)

Is it OK if I move these people to the Alumni section?

Regarding the gem metadata, do we want to list all current core team members under authors?  If so, I can update the gem metadata authors after fixing the README.

jeremyevans:
@rafaelfranca fixed the README in 256ea3e20993076f9749fd17f6fd91f0ded0e1e7.  The gemspec has not been updated, though.  Do we want to update the gemspec to reflect the core team in the README?

jeremyevans:
[...]

PR #1286, opened 2018-08-05
Let's keep a changelog

twitnithegirl:
Hi friends,
This PR consolidates and reformats the contents of HISTORY.md and NEWS.md into a new CHANGELOG.md.

I did my best to reformat the existing history closer to the format of Keep a Changelog. I went with this specific format so that we have a reference as we begin in earnest to maintain better release history going forward. I haven't done anything yet to reconstruct release history from _the lost days_ and it may be that those releases will just go undocumented like that semester you stopped writing in your diary in middle school. 

As discussed [here](https://github.com/rack/rack/issues/1263) along with maintaining a changelog I also move that we can go ahead and soft retire rack.github.io by taking the link out of the rack/rack repo description. After combing through to see if there is any content there that isn't reflected in github we can just kill the repo.

[Aaron](@tenderlove) and [Leah](@chneukirchen)mentioned [a while back](https://github.com/rack/rack/issues/1245) that it would be good to have more documentation types around here and I'm happy to help in that capacity.

I'd love to hear any feedback or questions about this change.


twitnithegirl:
Oops, sorry I didn't [ci skip]

twitnithegirl:
Also think we should have a list of all currently deprecated features and when we expect them to be deprecated. I could put that just before the break between the new and historic changelog and we can track things that become deprecated and their anticipated retirement versions in the changelog as they come. Alternatively we can track deprecation in its own file and just link it from the README.

tenderlove:
I have mixed feelings about changelogs.  I want to communicate overarching changes, but not necessarily every change.  IMO NEWS is better for that (Ruby recently removed changelogs in favor of just NEWS).  I guess, historically changelogs have been information for other developers of that library where NEWS is for the consumer of the library.  Since the Git history is available to everyone, changelogs seem irrelevant to me.  If we're willing to treat this changelog more like NEWS, targeted to *consumers* of rack, then I'm good with this.

twitnithegirl:
@tenderlove thanks for the feedback. Keeping it more consumer focused makes sense. What are your thoughts re: the historical changes? Keep them here and just make sure future updates are only items relevant to end-users?

tenderlove:
@twitnithegirl Ya, lets do that.  I think keeping it consumer focused going forward would be perfect.

tenderlove:
I'll merge this and then we can refer back to this ticket for guidance on changelog updates.  How does that sound?

twitnithegirl:
@tenderlove Sounds good to me. Thanks! 🎆

PR #1288, opened 2018-08-09
Multipart/form-data parser speed improvement for many parts

willnotwish:
Parsing a 3Mb multipart/form-data body with many parts (~15000) was taking around 30 seconds for my use case.

I managed to reduce this to 1.2 seconds by using a StringScanner (instead of a regular String) to track the current parsing position and reduce the amount of in-buffer modification.

I have described my findings here
[https://github.com/willnotwish/rack-performance](https://github.com/willnotwish/rack-performance) 

The changes I have made reduce parse time from 35s to 1.2s (on my laptop) for the size of body I was looking at initially.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,10 +3,11 @@ workflows:
 
   test:
     jobs:
-      - test-ruby-2.2
-      - test-ruby-2.3
-      - test-ruby-2.4
-      - test-ruby-2.5
+      - test-jruby
+      # - test-ruby-2.2
+      # - test-ruby-2.3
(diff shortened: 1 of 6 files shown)


tenderlove:
Yes, this is excellent.  Checking on the JRuby build, but this looks great.  Thank you!

tenderlove:
It seems like the JRuby suite is just timing out.  I wonder if we can get it to work on CircleCI

tenderlove:
I'm annoyed about these timeouts, but it's nothing to do with this PR so I'm just going to merge.  Thanks!

PR #1290, opened 2018-08-21
Add method for custom ip_filter and test to go with it

svcastaneda:
This pull request extracts the `ip_filter` into a lambda so that users can configure their own `ip_filter` (e.g., https://github.com/rack/rack/pull/1276).


tenderlove:
This looks great, thank you!

PR #1301, opened 2018-09-25
Add CircleCI badge to README [ci skip]

lostapathy:
This adds the CircleCI badge to the README since CRuby builds happen there instead of Travis.


lostapathy:
@rafaelfranca I apologize - I had a copy/paste failure and the image is broken.  I have corrected and force-pushed back to the same patch branch, but not sure the best way to have you re-merge the correction?

PR #1302, opened 2018-09-26
✨ add_header accepts an optional delimiter

### LINKED ISSUE #1298: Feature request: optional delimiter for add_header
Many headers are comma-delimited, so the existing `add_header` implementation (which always uses `','`) works great.

But some headers are space-delimited. It would be lovely to be able to provide a delimiter when appending to a header:

```rb
add_header 'My-Header', 'a-value', ' '
```

jamesarosen:
Previously, `Rack::Request#add_header` and `Rack::Response#add_header` both used `','` as the delimiter between multiple values. That works great for headers like `Accept`.

But Fastly's `Surrogate-Key` uses `' '`. And Mozilla uses `'\n'` for `Set-Cookie`, `WWW-Authenticate`, and `Proxy-Authenticate` and `', '` for other headers.

This commit adds a third optional argument to both `Rack::Request#add_header` and `Rack::Response#add_header` for the delimiter. It defaults to `','`, making the change fully backwards-compatible.

See https://docs.fastly.com/guides/api-caching/purging-api-cache-with-surrogate-keys
See https://github.com/bnoordhuis/mozilla-central/blob/28b39571c584530df4b0b74bf9d929942906c081/netwerk/protocol/http/nsHttpHeaderArray.h#L176-L198

Closes #1298

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -84,11 +84,11 @@ module Rack
       #   assert_equal 'image/png,*/*', request.get_header('Accept')
       #
       # http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2
-      def add_header key, v
+      def add_header key, v, delimiter = ','
         if v.nil?
           get_header key
         elsif has_header? key
-          set_header key, "#{get_header key},#{v}"
+          set_header key, "#{get_header key}#{delimiter}#{v}"
         else
           set_header key, v
         end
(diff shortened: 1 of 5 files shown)


matthewd:
You're not describing an "add header" operation: that's "add _to_ header", and that's not what this method is for.

If you need to append a value to an existing header in a non-RFC-compliant way, then you should replace the existing header using `set_header`.

This API represents an implementation of the RFC2616 header aggregation algorithm, and is not a tool for arbitrary manipulation of the contents of header strings.

> Multiple message-header fields with the same field-name MAY be present in a message if and only if the entire field-value for that header field is defined as a comma-separated list [i.e., #(values)]. It MUST be possible to combine the multiple header fields into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field-value to the first, each separated by a comma.

jamesarosen:
OK. I'm a little disappointed that this didn't come up on #1298.

Issue #1308, opened 2018-10-22
maybe endless loop in fileupload with firefox


aiomaster:
We have detected an issue using jquery file upload and a github version of rack.
Problem is, that larger fileuploads (larger than 50M) with firefox stop working in the rack middleware and produce 100% CPU load, so it looks like an infinit loop.
We have traced the commit that introduced the behaviour in rack: https://github.com/rack/rack/commit/534ef35bcc95b11719d2574a70757b61fa3eac33

The interesting stuff is, that the same files work in chrome but not in firefox.
So we compared the Request-Headers and the main difference is, that firefox uses other boundaries in the `Content-Type`-Header than Chrome.

Example-Header in Chrome:
`Content-Type: multipart/form-data; boundary=----WebKitFormBoundarydEXn76YwP9nKSKO3`

Example-Header in Firefox:
`Content-Type: multipart/form-data; boundary=---------------------------100673423314366417001080588648`

This is just a guessing, that the boundary parsing is involved here.

aiomaster:
Ok it boils down to this line, that does not return somehow: https://github.com/rack/rack/blob/5559676e7b5a3107d39552285ce8b714b672bde6/lib/rack/multipart/parser.rb#L268

The content of the @body_regex is:
```
/(.*?)(
)?\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-\-84208446911303861091947234149(
|--)/m
```
I don't know if that helps in a way.

jeremyevans:
Fixed by #1309

PR #1309, opened 2018-10-23
Fix multipart parser for large files #1308

aiomaster:
Try to avoid the lazy matching quantifier, cause it is causing serious trouble with large files.
We just use the string returned by `check_until` and remove the included boundary.
All tests pass and my problematic file uploads work as they should without performance issues.

The only problem is, that I don't know, if I should check in a large file just to demonstrate that it fails without that patch.

### DIFF
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -185,7 +185,7 @@ module Rack
         @collector = Collector.new tempfile
 
         @sbuf = StringScanner.new("".dup)
-        @body_regex = /(.*?)(#{EOL})?#{Regexp.quote(@boundary)}(#{EOL}|--)/m
+        @body_regex = /(?:#{EOL})?#{Regexp.quote(@boundary)}(?:#{EOL}|--)/m
         @rx_max_size = EOL.size + @boundary.bytesize + [EOL.size, '--'.size].max
         @head_regex = /(.*?#{EOL})#{EOL}/m
       end
@@ -268,8 +268,8 @@ module Rack
       end
 
       def handle_mime_body
-        if @sbuf.check_until(@body_regex) # check but do not advance the pointer yet
-          body = @sbuf[1]
+        if (body_with_boundary = @sbuf.check_until(@body_regex)) # check but do not advance the pointer yet
+          body = body_with_boundary.sub(/#{@body_regex}\z/m, '') # remove the boundary from the string
           @collector.on_mime_body @mime_index, body
           @sbuf.pos += body.length + 2 # skip \r\n after the content
           @state = :CONSUME_TOKEN
--- a/test/spec_multipart.rb
+++ b/test/spec_multipart.rb
@@ -6,6 +6,7 @@ require 'rack/multipart'
 require 'rack/multipart/parser'
 require 'rack/utils'
 require 'rack/mock'
+require 'timeout'
 
 describe Rack::Multipart do
   def multipart_fixture(name, boundary = "AaB03x")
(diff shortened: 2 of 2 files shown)


lib/rack/multipart/parser.rb
```diff
@@ -265,8 +265,8 @@ def handle_mime_head
       end
 
       def handle_mime_body
-        if @sbuf.check_until(@body_regex) # check but do not advance the pointer yet
-          body = @sbuf[1]
+        if (body_with_boundary = @sbuf.check_until(@body_regex)) # check but do not advance the pointer yet
```
matthewd: In looking at this issue, have you seen why this line uses check instead of scan? So far I can't see why `on_mime_body` cares about the `pos`, and we update it on the next line. :confused:

lib/rack/multipart/parser.rb
```diff
@@ -182,7 +182,7 @@ def initialize(boundary, tempfile, bufsize, query_parser)
         @collector = Collector.new tempfile
 
         @sbuf = StringScanner.new("".dup)
-        @body_regex = /(.*?)(#{EOL})?#{Regexp.quote(@boundary)}(#{EOL}|--)/m
+        @body_regex = /(?:#{EOL})?#{Regexp.quote(@boundary)}(?:#{EOL}|--)/m
```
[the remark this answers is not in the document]
aiomaster: I tried that, but it does not seem to work, maybe because of the `rx_max_size` calculation one line below. EOL could be "\r\n" and `^` just matches "\n" so there is some difference.

lib/rack/multipart/parser.rb
aiomaster: I don't understand the logic completely, but I guess it is the `else`-path of the check, that uses methods like `rest_size` and `peek`. I think the key question is, why this `rx_max_size` stuff is there.

I tried to make the changes as minimal as possible to solve the issue, but maybe we could refactor some of the other stuff, too.

aiomaster:
Ok, I've managed to make a PDF file that shows the problem.
See my commit https://github.com/webit-de/rack/commit/e95568cdf2fa04c064bc697e61c52c115a1d7507
I took a failing PDF from my customer (>250M, >100pages) and reduced it to the failing PDF added in the commit. I used master pdf editor so it added its watermark, but this is not the problem.
If I delete any of the remaining elements, the test passes. I think the filesize has at least to be more than 1M, but I don't really know.
If I delete randomly with an Hexeditor, sometimes the test fails and sometimes it passes.
I have no idea what exactly causes the issue, but it is a combination of the given PDF and the used boundary.
If you try a different boundary for example "foo" the test passes. This is the reason, why the same document works in a browser like chrome but not in firefox.
I think the problem is the number of "-" in the boundary.

Anyway, if you use my patch the problem has gone.

What I am unsure about is, if you want to add my failing PDF with the test to the repo or not.
It would be cool to know exactly which part in the PDF causes the problem to make a more minimalistic data snippet, but I was not able to produce such a PDF from scratch. Even if I replace one of the images with another image with the same size the test passes and the problem could not be reproduced.

For me I think, I can not do any more to the bug.
What do you think?

aiomaster:
Another thing that came to my mind is, that some logic in the c-code of the regex stuff is buggy.
I don't really know what the `strscan_do_scan` called by [`check_until`](https://ruby-doc.org/stdlib-2.5.3/libdoc/strscan/rdoc/StringScanner.html#method-i-check_until) is doing exactly, but maybe there is a bug somewhere.

matthewd:
This seems sufficient to fail for me, does it reproduce for you too?

```diff
diff --git a/test/spec_multipart.rb b/test/spec_multipart.rb
index cdd576b..ab561fc 100644
--- a/test/spec_multipart.rb
+++ b/test/spec_multipart.rb
@@ -157,7 +157,8 @@ def rd.rewind; end
     data.write("\r\n")
     data.write("Content-Type:application/pdf\r\n")
     data.write("\r\n")
-    data.write(File.read('test/test_1308.pdf'))
+    data.write("-" * (1024 * 1024))
+    data.write("\r\n")
     data.write("--#{boundary}--\r\n")

     fixture = {
```

aiomaster:
Yep, thats it. Now I added the test to my PR with your changes.

Now we can easily reproduce the issue and we have a workaround for rack.
Now we could try to refactor a little more or tell the ruby poeple, that there could be a problem in the c-code used by StringScanner with the given regex and the data.
If you like you can also just merge the workaround and fine.
What do you think?

aiomaster:
The latest released version of rack has this bug and so will hit all the users that allow larger fileuploads.
It occures using firefox and with files with parts of multiple dash signs (larger PDFs have a higher propability to trigger that bug)
I think this is an important issue and should get fixed soon.
I rebased my branch from the current master and hope that someone will merge it before others will have these kind of problems.

aiomaster:
The bug still exists in the current rack and is present with ruby 2.6.1, so it would really be cool if someone could have a look at this PR again and merge it maybe.
I rebased with the current master.

jeremyevans:
This pull request looks good to me and I plan to merge it unless there are objections.  The current behavior appears to be vulnerable to a denial of service attack, so after merging we should probably backport to 2.1 (this does not affect 2.0 and below).

aiomaster:
Thanks. Should I rebase on master?
The merge would allow us to use the upstream version of rack again otherwise our application goes down every day.
We had lots of PDF uploads and a user with firefox for example can trigger that bug and a denial of service even if not intended.
I just wonder why no one else had these problems for that long time and this issue is still open. I thought everyone in the ruby world uses rack for web applications.

jeremyevans:
Thanks for rebasing.   I'm guessing the reason this hasn't been widely reported is that the issue only appears in rack 2.1+, which was only released a few days ago.  I assume if we don't fix this soon, we'll start getting more reports.

PR #1314, opened 2018-11-10
Introduce Request.scheme_allowlist and deprecate existing constant

JuanitoFatas:
## Problem 

Update the constant to use clearer terminology. 

## Description:

Deprecates the constant, provide a new class method `Request.scheme_allowlist` for anyone who still needs it.

[Original motivation](https://github.com/rails/rails/issues/33677), other community efforts examples:

- https://github.com/rails/rails/pull/33681
- https://github.com/graphiti-api/graphiti/pull/10
- https://github.com/rails/rails/pull/33718
- https://github.com/ruby/psych/pull/378
- https://github.com/ruby/ruby/pull/2008
- https://github.com/ruby/ruby/pull/2009
- https://github.com/rubocop-hq/rubocop/pull/6464
- https://github.com/rubygems/rubygems/pull/2463
- https://github.com/dtao/safe_yaml/pull/93
- https://github.com/rubocop-hq/rubocop/pull/6466
- https://github.com/rubocop-hq/rubocop/pull/6467
- https://github.com/rspec/rspec-core/pull/2576
- https://github.com/rspec/rspec-expectations/pull/1083
- https://github.com/rspec/rspec-mocks/pull/1246
- https://github.com/rspec/rspec-support/pull/356
- https://github.com/flavorjones/loofah/pull/158
- https://github.com/pry/pry/pull/1874
- https://github.com/pry/pry/pull/1875

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file. For info on
 
 ### Changed
 - `Rack::Utils.status_code` now raises an error when the status symbol is invalid instead of `500`.
+- `Rack::Request::SCHEME_WHITELIST` has been renamed to `Rack::Request::ALLOWED_SCHEMES`
 
 ### Removed
 - HISTORY.md by @twitnithegirl
@@ -16,7 +17,7 @@ All notable changes to this project will be documented in this file. For info on
 #
 #
 # History/News Archive
-Items below this line are from the previously maintained HISTORY.md and NEWS.md files. 
+Items below this line are from the previously maintained HISTORY.md and NEWS.md files.
 #
 
 ## [2.0.0]
@@ -65,13 +66,13 @@ Items below this line are from the previously maintained HISTORY.md and NEWS.md 
 - Prevent extremely deep parameters from being parsed. CVE-2015-3225
 
 ## [1.6.1] 2015-05-06
-  - Fix CVE-2014-9490, denial of service attack in OkJson 
-  - Use a monotonic time for Rack::Runtime, if available 
+  - Fix CVE-2014-9490, denial of service attack in OkJson
+  - Use a monotonic time for Rack::Runtime, if available
   - RACK_MULTIPART_LIMIT changed to RACK_MULTIPART_PART_LIMIT (RACK_MULTIPART_LIMIT is deprecated and will be removed in 1.7.0)
 
 ## [1.5.3] 2015-05-06
   - Fix CVE-2014-9490, denial of service attack in OkJson
-  - Backport bug fixes to 1.5 series 
+  - Backport bug fixes to 1.5 series
 
 ## [1.6.0] 2014-01-18
   - Response#unauthorized? helper
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -18,7 +18,11 @@ module Rack
     end
 
     self.ip_filter = lambda { |ip| ip =~ /\A127\.0\.0\.1\Z|\A(10|172\.(1[6-9]|2[0-9]|30|31)|192\.168)\.|\A::1\Z|\Afd[0-9a-f]{2}:.+|\Alocalhost\Z|\Aunix\Z|\Aunix:/i }
-    SCHEME_WHITELIST = %w(https http).freeze
+    ALLOWED_SCHEMES = %w(https http).freeze
+    SCHEME_WHITELIST = ALLOWED_SCHEMES
+    if Object.respond_to?(:deprecate_constant)
+      deprecate_constant :SCHEME_WHITELIST
+    end
 
     def initialize(env)
       @params = nil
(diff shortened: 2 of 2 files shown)


tonytonyjan:
The term whitelist/blacklist is nothing to do with racist.

mikegee:
In my opinion, "allowed_schemes" is a bit clearer than "scheme_allowlist".

lib/rack/request.rb
```diff
@@ -19,6 +19,20 @@ class << self
...
     SCHEME_WHITELIST = %w(https http).freeze
+    if Object.respond_to?(:deprecate_constant)
+      deprecate_constant :SCHEME_WHITELIST
+    else
+      if $VERBOSE
+        warn("SCHEME_WHITELIST is deprecated and will be removed in a future version of Rack. Please use Rack::Request.scheme_allowlist instead.")
+      end
+    end
+
+    def self.scheme_allowlist
```
rafaelfranca: I think we don't need a class method for this. This constant was never made to be exposed as public. Just a public `ALLOWED_SCHEMAS` is fine.

mikegee:
I think these are called "schemes", not "schemas".

https://www.iana.org/assignments/uri-schemes/uri-schemes.xhtml

rafaelfranca:
Good point @mikegee. @JuanitoFatas can you rename it?

JuanitoFatas:
Updated and trailing spaces in Changelog got trimmed by my editor.

lib/rack/request.rb
```diff
@@ -18,7 +18,15 @@ class << self
...
 
     self.ip_filter = lambda { |ip| ip =~ /\A127\.0\.0\.1\Z|\A(10|172\.(1[6-9]|2[0-9]|30|31)|192\.168)\.|\A::1\Z|\Afd[0-9a-f]{2}:.+|\Alocalhost\Z|\Aunix\Z|\Aunix:/i }
-    SCHEME_WHITELIST = %w(https http).freeze
+    ALLOWED_SCHEMES = %w(https http).freeze
+    SCHEME_WHITELIST = ALLOWED_SCHEMES
+    if $VERBOSE
+      if Object.respond_to?(:deprecate_constant)
+        deprecate_constant :SCHEME_WHITELIST
+      else
+        warn("SCHEME_WHITELIST is deprecated and will be removed in a future version of Rack. Please use ALLOWED_SCHEMES instead.")
```
rafaelfranca: I just realized that this will always warn even if `SCHEME_WHITELIST` is not used by the app. Is that desired?

JuanitoFatas: I did some experiments:

```ruby
class Namespace
  ALLOWED_SCHEMES = %w(http https)

  SCHEME_WHITELIST = ALLOWED_SCHEMES
  if $VERBOSE
    if Object.respond_to?(:deprecate_constant)
      deprecate_constant :SCHEME_WHITELIST
    else
      warn("SCHEME_WHITELIST is deprecated. Please use ALLOWED_SCHEMES instead.")
    end
  end
end

puts "Reference ALLOWED_SCHEMES"
Namespace::ALLOWED_SCHEMES

puts "Reference SCHEME_WHITELIST"
Namespace::SCHEME_WHITELIST
```

=>


```
$ ruby -w namespace.rb
Reference ALLOWED_SCHEMES
Reference SCHEME_WHITELIST
1.rb:19: warning: constant Namespace::SCHEME_WHITELIST is deprecated
```

Only when we actually reference `SCHEME_WHITELIST` and `$VERBOSE` is on will emit warnings, loading the file will not emit warnings.

rafaelfranca: But is not that because `deprecate_constant` is defined? If you don't have it it happens as soon that like 27 is executed

JuanitoFatas: Ah I understand now. Yes, for Ruby < 2.3.0, if they turn on warning (`-w`, `--verbose`, `RUBYOPT=-w`), then a warning will show as soon as L27 executed. What do you suggest?

rafaelfranca: I'd say, let's remove the else branch for now. There is a chance that rack master will not support Ruby 2.2

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file. For info on
 
 ### Changed
 - `Rack::Utils.status_code` now raises an error when the status symbol is invalid instead of `500`.
+- `Rack::Request::SCHEME_WHITELIST` now deprecated, please use `Rack::Request::ALLOWED_SCHEMES` instead
```
matthewd: I think this would be clearer phrased as "has been renamed to"; the deprecation behaviour seems the less interesting part. (I'll also note we aren't helping anyone who's changing/overriding the constant, which feels at least as likely as someone using the constant to get `["http", "https"]`. But that's probably fine.)

JuanitoFatas:
Thanks everyone for comment and reviews 🙇

PR #1326, opened 2018-12-19
Fix multipart forms for Phusion Passenger in Rack 2.0

gsamokovarov:
The multipart parser expects its given Input Stream to respond to `#eof?`. However, this is not part of the [Rack SPEC](https://www.rubydoc.info/github/rack/rack/master/file/SPEC#label-The+Input+Stream). When `io.rewind` is called in Passenger, a `PhusionPassenger::Utils::TeeInput` is returned that does not respond to `#eof?` and _is_ SPEC compliant. This breaks all of the multipart forms in our main production application.

The fix has been in master since 2017 in df4a028a9c992db19ee50e7d8cc00d5d1f54f1a0. This is a backport of that change for the Rack 2 series as it has real-world implications.

### DIFF
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -39,8 +39,6 @@ module Rack
           str
         end
 
-        def eof?; @content_length == @cursor; end
-
         def rewind
           @io.rewind
         end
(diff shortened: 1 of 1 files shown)


gsamokovarov:
This was just backported. Thank you!

PR #1328, opened 2018-12-20
Performance: Transform `Utils::STATUS_WITH_NO_ENTITY_BODY` into a `Hash`

jodosha:
## Change

`Utils::STATUS_WITH_NO_ENTITY_BODY` is a `Set` instance and Rack uses `#include?`. My assumption is that `Set` is used to guarantee uniqueness of values and to have faster access than `Array`.

This is a proposal to change that constant from a `Set` into a `Hash`. It guarantees uniqueness of keys, and slightly faster access via `#key?`.

Please note that the use of `true` is just a dummy value. But it's important to use a singleton method to keep low the memory footprint.

```ruby
#!/usr/bin/env ruby
# frozen_string_literal: true

require "benchmark/ips"
require "set"

STATUS_WITH_NO_ENTITY_BODY_SET  = Set.new((100..199).to_a << 204 << 304)
STATUS_WITH_NO_ENTITY_BODY_HASH = Hash[((100..199).to_a << 204 << 304).product([true])]

Benchmark.ips do |x|
  x.report("set")  { STATUS_WITH_NO_ENTITY_BODY_SET.include?(304) }
  x.report("hash") { STATUS_WITH_NO_ENTITY_BODY_HASH.key?(304) }
  x.compare!
end

__END__
Result:

                 set   258.490k i/100ms
                hash   263.442k i/100ms
                 set      7.528M (± 4.9%) i/s -     37.740M in   5.025552s
                hash      8.345M (± 5.4%) i/s -     41.624M in   5.004100s

Comparison:
                hash:  8344683.5 i/s
                 set:  7528352.1 i/s - same-ish: difference falls within error

Ruby:

    ruby 2.5.3p105 (2018-10-18 revision 65156) [x86_64-darwin16]

Hardware:

    Hardware Overview:

      Model Name: MacBook Pro
      Model Identifier: MacBookPro12,1
      Processor Name: Intel Core i7
      Processor Speed: 3,1 GHz
      Number of Processors: 1
      Total Number of Cores: 2
      L2 Cache (per Core): 256 KB
      L3 Cache: 4 MB
      Memory: 16 GB
      Boot ROM Version: 180.0.0.0.0
      SMC Version (system): 2.28f7

Software:

    System Software Overview:

      System Version: macOS 10.12.6 (16G1618)
      Kernel Version: Darwin 16.7.0
      Time since boot: 17 days 28 minutes
```

### DIFF
--- a/lib/rack/chunked.rb
+++ b/lib/rack/chunked.rb
@@ -57,7 +57,7 @@ module Rack
       headers = HeaderHash.new(headers)
 
       if ! chunkable_version?(env[HTTP_VERSION]) ||
-         STATUS_WITH_NO_ENTITY_BODY.include?(status.to_i) ||
+         STATUS_WITH_NO_ENTITY_BODY.key?(status.to_i) ||
          headers[CONTENT_LENGTH] ||
          headers[TRANSFER_ENCODING]
         [status, headers, body]
--- a/lib/rack/content_length.rb
+++ b/lib/rack/content_length.rb
@@ -17,7 +17,7 @@ module Rack
       status, headers, body = @app.call(env)
       headers = HeaderHash.new(headers)
 
-      if !STATUS_WITH_NO_ENTITY_BODY.include?(status.to_i) &&
+      if !STATUS_WITH_NO_ENTITY_BODY.key?(status.to_i) &&
          !headers[CONTENT_LENGTH] &&
          !headers[TRANSFER_ENCODING] &&
          body.respond_to?(:to_ary)
--- a/lib/rack/content_type.rb
+++ b/lib/rack/content_type.rb
@@ -21,7 +21,7 @@ module Rack
       status, headers, body = @app.call(env)
       headers = Utils::HeaderHash.new(headers)
 
-      unless STATUS_WITH_NO_ENTITY_BODY.include?(status.to_i)
+      unless STATUS_WITH_NO_ENTITY_BODY.key?(status.to_i)
         headers[CONTENT_TYPE] ||= @content_type
       end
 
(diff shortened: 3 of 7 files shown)


rafaelfranca:
Given the benchmark shows that the difference falls within error it is hard to judge if there is any performance difference for this change. Were you able to see this in different benchmarks?

jodosha:
@rafaelfranca Not really, but I still proposed this change because the number of i/s is slightly better (+810K~) for `Hash` instead of `Set`.

I'll leave it to your judgment. Thanks for having a look at all of them. 👍

rafaelfranca:
I got uncomfortable with the values being just a dummy value.

Looking the implementation of `Set#include?` and `Hash#key?` there is only the overhead of a `Hash#[]` method call in `Set#include?` when the element is on the set, given `Hash#[]` and `Hash#key?` are almost the same for when the element exists in the hash. For the worst case (when the element is not in the set) the `Hash#key?` would be faster because `Hash#[]` checks for the default value while `key?` just return false.

So for the case where the response has a method body it is indeed 20% faster:

```
#!/usr/bin/env ruby
# frozen_string_literal: true

require "benchmark/ips"
require "set"

STATUS_WITH_NO_ENTITY_BODY_SET  = Set.new((100..199).to_a << 204 << 304)
STATUS_WITH_NO_ENTITY_BODY_HASH = Hash[((100..199).to_a << 204 << 304).product([true])]

Benchmark.ips do |x|
  x.report("set")  { STATUS_WITH_NO_ENTITY_BODY_SET.include?(500) }
  x.report("hash") { STATUS_WITH_NO_ENTITY_BODY_HASH.key?(500) }
  x.compare!
end

__END__
                 set   323.563k i/100ms
                hash   363.396k i/100ms
                 set     12.212M (± 3.5%) i/s -     61.153M in   5.014672s
                hash     15.116M (± 4.6%) i/s -     75.586M in   5.012996s

Comparison:
                hash: 15116435.9 i/s
                 set: 12211765.4 i/s - 1.24x  slower
```

PR #1332, opened 2019-01-14
Add status code: 103 Early Hints

egtra:
This is defined in RFC 8297.
https://tools.ietf.org/html/rfc8297


tenderlove:
Nice, thank you!

egtra:
Thank you!

PR #1337, opened 2019-01-29
Add Lambda handler

timoschilling:
This handler allows it to run a Rack app inside a AWS Lambda.

This is currently only an PR proposal, if it's gets a 👍, I will add some specs and some docs.

### DIFF
--- a/lib/rack/handler/lambda.rb
+++ b/lib/rack/handler/lambda.rb
@@ -1,0 +1,47 @@
+# frozen_string_literal: true
+
+require 'rack'
+
+module Rack
+  module Handler
+    class Lambda
+      def self.run(event:, context:)
+        @app ||= Rack::Builder.parse_file("#{ENV['LAMBDA_TASK_ROOT']}/config.ru").first
+        serve env(event, context)
+      end
+
+      def self.serve(env)
+        status, headers, body = @app.call(env)
+
+        {
+          'statusCode' => status,
+          'headers' => headers,
+          'body' => body.map(&:to_s)
+        }
+      end
+
+      def self.env(event, context)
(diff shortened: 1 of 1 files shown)


HoneyryderChuck:
just for the sake of documentation [a serverless rack plugin](https://github.com/logandk/serverless-rack/blob/master/lib/serverless_rack.rb) already exists. I don't know what rack itself gains of shipping with a lambda adapter though.

timoschilling:
@tenderlove I think the same reason why rack has handlers for puma, unicorn and co.

tenderlove:
> @tenderlove I think the same reason why rack has handlers for puma, unicorn and co.

Rack doesn't actually have handlers for Puma or Unicorn, they're defined in the webserver gems themselves.  The list of build-in handlers is [here](https://github.com/rack/rack/tree/master/lib/rack/handler).

I'm not opposed to putting handlers in Rack itself, but I think it's better to be put in a gem because

1. I don't use AWS lambda, so if there's a bug I don't think I'll be a good person to fix it
2. If there is a bug it won't get released until there is a Rack release (I don't release Rack that often)
3. Development speed will have to match Rack's development speed

This is why the popular web servers don't ship handlers in Rack.  Given there's already a serverless Rack solution, maybe adding an implementation as an example would be better.

What do you think?

timoschilling:
> > @tenderlove I think the same reason why rack has handlers for puma, unicorn and co.
> Rack doesn't actually have handlers for Puma or Unicorn, they're defined in the webserver gems themselves. The list of build-in handlers is [here](https://github.com/rack/rack/tree/master/lib/rack/handler).

Sorry I listed the wrong handlers 🤦‍♂️ 

Ok, I will move it into a separate gem.

PR #1344, opened 2019-02-20
Don't add headers when gzipped file returned 304

chooh:
Steps to reproduce:

1. use `Rack::Static` with `gzip: true`
2. make two sequential request for a static asset with a browser.

The first request will succeed with **200 OK** and `Last-Modified` header set.

The second request will be made with `If-Modified-Since` header and failed with `Rack::Lint::LintError: Content-Type header found in 304 response, not allowed`.

`Rack::File` correctly handles 304 response headers. But `Rack::Static` with gzip enabled adds `Content-Type` header whenever the response is not 404.

### DIFF
--- a/lib/rack/static.rb
+++ b/lib/rack/static.rb
@@ -130,6 +130,8 @@ module Rack
 
           if response[0] == 404
             response = nil
+          elsif response[0] == 304
+            # Do nothing, leave headers as is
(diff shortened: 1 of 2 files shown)


jeremyevans:
This looks good, thanks for the patch!

Issue #1306, opened 2018-10-21
sendfile with nginx/X-Accel-Redirect % or ? character encoding issue


dup2:
Using nginx with X-Accel-Redirect as documented within `lib/rack/sendfile.rb` does not seem to work for files which contain a `%` or `?` character in their name.

How to reproduce:

* Setup nginx with a location and X-Accel-Mapping as documented to allow sending files from `/data/`
* Create a file `/data/sample/Report90%.pdf`
* Trigger sendfile with the proper headers and file location pointing to `/data/sample/Report90%.pdf`

Nginx will accept the internal X-Accel-Redirect request but fail to read the file `/data/sample/Report90%.pdf` and return a 404 error to the client. This does not happen when the `%` is removed from the file name.

We tried to replace the `%` character with `%25` (and therefore triggering X-Accel-Mapping with a path of `/data/sample/Report90%25.pdf`) which seems to work. 

According to the changelog of nginx as of 2014 for version 1.5.9: `now nginx expects escaped URIs in "X-Accel-Redirect" headers.`

diskkid:
I face the same issue and want it to be fixed. I think rack should set `X-Accel-Redirect` to url-encoded path.

jeremyevans:
This appears to have been fixed by 78e837e4d0ef95bcf2dce8c38a1b411032240afa.

PR #1346, opened 2019-03-15
Set X-Accel-Redirect to percent-encoded path

diskkid:
This patch makes it possible to serve a file whose name includes % symbol or other non-ascii codes using X-Accel-Redirect and solves #1306.

As far as I searched, nginx expects X-Accel-Redirect is set to percent-encoded path after [this patch is merged](https://github.com/nginx/nginx/commit/f7ff5e65d0d20ba0425be7e3d8de4d04ceec9206) about 5 years ago.

I tested accessing to files under /var/www using nginx 1.14.0 and configs below.

```config.ru
require 'pathname'
require 'rack/sendfile'

use Rack::Sendfile

run ->(env) do
  path = URI.decode(env['REQUEST_PATH'])
  file = Pathname.new('/var/www' + path)
  [200, {}, file]
end
```

```/etc/nginx/sites-enabled/default
server {
    listen       80;
    server_name  localhost;

    location / {
        proxy_redirect     off;

        proxy_set_header   Host                $host;
        proxy_set_header   X-Real-IP           $remote_addr;
        proxy_set_header   X-Forwarded-For     $proxy_add_x_forwarded_for;

        proxy_set_header   X-Sendfile-Type     X-Accel-Redirect;
        proxy_set_header   X-Accel-Mapping     /var/www/=/foo/bar%/;

        proxy_pass http://127.0.0.1:9292;
    }

    location ~ /foo/bar%/(.*) {
        internal;
        alias /var/www/$1;
    }
}
```

### DIFF
--- a/lib/rack/sendfile.rb
+++ b/lib/rack/sendfile.rb
@@ -117,7 +117,8 @@ module Rack
           path = ::File.expand_path(body.to_path)
           if url = map_accel_path(env, path)
             headers[CONTENT_LENGTH] = '0'
-            headers[type] = url
+            # '?' must be percent-encoded because it is not query string but a part of path
+            headers[type] = ::Rack::Utils.escape_path(url).gsub('?', '%3F')
             obody = body
             body = Rack::BodyProxy.new([]) do
               obody.close if obody.respond_to?(:close)
--- a/test/spec_sendfile.rb
+++ b/test/spec_sendfile.rb
@@ -8,10 +8,10 @@ require 'rack/mock'
 require 'tmpdir'
 
 describe Rack::Sendfile do
-  def sendfile_body
-    FileUtils.touch File.join(Dir.tmpdir,  "rack_sendfile")
+  def sendfile_body(filename = "rack_sendfile")
+    FileUtils.touch File.join(Dir.tmpdir,  filename)
     res = ['Hello World']
-    def res.to_path ; File.join(Dir.tmpdir,  "rack_sendfile") ; end
+    res.define_singleton_method(:to_path) { File.join(Dir.tmpdir,  filename) }
     res
   end
 
@@ -71,6 +71,19 @@ describe Rack::Sendfile do
       response.body.must_be :empty?
       response.headers['Content-Length'].must_equal '0'
       response.headers['X-Accel-Redirect'].must_equal '/foo/bar/rack_sendfile'
+    end
+  end
+
+  it "sets X-Accel-Redirect response header to percent-encoded path" do
+    headers = {
+      'HTTP_X_SENDFILE_TYPE' => 'X-Accel-Redirect',
+      'HTTP_X_ACCEL_MAPPING' => "#{Dir.tmpdir}/=/foo/bar%/"
+    }
+    request headers, sendfile_body('file_with_%_?_symbol') do |response|
+      response.must_be :ok?
+      response.body.must_be :empty?
+      response.headers['Content-Length'].must_equal '0'
+      response.headers['X-Accel-Redirect'].must_equal '/foo/bar%25/file_with_%25_%3F_symbol'
     end
   end
 


vongruenigen:
:+1:

lib/rack/sendfile.rb
```diff
@@ -117,7 +118,7 @@ def call(env)
           path = ::File.expand_path(body.to_path)
           if url = map_accel_path(env, path)
             headers[CONTENT_LENGTH] = '0'
-            headers[type] = url
+            headers[type] = WEBrick::HTTPUtils.escape(url)
```
rafaelfranca: Does `Rack::Utils.escape` work here?

dup2:
👍

lib/rack/sendfile.rb
vongruenigen: No, since ``Rack::Utils.escape`` escapes the URL as a whole (i.e. it also escapes the frontslashes, colons etc.) whereas ``WEBrick::HTTPUtils.escape`` keeps the URL intact and only escapes the characters necessary:

```ruby
2.5.3 :001 > require 'rack'
2.5.3 :002 > require 'webrick'
2.5.3 :003 > url = 'http://test.com/hello%/world'
2.5.3 :004 > Rack::Utils.escape(url)
 => "http%3A%2F%2Ftest.com%2Fhello%25%2Fworld" 
2.5.3 :005 > WEBrick::HTTPUtils.escape(url)
 => "http://test.com/hello%25/world" 
```

diskkid: Thank you for your comments! 😃

It seems that nginx decodes whatever percent-encoded characters, so encoding '/' to '%2F' is not a problem. That works fine.
Instead, there is another problem that `Rack::Utils.escape` encodes ' ' (space) to '+'. Because nginx does not decode '+' to ' ', `Rack::Utils.escape` does not work if a path includes space characters.
`WEBrick::HTTPUtils.escape` works fine in the both cases.

rafaelfranca: We should not be using `WEBrick` though in rack. It is not a dependency of the library and I don't think it should be since users may opt out to a different web server.

dup2: Isn't `WEBrick` part of the standard library/default gems in ruby?

rafaelfranca: Yes, but is not required by default. You can only use it if you want. If use in rack you will have to load it, and I want to avoid this.

vongruenigen: How about using ``URI#escape`` then? I quickly tested it in irb and it seems to provide the same functionality:

```ruby
2.5.3 :001 > require 'uri'
 => true 
2.5.3 :002 > URI.escape('http://test.com/hello%/world')
 => "http://test.com/hello%25/world"
````

I'm not sure about availability in different ruby versions, though.

vongruenigen:
@rafaelfranca, it seems @diskkid is now escaping the path using ``Rack::Utils.escape_path``, I guess this make this PR good to merge?

diskkid:
Sorry for my late reply.
After I pushed the branch, I found that using `Rack::Utils.escape_path` is problematic because it could not handle '?' properly. I am still searching for an appropriate way.

I summarize results when I pass `/var/www/foo _%_?` to following methods.

|                           | Encoded path                             | Problem
----------------------------|------------------------------------------|---------------------------------
`Rack::Utils.escape`        |  `%2Fvar%2Fwww%2Ffoo+_%25_%3F`           | space to plus
`Rack::Utils.escape_path`   | `/var/www/foo%20_%25_?`                  | question remains
`WEBrick::HTTPUtils.escape` | `/var/www/foo%20_%25_?`                  | question remains/additional dependency
`URI.escape`                | `/var/www/foo%20_%25_?`                  | question remains/obsolete method
Should be                   | `/var/www/foo%20_%25_%3F` (`%2Fvar%2Fwww%2Ffoo%20_%25_%3F` is also OK) |

IMHO, it is reasonable that we use `Rack::Utils.escape` then replace '+' with '%20' or use `Rack::Utils.escape_path` then replace '?' with '%3F'. How do you think, @rafaelfranca?

diskkid:
I am still not sure which way is the best, but anyway I choose `Rack::Utils.escape_path` and add `gsub` to encode question mark.

BTW, It seems that CI is corrupted because rubocop installation fails... :(

vongruenigen:
@rafaelfranca, would you mind looking at this PR again, please? Also, do you know how we could fix the CI problems?

rafaelfranca:
ACK! Will look tomorrow. (on the CI problem too)

vongruenigen:
@rafaelfranca, thanks a lot for the quick response! :+1:

diskkid:
@rafaelfranca I pushed again and CI works well! Thanks.

diskkid:
@rafaelfranca Could you merge this patch if everything is ok?

vongruenigen:
@rafaelfranca, what's the path forward with this PR? It's now been open for over three months..

PR #1350, opened 2019-04-09
 Make logic more readable.

lanzhiheng:


lib/rack/conditional_get.rb
```diff
@@ -56,7 +56,7 @@ def fresh?(env, headers)
     end
 
     def etag_matches?(none_match, headers)
-      etag = headers['ETag'] and etag == none_match
+      headers['ETag'] == none_match
```
rafaelfranca: This changed the behavior. If both `headers['ETag']` and `none_match` are `nil` or false this will return true now, while it returned false with the previous code. If you want to improve this, to keep the same behavior it needs to be.

```ruby
etag = headers['ETag']
etag && etag == none_match
```

lib/rack/common_logger.rb
```diff
@@ -66,8 +66,8 @@ def log(env, status, header, began_at)
     end
 
     def extract_content_length(headers)
-      value = headers[CONTENT_LENGTH] or return '-'
-      value.to_s == '0' ? '-' : value
+      value = headers[CONTENT_LENGTH]
+      value.nil? || value.to_s == '0' ? '-' : value
```
rafaelfranca: While very unlikely to happen, this also changes the behavior when `headers[CONTENT_LENGTH]` is `false`.

```ruby
value = headers[CONTENT_LENGTH]
!value || value.to_s == '0' ? '-' : value
```

to keep the same behavior.

lanzhiheng: I think the value from `headers[CONTENT_LENGTH]` can be a `String` or `nil`. Is it possible to be true or false ?

lib/rack/conditional_get.rb
lanzhiheng: Thank you for your suggestion.

lib/rack/common_logger.rb
lanzhiheng: Like this warning "a header value must be a String, but the value of 'Content-Length' is a FalseClass"

PR #1351, opened 2019-04-09
 Remove `@has_app` from `Rack::Cascade`.

lanzhiheng:
I don't think we need `@has_app` in this class. 

1. `@apps` is an array we can use the `Array#include?` to check if target element in it.
2. if the `@apps` array not very large it is not a good choice to use `Hash` to store the status.

### DIFF
--- a/lib/rack/cascade.rb
+++ b/lib/rack/cascade.rb
@@ -11,7 +11,7 @@ module Rack
     attr_reader :apps
 
     def initialize(apps, catch = [404, 405])
-      @apps = []; @has_app = {}
+      @apps = []
       apps.each { |app| add app }
 
       @catch = {}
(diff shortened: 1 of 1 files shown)


kares:
... or just use a Set so there's no need to worry about O(n)

ioquatix:
A set is just a wrapper around `Hash`.

ioquatix:
This PR simplifies the implementation but also introduces a performance regression.

I don't know the real world implications of the performance regression - does anyone expect to call `#include?`?

If you can show that this code is not used (frequently), then I prefer the simpler code.

That being said, I have to ask @lanzhiheng to do the work here.

@Jeremy can you comment on your use case?

jeremy:
No idea. 2009 😂

ioquatix:
Thanks for your attention :)

PR #1352, opened 2019-04-10
 Add some unit testing for `condition_get.rb`.

lanzhiheng:


lanzhiheng:
@rafaelfranca  I got the jruby issue, can you rerun the CI. I don't think it is cause by my code.

rafaelfranca:
Thank you for the pull request. We don't test private methods by calling them directly. If you want to test those methods you need to go though the public API, and I believe they are already being tested in this file using the public API. If you think the current tests are not enough please add more tests using the public API.

Thanks

lanzhiheng:
@rafaelfranca Got you. Thank you for your comments.

PR #1353, opened 2019-04-16
Add method for custom ip_filter and test to go with it

larsxschneider:
This is a backport to `2-0-stable` of @svcastaneda's fix which was merged to `master` with https://github.com/rack/rack/pull/1290

---

This pull request extracts the `ip_filter` into a lambda so that users can configure their own ip_filter (e.g., #1276).

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -13,6 +13,12 @@ module Rack
   class Request
     SCHEME_WHITELIST = %w(https http).freeze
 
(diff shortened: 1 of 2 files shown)


rafaelfranca:
Thank you for the pull request but we can't add new features to a stable release.

larsxschneider:
@rafaelfranca thanks for your feedback. I understand your reasoning 👍

PR #1355, opened 2019-04-17
Use Regexp match? instead of =~

krzysiek1507:
`Regexp#match?` is way faster than `=~` so we can gain a little bit of speed.

Ruby introduced `Regexp#match?` in 2.4 so we need to polyfill it for Ruby < 2.4.

### DIFF
--- a/lib/rack/builder.rb
+++ b/lib/rack/builder.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative 'core_ext/regexp'
+
 module Rack
   # Rack::Builder implements a small DSL to iteratively construct Rack
   # applications.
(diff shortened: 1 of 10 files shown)


HoneyryderChuck:
Why not as a refinement?

krzysiek1507:
I've copied this from [ActiveSupport](https://github.com/rails/rails/blob/ac717d65a31d05458588b78ea7719b79f8ea69e5/activesupport/lib/active_support/core_ext/regexp.rb#L8-L10).

HoneyryderChuck:
if active support's implementation change (highly unlikely), you'll have to copy it again. If `match?` has been already defined, but it does something else, your code will break. 

Since rack officially supports ruby 2.2 or higher, and this would be an easy refinement, you could do it as easy as:

```ruby
module RegexpExtensions
  refine Regexp do
    def match?....

....

class Rack::Builder
  use RegexpExtensions if RUBY_VERSION > "2.3"
  ...

krzysiek1507:
@HoneyryderChuck you are right. Changed! Thanks!

krzysiek1507:
I think it's ready for review.

mikegee:
seems legit to me 👍

PR #1356, opened 2019-04-24
Add MIME type for .mjs javascript modules

eiko:
While it's possible to use the `.js` extension for javascript modules, NodeJS requires the `.mjs` extension. Adding the MIME type for `.mjs` ensures compatibility with javascript modules designed to run both on NodeJS and in the browser.

### DIFF
--- a/lib/rack/mime.rb
+++ b/lib/rack/mime.rb
@@ -331,6 +331,7 @@ module Rack
       ".mif"       => "application/vnd.mif",
       ".mime"      => "message/rfc822",
       ".mj2"       => "video/mj2",
+      ".mjs"       => "application/javascript",
       ".mlp"       => "application/vnd.dolby.mlp",
       ".mmd"       => "application/vnd.chipnuts.karaoke-mmd",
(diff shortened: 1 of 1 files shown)


IngusSkaistkalns:
I know its year 2022, but why was this closed?

eikobear:
> The person who opened it closed it the same day, so you would have to ask them. My guess is they realized they can just update `Rack::Mime::MIME_TYPES` directly in their application.

```javascript
Rack::Mime::MIME_TYPES.merge!({ '.mjs' => 'application/javascript' })
```

I did end up updating it directly, but IIRC I decided that use of `.mjs` in Node JS was too experimental / up in the air to merit propagating it across ecosystems at the time.

Since then, `.mjs` has become more of a standard, and its use in the browser is encouraged [in some places](https://v8.dev/features/modules#mjs), but I'm still on the fence about whether it should be built into Rack, or whether the user should opt in

ioquatix:
It would be fine to add this to mime types and it makes sense to me.

Issue #1322, opened 2018-12-14
New 2.0 release including X_Forwarded_For fixes from #1224?


jperville:
Hello,

PR #1224 has been merged to master last April but is not included in the latest release (2.0.6 as of December 2018). Is there a reason why this fix is not included?

## My problem

We have a Rails-based Rack application deployed at a customer site.

The application can only be accessed through a VPN, which assigns client addresses in the 172.17.0.0/24 range, which is considered "trusted" by this code https://github.com/rack/rack/blob/2.0.6/lib/rack/request.rb#L419-L421 .

As a consequence, the X_FORWARDED_FOR list is filtered to an empty list and Rack fallbacks to REMOTE_ADDR, which looks like 10.128.0.1

On this environment, it is impossible to read the client IP address in the applications logs since all clients requests are logged as coming from IP 10.128.0.1 when it is actually 172.17.0.7 or 123.

Having #1224 included in a new release of this gem would allow us to have the first IP in X_FORWARDED_FOR as the client ip instead of the REMOTE_ADDR , which would be correct in most cases.

mikegee:
Hi @jperville, we've been running the master branch of rack in production because we've needed this other PR (https://github.com/rack/rack/pull/1251) which changed `X-Forwarded-For` to drop ports. I'm also curious about getting the master branch in a release. The maintainers are trying to figure out the release plan as mentioned in my issue a few weeks back: https://github.com/rack/rack/issues/1312

ioquatix:
v2.1.0 was released.

PR #1358, opened 2019-05-09
Added support for SameSite=None cookie value

hennikul:
This value was added in revision 3 of rfc6265bis:
 https://tools.ietf.org/html/draft-ietf-httpbis-rfc6265bis-03#appendix-A.4

The value "None" indicates that cookie is used as a third party cookie. Chrome 76, due for release late July, will default to SameSite=Lax unless SameSite is set. The only way to set cookies used as third party cookies will be to set them with SameSite=None: 
https://web.dev/samesite-cookies-explained/

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -230,6 +230,8 @@ module Rack
           case value[:same_site]
           when false, nil
             nil
+          when :none, 'None', :None
+            '; SameSite=None'
           when :lax, 'Lax', :Lax
             '; SameSite=Lax'
           when true, :strict, 'Strict', :Strict
--- a/test/spec_response.rb
+++ b/test/spec_response.rb
@@ -127,6 +127,24 @@ describe Rack::Response do
     response["Set-Cookie"].must_equal "foo=bar"
   end
 
+  it "can set SameSite cookies with symbol value :none" do
+    response = Rack::Response.new
+    response.set_cookie "foo", { value: "bar", same_site: :none }
+    response["Set-Cookie"].must_equal "foo=bar; SameSite=None"
+  end
+
+  it "can set SameSite cookies with symbol value :None" do
+    response = Rack::Response.new
+    response.set_cookie "foo", { value: "bar", same_site: :None }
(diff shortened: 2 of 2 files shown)


jrgns:
Any idea when will this be released?

dbelling:
Is there a milestone or semantic version which these changes align to? Also curious as to when these will go live.

geordanr:
Any update on when this will become generally available, given the [upcoming changes to Chrome](https://www.chromium.org/updates/same-site)?

bnorton:
I realize that releases of rack are few and far between but must I maintain a fork of rack for the sake of setting `SameSite=None` cookie values?!

ioquatix:
Nope, let's try to get this I the next release.

ioquatix:
Just to check, this appears to be merged, so all we need is another release, right?

bnorton:
> Just to check, this appears to be merged, so all we need is another release, right?

@ioquatix that's what it looks like to me, if you were asking my opinion.

ioquatix:
Awesome, just wanted to check we weren't missing anything.

jrgns:
Yes. There's a couple of things that's been merged, but never released.

See also #1387 and #1322

bnorton:
For those who searched and found this thread/PR -- To make `SameSite=None` cookies work in a rails application (or any rack-based ruby app) you need to also FILTER OUT any clients that don't work nicely with this new flag, namely clients who mistreat the `None` value or drop the cookie when incompatible values are seen.

I have created this gist https://gist.github.com/bnorton/7dee72023787f367c48b3f5c2d71540f based on the rules and guidelines of the Chromium team => https://www.chromium.org/updates/same-site/incompatible-clients

outreach-soren:
Any update on when we can expect this change to be released? The breaking version of chrome is expected to be promoted to stable in a little over a month. 
https://www.chromestatus.com/features/schedule
I'd rather not have to create a fork to cherry-pick that commit in but I also don't want to have to wait til the last minute to implement this change.

jrgns:
@outreach-soren See #1387 , looks like they're aiming for a release next week.

outreach-soren:
Thanks!

PR #1360, opened 2019-05-12
More efficient mechanism for partial hijack.

ioquatix:
Existing method for partial hijack are inefficient. I would like to propose that we use a body that responds to `call` as the method for partial hijack. Additionally, `_nonblock` methods don't make sense for HTTP/2 streams or more advanced models of concurrency.

This could break existing systems where the response body responds to both `each` and `call`. In that situation, `each` should take priority.

### DIFF
--- a/SPEC
+++ b/SPEC
@@ -176,15 +176,12 @@ If rack.hijack? is true then rack.hijack must respond to #call.
 rack.hijack must return the io that will also be assigned (or is
 already present, in rack.hijack_io.
 rack.hijack_io must respond to:
-<tt>read, write, read_nonblock, write_nonblock, flush, close,
-close_read, close_write, closed?</tt>
+<tt>read, write, flush, close, close_read, close_write, closed?</tt>
 The semantics of these IO methods must be a best effort match to
 those of a normal ruby IO or Socket object, using standard
 arguments and raising standard exceptions. Servers are encouraged
 to simply pass on real IO objects, although it is recognized that
 this approach is not directly compatible with SPDY and HTTP 2.0.
-IO provided in rack.hijack_io should preference the
-IO::WaitReadable and IO::WaitWritable APIs wherever supported.
 There is a deliberate lack of full specification around
 rack.hijack_io, as semantics will change from server to server.
 Users are encouraged to utilize this API with a knowledge of their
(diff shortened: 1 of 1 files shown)


ioquatix:
The reason why it's inefficient is because response headers and `rack.*` metadata are mixed up and it requires processing the headers before streaming the body out. Additionally, I feel the use of a response `rack.hijack` key is semantically very weak. I believe for a partial hijack, it's more logical for the "body" of the response to go into the "body" of the response object. Additionally, it's relatively trivial to convert an response that does a partial hijack into one that generates data via `#each`. But knowing that you can directly provide the socket to the underlying body makes streaming responses more efficient.

ioquatix:
Here is a rough outline as to how it can work:

```
class BodyWrapper
	def each(&block)
		unless @thread
			@stream = FakeIO.new
			
			@thread = Thread.new do
				@body.call(@stream)
			end
		end
		
		@stream.each(&block)
	end
	
	def call(stream)
		@body.call(stream)
	end
end
```

ioquatix:
Also while we can make this change backwards compatible, I've made the PR with exactly what I think we should do, without considering backwards compatibility.

jeremyevans:
If the purpose behind this is efficiency, we'd probably want to see what difference this makes in benchmarks.  Especially if #1485 is merged and we can just look for the key in the response headers as opposed to iterating over them.  Even if we need to iterate over the headers, the partial hijack occurs after headers have been set, so iteration has already occurred before the body has been sent, and it seems unlikely a `== 'rack.hijack'` for each header would result in a significant slowdown.

Not to say that checking for `call` on body if it does not respond to `each` is a bad idea, it does seem simpler than having to check a special header.  However, your example has `each` and `call` defined on the same object, in which case I assume we would ignore `call` since `each` is defined.

boazsegev:
Personally I think `hijack` should be a reconsidered as a mainstream solution and become more of a last-resort approach.

`hijack`, at its core, allows Rack to get out of the way... but along with Rack, so does (in many ways) the server gets pushed aside.

This might be an important feature for some edge cases that need TCP/IP access... but for most Web applications this is a heavy burden, as they now need to handle their own IO and network stack.

I understand that [the `rack.upgrade?` extension](https://github.com/boazsegev/iodine/blob/master/SPEC-WebSocket-Draft.md) idea was benched, the discussion frozen and later closed.

But I do propose the we consider an interface that allows Rack to get out of the way **without** pushing the server out of the way, so that the server still controls the IO object and the application is decoupled from any network / IO / protocol data.

ioquatix:
This is a work in progress.

https://github.com/ioquatix/rack/tree/body-proc

PR #1364, opened 2019-05-18
Fix RuboCop version so that it supports Ruby 2.2

okuramasafumi:
Because RuboCop `0.69.0` doesn't support `2.2`.
However, Rack itself still supports `2.2`.

Should RuboCop be downgraded for now maybe?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -12,7 +12,7 @@ c_platforms = Bundler::Dsl::VALID_PLATFORMS.dup.delete_if do |platform|
   platform =~ /jruby/
 end
 
-gem "rubocop", require: false
+gem "rubocop", "0.68.1", require: false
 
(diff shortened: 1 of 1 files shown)


jeremy:
Perhaps we should only run RuboCop on 2.3+ ?

okuramasafumi:
@jeremy I modified CircleCI config not to run RuboCop on 2.2.
I found some weird error on CI that memcached is not found, but I guess it's not related to the change I made.

matthewd:
I think we must lock to a version of Rubocop that supports Ruby 2.2, as long as we support 2.2, because otherwise it may recommend / locally auto-correct to syntax constructs that aren't valid on 2.2.

okuramasafumi:
@matthewd OK, so I changed my change to simply fix RuboCop version.

PR #1365, opened 2019-05-20
Replace 'ETag' with ETAG constant

okuramasafumi:


rafaelfranca:
Thank you for the pull request but there is no advantage of using that constant now that all string are frozen on that file.

PR #1369, opened 2019-06-05
Move server names to a constant

krzysiek1507:


mikegee:
Is this line of code evaluated more than once in the lifetime of an app server process? I suspect the tradeoff between object allocation and readability isn't worth it in this case.

ioquatix:
There has been discussion about making it possible to add servers to this list. Maybe this PR is the start of that.

krzysiek1507:
I know nothing about such a discussion (I don't follow the gem discussions). I just wanted to move this list outside the method.

PR #1382, opened 2019-07-30
Base 64 encoding should be strict

yb66:
Hi there,

According to RFC 6265 [1]:

> To maximize compatibility with user agents, servers that wish to
store arbitrary data in a cookie-value SHOULD encode that data, for
example, using Base64 [RFC4648]

RFC 4648 uses m0 for the encoding (according to String#unpack in the
Ruby docs [2]). It's a SHOULD not a MUST but I don't see anything in
the logs to indicate why the advice isn't followed.

The encoding was introduced in 417ac6a and refactored in e6b6529.

[1] https://tools.ietf.org/html/rfc6265#section-4.1.1
[2] https://ruby-doc.org/core-2.6.3/String.html#method-i-unpack

Regards,
iain

### DIFF
--- a/lib/rack/session/cookie.rb
+++ b/lib/rack/session/cookie.rb
@@ -51,11 +51,11 @@ module Rack
       # Encode session cookies as Base64
       class Base64
         def encode(str)
-          [str].pack('m')
+          [str].pack('m0')
         end
 
         def decode(str)
-          str.unpack('m').first
+          str.unpack('m0').first
         end
 
         # Encode session cookies as Marshaled Base64 data
--- a/lib/rack/session/cookie.rb
+++ b/lib/rack/session/cookie.rb
@@ -6,6 +6,7 @@ require 'rack/request'
 require 'rack/response'
 require 'rack/session/abstract/id'
 require 'json'
+require 'base64'
 
 module Rack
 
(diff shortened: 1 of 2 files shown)


yb66:
Btw, I understand that the RFC can be read as "should encode in some way" or "should encode like 4648" so if it's not a good change, so be it.

What I would advocate for more is that the specs are changed to use data instead of the tests using programmatically packed and unpacked data as it's kind of pointless if that's what the code is doing. What do you think?

Regards,
iain

lib/rack/session/cookie.rb
```diff
@@ -51,11 +51,11 @@ class Cookie < Abstract::Persisted
       # Encode session cookies as Base64
       class Base64
         def encode(str)
-          [str].pack('m')
+          [str].pack('m0')
         end
 
         def decode(str)
```
ahorek: why not just delegate it to `Base64.strict_decode64(str)` which is in the stdlib?

https://github.com/ruby/ruby/blob/c5eb24349a4535948514fe765c3ddb0628d81004/lib/base64.rb#L73

yb66: Hi,

That's what used to happen but was removed in https://github.com/rack/rack/commit/e6b65296a247e23742d85e47cf31582b5883918a for "1.9 support", whatever that means. I don't know if Rack still supports v1.9. The gemspec says not but I'm just following convention. Happy to change.

Regards,
iain

ahorek: by delagating to Base64 `str.unpack('m0').first` could be implemented as `str.unpack1("m0")` on modern rubies. If the original intention was 1.9 support maybe we could reconsider to restore the behaviour, the current version of rack supports at least ruby 2.2.

leahneukirchen:
For encoding, fine, but this just drops the final newline afaics.

For decoding, it will break the existing cookies, no?

yb66:
@leahneukirchen I don't know until I try it, but that does underline my point about using data in the tests instead of mirroring the implementation. I'll try it out a bit and see if it breaks.

Regards,
iain

lib/rack/session/cookie.rb
yb66: Looking again, Rack's Base64 is a superclass to Marshal, JSON and ZipJSON. Any idea where ZipJSON is used? I don't see it being called anywhere in this lib.

yb66: I also had a search through Github and beyond, I don't see anyone using those encoders.

yb66:
I checked, it does break them so I've updated the code to take that into account.

yb66:
I'll fix the rubocop "violations" later.

lib/rack/session/cookie.rb
```diff
@@ -48,55 +49,36 @@ module Session
...
-          rescue
+        def _decode( str, meth = :strict_decode64 )
+          ::Marshal.load( Base64.send meth, str )
+        rescue ArgumentError
+          # Try non-strict Base64 before failing gracefully
+          if meth == :strict_decode64
+            meth = :decode64
+            retry
+          else
+            warn e.message
```
ahorek: undefined local variable or method `e' ?

yb66: That is a very good point. I've fixed it but an accompanying test might be an idea.

yb66: I tried my best but Base64.decode64 will not throw an ArgumentError under any circumstances which means the else branch of line 72 never gets hit. However, I'm loathe to remove it because it so obviously could become an infinite loop if that behaviour ever changes. Let me know what you think.

Circle-CI doesn't seem to like my force push, sorry.

ioquatix:
https://github.com/rack/rack/pull/1460 was merged into master, can you please rebase on master and squash all your commits.

yb66:
@ioquatix Sure. Done that locally but some tests fail for reasons I can't fathom. I'll look into it then push.

Regards,
iain

jeremyevans:
If the current behavior is not causing any problems, I would rather keep it and close this.  Are there any actual problems with the current behavior?

ioquatix:
@jeremyevans We already merged the most important elements of this PR in a previous PR, but if there were other parts that were still relevant, I wanted to give @yb66 the option to rebase his proposed changes. However, it's been over two weeks, so I think we can close this.

@yb66 if you do get around to rebasing and think there are still relevant changes to be made, please feel free to open a new PR.

PR #1383, opened 2019-08-01
Refactor QueryParser::Params#to_params_hash for readability and performance

Drenmi:
This change focuses on making this single method a little bit less cryptic, and a little bit more performant.

In summary, the changed version:

- Adds documentation about what the method does together with a mutation warning. (Neither the old nor the new version is thread safe.)
- Adds an alias (`#to_h`) for the method that doesn't contain the redundant fragment.
- Flattens the conditional branches to make them easier to follow.
- Avoids allocating an array of strings for each params object.
- Improves performance by 12% for an average query string, and by 40% for an empty one. (Average of 5 runs with `benchmark-ips` using MRI 2.6.0.)

Feel free to pick and choose from the proposed changes.

### DIFF
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -55,7 +55,7 @@ module Rack
         end
       end
 
-      return params.to_params_hash
+      return params.to_h
     end
 
     # parse_nested_query expands a query string into structural types. Supported
@@ -73,7 +73,7 @@ module Rack
         normalize_params(params, k, v, param_depth_limit)
       end
 
-      return params.to_params_hash
+      return params.to_h
     rescue ArgumentError => e
       raise InvalidParameterError, e.message
     end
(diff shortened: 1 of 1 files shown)


lib/rack/query_parser.rb
```diff
@@ -177,22 +177,42 @@ def key?(key)
         @params.key?(key)
       end
 
-      def to_params_hash
-        hash = @params
```
Drenmi: I found this assignment mostly misleading. It seems to suggest that `hash` is a copy of `@params`, when in reality it is a reference.

lib/rack/query_parser.rb
```diff
@@ -177,22 +177,42 @@ def key?(key)
...
-        hash.keys.each do |key|
```
Drenmi: I saved allocating this array of strings for each `Params` object. Since we're not adding keys in the iteration, it's okay to enumerate the hash directly.

lib/rack/query_parser.rb
```diff
@@ -177,22 +177,42 @@ def key?(key)
...
-          value = hash[key]
```
Drenmi: The performance improvement is likely due to not having to allocate an array, together with the replacement of this hash index access.

Drenmi:
The CI failure for Ruby 2.2 seems unrelated. Or at least it's not failing with 2.2.2 locally.

lib/rack/query_parser.rb
```diff
@@ -177,22 +177,42 @@ def key?(key)
...
+      #
+      #   Mutation warning!
+      #
+      #   1. This method mutates the internal representation of the `Params`
+      #      objects in order to save object allocations.
+      #
+      #   2. The value you get back is a reference to the internal hash
+      #      representation, not a copy.
+      #
+      #   3. Because the `Params` object's internal representatoin is mutable
```
mikegee:
```suggestion
      #   3. Because the `Params` object's internal representation is mutable
```

Drenmi: Good catch, Mike! This also gave me a way to re-trigger the CI. 😄

Issue #1243, opened 2018-03-20
Frozen String LIteral changes


oniofchaos:
Looks like there have been multiple PRs opened relating to migrating the codebase to using `#frozen_string_literal: true` in order to prevent unnecessary string allocations. Unfortunately, only 1 that I saw has been merged and it only updated 1 file. Can any of the maintainers speak to why the others haven't gotten merged? I'm willing to help out with any extra changes desired to make this happen.

https://github.com/rack/rack/pull/1182
https://github.com/rack/rack/pull/1023

cc: @pat @twalpole (thanks for your PRs!)

oniofchaos:
I see https://github.com/rack/rack/pull/1203 got merged but it doesn't look like it's in any of the tagged versions?

twalpole:
@oniofchaos Maintainers just didn't seem interested

oniofchaos:
@tenderlove thanks for merging #1182, would there be interest in a PR that adds the pragma to more files? If so I am happy to submit.

tenderlove:
@oniofchaos yes, I'm interested in adding the pragma directive to all of Rack.

tenderlove:
I think everything has the frozen string directive.  Can this be closed?

PR #1385, opened 2019-08-20
Remove redundant freeze method

Sean0628:
Frozen string literal pragma has been introduced to all of Rack. cf. https://github.com/rack/rack/issues/1243
Therefore, string literals do not need to be frozen with `Object#freeze` method anymore.
This PR tends to remove redundant `"str".freeze`.
Thanks.

lib/rack/utils.rb:
https://github.com/rack/rack/blob/29e9aeb63cdebe49682ea0cf9ef423007ac1d738/lib/rack/utils.rb#L2
test/spec_deflater.rb:
https://github.com/rack/rack/blob/29e9aeb63cdebe49682ea0cf9ef423007ac1d738/test/spec_deflater.rb#L1

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -447,7 +447,7 @@ module Rack
       end
 
       def []=(k, v)
-        canonical = k.downcase.freeze
+        canonical = k.downcase
         delete k if @names[canonical] && @names[canonical] != k # .delete is expensive, don't invoke it unless necessary
         @names[canonical] = k
         super k, v
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -447,7 +447,7 @@ module Rack
       end
 
       def []=(k, v)
-        canonical = k.downcase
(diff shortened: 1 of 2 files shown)


lib/rack/utils.rb
```diff
@@ -447,7 +447,7 @@ def [](k)
       end
 
       def []=(k, v)
-        canonical = k.downcase.freeze
```
ahorek: the intention here is to avoid mutations, frozen string literal pragma doesn't really do the same thing

```
# frozen_string_literal: true

canonical = 'test'.downcase
puts canonical.frozen?
=> false
```

Sean0628: Thank you for the comment.
My mistake. I was misunderstanding the behaviour of frozen string literal pragma.
As you mentioned, frozen string literal pragma does not freeze the var `canonical` in this case.
Reverted the modification referred by your comment.

Thanks.

PR #1388, opened 2019-09-09
Fix deflater zlib buffer errors on empty body part

### LINKED ISSUE #959: Rack::Deflater error when writing empty string
Hi.  I think I found a problem with Rack::Deflater.

This rackup config will throw a Zlib::DataError exception here (and I was getting Zlib::BufError in a more complex application before):

```
require 'zlib'
require 'rack'
use Rack::Deflater
run lambda { |env| [200, { }, ['test', '']] }
```

It seems to be a Zlib problem:

```
require 'zlib'
z = Zlib::GzipWriter.new open('/dev/null', 'w')
z.write 'test'
z.flush
z.write ''
z.flush
z.close
```

So flushing an empty string seems to be a no-no.  (But an empty flush doesn't error out if it's the first one for the stream somehow.)

Maybe there should be a guard in Rack::Deflater::GzipStream#each like this?

```
    ...
    @body.each { |part|
      unless part.empty?
        gzip.write(part)
        gzip.flush
      end
    }
    ...
```

Even if this is a Zlib bug, it could be nice to protect older ruby versions from it.

felixbuenemann:
This fixes a `Zlib::BufError` exception in `Rack::Deflater` that happens with enumerable response bodies if a body part is an empty string and sync mode is enabled (default).

Fixes #959

### DIFF
--- a/lib/rack/deflater.rb
+++ b/lib/rack/deflater.rb
@@ -88,8 +88,9 @@ module Rack
         gzip = ::Zlib::GzipWriter.new(self)
         gzip.mtime = @mtime if @mtime
         @body.each { |part|
-          gzip.write(part)
-          gzip.flush if @sync
+          len = gzip.write(part)
+          # Flushing empty parts would raise Zlib::BufError.
+          gzip.flush if @sync && len > 0
         }
       ensure
         gzip.close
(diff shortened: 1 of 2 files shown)


felixbuenemann:
Not sure why the specs are failing, but it seems unrelated, `bundle exec rake ci` works fine on my machine (macOS 10.14.6, ruby 2.6.1).

felixbuenemann:
Fixed unrelated spec failures in #1389.

felixbuenemann:
Rebased on master to make tests pass.

PR #1395, opened 2019-10-06
Allow direct access to response body through SimpleBodyProxy.

NikolayRys:
A faster body proxy was introduced in Rack: https://github.com/rack/rack/pull/1327 in order to achieve performance benefit and get rid of a circular dependency between response and body wrapper.

However, it breaks some clients that rely on the old behavior, because the Response method `#body` (https://github.com/rack/rack/blob/master/lib/rack/response.rb#L25) always has been accessible before through the body wrapper.

I personally have discovered it with the popular API Ruby library: https://github.com/ruby-grape/grape. If uses rack inside https://github.com/ruby-grape/grape/blob/master/lib/grape/middleware/formatter.rb#L47 but expect the `#body` to be accessible https://github.com/ruby-grape/grape/blob/master/spec/grape/middleware/formatter_spec.rb#L197 
As I see, Rack joins the body content through the `each` method(https://github.com/rack/rack/blob/master/test/spec_response.rb#L269) in order to access it, but I believe there are many other cases in the wild besides Grape that would expect the presence of "#body" method.

Thus, I propose to add a getter to the raw body object on `SimpleBodyProxy`. Bringing this method back won't jeopardize any goals why this new wrapper has been introduced.  Namely, the performance and avoiding a circular dependency between response and the wrapper.

### DIFF
--- a/lib/rack/simple_body_proxy.rb
+++ b/lib/rack/simple_body_proxy.rb
@@ -2,6 +2,8 @@
 
 module Rack
   class SimpleBodyProxy
+    attr_reader :body
+
     def initialize(body)
       @body = body
     end
--- a/test/spec_response.rb
+++ b/test/spec_response.rb
@@ -287,6 +287,12 @@ describe Rack::Response do
     r.status.must_equal 200
   end
 
+  it "allows direct access to body" do
+    r = Rack::Response.new(["foo", "bar"])
+    _status, _header, body_proxy = r.finish
+    body_proxy.body.must_equal ["foo", "bar"]
+  end
+
   it "has a constructor that can take a block" do
     r = Rack::Response.new { |res|
       res.status = 404


NikolayRys:
Disregard failed tests, please, it'll become green after 
https://github.com/rack/rack/pull/1394 is merged.

NikolayRys:
@jodosha and @rafaelfranca you probably would like to have a look

rafaelfranca:
Was that change released? I don't think it was. I only see it in the master branch.

NikolayRys:
Hi, @jodosha,

In Grape it hasn't been discovered before because build with a recent rack is actually excluded from CI: https://github.com/ruby-grape/grape/blob/master/.travis.yml#L50

Though this issue seems to be limited to the Grape's test suite, so it won't be too hard to update it to the new interface of Rack. Namely, the `#finish` method is explicitly called there. I will change this behavior if we decide that there is no problem with the Rack.

However, I would argue that the chosen approach does not make much sense, and here's why:

1. The #body method has been available there for 9 years, as long as BodyProxy existed.
It is still not properly released. The newest available version is 2.0.7: https://rubygems.org/gems/rack/versions, where it is still as before: https://github.com/rack/rack/blob/2.0.7/lib/rack/response.rb#L69 For sure, I expect people to become vocal when this change goes live. I believe that this alone should be comfortably disqualifying on its own, but there is more to it.

2. The `#body` method is actually still available there, just only sometimes. It depends on the block for `#finish` to be present, in which case the different type of wrapper is returned. Generally, it's a controversial idea to return objects of different data types from the same method, but here it happens also for an unclear and surprising reason, at least from the client perspective.

3. It's also pretty convenient to have the easily accessible body from the wrapper, instead of -re-assembling it using "#each": https://github.com/rack/rack/blob/master/test/spec_response.rb#L269

Please let me know what you think.

rafaelfranca:
> 2\. It's also pretty convenient to have the easily accessible body from the wrapper, instead of re-assembling it using "#each": https://github.com/rack/rack/blob/master/test/spec_response.rb#L269

I think this an incorrect assumption. The rack spec only requires the body to respond to `#each` and you are assuming it also needs to respond to `#body` as well.

Yes, the previous `BodyProxy` was exposing all the methods of the response, but I don't think we should continue to do it.

NikolayRys:
Okay, @rafaelfranca, you are right about it, let's disregard the last bullet point.

Though, it's still true that it's going to break stuff when released. In this case, I suggest to delay the removal of `#body` until next major release instead of doing it on a minor one between 2.0.7 and 2.1.0.

rafaelfranca:
But why only `body`? That was not the only method that stoped to work. All the methods in Response used to work.

@matthewd @tenderlove what do you think?

NikolayRys:
Because this wrapper doesn't contain anything else, so it cannot mimic much more than that. 
But indeed this logic may also apply to a couple of smaller helpers like `#empty?`, I guess. 

If we decide to keep it, I'll look for other relevant methods.

jodosha:
@NikolayRys My apologize, I thought that https://github.com/rack/rack/pull/1327 was already released.

---

Given `BodyProxy` relies on `#method_missing` to forward unknown method invocations to the wrapped object, it didn't use to respond to `#body` per se, but because the inner object did.

For instance, if you pick latest master commit (5f995a5), and apply this diff, you'll see a bunch of errors:

```diff
diff --git a/test/spec_body_proxy.rb b/test/spec_body_proxy.rb
index 6be79f8..7afe2f9 100644
--- a/test/spec_body_proxy.rb
+++ b/test/spec_body_proxy.rb
@@ -8,6 +8,7 @@ describe Rack::BodyProxy do
   it 'call each on the wrapped body' do
     called = false
     proxy  = Rack::BodyProxy.new(['foo']) { }
+    proxy.body
     proxy.each do |str|
       called = true
       str.must_equal 'foo'
@@ -18,6 +19,7 @@ describe Rack::BodyProxy do
   it 'call close on the wrapped body' do
     body  = StringIO.new
     proxy = Rack::BodyProxy.new(body) { }
+    proxy.body
     proxy.close
     body.must_be :closed?
   end
@@ -25,12 +27,14 @@ describe Rack::BodyProxy do
   it 'only call close on the wrapped body if it responds to close' do
     body  = []
     proxy = Rack::BodyProxy.new(body) { }
+    proxy.body
     proxy.close.must_be_nil
   end
 
   it 'call the passed block on close' do
     called = false
     proxy  = Rack::BodyProxy.new([]) { called = true }
+    proxy.body
     called.must_equal false
     proxy.close
     called.must_equal true
@@ -43,6 +47,7 @@ describe Rack::BodyProxy do
 
     begin
       proxy = Rack::BodyProxy.new(object) { called = true }
+      proxy.body
       called.must_equal false
       proxy.close
     rescue RuntimeError => e
@@ -55,6 +60,7 @@ describe Rack::BodyProxy do
   it 'allow multiple arguments in respond_to?' do
     body  = []
     proxy = Rack::BodyProxy.new(body) { }
+    proxy.body
     proxy.respond_to?(:foo, false).must_equal false
   end
 
@@ -63,6 +69,7 @@ describe Rack::BodyProxy do
     body.respond_to?(:to_ary).must_equal true
 
     proxy = Rack::BodyProxy.new(body) { }
+    proxy.body
     proxy.respond_to?(:to_ary).must_equal false
     proxy.respond_to?("to_ary").must_equal false
   end
@@ -70,6 +77,7 @@ describe Rack::BodyProxy do
   it 'not close more than one time' do
     count = 0
     proxy = Rack::BodyProxy.new([]) { count += 1; raise "Block invoked more than 1 time!" if count > 1 }
+    proxy.body
     2.times { proxy.close }
     count.must_equal 1
   end
@@ -77,6 +85,7 @@ describe Rack::BodyProxy do
   it 'be closed when the callback is triggered' do
     closed = false
     proxy = Rack::BodyProxy.new([]) { closed = proxy.closed? }
+    proxy.body
     proxy.close
     closed.must_equal true
   end
```

This because the wrapped objects don't respond to `#body`.

---

Before https://github.com/rack/rack/pull/1327, `Rack::Response` always used to pass `self` to `Rack::BodyProxy`, so this last object was able respond to `#body` because the response itself was able too.

The real problem isn't the lack of `#body` for `Rack::SimpleBodyProxy`, but the fact that we're now just passing a `@body` string as third element of the serialization (`[code, header, body]`), instead of the full blown `Rack::Response`.

As @rafaelfranca said, it isn't just a problem of method `#body`, but now all the methods of `Rack::Response` are gone, because we don't pass it anymore to a proxy object.

---

The problem that I see with that trick (passing `self` to `Rack::BodyProxy`) is that it led to odd syntaxes like the one you [linked](https://github.com/ruby-grape/grape/blob/master/spec/grape/middleware/formatter_spec.rb#L196-L197):

```ruby
_, _, body = subject.call('PATH_INFO' => '/info.custom')
expect(body.body).to eq(['CUSTOM FORMAT'])
```

I'm pointing to the `body.body` bit. The `body` object is a `Rack::Response` instance (POLS violation?).

The response returns itself as third element of the array, instead of returning the body. 😱

I don't know if this is contemplated or not by the [Rack SPEC](https://www.rubydoc.info/github/rack/rack/file/SPEC#label-The+Response), which is a bit cryptic about this point:

> The Body commonly is an Array of Strings, the application instance itself, or a File-like object.

I don't understand what "the application instance itself" means. Is the response itself allowed? If **yes** then we should rollback https://github.com/rack/rack/pull/1327, if **no** we had `rack` (the gem) don't following the Rack SPEC.

---

From a community perspective, we could opt to revert https://github.com/rack/rack/pull/1327. Because that PR was merged to bring a perf gain, but if it has to break an entire ecosystem, the perf gain isn't worth.

NikolayRys:
Hi, thank for the comprehensive response, but I have been aware that `#body` worked through `#missing_method` :)

I would support the revert option too instead of this PR, but I would ask you to allow me to do it.

NikolayRys:
Created an alternative PR: https://github.com/rack/rack/pull/1407

PR #1396, opened 2019-10-08
Remove unnecessary super on Rack::Session::Abstract::ID.inherited

styd:
Currently there is no ancestors of Rack::Session::Abstract::ID that implement `.inherited`.
So, I guess this `super` is unnecessary.


rafaelfranca:
Class does implement it, it is just a best practice to call super if you are overriding a method defined in the superclass.

Thank you for the pull request

PR #1391, opened 2019-09-29
Make stringify_keys use transform_keys

styd:
Since Ruby 2.5, there is `Hash#transform_keys` that runs faster than the traditional way of transforming keys. From my benchmark, it's around 11-13% faster on 2.5 and above.
So, I'm thinking of turning `stringify_keys` to use `transform_keys` to take the performance benefit and I also make the polyfill for Ruby less than 2.5.

Here is my benchmark:
```ruby
require 'benchmark/ips'

def stringify_keys(other)
  hash = {}
  other.each do |key, value|
    hash[key.to_s] = value
  end
  hash
end

def stringify_keys2(other)
  other.transform_keys(&:to_s)
end

hash = {a: 'an a', b: 'a b', az: 'a and z', zzz: 'sleeping..'}

Benchmark.ips do |x|
  x.report('traditional') do
    stringify_keys(hash)
  end

  x.report('transform') do
    stringify_keys2(hash)
  end

  x.compare!
end
```
```
Warming up --------------------------------------
         traditional    85.231k i/100ms
           transform    93.393k i/100ms
Calculating -------------------------------------
         traditional      1.005M (± 2.3%) i/s -      5.029M in   5.004178s
           transform      1.132M (± 0.3%) i/s -      5.697M in   5.032428s

Comparison:
           transform:  1132060.4 i/s
         traditional:  1005440.7 i/s - 1.13x  slower
```

PR #1401, opened 2019-10-17
Fix stringify_keys error when other is SessionHash

styd:
Handle case when other is also SessionHash.
Should fix errors in Sinatra as reported in [here](https://github.com/rack/rack/pull/1391#issuecomment-542980651).
I'm not sure if the test I add is the correct way to test `stringify_keys`.

### DIFF
--- a/lib/rack/session/abstract/id.rb
+++ b/lib/rack/session/abstract/id.rb
@@ -29,6 +29,14 @@ module Rack
           end
         } unless {}.respond_to?(:transform_keys)
 
+        def transform_keys(&block)
+          hash = dup
+          each do |key, value|
+            hash[block.call(key)] = value
+          end
+          hash
+        end
+
         include Enumerable
         attr_writer :id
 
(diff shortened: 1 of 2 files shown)


styd:
I modified `stringify_keys` into this in my branch:
```ruby
def stringify_keys(other)
  if other.is_a? Hash
    puts "stringifying hash keys"
  else
    puts "stringifying not hash keys"
  end
  other.transform_keys(&:to_s)
end
```
Then I run the test:
```sh
$ bundle exec rake 2>/dev/null | grep -c "stringifying hash keys"
76
$ bundle exec rake 2>/dev/null | grep -c "stringifying not hash keys"
1
```
I think only the test I added that doesn't use hash for `other`.

Then, I did the same on sinatra master with rack pointing to this rack branch.
```sh
$ bundle exec rake 2>/dev/null | grep -c "stringifying hash keys"
6
$ bundle exec rake 2>/dev/null | grep -c "stringifying not hash keys"
1
```
Apparently hash is stringified far more often than session hash.

rafaelfranca:
👍

PR #1407, opened 2019-10-28
Revert SimpleBodyProxy

NikolayRys:
This is an alternative solution to the issue is outlined in https://github.com/rack/rack/pull/1395

As suggested by @jodosha, this commit reverts SimpleBodyProxy, introduced in https://github.com/rack/rack/pull/1327/
It's not a regular git-revert because the PR also contained a fix for the tests, which is an independent change. and should be preserved. 

If this gets merged, then https://github.com/rack/rack/pull/1395 can be closed and vice versa.


jodosha:
@NikolayRys Thanks for taking care of this. 👍 
@rafaelfranca @tenderlove https://github.com/rack/rack/pull/1327 can be considered for Rack 3. Thanks. 😸

PR #1409, opened 2019-11-08
Add NGINX Unit web server to list on README.md

VBart:
https://unit.nginx.org/configuration/#ruby


ioquatix:
Wondering if we should add the link too?

VBart:
@ioquatix Thanks for merging. I think it will be convenient for a user to be able to click on, instead of googling some technology. Should I create a pull request with a link to "https://unit.nginx.org/"?

ioquatix:
Yes please submit a PR, we should add links for all of them.

PR #1413, opened 2019-11-16
WIP: fix tests

ahorek:


ioquatix:
Did any of this actually work? I couldn't get lighttpd to work locally.

ahorek:
yeah, it works locally, but it still fails on CI, even if all necessary libraries should be there.

Errno::ECONNREFUSED: Failed to open TCP connection to 127.0.0.1:9203 (Connection refused - connect(2) for "127.0.0.1" port 9203)

the main problem is `#!/usr/bin/env ruby` doesn't pick the right ruby version.

I would rather support to remove those spec, as you did in https://github.com/rack/rack/pull/1414

Issue #1415, opened 2019-11-20
etag not generated when Cache-Control: no-cache is present


henm:
In `Rack::ETag`, the etag is not generated when `Cache-Control: no-cache` is present (https://github.com/rack/rack/blob/master/lib/rack/etag.rb#L60).

According to the spec, the etag can be used for validation when `no-cache` is present, so does it make sense to not generate the etag in this case?

Documents and specifications:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Directives
https://tools.ietf.org/html/rfc7234#section-5.2.1.4
https://tools.ietf.org/html/rfc7234#section-4.3

Thanks!

ioquatix:
I think it's up to you to make the case that it does or does not. Then, make a PR to add this functionality, maybe behind an option, so that existing code using `Rack::ETag` doesn't suddenly change behaviour unexpectedly (or make the case that it should be the default).

Thanks in advance for your effort.

jeremyevans:
As this isn't a bug (even if changing the behavior could be useful), I'm going to close this. Let's continue the discussion in #1416.

PR #1416, opened 2019-11-20
Generate etag even if Cache-Control: no-cache is set

henm:
Currently an etag is not generated if `Cache-Control: no-cache` is set. Imo this is not ideal, as `no-cache` explicitly requires a validation of the content, in which case having an etag is very helpful.

Therefore this PR proposes to generate the etag whether `no-cache` is present or not.

Some background:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control#Directives
https://tools.ietf.org/html/rfc7234#section-5.2.1.4
https://tools.ietf.org/html/rfc7234#section-4.3

See also https://github.com/rack/rack/issues/1415

Thanks!

### DIFF
--- a/lib/rack/etag.rb
+++ b/lib/rack/etag.rb
@@ -57,8 +57,7 @@ module Rack
       end
 
       def skip_caching?(headers)
-        (headers[CACHE_CONTROL] && headers[CACHE_CONTROL].include?('no-cache')) ||
-          headers.key?(ETAG_STRING) || headers.key?('Last-Modified')
+        headers.key?(ETAG_STRING) || headers.key?('Last-Modified')
       end
 
       def digest_body(body)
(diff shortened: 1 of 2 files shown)


ioquatix:
Can you please rebase on master.

jeremyevans:
I think it we are going to make this change, we should keep the specs for it and update them for the new behavior (i.e. `"not set ETag if no-cache is given"` -> `"sets ETag even if no-cache is given"`).  Other than that, the change looks fine to me.

henm:
I've been traveling and was away from my computer, thanks for taking care of adding the test in https://github.com/rack/rack/pull/1513!

rmoehn:
This broke our production code, as others detailed in #1619. I wish you would be more circumspect about changes like this, especially when:

- The justification is that the previous behaviour was ‘not ideal’.
- Nobody has complained for nine years (https://github.com/rack/rack/commit/cd73f8f0db954659286abe006ad803603d261b55).
- The resulting behaviour change is hard to catch with automated tests. – Or do you have a good approach for testing whether or not streaming is actually happening in a Rails controller or integration test?

PR #1417, opened 2019-11-25
Rack::Directory : allow directory trasversal inside root directory

ThomasSevestre:
This PR allows use of `..` in HTTP request to retreive files with Rack::Directory middleware. This is allowed only if the file is accessible with a direct path.

For instance with this directory structure :

    /root/a/a.txt
    /root/b/b.txt
    /c/c.txt

And the following Rack application

    Rack::Directory.new("/root")

Responses to the following requests are:

    GET a/a.txt
    Before PR : 200, After PR : 200
    
    GET a/../b/b.txt
    Before PR : 403, After PR : 200
    
    GET a/../../c/c.txt
    Before PR : 403, After PR : 403

### DIFF
--- a/lib/rack/directory.rb
+++ b/lib/rack/directory.rb
@@ -96,6 +96,7 @@ table { width:100%%; }
 
     def check_forbidden(path_info)
       return unless path_info.include? ".."
+      return if ::File.expand_path(::File.join(@root, path_info)).start_with?(@root)
 
       body = "Forbidden\n"
       size = body.bytesize
--- a/test/spec_directory.rb
+++ b/test/spec_directory.rb
@@ -72,14 +72,26 @@ describe Rack::Directory do
     res.must_be :bad_request?
   end
 
+  it "allow directory traversal inside root directory" do
+    res = Rack::MockRequest.new(Rack::Lint.new(app)).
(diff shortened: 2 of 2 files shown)


lib/rack/directory.rb
```diff
@@ -96,6 +96,7 @@ def check_bad_request(path_info)
 
     def check_forbidden(path_info)
       return unless path_info.include? ".."
+      return if ::File.expand_path(::File.join(@root, path_info)).start_with?(@root)
```
ioquatix: On windows, this will assume "\" as a path separator. Is that what you are expecting?

ThomasSevestre: It looks like it is safe to use File.join on Windows but I have no way to test it...

https://stackoverflow.com/questions/597488/how-to-do-a-safe-join-pathname-in-ruby

jeremyevans: This should be safe on Windows as `@root` is set to the result of `::File.expand_path`.

jeremyevans:
@ioquatix I don't think there is a security issue here, and this looks useful and I would like to merge it.  Are you OK with that?

ioquatix:
LGTM... Enforcing the root directory is sufficient.

PR #1420, opened 2019-11-28
Support multipart range requests

fatkodima:
Closes #559

### DIFF
--- a/lib/rack/files.rb
+++ b/lib/rack/files.rb
@@ -18,6 +18,7 @@ module Rack
   class Files
     ALLOWED_VERBS = %w[GET HEAD OPTIONS]
     ALLOW_HEADER = ALLOWED_VERBS.join(', ')
+    MULTIPART_BOUNDARY = 'AaB03x'
 
     attr_reader :root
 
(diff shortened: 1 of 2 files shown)


lib/rack/files.rb
```diff
@@ -70,75 +71,111 @@ def serving(request, path)
...
+        status = 206
         range = ranges[0]
-        response[0] = 206
-        response[1]["Content-Range"] = "bytes #{range.begin}-#{range.end}/#{size}"
+        headers["Content-Range"] = "bytes #{range.begin}-#{range.end}/#{size}"
         size = range.end - range.begin + 1
+      else
+        # Partial content, multiple ranges:
+        status = 206
+        headers[CONTENT_TYPE] = "multipart/byteranges; boundary=#{MULTIPART_BOUNDARY}"
```
ioquatix: Sorry, I only have a short time to review this, but I'm excited to see this PR.

Is `AaB03x` sufficient as a boundary?

lib/rack/files.rb
```diff
@@ -70,75 +71,111 @@ def serving(request, path)
...
-      if request.head?
-        []
-      else
-        Iterator.new path, range
+      def multipart?
+        ranges.size > 1
+      end
+
+      def multipart_heading(range)
+<<-EOF
```
ioquatix: Can we encode this as a string rather than a heredoc.

fatkodima: I tried to stick here with existing code styles, like
https://github.com/rack/rack/blob/93dfcdf46084760079514c858ae7391e90a7821e/lib/rack/multipart/generator.rb#L71-L89

and like examples in many test cases. 
So should I change this to string?

lib/rack/files.rb
fatkodima: Yes, why not? The same boundary is used for multipart forms https://github.com/rack/rack/blob/93dfcdf46084760079514c858ae7391e90a7821e/lib/rack/multipart.rb#L14

jeremyevans:
@fatkodima I think the implementation looks pretty good.  I think you can keep using the existing static boundary. I also think you can keep using the heredoc, you don't need to change it to a string.  Can you please rebase this against master?  Please make sure that it works for multiple ranges the same as for single ranges (which now use `BaseIterator`, see 3936b90697fcaf02751e596bd9dfd0e24add9b29).

fatkodima:
@jeremyevans Updated.

jeremyevans:
Looks good.  I tested locally with `curl -r` and got the expected results.  I plan on merging in a few days unless other committer objects.

PR #1422, opened 2019-11-29
Robust separation of Content-Disposition fields

### LINKED ISSUE #1386: Robust multipart field separation
As a maintainer of the [Pharo](https://pharo.org) track of the [Exercism](https://exercism.io) project, we wrote our own IDE-integrated-client to interface the Exercism REST API.  IIUC, Exercism uses Sinatra which uses Rack which has led me here.  I'm not a Ruby programmer and never used Sinatra or Rack before, so please bear with me as I try to provide anything you need from me.

One of the API features is multipart uploading of files.  Out of the box this was broken with the standard Pharo library due to lack of a space between multipart fields.  Specifically I discovered that the Pharo library sent... 
```
Content-Disposition: form-data;name="files[]";filename="myfile.txt"
```
which failed with **"500 INTERNAL SERVER ERROR"**, 
while Exercism's default client (i.e. library of another language) worked sending...
```
Content-Disposition: form-data; name="files[]"; filename="myfile.txt"
```
with the difference being the space after the semi-colon between parameters.

For our own application we have already solved this by monkey-patching the Pharo library, and am following up other channels to change the Pharo library, but its a subtle error (especially wrapped by SSL) and discovery took about 30 hours focused investigation.  So I'm seeking to spare future-others that trouble. 

First, a side-query... *Is it a "requirement" of an RFC somewhere the there "MUST" be a space between multipart fields?*  But in any case, even if it "must" have a space, the [Robustness Principal](https://en.wikipedia.org/wiki/Robustness_principle) suggests parsing input sans-space would be useful.

fatkodima:
Closes #1386

### DIFF
--- a/lib/rack/multipart.rb
+++ b/lib/rack/multipart.rb
@@ -16,10 +16,10 @@ module Rack
     TOKEN = /[^\s()<>,;:\\"\/\[\]?=]+/
     CONDISP = /Content-Disposition:\s*#{TOKEN}\s*/i
     VALUE = /"(?:\\"|[^"])*"|#{TOKEN}/
-    BROKEN_QUOTED = /^#{CONDISP}.*;\sfilename="(.*?)"(?:\s*$|\s*;\s*#{TOKEN}=)/i
-    BROKEN_UNQUOTED = /^#{CONDISP}.*;\sfilename=(#{TOKEN})/i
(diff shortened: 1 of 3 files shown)


ioquatix:
Thanks for your effort. This looks great.

PR #1434, opened 2019-12-24
Lazily initialize the response body and only buffer it if required.

ioquatix:
The implementation of `Rack::Response` will double buffer every body even
if the user does not call `#write`. This is inefficient and additionally
imposes some type checking overhead even if none is wanted.

This should reduce allocations and improve performance for situations where `Rack::Response` is used without manipulating the response body.

Fixes <https://github.com/socketry/falcon/issues/93>.

### DIFF
--- a/test/spec_response.rb
+++ b/test/spec_response.rb
@@ -480,6 +480,15 @@ describe Rack::Response do
     b.wont_equal res.body
   end
 
+  it "doesn't call close on #body when 205" do
+    res = Rack::Response.new
+
+    res.body = StringIO.new
+    res.status = 205
+    _, _, b = res.finish
+    res.body.wont_be :closed?
+  end
+
   it "flatten doesn't cause infinite loop" do
     # https://github.com/rack/rack/issues/419
     res = Rack::Response.new("Hello World")
--- a/test/spec_response.rb
+++ b/test/spec_response.rb
@@ -307,9 +307,9 @@ describe Rack::Response do
     header['Content-Length'].must_be_nil
 
     lambda {
-      Rack::Response.new(Object.new)
-    }.must_raise(TypeError).
-      message.must_match(/stringable or iterable required/)
+      Rack::Response.new(Object.new).each{}
+    }.must_raise(NoMethodError).
+      message.must_match(/undefined method .each. for/)
   end
 
   it "knows if it's empty" do
@@ -433,6 +433,28 @@ describe Rack::Response do
     res.headers["Content-Length"].must_equal "8"
   end
 
+  it "does not wrap body" do
+    body = Object.new
+    res = Rack::Response.new(body)
+
+    # It was passed through unchanged:
+    res.finish.last.must_equal body
+  end
+
+  it "does wraps body when using #write" do
+    body = ["Foo"]
+    res = Rack::Response.new(body)
+
+    # Write something using the response object:
+    res.write("Bar")
+
+    # The original body was not modified:
+    body.must_equal ["Foo"]
+
+    # But a new buffered body was created:
+    res.finish.last.must_equal ["Foo", "Bar"]
+  end
+
   it "calls close on #body" do
     res = Rack::Response.new
     res.body = StringIO.new
@@ -449,12 +471,6 @@ describe Rack::Response do
     res.status = 204
     _, _, b = res.finish
     res.body.must_be :closed?
-    b.wont_equal res.body
-
-    res.body = StringIO.new
-    res.status = 205
-    _, _, b = res.finish
-    res.body.wont_be :closed?
     b.wont_equal res.body
 
     res.body = StringIO.new
--- a/lib/rack/mock.rb
+++ b/lib/rack/mock.rb
@@ -169,6 +169,8 @@ module Rack
       @cookies = parse_cookies_from_header
 
       super(body, status, headers)
+
+      buffered_body!
     end
 
     def =~(other)
@@ -190,7 +192,13 @@ module Rack
       #     ...
       #     res.body.should == "foo!"
       #   end
-      super.join
+      buffer = String.new
+
+      super.each do |chunk|
+        buffer << chunk
+      end
+
+      return buffer
     end
 
     def empty?
--- a/lib/rack/response.rb
+++ b/lib/rack/response.rb
@@ -62,7 +62,7 @@ module Rack
     end
 
     def finish(&block)
-      if STATUS_WITH_NO_ENTITY_BODY.key?(status.to_i)
+      if STATUS_WITH_NO_ENTITY_BODY[status.to_i]
         delete_header CONTENT_TYPE
         delete_header CONTENT_LENGTH
         close
--- a/lib/rack/response.rb
+++ b/lib/rack/response.rb
@@ -26,11 +26,7 @@ module Rack
     alias headers header
 
     CHUNKED = 'chunked'
-    STATUS_WITH_NO_ENTITY_BODY = {
-      204 => true,
-      205 => true,
-      304 => true,
-    }.freeze
+    STATUS_WITH_NO_ENTITY_BODY = Utils::STATUS_WITH_NO_ENTITY_BODY
 
     def initialize(body = nil, status = 200, header = {})
       @status = status.to_i
--- a/lib/rack/response.rb
+++ b/lib/rack/response.rb
@@ -28,30 +28,32 @@ module Rack
     CHUNKED = 'chunked'
     STATUS_WITH_NO_ENTITY_BODY = {
       204 => true,
-      304 => true
+      205 => true,
+      304 => true,
     }.freeze
 
-    def initialize(body = [], status = 200, header = {})
+    def initialize(body = nil, status = 200, header = {})
       @status = status.to_i
       @header = Utils::HeaderHash.new(header)
 
-      @writer  = lambda { |x| @body << x }
-      @block   = nil
-      @length  = 0
+      @writer = self.method(:append)
 
-      @body = []
+      @block = nil
+      @length = 0
 
-      if body.respond_to? :to_str
-        write body.to_str
-      elsif body.respond_to?(:each)
-        body.each { |part|
-          write part.to_s
-        }
+      # Keep track of whether we have expanded the user supplied body.
+      if body.nil?
+        @body = []
+        @buffered = true
+      elsif body.respond_to? :to_str
+        @body = [body]
+        @buffered = true
       else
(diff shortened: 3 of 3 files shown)


ioquatix:
@tenderlove what do you think?

@senid231 maybe this can avoid the issues you mentioned, because we don't try to wrap the body unless it's explicitly required, which it isn't in most cases, especially in the case of rack-protection, which is only manipulating headers.

ioquatix:
It would be great if we could release this as part of 2.1.0

senid231:
@ioquatix @tenderlove I've tested this branch and it works correctly when using websocket on `falcon` and `Rack::Protection::CookieTossing` which wraps response into `Rack::Response`.

ioquatix:
@senid231 thanks for testing!!! That’s great news.

test/spec_response.rb
```diff
@@ -451,24 +473,12 @@ def object_with_each.each
     res.body.must_be :closed?
     b.wont_equal res.body
 
-    res.body = StringIO.new
-    res.status = 205
-    _, _, b = res.finish
-    res.body.wont_be :closed?
-    b.wont_equal res.body
-
```
tenderlove: Is this because of changes to `mock.rb`?

ioquatix: It's testing status 205 but that's not part of the spec "calls close on #body when 204 or 304". So, what's wrong, the spec, or the name of the spec?

test/spec_response.rb
```diff
@@ -451,24 +473,12 @@ def object_with_each.each
...
     _, _, b = res.finish
     res.body.must_be :closed?
     b.wont_equal res.body
   end
-
-  it "wraps the body from #to_ary to prevent infinite loops" do
-    res = Rack::Response.new
-    res.finish.last.wont_respond_to(:to_ary)
-    lambda { res.finish.last.to_ary }.must_raise NoMethodError
-  end
```
ioquatix: @tenderlove does this concern you too? Can you explain what kind of infinite loops it's trying to prevent?

tenderlove: No, this test makes no sense to me.  Let's ✂️

tenderlove: It looks like this test was introduced in 5b251a986962f3577e14c01a698fab9dd21383ba to fix #419. It seems like we shouldn't delete it, but maybe we should call `flatten` rather than checking for `to_ary`?

It seems OK for something to respond to `to_ary`, but it doesn't seem OK to produce a circular data structure.

ioquatix: Fixed in https://github.com/rack/rack/pull/1453

ioquatix:
```
#!/usr/bin/env ruby

require 'benchmark/ips'

def no_entity_body?(status)
	(100..199).include?(status) || status == 204 || status == 304
end

# Responses with HTTP status codes that should not have an entity body
STATUS_WITH_NO_ENTITY_BODY = Hash[((100..199).to_a << 204 << 304).product([true])]


Benchmark.ips do |x|
  x.report("no_entity_body?(100)", "no_entity_body?(100)")
  x.report("no_entity_body?(200)", "no_entity_body?(200)")
  x.report("no_entity_body?(300)", "no_entity_body?(300)")
  x.report("no_entity_body?(304)", "no_entity_body?(304)")

  x.report("STATUS_WITH_NO_ENTITY_BODY.key?(100)", "STATUS_WITH_NO_ENTITY_BODY.key?(100)")
  x.report("STATUS_WITH_NO_ENTITY_BODY.key?(200)", "STATUS_WITH_NO_ENTITY_BODY.key?(200)")
  x.report("STATUS_WITH_NO_ENTITY_BODY.key?(300)", "STATUS_WITH_NO_ENTITY_BODY.key?(300)")
  x.report("STATUS_WITH_NO_ENTITY_BODY.key?(304)", "STATUS_WITH_NO_ENTITY_BODY.key?(304)")

  x.report("STATUS_WITH_NO_ENTITY_BODY[100]", "STATUS_WITH_NO_ENTITY_BODY[100]")
  x.report("STATUS_WITH_NO_ENTITY_BODY[200]", "STATUS_WITH_NO_ENTITY_BODY[200]")
  x.report("STATUS_WITH_NO_ENTITY_BODY[300]", "STATUS_WITH_NO_ENTITY_BODY[300]")
  x.report("STATUS_WITH_NO_ENTITY_BODY[304]", "STATUS_WITH_NO_ENTITY_BODY[304]")

  # Compare the iterations per second of the various reports!
  x.compare!
end
```

```
no_entity_body?(100)   414.980k i/100ms
no_entity_body?(200)   412.385k i/100ms
no_entity_body?(300)   408.580k i/100ms
no_entity_body?(304)   404.636k i/100ms
STATUS_WITH_NO_ENTITY_BODY.key?(100)
                       570.662k i/100ms
STATUS_WITH_NO_ENTITY_BODY.key?(200)
                       549.375k i/100ms
STATUS_WITH_NO_ENTITY_BODY.key?(300)
                       551.035k i/100ms
STATUS_WITH_NO_ENTITY_BODY.key?(304)
                       559.606k i/100ms
STATUS_WITH_NO_ENTITY_BODY[100]
                       574.552k i/100ms
STATUS_WITH_NO_ENTITY_BODY[200]
                       571.062k i/100ms
STATUS_WITH_NO_ENTITY_BODY[300]
                       573.254k i/100ms
STATUS_WITH_NO_ENTITY_BODY[304]
                       593.030k i/100ms
no_entity_body?(100)     12.185M (± 3.5%) i/s -     61.002M in   5.013446s
no_entity_body?(200)      9.685M (± 7.0%) i/s -     48.249M in   5.012963s
no_entity_body?(300)      9.880M (± 3.5%) i/s -     49.438M in   5.010383s
no_entity_body?(304)      9.981M (± 2.2%) i/s -     50.175M in   5.029717s
STATUS_WITH_NO_ENTITY_BODY.key?(100)
                         24.819M (± 1.6%) i/s -    124.404M in   5.013830s
STATUS_WITH_NO_ENTITY_BODY.key?(200)
                         24.207M (± 3.2%) i/s -    120.862M in   5.000023s
STATUS_WITH_NO_ENTITY_BODY.key?(300)
                         24.482M (± 0.7%) i/s -    122.881M in   5.019453s
STATUS_WITH_NO_ENTITY_BODY.key?(304)
                         23.185M (± 1.9%) i/s -    116.398M in   5.022572s
STATUS_WITH_NO_ENTITY_BODY[100]
                         29.032M (± 4.0%) i/s -    145.362M in   5.019414s
STATUS_WITH_NO_ENTITY_BODY[200]
                         26.724M (± 2.5%) i/s -    133.629M in   5.004566s
STATUS_WITH_NO_ENTITY_BODY[300]
                         27.777M (± 2.5%) i/s -    139.301M in   5.019331s
STATUS_WITH_NO_ENTITY_BODY[304]
                         27.594M (± 0.9%) i/s -    138.176M in   5.007893s

Comparison:
STATUS_WITH_NO_ENTITY_BODY[100]: 29031901.2 i/s
STATUS_WITH_NO_ENTITY_BODY[300]: 27776583.5 i/s - same-ish: difference falls within error
STATUS_WITH_NO_ENTITY_BODY[304]: 27593955.8 i/s - 1.05x  slower
STATUS_WITH_NO_ENTITY_BODY[200]: 26724045.2 i/s - 1.09x  slower
STATUS_WITH_NO_ENTITY_BODY.key?(100): 24818995.2 i/s - 1.17x  slower
STATUS_WITH_NO_ENTITY_BODY.key?(300): 24482035.6 i/s - 1.19x  slower
STATUS_WITH_NO_ENTITY_BODY.key?(200): 24206584.9 i/s - 1.20x  slower
STATUS_WITH_NO_ENTITY_BODY.key?(304): 23184852.7 i/s - 1.25x  slower
no_entity_body?(100): 12185477.6 i/s - 2.38x  slower
no_entity_body?(304):  9980576.9 i/s - 2.91x  slower
no_entity_body?(300):  9880238.1 i/s - 2.94x  slower
no_entity_body?(200):  9684744.5 i/s - 3.00x  slower
```

test/spec_response.rb
ioquatix: Okay I restored parts of this spec, except the last line, since the body is passed through unchanged.

ioquatix:
Okay, I'm happy to merge this now.

ioquatix:
@tenderlove if you are happy with this can you merge it?

AlexWayfer:
Note: it introduces break-changes for minor version (`2.1`).

You can see https://github.com/AlexWayfer/flame/pull/115, especially https://github.com/AlexWayfer/flame/pull/115/checks?check_run_id=386045064

So… the type (class) of (the last) element in Array, returning from `Response#finish`, has been changed in this MR. And it's critical (the method is public).

jeremyevans:
@AlexWayfer It is a behavior change compared to previous versions, but the previous behavior was not specified, so if you were relying on it, you were relying on undefined behavior/implementation details.

AlexWayfer:
> but the previous behavior was not specified

Hm… okay. But what the "specified behavior"? Specs/tests (for `finish` method in this case)?

ioquatix:
`body` should respond to `#each` and yield 0 or more strings.

AlexWayfer:
> `body` should respond to `#each` and yield 0 or more strings.

There was `self` (`Rack::Response`) as the last item of array in `Response#finish` result, with `#status`, `#body` and other methods (array, obviously, doesn't have them). That's the change. I don't understand why you're talking about `body`.

ioquatix:
`Rack::Response#finish` is to produce the array as defined by the SPEC.

AlexWayfer:
> `Rack::Response#finish` is to produce the array as defined by the SPEC.

Okay, but I still have the question: what is "SPEC"? I'm sorry, I just want to understand (especially as rack-dependent-gem maintainer).

ioquatix:
I added links.

AlexWayfer:
Okay! Thank you.

But… the specs (10–12 lines) are about `call`, not `finish`. 

And your `finish` implementation still has `self` as the last element of array in some cases (like `if block_given?`).

But @jeremyevans wrote above, that `finish` is not in `SPEC` (file), so that's the argument. 

But… the response of `finish` is still inconsistent and I ask to specify it, please. I have to use it, as I remember from documentation, and it's public, but its result is undefined behavior! Strange and unfriendly (by API).

ioquatix:
The response of `#finish` will match the documented requirements for the response array which is:

`[status, headers, body]`

I am happy to add documentation to this effect.

AlexWayfer:
> I am happy to add documentation to this effect.

Thank you, I appreciate it!

PR #1435, opened 2019-12-30
Improve `Request#host_with_port`

AlexWayfer:
Don't return empty port, like `domain:`.

And don't return `:0` port when it's absent.

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -238,16 +238,13 @@ module Rack
       end
 
       def host_with_port
-        if forwarded = get_header(HTTP_X_FORWARDED_HOST)
-          forwarded.split(/,\s?/).last
-        else
-          get_header(HTTP_HOST) || "#{get_header(SERVER_NAME) || get_header(SERVER_ADDR)}:#{get_header(SERVER_PORT)}"
-        end
+        port = self.port
+        port.nil? || port == DEFAULT_PORTS[scheme] ? host : "#{host}:#{port}"
       end
 
       def host
         # Remove port number.
-        h = host_with_port
+        h = hostname.to_s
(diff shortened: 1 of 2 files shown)


ioquatix:
This function got more complicated so can you please use explicit `return`. Other than that it looks okay.

AlexWayfer:
> This function got more complicated so can you please use explicit `return`. Other than that it looks okay.

I can, but there are similar `elsif` in `request.rb` for `scheme` and `port`… and `return` is redundant. Are you sure?

ioquatix:
There is already one conditional return and four ways the function can exit. As soon as functions have more than 1 or 2 exit paths, I prefer explicit return. Feel free to propose some alternative.

ioquatix:
One option would be to refactor this into a hostname method and this one which adds the port if it exists.

AlexWayfer:
> One option would be to refactor this into a hostname method and this one which adds the port if it exists.

I thought about this, that `host_with_port` should use `host` and `port` inside, but then I tried and faced with complex logic like already concatenated string in headers `HTTP_HOST` or `HTTP_X_FORWARDED_HOST`. And I've decided not to change this logic, I didn't come up with a good name.

But I like `hostname`. I'll try.

jeremyevans:
I think the change in this pull request makes sense, so I plan to merge it unless there are objections.  We can always refactor into a `hostname` method later if needed.

AlexWayfer:
I forgot about this, sorry. Give me 24 hours, please, I'm trying to refactor right now.

AlexWayfer:
Done, you can check the refactor with `hostname` (private for now) method.

Thank you for suggestion.

Yeah, `port` method is affected too.

PR #1453, opened 2020-01-07
Remove `to_ary` from Response

tenderlove:
Responses are not arrays, so we should not allow them to be implicitly
coerced to an array.  If you would like you response to be converted to
an array, you must explicitly do it with the `to_a` method.

This also prevents creation of unnecessary body proxies

refs #419

### DIFF
--- a/lib/rack/body_proxy.rb
+++ b/lib/rack/body_proxy.rb
@@ -9,10 +9,6 @@ module Rack
     end
 
     def respond_to?(method_name, include_all = false)
-      case method_name
-      when :to_ary, 'to_ary'
-        return false
-      end
       super or @body.respond_to?(method_name, include_all)
     end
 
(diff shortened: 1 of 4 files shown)


tenderlove:
I was able to confirm that without removing [this line](https://github.com/rack/rack/pull/1453/files#diff-10b933d2c1fdc82ceecade456c64e1c2L79) the `flatten` tests I added would go in to an infinite loop.

dblock:
Looks like this will break Grape, https://github.com/ruby-grape/grape/issues/1955. I haven't dug deep here, @dnesteryuk can probably provide an explanation.

delner:
It does indeed break Grape; [`middleware/base.rb`](https://github.com/ruby-grape/grape/blame/v1.2.5/lib/grape/middleware/base.rb#L43) doesn't coerce the response it gets to an array which causes [`Router#cascade?`](https://github.com/ruby-grape/grape/blob/v1.2.5/lib/grape/router.rb#L163) to receive an Array with the `Response` object and `nil` values.

This issue explains the effect on Grape: https://github.com/ruby-grape/grape/issues/1966

PR #1456, opened 2020-01-08
Response body compact

ioquatix:


ioquatix:
Merging to test grape on edge rails.

ioquatix:
https://github.com/ruby-grape/grape/pull/1956

Here are the changes required to get grape to pass. It is mostly incorrect usage of Rack.

PR #1458, opened 2020-01-08
Add changes to Changelog

p8:
I've updated the Changelog with changes for the following commits: `git log 2-0-stable..master`

PR #1461, opened 2020-01-09
Remove changes from CHANGELOG already present in older versions [ci s…

p8:
In #1458 I added two changes were added that are already present in 2-0-stable.


ioquatix:
Thanks!

PR #1462, opened 2020-01-09
Make SessionId#to_s be an alias of #public_id

jeremyevans:
This restores backwards compatibility with code that uses the
session id.

No justification was given for making to_s raise in the commit
that changed the behavior (cc1d162d28396b6a71f266e6a40ffc19a258792b).

If there is a reason to keep raising an exception, a specific
exception class and explicit error message should be used.

### DIFF
--- a/lib/rack/session/abstract/id.rb
+++ b/lib/rack/session/abstract/id.rb
@@ -28,9 +28,9 @@ module Rack
       end
 
       alias :cookie_value :public_id
+      alias :to_s :public_id
 
       def empty?; false; end
-      def to_s; raise; end
       def inspect; public_id.inspect; end
 
       private


ioquatix:
@jeremyevans I assume you want to backport this to 2.0.x?

jeremyevans:
Yes, since 2.0.8 is affected.

ioquatix:
@tenderlove I'm in agreement with @jeremyevans, so can you merge this, backport it, release 2.0.9 and 2.1.0 tomorrow? Thanks.

tenderlove:
> We thought about doing it, but I think we decided to make it raise so third-party session adapters could upgrade their code.

Yes, this is the justification.  As I said in another thread I should have raised a more specific exception and I really apologize.

tanelsuurhans:
It looks like the 2.0.9 release was missed with this one.

jeremyevans:
2.0.9 hasn't been released yet.  When it is released, it will include this fix.  See https://github.com/rack/rack/commits/2-0-stable

jbcrail:
Can this be backported to 1.6.x? I upgraded recently to address a CVE, but was affected by this issue. Unfortunately, upgrading to 2.x isn't feasible in the short term.

jeremyevans:
@jbcrail This was already backported to 1.6 at 698a060133940a7d9a35c538890ab907f853250e, there just hasn't been another release of 1.6 yet.

PR #1465, opened 2020-01-10
Fix ActiveStorage use-case and add test case. Fixes #1464.

### LINKED ISSUE #1464: Rack 2.1.0 breaks ActiveStorage
I submitted [this issue](https://github.com/rails/rails/issues/38214) to Rails and they sent me here.

Rack::File (now Rack::Files) no longer accepts nil as a first parameter per this commit: https://github.com/rack/rack/commit/8f85307711ca5e7a4729641fda1552890ffa129a. `File.expand_path` throws a TypeError.

This breaks the ActiveStorage file server: https://github.com/rails/rails/blob/master/activestorage/app/controllers/concerns/active_storage/file_server.rb#L6

ioquatix:


ioquatix:
I'll backport this to `2-1-stable`.

ioquatix:
Done...

Issue #1094, opened 2016-07-11
Inconsistency in the ordering of (status, headers, body)


ma11hew28:
The argument order for `Response#new` (body, status, headers) is inconsistent with that of `MockResponse#new` and the order of elements in the array returned by a Rack application's `call` method (status, headers, body). Related: #1095

TechRsch:
Just ran across this today and my head exploded.  :) Positional parameters are problem enough without randomly reordering them.  I certainly appreciate all the work involved here.  I am currently opening an API to my Rails application for my Rack Middleware application and found these parameters out of order.  I expect there is a problem now, with legacy code.  Perhaps keyword arguments could be an option?  Thanks.

ioquatix:
I've added `Rack::Response[status, headers, body]` and `Rack::MockResponse[status, headers, body]` as a way to mitigate this issue and perhaps in Rack 3.0 we can break this.

PR #1466, opened 2020-01-10
Add new `Response.[]` and `MockResponse.[]` which are very similar. F…

ioquatix:
…ixes #1094.

### DIFF
--- a/lib/rack/response.rb
+++ b/lib/rack/response.rb
@@ -21,7 +21,7 @@ module Rack
   # Your application's +call+ should end returning Response#finish.
 
   class Response
-    def self.[] (status, headers, body)
+    def self.[](status, headers, body)
       self.new(body, status, headers)
     end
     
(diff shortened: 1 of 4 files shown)


ioquatix:
@tenderlove @rafaelfranca what do you think? It's a first step to trying to fix the inconsistencies of the interfaces.

`MockResponse#body` returns a string but `Response#body` is whatever was supplied, or an array, depending on whether it gets buffered. I can see how this is confusing, but it's also probably embedded in millions of tests by now...

lib/rack/response.rb
```diff
@@ -21,6 +21,10 @@ module Rack
   # Your application's +call+ should end returning Response#finish.
 
   class Response
+    def self.[] (status, headers, body)
```
rafaelfranca:
```suggestion
    def self.[](status, headers, body)
```

PR #1469, opened 2020-01-11
Handle case where session id key is requested but it is missing

### LINKED ISSUE #1433: Rack 1.6.12 gives an error when getting the session ID
Upgrading to Rack 1.6.12 is giving us an error in some code related to logging – we use the session ID from Rack to help group errors relating to the same session together across services.

The line giving the error in Rack is here: https://github.com/rack/rack/blob/1.6.12/lib/rack/session/abstract/id.rb#L434

And the error is 
```
undefined method `public_id' for nil:NilClass
```

The line in our code that calls it is an `if` statement checking if the session ID exists:
```
if env['rack.session'] && env['rack.session']['session_id']
  ... do stuff with env['rack.session']['session_id']
end
```

From the look of it, this used to return nil if the underlying hash key was absent, but now it seems to assume there _will_ be an `id` object if `["session_id"]` is getting called. Is there something else we should be doing to determine if the session exists now?

I think this might be related to this as well, if it’s all to do with what the type of the `id` object is now: https://github.com/rack/rack/issues/1432

jeremyevans:
Use historical behavior of returning nil in this case.

Add tests for Rack::Session::Abstract::PersistedSecure::SecureSessionHash,
mostly based on the existing ones for Rack::Session::Abstract::SessionHash.

Fixes #1433.  Needs backport to 1.6 and 2.0.

### DIFF
--- a/lib/rack/session/abstract/id.rb
+++ b/lib/rack/session/abstract/id.rb
@@ -460,7 +460,7 @@ module Rack
           def [](key)
             if key == "session_id"
               load_for_read!
-              id.public_id
+              id.public_id if id
             else
               super
             end
--- a/test/spec_session_persisted_secure_secure_session_hash.rb
+++ b/test/spec_session_persisted_secure_secure_session_hash.rb
@@ -1,0 +1,73 @@
+# frozen_string_literal: true
+
+require 'minitest/global_expectations/autorun'
+require 'rack/session/abstract/id'
+
+describe Rack::Session::Abstract::PersistedSecure::SecureSessionHash do
+  attr_reader :hash
+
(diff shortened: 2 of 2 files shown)


ioquatix:
Can you rebase on master and I will review.

jeremyevans:
Done, and CI is green.

ioquatix:
@tenderlove can you backport to 1.6 and 2.0 and release?

rafaelfranca:
I backported this commit to all branches

dee-see:
Hello! Do you have any ETA for 2.0.9?

Thanks

Issue #1472, opened 2020-01-11
2.1.0 - Content-Length no longer present in response's specs


erwanlr:
After updating to 2.1.0, all my tests checking for the `Content-Length` header in responses are failing. It seems like the header is not set anymore somehow.

Here is an example:

```ruby
# config.ru
# frozen_string_literal: true

run lambda { |_| [200, {}, ['Hello, world!']] }
```

```ruby
# spec.rb
require 'rspec'
require 'rspec/its'
require 'rack'
require 'rack/test'

RSpec.configure do |config|
  config.include Rack::Test::Methods
end

describe 'App' do
  def app
    Rack::Builder.parse_file('config.ru').first
  end

  context 'when GET /' do
    subject(:res) { get('/') }

    its(:status) { is_expected.to be 200 }
    its(:headers) { are_expected.to include('Content-Length' => '13') }
    its(:body) { is_expected.to eql 'Hello, world!' }
  end
end
```

```ruby
# test.rb
require 'minitest/global_expectations/autorun'
require 'rack/mock'

describe 'App' do
  def app
    Rack::Builder.parse_file('config.ru').first
  end

  def request
    Rack::MockRequest.new(app)
  end

  it 'GET /' do
    response = request.get('/')

    response.status.must_equal 200
    response.header['Content-Length'].must_equal '13'
    response.body.must_equal 'Hello, world!'
  end
end
```

Gemfile
```ruby
# frozen_string_literal: true

source 'https://rubygems.org'

gem 'rack', '~> 2.1.1'
#gem 'rack', git: 'https://github.com/rack/rack.git'

gem 'rack-test', '~> 1.1.0'
gem 'rspec', '~> 3.9.0'
gem 'rspec-its', '~> 1.3.0'

gem 'minitest', '~> 5.14.0'
gem 'minitest-global_expectations', '~> 1.0.1'
```

Put these files in a same directory and run` rspec spec.rb` if using RSpec or `bundle exec ruby test.rb` to use Minitest

Output with Rack 2.1.0, and Rack-Test 1.1.0:
```
$ rspec spec.rb 
.F.

Failures:

  1) App when GET / headers is expected to include {"Content-Length" => "10"}
     Failure/Error: its(:headers) { are_expected.to include('Content-Length' => '10') }
     
       expected {} to include {"Content-Length" => "10"}
       Diff:
       @@ -1,2 +1 @@
       -"Content-Length" => "10",
       
[...]

Issue #1266, opened 2018-04-20
Rack::Chunked middleware conflicts with Rack::Sendfile


boazsegev:
The Rack::Chunked middleware conflict with the Rack::Sendfile middleware / approach.

The [Rack::Chunked doesn't test if the `'X-Sendfile'` header family is set](https://github.com/rack/rack/blob/377c54610c6849259e7915c2860fa58835492566/lib/rack/chunked.rb#L59-L63), adding the chunked transfer encoding header.

This results in two possible approaches:

1. The server ignores any headers set by the application (losing any cookie data or other custom headers).

2. The encoding header remains intact\*, causing the response to fail.

Both approaches result in loss of data (either the cookie data or a failed response).

The best way to fix the issue might be to add tests to the Rack::Chunked middleware, testing for the `'X-Sendfile'` header family.

\* Consider that the header might have been `Transfer-Encoding: gzip` and the file `X-sendfile: style.css.gz`, which means that the header should have remained intact.

This issue [originated here](https://github.com/boazsegev/plezi/issues/24).

ioquatix:
I've actually been thinking about this, and I believe we should remove `Rack::Chunked` because it's something that should be handled by the web server, not the application.

boazsegev:
@ioquatix, I agree.

Transport protocol details should probably be considered a server concern.

acmh:
Is this issue up for grabs?

ioquatix:
@acmh I guess yes - what do you think should be done? My suggested plan is to remove it in Rack 3.0.

boazsegev:
@acmh - I already mentioned this, but I second the approach offered by @ioquatix. There are a number of reasons why deprecating the `Rack::Chunked` middleware would be a good idea, and this issue is just one of them. For example:

* Message formatting is a Server concern (Protocol details) and not an application concern. Message formatting might (and did/will) change with HTTP/2 and HTTP/3.

* By returning this concern to the server layer, servers will be able to automatically use chunked encoding when appropriate (i.e., no `content-length` header and a response stream that routes to `#each` ( not a simple Array with a single String member). This could be performed according to the correct protocol specifications and according to the correct protocol version.

* By avoiding a `chunked` response, testing might be simplified and the response (payload) could be tested more easily, bypassing the need for a client emulation layer. This will also improve performance.

acmh:
[...]

PR #1475, opened 2020-01-11
Remove chunked middleware from default server stack.

ioquatix:
The chunked middleware assumes details about the protocol level
implementation. Due to recent changes from HTTP_VERSION to SERVER_PROTOCOL,
this dormant middleware now appears to be breaking a lot of assumptions.

https://github.com/rack/rack/issues/1472
https://github.com/rack/rack/issues/1266

### DIFF
--- a/lib/rack/server.rb
+++ b/lib/rack/server.rb
@@ -262,13 +262,11 @@ module Rack
         m = Hash.new {|h, k| h[k] = []}
         m["deployment"] = [
           [Rack::ContentLength],
-          [Rack::Chunked],
           logging_middleware,
           [Rack::TempfileReaper]
         ]
         m["development"] = [
           [Rack::ContentLength],
-          [Rack::Chunked],
           logging_middleware,
           [Rack::ShowExceptions],
           [Rack::Lint],


ioquatix:
@tenderlove I think we need to backport and release 2.1.1 as this change is causing a lot of potential issues upstream.

Because we fixed the version detection, the previously non-working `Rack::Chunked` is now mutilating responses in full force.

JuanitoFatas:
May I ask what is the replacement for `Rack::Chunked` (trying to update [this test](https://github.com/puma/puma/blob/2039e616b6b281a8c24ef6d5eef84b63bed394ad/test/test_rack_server.rb#L234) so no deprecation warning) 😅? Thank you.

jeremyevans:
It was deprecated and removed without replacement.  Servers are welcome to use the previous code for HTTP 1.1 responses if it meets their needs.

ioquatix:
@JuanitoFatas I'm hoping that Puma supports chunked encoding without using `Rack::Chunked` - I advise you to only run the test on Rack < 3, or test that Puma itself supports chunked encoding, not that `Rack::Chunked` is working. The actual test should work irrespective of whether `Rack::Chunked` is used.

PR #1479, opened 2020-01-13
Ensure `Rack::Lint` to work with `Rack::Response` on 2.1

jodosha:
When you use `Rack::MockRequest#request` the returning value is a `Rack::MockResponse`, which is a subclass of `Rack::Response`.

Given Rack 2.1 [removed](https://github.com/rack/rack/commit/72959ebc2f300f3b2ccb7ae2aae9f199e611dfb6) `Rack::Response#to_ary`, `Rack::Lint` should explicitly invoke `#to_a` to extract status, headers, and body from current response.

Without this patch, you'll get errors like the following:

```shell
Failures:

  1) Hanami::Router#root path recognition block recognizes
     Failure/Error: actual = app.request("GET", "/", lint: true)

     NoMethodError:
       undefined method `to_i' for #<Rack::MockResponse:0x00007f9212d970c8>
       Did you mean?  to_s
                      to_a
     # /Users/luca/.gem/ruby/2.7.0/gems/rack-2.1.1/lib/rack/lint.rb:621:in `block in check_status'
     # /Users/luca/.gem/ruby/2.7.0/gems/rack-2.1.1/lib/rack/lint.rb:21:in `assert'
     # /Users/luca/.gem/ruby/2.7.0/gems/rack-2.1.1/lib/rack/lint.rb:621:in `check_status'
     # /Users/luca/.gem/ruby/2.7.0/gems/rack-2.1.1/lib/rack/lint.rb:53:in `_call'
     # /Users/luca/.gem/ruby/2.7.0/gems/rack-2.1.1/lib/rack/lint.rb:39:in `call'
     # /Users/luca/.gem/ruby/2.7.0/gems/rack-2.1.1/lib/rack/mock.rb:77:in `request'
     # ./spec/integration/hanami/router/routing_spec.rb:204:in `block (5 levels) in <top (required)>'
```

### DIFF
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -48,7 +48,7 @@ module Rack
       env[RACK_ERRORS] = ErrorWrapper.new(env[RACK_ERRORS])
 
       ## and returns an Array of exactly three values:
-      status, headers, @body = @app.call(env)
+      status, headers, @body = @app.call(env).to_a
       ## The *status*,
       check_status status
       ## the *headers*,
--- a/test/spec_lint.rb
+++ b/test/spec_lint.rb
@@ -17,6 +17,12 @@ describe Rack::Lint do
                    }).call(env({})).first.must_equal 200
   end
 
+  it "works with Rack::Response" do
+    Rack::Lint.new(lambda { |env|
+                     Rack::Response.new(["foo"], 200, { "Content-type" => "test/plain", "Content-length" => "3" })
+                   }).call(env({})).first.must_equal 200
+  end
+
   it "notice fatal errors" do
     lambda { Rack::Lint.new(nil).call }.must_raise(Rack::Lint::LintError).
       message.must_match(/No env given/)


rafaelfranca:
`status, headers, @body = @app.call(env)` is a very common idiom on rack applications, I'm afraid that change on 2.1 will break that like we are see here in `Lint`. @tenderlove @ioquatix were we expecting this change?

ioquatix:
It’s not a valid use case according to the SPEC which requires an array of three elements. That being said, if we can fix the rails use case which is calling `flatten` incorrectly on the response object, maybe we can consider adding back `to_ary`. I don’t mind supporting `to_ary` but we need to decide which bug we want and whether we need to adjust the SPEC.

ioquatix:
Looking specifically at this PR and the SPEC, we don’t even guarantee we can call `to_a`.

jodosha:
@ioquatix What is surprising is that this breaking change happened in a minor release.

Even if `#to_a` is gonna get removed, we should at least try to fix `Rack::Lint`. For instance:

```ruby
response = @app.call(env)
status = response.status
headers = response.headers
@body = response.body
```

jodosha:
After temporarily monkey-patching `Rack::Lint` to see if my fix would have solved the problems, I discovered that there are several places in Rack codebase itself where the expansion of response is done (`status, headers, body = app.call(env)`).

At the end of the day, instead of fixing `Rack::Lint` and all the other expansion occurrencies, it's worth to let `Rack::MockResponse` to implement `#to_ary`.

In this way when `Rack::MockRequest#request` is used specs won't crash for people.

@ioquatix @rafaelfranca What do you think?

ioquatix:
Removing `Response#to_ary` which was a public interface might have been a mistake. But the alternative is adjusting the SPEC. Then fixing places where users are calling `flatten` and/or changing `Rack#finish` to return a wrapper around itself so that `flatten` doesn’t cause infinite recursion.

I guess if it was up to me, I’d suggest:

- Updating the spec to allow `to_ary` on response.
- Revert the removal of `to_ary` and add better specs to cover these use cases.

I don’t really know if there is any good solution here. But clearly we’ve broken something which is causing pain for users. Putting a wrapper around `self` in `Response#finish` is relatively straight forward... it’s not as efficient but 🙃

jodosha:
@ioquatix @rafaelfranca I believe that the SPEC is too concerned by the Ruby details.

IMO specs are like cross-language specifications (like RFCs). To me, anyone should be able to pick the Rack SPEC and implement into another programming language.

In this prospective, Rack SPEC shouldn't enter into the details of `#to_a` **and** at the same time we should guarantee backward compat, even if you still think there is a contradiction with the SPEC.

Maybe removing `#to_a` can be a gola for 3.0?

rafaelfranca:
@ioquatix I like your suggestions. The SPEC talks about `to_s` and other things that are specific to ruby so I don't see a problem on talking about `to_ary`. If we want to remove to_ary at long term we should do in the major release and fix Rails and all other places that are doing that.

jeremyevans:
@jodosha I disagree regarding SPEC.  SPEC is specifically for Ruby, discussing Ruby classes and Ruby methods, since its introduction in 2008 (https://github.com/rack/rack/commit/7ffddff24442685bb5ced7bb787aad0ff7aa9c73#diff-a9462190f9ccc9f1ced64667af619aed).

It is impossible to guarantee backwards compatibility when changing SPEC to make SPEC stricter.  Make SPEC stricter would improve things in many cases.  So to guarantee backwards compatibility would be to limit the possible improvements we could make to SPEC.  Consider #1485. That is backwards incompatible, but not modifying SPEC for it leads to a situation where a valid Rack application would not work with almost all middleware that modifies response headers.   

Regarding `to_a`/`to_ary`, I don't think we should modify SPEC for those.  SPEC states the response should be `an Array of exactly three values`.  Not an object that responds to `to_ary` or `to_a` and returns an array of exactly three values.  Recently I made `Rack::Lint` more strict so it correctly checks this, which is what is causing the conflict.  Any rack application that isn't returning an Array of 3 values should be fixed, IMO.

jodosha:
@jeremyevans @ioquatix Thanks for the explanations. I believe this PR isn't required anymore.

PR #1480, opened 2020-01-13
Adds specs to reproduce #1472

erwanlr:
See #1472


jeremyevans:
We determined that #1472 wasn't a bug, so I'm going to close this.

PR #1481, opened 2020-01-14
Respect preexisting `:BindAddress` options passed to `Rack::Handler::WEBrick.run`

postmodern:
I have noticed that other projects which use `Rack::Handler::WEBrick` directly will pass in WEBrick options (ex: `:BindAddress`) as opposed to the standard Rack options (ex: `:Host`). This change prevents overriding any `:BindAddress` options that were explicitly passed in.


jeremyevans:
I fixed this in a slightly more backwards compatible way in 0d8838041f4b91e88a10feefa2ca40b8dac4a2dc.

ioquatix:
Nice work @jeremyevans & @postmodern!

PR #1484, opened 2020-01-14
Make body not respond to to_path to Rack::Files range requests

### LINKED ISSUE #1235: Broken range requests with Rack::File and WEBrick
The changes in e0ac329 and #753 mean that when `Rack::File` processes a range request the file object gets returned to WEBrick and the offset position is lost, resulting in the file being read from the beginning again.

jeremyevans:
If the body responds to to_path, the server is allowed to serve
the file at that path and ignore the body.  That works well for
most requests, but not for range requests, since the correct
byte range won't be served.

Work around this issue by using a separate body class for the
partial content responses that does not respond to to_path, and
the current body class for non-partial content responses.

Fixes #1235

### DIFF
--- a/lib/rack/files.rb
+++ b/lib/rack/files.rb
@@ -90,6 +90,7 @@ module Rack
         return response
       else
         # Partial content:
+        partial_content = true
         range = ranges[0]
         response[0] = 206
         response[1]["Content-Range"] = "bytes #{range.begin}-#{range.end}/#{size}"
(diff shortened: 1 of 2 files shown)


ioquatix:
Falcon explicitly handles this case:

https://github.com/socketry/falcon/blob/21d383fd04b97ad2d82bf064481570cd9d489e35/lib/falcon/adapters/output.rb#L41-L43

Not sure the solution merged here is the right one, if you assume that servers can handle range requests better than the middleware. That being said, if servers are currently broken... then maybe this is the right approach.

jeremyevans:
I'm not sure if this approach is the best approach either.  But unless we are sure the server will correctly support range requests in this scenario, it seems a safer approach.

PR #1485, opened 2020-01-15
Require the response headers be a hash in SPEC

### LINKED ISSUE #1222: Headers specs are not clear enough
Most Rack middlewares expect the returned headers to act like a Hash but this is not stated in the specs:

https://github.com/rack/rack/blob/master/SPEC#L226

```
=== The Headers
The header must respond to +each+, and yield values of key and value.
Special headers starting "rack." are for communicating with the
server, and must not be sent back to the client.
The header keys must be Strings.
The header must not contain a +Status+ key.
The header must conform to RFC7230 token specification, i.e. cannot
contain non-printable ASCII, DQUOTE or "(),/:;<=>?@[\]{}".
The values of the header must be Strings,
consisting of lines (for multiple header values, e.g. multiple
<tt>Set-Cookie</tt> values) separated by "\\n".
The lines must not contain characters below 037.
```

While reading this specs, I get the impression that something like this would meet the requirements:

```ruby
headers = [
  [ 'Server-Timing', 'app;dur=100' ],
  [ 'Server-Timing', 'db;dur=60' ],
]
```

It responds to `each` and yields values of key and value, right?

However, if we take a look at the Runtime middleware we can see it would not work with such a header:

https://github.com/rack/rack/blob/master/lib/rack/runtime.rb#L25

``` ruby
unless headers.has_key?(@header_name)
```

An array doesn't respond to `has_key?`. By the way, Matz prefers `key?` to be used over `has_key?`.

Shouldn't the specs be clear about headers having to act as hashes, or at least list which methods it should respond to?

jeremyevans:
This is stricter than what was previously required.  However,
non-hash response headers would break most of the middleware
that accesses response headers.

Fixes #1222

### DIFF
--- a/SPEC
+++ b/SPEC
@@ -224,8 +224,8 @@ if the request env has <tt>rack.hijack?</tt> <tt>true</tt>.
 This is an HTTP status. When parsed as integer (+to_i+), it must be
 greater than or equal to 100.
 === The Headers
-The header must respond to +each+, and yield values of key and value.
-The header keys must be Strings.
+The header must be a unfrozen Hash.
+The keys of the header must be Strings.
 Special headers starting "rack." are for communicating with the
 server, and must not be sent back to the client.
 The header must not contain a +Status+ key.
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -631,13 +631,16 @@ module Rack
 
     ## === The Headers
     def check_headers(header)
-      ## The header must respond to +each+, and yield values of key and value.
-      assert("headers object should respond to #each, but doesn't (got #{header.class} as headers)") {
-         header.respond_to? :each
+      ## The header must be a unfrozen Hash.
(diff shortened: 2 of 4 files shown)


rafaelfranca:
This looks good to me but before we make it more strict I'd like to make sure rails will not break with this, unless we are planning to only make the change on 3.0

ioquatix:
This is probably a good change. But it is also a major breaking change to anyone who uses a rich data type for the response headers, which includes Falcon I believe.

I'll review it.

However, I generally understand and agree with this change, except that we should also probably ensure that users aren't modifying the header hash directly, or if they are, the SPEC needs to also state the mutability of the header hash and how to deal with mutations.

jeremyevans:
I made some minor tweaks.  Based on @ioquatix's feedback, SPEC now requires that the header hash be unfrozen so it can be efficiently modified as needed in middleware.

ioquatix:
I've rebased on master in https://github.com/rack/rack/pull/1597

PR #1486, opened 2020-01-15
Fix incorrect MockResponse#body String encoding

sikachu:
This is a regression introduced in Rack 2.1.0, possibly after 8c62821f4a464858a6b6ca3c3966ec308d2bb53e was introduced.

Before the aforementioned commit, `MockResponse#body` would return a UTF-8 string as it joins all the response parts together. However, because the aforementioned commit uses `String.new` without specifying an encoding, Ruby would [create a new String with `ASCII-8BIT` encoding
by default](https://ruby-doc.org/core-2.7.0/String.html#method-c-new).

This commit adds failing test cases to prevent future regression to ensure that we return UTF-8 String for all string body, and also making sure that we return binary string if any of the response part is a binary string (with `ASCII-8BIT` encoding).

---

As a side note, I found out about this regression after upgrading to Rack 2.1.1 and found out that API documentation, which is generated using [rspec_api_documentation](https://github.com/zipmark/rspec_api_documentation), no longer show JSON response body. After digging through the code, it turns out they check for [string encoding](https://github.com/zipmark/rspec_api_documentation/blob/560c3bdc7bd5581e7c223334390221ecfc910be8/lib/rspec_api_documentation/client_base.rb#L90-L95) to see if they should include the response body in the documentation or not. After than, I dug through the code and found 8c62821f4a464858a6b6ca3c3966ec308d2bb53e which uses `String.new` without specifying the encoding.

Please let me know if there's any further work on this PR needed to get it merged, or if I'm wrong to assume that I should get UTF-8 String back from `MockResponse#body`. Thank you.

### DIFF
--- a/lib/rack/mock.rb
+++ b/lib/rack/mock.rb
@@ -196,9 +196,13 @@ module Rack
       #     ...
       #     res.body.should == "foo!"
       #   end
-      buffer = String.new
+      buffer = String.new(encoding: Encoding::UTF_8)
 
       super.each do |chunk|
+        if chunk.encoding == Encoding::ASCII_8BIT
+          buffer.encode!(Encoding::ASCII_8BIT)
+        end
+
         buffer << chunk
       end
 
--- a/test/spec_mock.rb
+++ b/test/spec_mock.rb
@@ -348,6 +348,19 @@ describe Rack::MockResponse do
     res.body.must_equal 'hi'
   end
 
+  it "return non-binary response body as UTF-8 string" do
+    body = ["utf8 body", "iso-8859-1 body".encode(Encoding::ISO_8859_1)]
+    res = Rack::MockResponse.new(200, {}, body)
+    res.body.encoding.must_equal Encoding::UTF_8
+  end
+
+  it "return binary response body as binary string" do
+    fixture = File.join(File.dirname(__FILE__), "multipart", "rack-logo.png")
+    body = [File.binread(fixture)]
+    res = Rack::MockResponse.new(200, {}, body)
+    res.body.encoding.must_equal Encoding::ASCII_8BIT
+  end
+
   it "optionally make Rack errors fatal" do
     lambda {
       Rack::MockRequest.new(app).get("/?error=foo", fatal: true)


sikachu:
It seems like failure in test-ruby-2.5 does not relate to this change?

jeremyevans:
I think making this change would break non-UTF8 non-ASCII8BIT response bodies.  There should be no problem with `String.new` without encoding, because as soon as you append a UTF-8 string to it, it will change to UTF-8 encoding (likewise with other ASCII compatible encodings):

```ruby
String.new.encoding
# => #<Encoding:ASCII-8BIT>
(String.new << "\u1234").encoding
# => #<Encoding:UTF-8>
```

Can you provide example code showing how the current behavior is a problem?

tenderlove:
I think the response body should probably always be ASCII-8BIT.  Rack can't really know anything about the encoding of the bytes that users want to send.  At the end of the day, it's just bytes written to the socket, and I don't think Rack should have any opinion about the encoding the user sends.  IOW, the buffer should just be `''.b`

tenderlove:
Just to be incredibly specific, I've written some sample code:

```
x = "四\u2160"
y = "Hello \x93\xfa\x96\x7b".force_encoding("Shift_JIS")

p x.encoding        # Encoding:UTF-8
p x.valid_encoding? # true

p y.encoding        # Encoding: Shift_JIS
p y.valid_encoding? # true

begin
  x + y
rescue
  # They can't be concatenated
  puts "oh no"
end

# Rack response
response =  [200, {}, [x, y]]

require "socket"
lock = Queue.new

# Fake Rack Compatible Webserver
server = Thread.new do
  server = TCPServer.new 2000 # Server bind to port 2000
  lock << true
  client = server.accept    # Wait for a client to connect
  response.last.each do |body|
    client.write body
  end
  client.close
end

# Fake Web Browser
client = Thread.new do
  lock.pop
  s = TCPSocket.new 'localhost', 2000

  while line = s.gets # Read lines from socket
    puts line         # and print them
  end
end

server.join
client.join
```

The two strings are UTF-8 and Shift_JIS.  They are not compatible encodings, but the individual strings contain valid bytes for their respective encodings.  Since the strings don't have compatible encodings, they can't be concatenated in Ruby.  But it's totally fine to write both of them to a socket.  Of course this results in mixed UTF-8 and Shift_JIS being written to the client, but it is *not* Rack's responsibility to ensure that the user didn't make this kind of mistake.

Also, we can't assume that all responses should be UTF-8 (it's totally fine for people to send Shift_JIS encoded web pages).

Since we can't know the correct encoding, and the socket doesn't care anyway, I think the mock body should always be ASCII-8BIT.

Hopefully that makes sense!

jeremyevans:
@tenderlove That makes sense to me.  I think it requires changes to `MockResponse#body`: `buffer << chunk.to_str.b`.  I think this change would break tests using `MockResponse#body` and comparing it to an encoded string (`"\u1234" == "\u1234".b # => false`), but that may be considered acceptable.

tenderlove:
One thought I had is we could examine the `Content-Type` header of the response and set the encoding based on that.  Again though, that's not something a webserver would actually do, so I'm not sure :/

sikachu:
Thanks Aaron, I think that totally makes a perfect sense.

@jeremyevans the code that ran into this problem is at https://github.com/zipmark/rspec_api_documentation/blob/560c3bdc7bd5581e7c223334390221ecfc910be8/lib/rspec_api_documentation/client_base.rb#L90-L95. After reading @tenderlove's response I feel like maybe they're using Ruby String encoding for a wrong purpose, so maybe the fix need to be done there ...

I guess the old behavior (that we'd get UTF-8 string) is just a side-effect of using `.join`, and we should not rely on String encoding to determine the result encoding.

I can submit PR to rspec_api_documentation to check for content type and print out if it's a supported content type insead. Maybe that'd work.

On Rack side, do you think it's necessary to add a CHANGELOG or a test case regarding `MockResponse#body` now return ASCII-8BIT string? I can submit a PR to do that. Please let me know.

jeremyevans:
@sikachu I think it's fine to send an update to the CHANGELOG for 2.1.0 in the master branch mentioning the MockResponse#body may now return non-UTF8 strings.  I don't think you can say it always returns an ASCII-8BIT string, because that change hasn't been made yet. A test case for the current behavior is fine to include as well.

As it appears we won't be merging this PR, I'm going to close this now.

PR #1488, opened 2020-01-15
Update SPEC and Lint to require ASCII-8BIT encoding for CGI values wih non-ASCII characters

### LINKED ISSUE #1183: warning: regexp match /.../n against to UTF-8 string
When I upgrade my ruby version from 2.4.0 to 2.4.1, and rails version from 5.0.2 to 5.1.1, I got this warning during I ran my integration tests.
`.rvm/gems/ruby-2.4.1/gems/rack-2.0.3/lib/rack/query_parser.rb:64: warning: regexp match /.../n against to UTF-8 string
`

It seems the same with **[this](https://github.com/rails/rails/issues/3927)**

jeremyevans:
Fixes warning when binary Regexp is used to match against a
CGI value (QUERY_STRING).

ASCII-8BIT encoding is already required for response bodies.
Requiring this encoding for all CGI values would break usage
with puma and potentially other rack servers. So it seems
safest to only require ASCII-8BIT encoding when the value
actually contains non-ASCII.  We could further limit it to
just QUERY_STRING, but I'm not sure we want to do that.

Testing with a few rack servers:

* Unicorn always uses ASCII-8BIT for all CGI values.
* Puma will use UTF-8 for some values, but uses ASCII-8BIT
  for CGI values contianing non-ASCII characters.
* Webrick doesn't support non-ASCII chracters, at least for
  PATH_INFO or QUERY_STRING.

If we don't do this, the only way to avoid the warning would be
to duplicate QUERY_STRING if it is not already binary.  Checking
encoding will slow down parsing even if it already uses
ASCII-8BIT encoding.

Fixes #1183

### DIFF
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -278,10 +278,16 @@ module Rack
       }
 
       ## The CGI keys (named without a period) must have String values.
+      ## If the string values for CGI keys contain non-ASCII characters,
+      ## they should use ASCII-8BIT encoding.
       env.each { |key, value|
         next  if key.include? "."   # Skip extensions
         assert("env variable #{key} has non-string value #{value.inspect}") {
           value.kind_of? String
+        }
+        next if value.encoding == Encoding::ASCII_8BIT
+        assert("env variable #{key} has value containing non-ASCII characters and has non-ASCII-8BIT encoding #{value.inspect} encoding: #{value.encoding}") {
+          value.b !~ /[\x80-\xff]/n
         }
       }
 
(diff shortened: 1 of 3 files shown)


lib/rack/lint.rb
```diff
@@ -278,11 +278,17 @@ def check_env(env)
...
 
       ## The CGI keys (named without a period) must have String values.
+      ## If the string values contain non-ASCII characters, they should
+      ## use a ASCII-8BIT encoding.
       env.each { |key, value|
         next  if key.include? "."   # Skip extensions
         assert("env variable #{key} has non-string value #{value.inspect}") {
           value.kind_of? String
         }
+        next if value.encoding.to_s == "ASCII-8BIT"
```
kamipo: How about avoid to_s?

```suggestion
        next if value.encoding == Encoding::ASCII_8BIT
```

jeremyevans: That should be fine.  I figured this is Rack::Lint, micro optimization is not really necessary, but there's no harm in doing so.

ioquatix:
LGTM!

PR #1493, opened 2020-01-17
Support providing IO-like object to Rack::Multipart::UploadedFile

### LINKED ISSUE #1028: Adding data to multipart request not from file
I've been working using Rack to submit multipart requests to the Apple News API. This works fine when just submitting data from a JSON file, but there are cases in which an additional `metadata` part is required and is not expected to have the `filename` field in the `Content-Disposition` header. Here is the example from the [Apple News Documentation](https://developer.apple.com/library/ios/documentation/General/Conceptual/News_API_Ref/UpdateArticle.html#//apple_ref/doc/uid/TP40015409-CH15-SW1):

```
--535e329ca936f79a19ac9a251f7d48f7
Content-Type: application/json
Content-Disposition: form-data; name=metadata

{
   "data": {
        "revision": "AAAAAAAAAAAAAAAAAAAAev=="
    }
}
--535e329ca936f79a19ac9a251f7d48f7
```

The issue that I was running into when using `Multipart::build_multipart` is that it if there is no `original_filename` attribute, then it uses `Generator#content_for_other` to create this part of the request, which does not specify `Content-Type` (lines 83-90 of multipart/generator.rb). If I were to use a file (or file like object) and build this with Multipart::UploadedFile, then the filename is included in the `Content-Disposition`, which I do not want.

What I was hoping to see (and I'd be willing to help contribute to) is a class similar to `Multipart::UploadedFile` that instead of receiving the path to a file upon initialization could simply receive a string of the content, and corresponding changes to `Multipart::Generator` that would allow for an instance of this class to be passed to the generator and build the part of the request correctly.

I'm not sure if this is an issue with other APIs that expect multipart requests, but for the Apple News API there are several different metadata fields that we will need to be submitting, so this functionality would be very useful. Thanks for all the hard work on Rack and let me know if I can provide any clarification or help make these changes.

jeremyevans:
This changes the UploadedFile API to use keyword arguments, being
backwards compatible with the previous API by keeping the positional
arguments and making the keyword defaults be the positional argument
values.

The two new arguments to UploadedFile#initialize are filename and io.
filename allows overriding the filename that will be used, instead of
always using the base name of the path.  io specifies an IO-like
object that can be read from to get the data, overridding any
provided path.  This allows you to generate UploadedFile instances
that previously were not possible without accessing the file system.

This changes the multipart generator to only include a Content-Length
for files with paths, and only include a filename in
Content-Disposition if the UploadedFile has one.

Fixes #1028.

### DIFF
--- a/lib/rack/multipart/generator.rb
+++ b/lib/rack/multipart/generator.rb
@@ -17,9 +17,13 @@ module Rack
 
         flattened_params.map do |name, file|
           if file.respond_to?(:original_filename)
-            ::File.open(file.path, 'rb') do |f|
-              f.set_encoding(Encoding::BINARY)
-              content_for_tempfile(f, file, name)
+            if file.path
+              ::File.open(file.path, 'rb') do |f|
+                f.set_encoding(Encoding::BINARY)
+                content_for_tempfile(f, file, name)
+              end
+            else
+              content_for_tempfile(file, file, name)
             end
           else
             content_for_other(file, name)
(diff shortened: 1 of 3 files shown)


lib/rack/multipart/generator.rb
```diff
@@ -69,12 +73,13 @@ def flattened_params
       end
 
       def content_for_tempfile(io, file, name)
+        length = ::File.stat(file.path).size if file.path
+        filename = "; filename=\"#{Utils.escape(file.original_filename)}\"" if file.original_filename
```
prathe, line 77: Does this create a regression? Filenames with spaces will now have a `+` instead of the space.

Does that means that clients will need to unescape de filename prior to use?

PR #1499, opened 2020-01-17
Escape [ and ] in Rack::Utils.escape

### LINKED ISSUE #792: Rack::Utils.build_nested_query should use escaped square brackets
According to validator.w3.org and http://www.ietf.org/rfc/rfc3986.txt square brackets are not valid characters in query part of url and should be escaped as %5B resp %5D.

Rack::Utils.build_nested_query is using unescaped ones.

I must admit I'm not entirely sure what is you supposed to do if you want to use square brackets as hash key, but seems that input part of rack is already parsing encoded square brackets according to http://www.ietf.org/rfc/rfc3986.txt ...

jeremyevans:
These should be escaped per RFC 3986 section 3.4, which does not
allow them in query strings.

Rack already handled these correctly when processing input, the
test changes are places where the values were checked directly
without running it through an input processor.

Fixes #792

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -32,7 +32,7 @@ module Rack
 
     # URI escapes. (CGI style space to +)
     def escape(s)
-      URI.encode_www_form_component(s)
+      URI.encode_www_form_component(s).gsub('[', '%5B').gsub(']', '%5D')
     end
     module_function :escape
 
@@ -119,11 +119,11 @@ module Rack
       case value
       when Array
         value.map { |v|
-          build_nested_query(v, "#{prefix}[]")
+          build_nested_query(v, "#{prefix}%5B%5D")
         }.join("&")
       when Hash
         value.map { |k, v|
-          build_nested_query(v, prefix ? "#{prefix}[#{escape(k)}]" : escape(k))
+          build_nested_query(v, prefix ? "#{prefix}%5B#{escape(k)}%5D" : escape(k))
         }.delete_if(&:empty?).join('&')
       when nil
         prefix
--- a/test/spec_mock.rb
+++ b/test/spec_mock.rb
@@ -167,7 +167,7 @@ describe Rack::MockRequest do
     env = YAML.load(res.body)
     env["REQUEST_METHOD"].must_equal "GET"
     env["QUERY_STRING"].must_include "baz=2"
-    env["QUERY_STRING"].must_include "foo[bar]=1"
+    env["QUERY_STRING"].must_include "foo%5Bbar%5D=1"
     env["PATH_INFO"].must_equal "/foo"
     env["mock.postdata"].must_equal ""
   end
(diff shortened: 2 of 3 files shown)


ioquatix:
Wow, I did not know about this.

I feel this is a major breaking change.

So the query string can't contain square brackets?

I need to review RFC...

ioquatix:
https://stackoverflow.com/questions/11490326/is-array-syntax-using-square-brackets-in-url-query-strings-valid

ioquatix:
https://url.spec.whatwg.org/#query-state implicitly seems to allow it.

ioquatix:
I feel like this could break tools like webmock. @bblimke WDYT?

matthewd:
I'm obviously biased, but while I'd generally prefer to follow a proper RFC, in this instance I'm inclined to defer to the whatwg pseudo-spec as a description of current behaviour-in-practice, and by implication what behaviour is reasonable to rely upon.

I think this substantially uglifies a commonly-used URL pattern (did I mention some bias? 😅) for strict conformance with a rule that's been widely ignored for years (e.g. via that SO link, we see that [browsers agree it doesn't get escaped](https://bugzilla.mozilla.org/show_bug.cgi?id=1152455#c6)).

Finally, at risk of sending the question to a forum where I have less voice: given the implementation, if this is something that should change, isn't it a matter for upstream `URI.encode_www_form_component`? ([related](https://bugs.ruby-lang.org/issues/12235), which links to https://github.com/whatwg/url/issues/112, which seems to ~explicitly support the current behaviour, by my reading)

bblimke:
@ioquatix thanks for mentioning me. I'm not concerned about webmock as this would only affect webmock test suite and not the behaviour of the webmock api, though I'm sure this would affect many other gems and apps.

I also didn't know about this issue. If I can have an opinion, I'd keep the current behaviour, since based on the linked threads it seems the community is used to the fact that these characters don't need to be escaped.
`x[y][][z]=1&x[y][][w]=a&x[y][][z]=2&x[y][][w]=3` also looks much nicer than
`x%5By%5D%5B%5D%5Bz%5D=1&x%5By%5D%5B%5D%5Bw%5D=a&x%5By%5D%5B%5D%5Bz%5D=2&x%5By%5D%5B%5D%5Bw%5D=3` ;)

@jeremyevans have you created the PR to comply with the RFC or was there another reason, e.g current behaviour caused some issues?

ioquatix:
I agree it's a tricky issue and normally I'd follow RFCs... that being said, this makes URLs super ugly for what is a really common case.

I think this is a case where the RFC should probably be updated. Or at least clarified. The questions I guess I'd be asking is, 1/ can the current way it works using square brackets actually cause problems and 2/ does the usage of square brackets actually prevent some other use case that we'd prefer instead?

In any case, it would be hard for the RFCs to back out of the informal standard that's been established.

One question I don't know the answer to, is when you have elements like `<input name="myarray[]" .../>`. Does the name attribute value need to be escaped too? Or is that up to the browser? And what do browsers do at the moment?

jeremyevans:
All browsers I tested escape `[` and `]` when submitting requests per the RFC.  Firefox displays the unescaped version in the address bar, but submits the escaped version.

That said, I don't feel strongly about this.  If we don't want to follow the RFC because it results in uglier URLs, we can close this and #792.  We should probably document that we purposely chose not to follow the RFC in this case, though.

tenderlove:
Lets document that we don't follow RFC.  I _really_ don't think this change is worth it.  I can imagine tons of tests breaking out there because suddenly `[` is escaped differently.  Being "correct" WRT the RFC doesn't buy us anything and also would cause a lot of pain.

PR #1501, opened 2020-01-19
Added support for SameSite=None cookie for 1.6

7a6163:
I have some projects with Rails 4.
Could you backport support of same site cookie is none for 1.6?

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -315,6 +315,8 @@ module Rack
           case value[:same_site]
           when false, nil
             nil
+          when :none, 'None', :None
+            '; SameSite=None'.freeze
           when :lax, 'Lax', :Lax
             '; SameSite=Lax'.freeze
           when true, :strict, 'Strict', :Strict
--- a/test/spec_response.rb
+++ b/test/spec_response.rb
@@ -97,6 +97,24 @@ describe Rack::Response do
     response["Set-Cookie"].should.equal "foo=bar"
   end
 
+  it "can set SameSite cookies with symbol value :none" do
+    response = Rack::Response.new
+    response.set_cookie "foo", { value: "bar", same_site: :none }
+    response["Set-Cookie"].must_equal "foo=bar; SameSite=None"
+  end
+
+  it "can set SameSite cookies with symbol value :None" do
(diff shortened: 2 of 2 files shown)


sharang-d:
I would love to have this merged in!

rafaelfranca:
Unfortunately not. Rack 1.6 as Rails 4 don't receive any changes that are not related to security.

willianveiga:
As a workaround you can create a `config/initializers/rake.rb` file with the following content:

```
# frozen_string_literal: true

require 'rack/utils'

module Rack::Utils
  def set_cookie_header!(header, key, value)
    case value
    when Hash
      domain  = "; domain="  + value[:domain]       if value[:domain]
      path    = "; path="    + value[:path]         if value[:path]
      max_age = "; max-age=" + value[:max_age].to_s if value[:max_age]
      expires = "; expires=" +
        rfc2822(value[:expires].clone.gmtime) if value[:expires]
      secure = "; secure"  if value[:secure]
      httponly = "; HttpOnly" if (value.key?(:httponly) ? value[:httponly] : value[:http_only])
      same_site =
        case value[:same_site]
        when false, nil
          nil
        when :none, 'None', :None
          '; SameSite=None'
        when :lax, 'Lax', :Lax
          '; SameSite=Lax'.freeze
        when true, :strict, 'Strict', :Strict
          '; SameSite=Strict'.freeze
        when true, :strict, 'Strict', :Strict
          '; SameSite=Strict'.freeze
        else
          raise ArgumentError, "Invalid SameSite value: #{value[:same_site].inspect}"
        end
      value = value[:value]
    end
    value = [value] unless Array === value
    cookie = escape(key) + "=" +
      value.map { |v| escape v }.join("&") +
      "#{domain}#{path}#{max_age}#{expires}#{secure}#{httponly}#{same_site}"
    case header["Set-Cookie"]
    when nil, ''
      header["Set-Cookie"] = cookie
    when String
      header["Set-Cookie"] = [header["Set-Cookie"], cookie].join("\n")
    when Array
      header["Set-Cookie"] = (header["Set-Cookie"] + [cookie]).join("\n")
    end
    nil
  end
  module_function :set_cookie_header!
end
```

PR #1505, opened 2020-01-22
Fix `use` with kwargs

kamipo:
Rails also has similar issue and is already fixed in
https://github.com/rails/rails/pull/38091.


ioquatix:
Thanks!

PR #1509, opened 2020-01-22
Set Content-Length response header even for bodies not responding to to_ary

### LINKED ISSUE #734: Rack::BodyProxy with Rack::ContentLength middleware

jeremyevans:
This to_ary check was previously there to work around an infinite
loop issue due to Rack::Response#to_ary being defined.
Rack::Response#to_ary has been removed, so this check can be removed,
and the middleware can correctly calculate Content-Length for all
valid rack responses.

Fixes #734

### DIFF
--- a/lib/rack/content_length.rb
+++ b/lib/rack/content_length.rb
@@ -19,8 +19,7 @@ module Rack
 
       if !STATUS_WITH_NO_ENTITY_BODY.key?(status.to_i) &&
          !headers[CONTENT_LENGTH] &&
-         !headers[TRANSFER_ENCODING] &&
-         body.respond_to?(:to_ary)
+         !headers[TRANSFER_ENCODING]
 
         obody = body
         body, length = [], 0
--- a/test/spec_content_length.rb
+++ b/test/spec_content_length.rb
@@ -20,13 +20,13 @@ describe Rack::ContentLength do
     response[1]['Content-Length'].must_equal '13'
   end
 
-  it "not set Content-Length on variable length bodies" do
+  it "set Content-Length on variable length bodies" do
     body = lambda { "Hello World!" }
     def body.each ; yield call ; end
 
     app = lambda { |env| [200, { 'Content-Type' => 'text/plain' }, body] }
     response = content_length(app).call(request)
-    response[1]['Content-Length'].must_be_nil
+    response[1]['Content-Length'].must_equal '12'
   end
 
   it "not change Content-Length if it is already set" do


wjordan:
The `to_ary` check in `ContentLength` middleware removed by this PR was _not_ previously there to work around the infinite-loop issue related to `Rack::Response` (#419, which was ultimately fixed by the API change in #1453).

That infinite-loop issue was only indirectly related to `ContentLength` because the initial workaround for that issue (5b251a986962f3577e14c01a698fab9dd21383ba) caused _any_ object wrapped by `BodyProxy` to no longer respond to `#to_ary`, which in turn caused `ContentLength` middleware to incorrectly skip any body responses wrapped by `BodyProxy`.

In fact, the `to_ary` check was originally added (7a0476be9289b6e38dbc27946168d41894439f42) in order for `ContentLength` to only eagerly-load Rack bodies that were Arrays (or objects that could be coerced into Arrays), and pass through any other potentially-streaming bodies without blocking.

See #166 and 94952077baa9cb6a8712064baab51eba6f43adb1 for further related discussion.

This decade-old discussion on how best to handle streaming response bodies in the Rack interface while also supporting full-response-reading middleware such as `ContentLength` is still ongoing (e.g., #1745).

jeremyevans:
@wjordan Thanks for the providing the additional context.  I guess I was misled by #734 which said the code was added to address #419 (the infinite loop issue).

I'm not sure if we want to revert the behavior.  We'd have to come up with another way to address #734 if we did revert it.

If we decide to treat some bodies differently than other bodies, than we may want to revisit this.  However, we should probably make a consistent decision that is applied across all middleware that ship with Rack.  Just checking for `to_ary` doesn't seem to be a good choice for that. Personally, I think the current behavior is fine, where users can set `Transfer-Encoding` if they don't want want the `Content-Length` calculated.  However, I don't use the middleware personally and I don't have strong feelings for how it should work.

ioquatix:
No one should be setting the `transfer-encoding` header since it's invalid for HTTP/2.

PR #1511, opened 2020-01-23
Require rack.session entry to respond to to_hash and return Hash instance

### LINKED ISSUE #461: SPEC does not require `to_hash` on `rack.session` interface

jeremyevans:
This is already required by rack's session support, so it should
be in SPEC/Lint.

Add some more rack.session Lint tests while here.

Fixes #461

### DIFF
--- a/SPEC
+++ b/SPEC
@@ -104,7 +104,7 @@ be implemented by the server.
                         fetch(key, default = nil) (aliased as []);
                         delete(key);
                         clear;
-                        to_hash (returning Hash instance);
+                        to_hash (returning unfrozen Hash instance);
 <tt>rack.logger</tt>:: A common object interface for logging messages.
                        The object must implement:
                         info(message, &block)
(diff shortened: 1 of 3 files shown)


ioquatix:
Do you want to indicate mutability on the return value?

jeremyevans:
@ioquatix That seems like a good idea to me, I'll add a commit for that.

PR #1516, opened 2020-01-23
Do more exact matching of domain and path when deleting cookies

jeremyevans:
Also, correctly handle deleting with both a domain and path
provided.

Fixes #1234


jeremyevans:
I think these changes are correct, but I would appreciate a review by another maintainer before merging.

PR #1517, opened 2020-01-23
Document `Rack::Response#initialize` and `#finish`.

ioquatix:
In response to https://github.com/rack/rack/pull/1434

I used yard style documentation but maybe it's wrong.

@jeremyevans @tenderlove what are your feelings on improving the documentation, e.g. using YARD style syntax, or sticking with RDoc (which I feel is a bit more clunky).

### DIFF
--- a/lib/rack/response.rb
+++ b/lib/rack/response.rb
@@ -19,22 +19,27 @@ module Rack
   # +write+ are synchronous with the Rack response.
   #
   # Your application's +call+ should end returning Response#finish.
-
   class Response
     def self.[](status, headers, body)
       self.new(body, status, headers)
     end
 
-    attr_accessor :length, :status, :body
-    attr_reader :header
-    alias headers header
-
     CHUNKED = 'chunked'
     STATUS_WITH_NO_ENTITY_BODY = Utils::STATUS_WITH_NO_ENTITY_BODY
 
-    def initialize(body = nil, status = 200, header = {})
+    # Initialize the response object with the specified body, status
+    # and headers.
+    #
+    # @param body [nil | #each | #to_str] the response body.
+    # @param status [Integer] the integer status as defined by the
+    # HTTP protocol RFCs.
+    # @param headers [#each] a list of key-value header pairs which
+    # conform to the HTTP protocol RFCs.
+    #
+    # Providing a body which responds to #to_str is legacy behaviour.
+    def initialize(body = nil, status = 200, headers = {})
       @status = status.to_i
-      @header = Utils::HeaderHash.new(header)
+      @headers = Utils::HeaderHash.new(headers)
 
       @writer = self.method(:append)
 
@@ -56,6 +61,12 @@ module Rack
       yield self if block_given?
     end
 
+    attr_accessor :length, :status, :body
+    attr_reader :headers
+
+    # @deprecated Use {#headers} instead.
+    alias header headers
+
     def redirect(target, status = 302)
       self.status = status
       self.location = target
@@ -65,6 +76,8 @@ module Rack
       CHUNKED == get_header(TRANSFER_ENCODING)
     end
 
+    # @return [Array] a 3-tuple suitable of `[status, headers, body]`
+    # which is a suitable response from the middleware `#call(env)` method.
     def finish(&block)
       if STATUS_WITH_NO_ENTITY_BODY[status.to_i]
         delete_header CONTENT_TYPE


ioquatix:
cc @AlexWayfer

lib/rack/response.rb
```diff
@@ -56,6 +61,12 @@ def initialize(body = nil, status = 200, header = {})
       yield self if block_given?
     end
 
+    attr_accessor :length, :status, :body
+    attr_reader :headers
+
+    # Deprecated.
```
rafaelfranca: If we are using yard, should this be `@deprecated`?

ioquatix: Yes.

ioquatix:
Okay, so with this commit, I think we are committed to using YARD going forward, and refactoring existing documentation is now on the cards for PRs.

jeremyevans:
I much prefer sticking with standard RDoc syntax.  I don't want to have to use a tool that isn't shipped with Ruby to maintain rack.

ioquatix:
@jeremyevans I am easy for either option, but what would this PR look like using RDoc? My understanding is that RDoc is pretty limited.

jeremyevans:
RDoc doesn't have explicit support for parameter names, so usually you would just document what each argument/option does in text form.  Something like (modulo whitespace):

```
    # Arguments:
    # body :: The response body.  Should be an object that responds to each and yields strings, a string instance, or nil.
    # status :: The integer status as defined by the HTTP protocol RFCs.
    # header :: A list of key-value header pairs which conform to the HTTP protocol RFCs.
```

lib/rack/response.rb
```diff
@@ -19,22 +19,27 @@ module Rack
...
-    alias headers header
-
     CHUNKED = 'chunked'
     STATUS_WITH_NO_ENTITY_BODY = Utils::STATUS_WITH_NO_ENTITY_BODY
 
-    def initialize(body = nil, status = 200, header = {})
+    # Initialize the response object with the specified body, status
+    # and headers.
+    #
+    # @param body [nil | #each | #to_str] the response body.
```
AlexWayfer: As I see in https://rubydoc.info/gems/yard/file/README.md, there should be `,` instead of `|`, but I'm not sure.

lib/rack/response.rb
```diff
@@ -56,6 +61,12 @@ def initialize(body = nil, status = 200, header = {})
       yield self if block_given?
     end
 
+    attr_accessor :length, :status, :body
+    attr_reader :headers
+
+    # @deprecated Use {#headers} instead.
+    alias header headers
```
AlexWayfer: I think this all should place above, as it was, and as Ruby Style (Community) Guide suggest here: https://rubystyle.guide/#consistent-classes

lib/rack/response.rb
```diff
@@ -65,6 +76,8 @@ def chunked?
       CHUNKED == get_header(TRANSFER_ENCODING)
     end
 
+    # @return [Array] a 3-tuple suitable of `[status, headers, body]`
+    # which is a suitable response from the middleware `#call(env)` method.
```
AlexWayfer: This is not true, as I understand, because this method is still returning `self` instead of `body` in the `block_given?` case.

ioquatix: And `self` responds to `#each` which is the requirement for `body`.

lib/rack/response.rb
ioquatix: Okay, if that's how it's being handled, I'll investigate & revert that change. I always like to have the `#initialize` first, just from the POV that it defines the shape of the class in a pretty major way.

lib/rack/response.rb
ioquatix: Fixed! Thanks.

lib/rack/response.rb
ioquatix: Reverted.

lib/rack/response.rb
AlexWayfer: `Response#each` doesn't seem regular…

If you're calling `#finish` with a block, in which violating:

> and must only yield String values

(you can to do so), then SPEC is broken.

I may make mistakes, but… it's hard.

Especially at the end of `body`'s SPEC:

> The Body commonly is an Array of Strings, the application instance itself, or a File-like object.

What of this is a `Response` instance? I know, "commonly", but there is a set…

lib/rack/response.rb
AlexWayfer: Maybe community want to separate state and behavior in classes. And `attr`s are more like state, like constants, and `initialize` is a method as behavior. I'm not sure. Or just instance-method-to-instance-method, and (like-)class-method to class-method.

lib/rack/response.rb
ioquatix: It only needs to respond to `#each` and yield strings. `#finish` with a block yields strings that the user provides. If the user doesn't provide strings, it's fault of user.

ioquatix: The correct way to read response body is something like this:

```ruby
status, headers, body = app.call(env)
buffer = String.new
body.each{|chunk| buffer << chunk}
```

I think we should add helper for this in `MockResponse` which has very weird semantics for `#body` and has caused a lot of confusion too.

AlexWayfer: If you're writing something like…

```ruby
result = response.finish { |string| string.upcase! }
result.last.find(&:nil?)
```

You'll get `true` in some (possible unpredictable) cases.

So… it's user's fault to write block (without direct documentation) which returns `nil` sometimes (maybe with one-lined `if`)? I can agree with it, but it looks not enough strict for me.

ioquatix: `Rack::Lint` is the place to detect issues like this and it's done this way because it's slower to enforce checks in production in normal code path.

PR #1518, opened 2020-01-23
Response documentation

ioquatix:
Follow up to feedback from https://github.com/rack/rack/pull/1517

### DIFF
--- a/lib/rack/response.rb
+++ b/lib/rack/response.rb
@@ -30,7 +30,7 @@ module Rack
     # Initialize the response object with the specified body, status
     # and headers.
     #
-    # @param body [nil | #each | #to_str] the response body.
+    # @param body [nil, #each, #to_str] the response body.
     # @param status [Integer] the integer status as defined by the
     # HTTP protocol RFCs.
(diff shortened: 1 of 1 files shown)


ioquatix:
@AlexWayfer all issues should be resolved here.

AlexWayfer:
> @AlexWayfer all issues should be resolved here.

Wow, thank you for the work! :clap: 

I'm still not sure about `Response`'s instance in a `finish` result, but existing changes looks good.

ioquatix:
@AlexWayfer I'm happy to add more clarification as to the result, but maybe at this point do you want to think about what would best satisfy you and propose it here in a comment so I can understand more what we can do to improve it?

AlexWayfer:
> @AlexWayfer I'm happy to add more clarification as to the result, but maybe at this point do you want to think about what would best satisfy you and propose it here in a comment so I can understand more what we can do to improve it?

Thank you for this approach. I've created #1519 to do what I mean, want to see.

jeremyevans:
These changes look fine to merge to me.

PR #1519, opened 2020-01-23
Simplify and make more consistent `Response#finish`

AlexWayfer:
Reduce a number of types of the last element in the response array.

Remove `Response#each` at all.

PLEASE! Be careful.

Tests are passed, but I don't understand Rack API completely, I'm a newbie in this.

### DIFF
--- a/lib/rack/response.rb
+++ b/lib/rack/response.rb
@@ -87,24 +87,17 @@ module Rack
       else
         if block_given?
           @block = block
-          [status.to_i, header, self]
-        else
-          [status.to_i, header, @body]
+          response = self
+          @body.define_singleton_method(:each) do |&each_block|
+            block.call(response)
+            super(&each_block)
+          end
         end
+        [status.to_i, header, @body]
       end
     end
 
     alias to_a finish           # For *response
-
-    def each(&callback)
-      @body.each(&callback)
-      @buffered = true
-
-      if @block
-        @writer = callback
-        @block.call(self)
-      end
-    end
 
     # Append to body and update Content-Length.
     #
--- a/test/spec_response.rb
+++ b/test/spec_response.rb
@@ -66,6 +66,27 @@ describe Rack::Response do
     parts.must_equal ["foo", "bar", "baz"]
   end
 
+  it "calls finish block when body is iterated" do
+    response = Rack::Response.new
+
+    finished = false
+    _, _, body = response.finish do
+      finished = true
+      response.write "foo"
+      response.write "bar"
+      response.write "baz"
+    end
+
+    parts = []
+    finished.must_equal false
+    body.each do |part|
+      finished.must_equal true
+      parts << part
+    end
+
+    parts.must_equal ["foo", "bar", "baz"]
+  end
+
   it "can set and read headers" do
     response = Rack::Response.new
     response["Content-Type"].must_be_nil


jeremyevans:
While I like the idea of using the `@body` directly, it changes behavior when a callback is provided, as it changes when the callback is called.  With the current code, the block passed to finish is not called until the server iterates over the body (probably after the status and headers are written).  With this change, the block passed to finish is called before finish returns.

Here's a spec that passes currently but would fail with this change:

```ruby
  it "calls finish block when body is iterated" do
    response = Rack::Response.new

    finished = false
    _, _, body = response.finish do
      finished = true
      response.write "foo"
      response.write "bar"
      response.write "baz"
    end

    parts = []
    finished.must_equal false
    body.each do |part|
      finished.must_equal true
      parts << part
    end

    parts.must_equal ["foo", "bar", "baz"]
  end
```

I'm not sure how important the current order of operations is here, but I lean on the side of closing this unless another core team member objects.

AlexWayfer:
@jeremyevans thank you! I've added your test and now this PR passes it. Also, `Response#each` has been removed.

jeremyevans:
This approach breaks if `@body` is frozen.  Additionally, adding methods to specific objects at runtime is usually considered bad style.

To be honest, I don't think the current code causes a problem, so I'm not sure we should change it.  It would be better if `finish` didn't take a block, but that ship has sailed.

AlexWayfer:
> This approach breaks if `@body` is frozen.

Is it real? I can take a tests for it and try to pass.

> Additionally, adding methods to specific objects at runtime is usually considered bad style.

Yes, as when the output of public method is not consistent and not documented / tested (it's resolving in a neighbor issue).

> To be honest, I don't think the current code causes a problem

It already caused.

> so I'm not sure we should change it

I too, just to show my vision and try to simplify a roulette between types (classes), which changes in patch-versions.

jeremyevans:
Please describe what problem this approach solves, preferably with a failing test.

AlexWayfer:
> Please describe what problem this approach solves, preferably with a failing test.

You can see the chain: this PR ⇒ https://github.com/rack/rack/pull/1518#issuecomment-577933742 ⇒ https://github.com/rack/rack/pull/1517#discussion_r370398868 ⇒ https://github.com/rack/rack/pull/1434#issuecomment-577675657

So, the last is the original problem.

I heard that "this public interface is not documented, so it's your fault", but I want to prevent such errors in the future and simplify things.

jeremyevans:
#1434 was merged in the master branch, presumably fixing the issue described in it.  So I'll ask again, in slightly different terms, please describe what problem this approach solves compared to the current master branch, preferably with a failing test.  I get that the approach is simpler, and I like simpler, but I'm not sure it solves an actual problem.  I'm not saying it doesn't solve a problem, I just haven't been convinced it does solve a problem yet.

AlexWayfer:
> #1434 was merged in the master branch, presumably fixing the issue described in it. So I'll ask again, in slightly different terms, please describe what problem this approach solves compared to the current master branch, preferably with a failing test. I get that the approach is simpler, and I like simpler, but I'm not sure it solves an actual problem. I'm not saying it doesn't solve a problem, I just haven't been convinced it does solve a problem yet.

Oh… it's hard to explain in other words.

The test can looks like:

```ruby
results = []
combinations_of_finish_args.each { |args| results << response.finish(*args).map(&:class) }
results.uniq.should_equal 1
```

The problem: I'm the maintainer of the Rack-compatible (infamous, yeah) web-framework Flame.

And the dependencies-integration (Depfu) reported (lucky) about errors in Rack minor-update (I've locked major updates, like `~> 2.0`): https://github.com/AlexWayfer/flame/pull/114

And… I've fixed it in https://github.com/AlexWayfer/flame/pull/116, but it dangerous, and I want to avoid this in the future.

jeremyevans:
Honestly, that test seems brittle and unnecessary.  In Ruby, we generally don't care about what class an object is, only that the object responds to the methods we are sending it.  If the example you gave is the best reason to accept this PR, I think we should close it.

AlexWayfer:
At least, it has +6 & -13 lines in the `Response` source and unification of classes. 😬 

But… as you wish.

AlexWayfer:
Current (`master`) `Response` has `Array` or `Response` as the last element of `#finish` result.

Then all of them have `#each`, but the `Response#each` can return `nil` or result of block from `#finish`… 

So inconsistency.

jeremyevans:
`each` on the 3rd element is called for side effects (the strings it yields), not for the return value.  So it is inconsistent, but also inconsequential.

ioquatix:
This has been a good discussion but I think we can close this. As long as body repsonds to each, it's clearly defined as part of the spec. I don't think we can or want to change assumptions about how this works.

ioquatix:
What I think would make sense is improving documentation around this so there is less confusion.

PR #1528, opened 2020-01-27
Only use ::Rack::RegexpExtensions on Ruby 2.3

jeremyevans:
Ruby 2.4+ support Regexp#match? and does not require loading a
refinement.

### DIFF
--- a/lib/rack/deflater.rb
+++ b/lib/rack/deflater.rb
@@ -19,7 +19,7 @@ module Rack
   # directive of 'no-transform' is present, or when the response status
   # code is one that doesn't allow an entity body.
   class Deflater
-    using ::Rack::RegexpExtensions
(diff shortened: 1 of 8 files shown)


rafaelfranca:
The refinement itself don't do anything in 2.4. Is there any overhead on calling `using` in a module that does nothing?

jeremyevans:
Yes, there is overhead in CRuby, and I think also in JRuby, though someone with more JRuby experience could probably confirm.

rafaelfranca:
👍

ioquatix:
We will probably drop Ruby 2.3 in rack 2.3/3.0 depending on how the timeline works out. It's far out there in terms of EOL, and even 2.4 is almost EOL.

PR #1529, opened 2020-01-27
DRY up tests using the test helper

jeremyevans:
Rely on autoload to load the parts being tested, which also tests the autoloads are working.

Also, add a few autoloads that were missing, for consistency.

I much prefer this style of testing.  Having the necessary setup code in one place makes it easier to make global changes when testing, and add things such as code coverage.


jeremyevans:
I'd appreciate a review of this by another core team member.  It's a fairly significant change, mostly to the tests, but I'm not sure there is agreement in the core team for moving in this direction.

lib/rack/reloader.rb
```diff
@@ -24,7 +22,7 @@ module Rack
   # It is performing a check/reload cycle at the start of every request, but
   # also respects a cool down time, during which nothing will be done.
   class Reloader
-    using ::Rack::RegexpExtensions if RUBY_VERSION < '2.4'
+    (require_relative 'core_ext/regexp'; using ::Rack::RegexpExtensions) if RUBY_VERSION < '2.4'
```
ioquatix: For the next release of Rack, why don't we drop < 2.4? @tenderlove @jeremyevans

ioquatix:
It looks like a good change to me.

lib/rack/reloader.rb
jeremyevans: Not sure when we plan to release Rack 2.2, but that sounds OK to me.

PR #1538, opened 2020-01-30
Fix `Response` methods (`host`, `port`, etc.) for IPv6

AlexWayfer:
I'm not sure about IPv6 format, you can correct tests and I'll adopt the code.


lib/rack/request.rb
```diff
@@ -234,17 +234,19 @@ def xhr?
 
       def host_with_port
         port = self.port
-        port.nil? || port == DEFAULT_PORTS[scheme] ? host : "#{host}:#{port}"
+        if port.nil? || port == DEFAULT_PORTS[scheme]
+          host
+        else
+          host = self.host
+          # If host is IPv6
+          host = "[#{host}]" if host.include?('::')
```
jeremyevans: This appear wrong, as you can have valid IPv6 addresses without `::`.  If the host contains any colons at all, it should be considered an IPv6 address.

lib/rack/request.rb
```diff
@@ -501,23 +503,23 @@ def hostname
         end
       end
 
-      def strip_port(ip_address)
+      def strip_port(uri)
```
jeremyevans: `uri` is a misleading parameter name, it's really hostname + port.  As it doesn't appear that you are actually changing any logic in this method, I won't modify it at all.

AlexWayfer: > uri is a misleading parameter name, it's really hostname + port.

Maybe.

> As it doesn't appear that you are actually changing any logic in this method, I won't modify it at all.

Now `strip_port` used for such data as `example.org`, and it doesn't look like `ip_address`.

lib/rack/request.rb
AlexWayfer: Thank you, changed to `:`.

lib/rack/request.rb
AlexWayfer: I've renamed to `hostname`. But I can revert this at all, if you're sure, but, again: domains don't look like IP addresses.

jeremyevans: I think it should be reverted.  `hostname` is more accurate (not completely as hostnames don't have ports), but not worth making git blame more difficult.

AlexWayfer: OK, I've returned `ip_address` back. But, if you (Rack collaborators) will decide to rename it — I can.

PR #1543, opened 2020-02-01
Allow passing through same_site option to session cookie when using Rack::Session::Cookie middleware

alexspeller:
Recently, rack [added support for SameSite=None cookies](https://github.com/rack/rack/pull/1358)

However there is currently no way to set these cookies using the Rack::Session::Cookie
middleware without monkeypatching.

This pull request allows setting the SameSite value either by direct, literal
passthrough to the add_cookie_to_header method, or by passing a callable.

The callable option is required because some browsers are incompatible with
some values of the header, so it needs to be [different per-browser](https://www.chromium.org/updates/same-site/incompatible-clients).

Static usage:

```ruby
use Rack::Session::Cookie, secret: 'supersecret', same_site: :none
```

Dynamic usage:

```ruby
use Rack::Session::Cookie,
  secret: 'supersecret',
  same_site: lambda { |req, res| MyUtilClass.correct_value_for_samesite(req.user_agent) }
```

### DIFF
--- a/test/spec_session_cookie.rb
+++ b/test/spec_session_cookie.rb
@@ -196,6 +196,23 @@ describe Rack::Session::Cookie do
     response.body.must_equal '{"counter"=>1}'
   end
 
+  it "passes through same_site option to session cookie" do
+    response = response_for(app: [incrementor, same_site: :none])
+    response["Set-Cookie"].must_include "SameSite=None"
+  end
+
+  it "allows using a lambda to specify same_site option, because some browsers require different settings" do
+    # Details of why this might need to be set dynamically:
+    # https://www.chromium.org/updates/same-site/incompatible-clients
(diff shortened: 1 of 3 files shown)


test/spec_session_cookie.rb
```diff
@@ -406,6 +423,11 @@ def call(env)
     response.body.must_equal '"foo"'
   end
 
+  it "Sets same_site option in cookie if session is " do
```
jeremyevans: Not sure this spec actually tests the same_site option.

alexspeller: Ah you're right, I left that spec in by accident after I moved stuff around to what I thought was a more sensible place in the file nearer to other more related functionality. I removed it now. Thanks for looking!

PR #1544, opened 2020-02-01
Prefer `instance_eval` rather than `eval`.

ioquatix:
The reason for this change is to support `# frozen_string_literal: true` at the top of the `config.ru` file. Right now it's ineffectual and generates a warning.

### DIFF
--- a/lib/rack/builder.rb
+++ b/lib/rack/builder.rb
@@ -109,8 +109,11 @@ module Rack
     # Evaluate the given +builder_script+ string in the context of
     # a Rack::Builder block, returning a Rack application.
     def self.new_from_string(builder_script, file = "(rackup)")
-      eval "Rack::Builder.new {\n" + builder_script + "\n}.to_app",
-        TOPLEVEL_BINDING, file, 0
+      builder = self.new
+
+      builder.instance_eval(builder_script, file)
+
+      return builder.to_app
     end
 
     # Initialize a new Rack::Builder instance.  +default_app+ specifies the


ioquatix:
@tenderlove @jeremyevans WDYT?

Also, does anyone ever use the server options?

jeremyevans:
This probably would break constant definition as well as method definition inside the script, that's the reason that `TOPLEVEL_BINDING` is used.  We should probably add a spec for both of those cases.

ioquatix:
Should we allow constants and methods to be defined in config.ru?

jeremyevans:
I think we should.  Not doing so is likely to break quite a few config.ru files.

ioquatix:
How should we handle `# frozen_string_literal: true` at the top of the file?

jeremyevans:
We could scan for it and insert it at the top of the evaled script, if we want to support that.

ioquatix:
https://bugs.ruby-lang.org/issues/16602

I asked here too.

I already thought about moving comments around, but it seems messy.

I'd like to know what @tenderlove thinks.

I think we need to decide whether to use `eval` or `instance_eval`. Clearly what we are doing with `eval` is a bit of a hack. But I can understand why we do it and I guess it will break behaviour if people are expecting to be in the global scope.

That being said, defining global constants/code in `config.ru` just seems like a horrible idea in general, so maybe it's not bad if we aim to change that?

ioquatix:
Matz already said he won't make frozen string literals the default in Ruby 3 so I believe we need to support it.

eregon:
How about scanning for magic encoding comments at the start of builder_script, and then just copy those before the `Rack::Builder.new {\n`?
Even if there was `eval(str, frozen_string_literal: true)`, you'd still want to enable only if the user did in `config.ru`, otherwise you would break scripts assuming mutable string literals.

For context, what kind of code do you expect in `config.ru` that would make a significant difference with `frozen_string_literal`?

eregon:
Maybe it would be useful to be able to do something like `TOPLEVEL_BINDING.with_self(Rack::Builder.new)`.
Not sure if there is a way to do that currently.

I think this should work and be compatible without requiring to scan magic comments:
```ruby
    def self.new_from_string(builder_script, file = "(rackup)")
      binding_with_builder = TOPLEVEL_BINDING.eval("Rack::Builder.new { binding }")
      eval builder_script, binding_with_builder, file, 0
      binding_with_builder.eval("to_app")
    end
```

jeremyevans:
I'm strongly against changing `eval` to `instance_eval` in this case, as I think it will break tons of code, and the advantage of supporting frozen string literals is not even close to worth it, especially considering there is a simple work around for it.

Note the issue is not just defining global constants, but reopening existing classes:

```ruby
class A; end
class B; end
a = A.new
a.instance_eval("class B; def c; 3 end; end")
B.new.c # NoMethodError
a.singleton_class::B.new.c # => 3
```

eregon:
@jeremyevans My approach still uses `eval`, so I think it's fully compatible, what do you think of it?

jeremyevans:
@eregon It's the same approach I mentioned earlier in  https://github.com/rack/rack/pull/1544#issuecomment-581108563, so yes, I am in favor.

eregon:
@jeremyevans No, I mean https://github.com/rack/rack/pull/1544#issuecomment-581123830. That does not require scanning and I think is fully compatible.

jeremyevans:
@eregon  Sorry, I thought you were referring to your earlier comment (https://github.com/rack/rack/pull/1544#issuecomment-581123314). The double-eval approach you demonstrated should be fine.  What are your thoughts on that @ioquatix ?

ioquatix:
@eregon you should take the credit for that genius contraption. Can you make a PR?

eregon:
Here you go: https://github.com/rack/rack/pull/1545

PR #1545, opened 2020-02-02
Support magic comments in .ru files

eregon:
* Such as `# frozen_string_literal: true`.

This seems actually like a relevant bug fix because many files under `example` and `test` use `# frozen_string_literal: true` but actually it has not effect before this fix.

From the idea in https://github.com/rack/rack/pull/1544#issuecomment-581123830
cc @ioquatix @jeremyevans

### DIFF
--- a/lib/rack/builder.rb
+++ b/lib/rack/builder.rb
@@ -109,8 +109,11 @@ module Rack
     # Evaluate the given +builder_script+ string in the context of
     # a Rack::Builder block, returning a Rack application.
     def self.new_from_string(builder_script, file = "(rackup)")
-      eval "Rack::Builder.new {\n" + builder_script + "\n}.to_app",
-        TOPLEVEL_BINDING, file, 0
+      # We want to build a variant of TOPLEVEL_BINDING with self as a Rack::Builder instance.
+      # We cannot use instance_eval(String) as that would resolve constants differently.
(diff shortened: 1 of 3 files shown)


ioquatix:
Another reason why we should have `--warnings` on by default during test runs.

ioquatix:
Thanks for your awesome implementation! It's quite possibly the most awesome thing in Ruby I've seen in several months at least.

eregon:
Thank you for the kind words :)

shime:
I'm writing a deep dive post about `rackup` and I have to say, this is the most amazing piece of code I've stumbled upon. Pure wizardry.

I don't fully understand why `TOPLEVEL_BINDING` is necessary here, though. What would be the effect of not having it?

```ruby
# Why not something like this?
def self.new_from_string(builder_script, file = "(rackup)")
  bind, builder = Rack::Builder.new.instance_eval { [binding, self] }
  eval builder_script, bind, file
  builder.to_app
end
```

eregon:
That would inherit the `Module.nesting`, which is `[Rack::Builder, Rack]` in that case. That would imply a `.ru` file with `p Builder` would print `Rack::Builder` and so defining your own building class would actually instead reopen Rack's Builder class which is most likely unintended.

`TOPLEVEL_BINDING` is used to preserve an empty `Module.nesting` just like regular top-level code in a file.

eregon:
Some more details in my tweet: https://twitter.com/eregontp/status/1232035098570215424

shime:
Thank you! 🙏

PR #1547, opened 2020-02-03
Add support for SameSite cookie to 1.4.7

ashawley:
We have a Rails 3 app that is still using Rack 1.4.7 that needs `SameSite` cookie flag.

This is not a security fix, but there was a significant [policy change in Chrome 80](https://blog.chromium.org/2019/10/developers-get-ready-for-new.html), released 4 February 2020, that requires this capability with cookie flags in applications that are third-party resources.

Backports of #1033, #1051 and #1358.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,15 @@
+dist: trusty
+language: ruby
 before_install: sudo apt-get install lighttpd libfcgi-dev libmemcache-dev memcached
 install:
   - gem env version | grep '^\(2\|1.\(8\|9\|[0-9][0-9]\)\)' || gem update --system
-  - gem install --conservative rake
+  - gem install rake -v '~> 0.9.6'
   - rake deps
 script: rake ci
 rvm:
   - 1.8.7
   - 1.9.2
   - 1.9.3
-  - rbx-2.0
-  - jruby
+  - 2.0.0
+  - jruby-1.7
   - ree
-branches:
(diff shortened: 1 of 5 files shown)


ashawley:
According to the security policy, 1.4 is no longer a supported release, so no new versions will be released.  However, certain fixes could be backported to git.  Since github can be used as a source in bundler, I figured it's worth sharing a low-risk improvement like this.  There may be a need for this by others.

ashawley:
Also includes

1. incrementing the version to 1.4.7.1,
2. fixing the Travis build and 
3. using #476 to fix a unit test on Ruby 2.0.

jeremyevans:
Sorry, not supported means not supported.  I need to update security policy, now that 2.1 has released, but even the current security policy doesn't list 1.4 as supported.

I recommend you backport the commits to your own fork and then use that fork.

ashawley:
I'm disappointed to hear that, and others will be as well.  Hopefully, you'll reconsider.

If not, yes, please update your security policy.  I read it that git pulls would be accepted.  Below was the section I was reading.  It doesn't say "not supported means not supported".

> _When a release series is no longer supported, it’s your own responsibility to deal with bugs and security issues. We may provide back-ports of the fixes and publish them to git, however there will be no new versions released. If you are not comfortable maintaining your own versions, you should upgrade to a supported version._

jeremyevans:
@ashawley I'm proposing that we remove that from the security policy in #1548.

PR #1549, opened 2020-02-03
Memoize header hash

ioquatix:
The motivation/measurements/benchmarks are in `benchmark/header-hash/README.md`.

Fixes the main performance issues outlined by https://github.com/rack/rack/issues/738

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -405,7 +405,7 @@ module Rack
     #
     # @api private
     class HeaderHash < Hash # :nodoc:
-      def self.[] headers
+      def self.[](headers)
         if headers.is_a?(HeaderHash)
           return headers
         else
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -405,6 +405,14 @@ module Rack
     #
     # @api private
     class HeaderHash < Hash # :nodoc:
+      def self.[] headers
+        if headers.is_a?(HeaderHash)
+          return headers
+        else
+          return self.new(headers)
+        end
+      end
+
       def initialize(hash = {})
         super()
         @names = {}
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -406,7 +406,7 @@ module Rack
     # @api private
     class HeaderHash < Hash # :nodoc:
       def self.[](headers)
-        if headers.is_a?(HeaderHash)
+        if headers.is_a?(HeaderHash) && !headers.frozen?
           return headers
         else
           return self.new(headers)
--- a/test/spec_utils.rb
+++ b/test/spec_utils.rb
@@ -727,6 +727,35 @@ describe Rack::Utils::HeaderHash do
     h['foo'].must_be_nil
     h.wont_include 'foo'
   end
+
+  it "uses memoized header hash" do
+    env = {}
+    headers = Rack::Utils::HeaderHash.new({ 'content-type' => "text/plain", "content-length" => "3" })
+
+    app = lambda do |env|
+      [200, headers, []]
+    end
+
+    app = Rack::ContentLength.new(app)
+
+    response = app.call(env)
+    assert_same response[1], headers
+  end
+
+	it "duplicates header hash" do
+    env = {}
+    headers = Rack::Utils::HeaderHash.new({ 'content-type' => "text/plain", "content-length" => "3" })
+		headers.freeze
+
+    app = lambda do |env|
+      [200, headers, []]
+    end
+
+    app = Rack::ContentLength.new(app)
+
+    response = app.call(env)
+    refute_same response[1], headers
+  end
 end
 
 describe Rack::Utils::Context do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -30,6 +30,7 @@ All notable changes to this project will be documented in this file. For info on
 - `Files` handling of range requests no longer return a body that supports `to_path`, to ensure range requests are handled correctly. ([@jeremyevans](https://github.com/jeremyevans))
 - `Multipart::Generator` only includes `Content-Length` for files with paths, and `Content-Disposition` `filename` if the `UploadedFile` instance has one. ([@jeremyevans](https://github.com/jeremyevans))
 - `Request#ssl?` is true for the `wss` scheme (secure websockets). ([@jeremyevans](https://github.com/jeremyevans))
+- `Rack::HeaderHash` is memoized by default. ([#1549](https://github.com/rack/rack/pull/1549), [@ioquatix](https://github.com/ioquatix))
 
 ### Removed
 
--- a/lib/rack/chunked.rb
+++ b/lib/rack/chunked.rb
@@ -69,7 +69,7 @@ module Rack
 
     def call(env)
       status, headers, body = @app.call(env)
-      headers = HeaderHash.new(headers)
+      headers = HeaderHash[headers]
 
       if ! chunkable_version?(env[SERVER_PROTOCOL]) ||
          STATUS_WITH_NO_ENTITY_BODY.key?(status.to_i) ||
--- a/lib/rack/common_logger.rb
+++ b/lib/rack/common_logger.rb
@@ -30,10 +30,10 @@ module Rack
 
     def call(env)
       began_at = Utils.clock_time
-      status, header, body = @app.call(env)
-      header = Utils::HeaderHash.new(header)
-      body = BodyProxy.new(body) { log(env, status, header, began_at) }
-      [status, header, body]
+      status, headers, body = @app.call(env)
+      headers = Utils::HeaderHash[headers]
+      body = BodyProxy.new(body) { log(env, status, headers, began_at) }
+      [status, headers, body]
     end
 
     private
--- a/lib/rack/conditional_get.rb
+++ b/lib/rack/conditional_get.rb
@@ -23,7 +23,7 @@ module Rack
       case env[REQUEST_METHOD]
       when "GET", "HEAD"
         status, headers, body = @app.call(env)
-        headers = Utils::HeaderHash.new(headers)
+        headers = Utils::HeaderHash[headers]
         if status == 200 && fresh?(env, headers)
           status = 304
           headers.delete(CONTENT_TYPE)
--- a/lib/rack/content_length.rb
+++ b/lib/rack/content_length.rb
@@ -12,7 +12,7 @@ module Rack
 
     def call(env)
       status, headers, body = @app.call(env)
-      headers = HeaderHash.new(headers)
+      headers = HeaderHash[headers]
 
       if !STATUS_WITH_NO_ENTITY_BODY.key?(status.to_i) &&
          !headers[CONTENT_LENGTH] &&
--- a/lib/rack/content_type.rb
+++ b/lib/rack/content_type.rb
@@ -17,7 +17,7 @@ module Rack
 
     def call(env)
       status, headers, body = @app.call(env)
-      headers = Utils::HeaderHash.new(headers)
+      headers = Utils::HeaderHash[headers]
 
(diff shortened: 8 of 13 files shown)


ioquatix:
https://github.com/rack/rack/issues/738

ioquatix:
cc @schneems

lib/rack/utils.rb
```diff
@@ -405,6 +405,14 @@ def context(env, app = @app)
     #
     # @api private
     class HeaderHash < Hash # :nodoc:
+      def self.[] headers
```
rafaelfranca:
```suggestion
      def self.[](headers)
```

tenderlove:
> This benchmarks are nice and an abbreviated version could go in the commit message, but I don't think we need them in the repository. Could you remove them and squash?

I agree with this, but can you put the benchmarks and results in this PR?  We can refer back to the PR to see the results, and the PR will be linked to from the merge commit (another reason I like backporting merge commits rather than individual cherry picking).

ioquatix:
I am happy to remove the benchmarks directory. However:

- It doesn't cost anyone anything.
- It provides a framework for future benchmarks so people don't have to reinvent the wheel (I tried to find the original benchmarks done by @schneems in #738 but it doesn't seem to exist anywhere).
- Assumptions based on benchmarks can be checked again in the future.

Putting the benchmark data into the GitHub PR ties it to GitHub. I often work offline and if I don't have all the data it's a pain.

So, unless there is a strong practical reason not to include the benchmarks in the repo, I'd prefer to leave it. However, I welcome further discussion. I'm easy either way.

jeremyevans:
It does cost something.  Everything added to the repository costs something and requires time/bandwidth to clone and store.  Even in 10 years where the data is no longer needed, it'll still be in the .git directory taking up space.

If you do want to keep the benchmark, keep just the benchmark and not the results/gemfiles.  That way people can use and learn from the benchmark without taking up much space, and it still allows rechecking assumptions later.  I would especially remove the gemfiles as we would want to test with with not just Falcon, but also Puma and Unicorn to judge the real world impact.

Also, I think the benchmark should be made more realistic.  You could consider worst case to be every middleware that Rack ships with that uses `HeaderHash`, and use that along with `Rack::Response`.  100 `Rack::ContentLength` middlewares doesn't tell us much about how this affects real world cases.  I don't even think Rails uses that many middlewares. :)

tenderlove:
It would be pretty cool if we had a benchmark suite and could keep track of the speeds over time.

ioquatix:
> It does cost something. Everything added to the repository costs something and requires time/bandwidth to clone and store. Even in 10 years where the data is no longer needed, it'll still be in the .git directory taking up space.

Understood. 

We could create a separate repo for benchmarks. But that is very cumbersome when they are tied to PRs.

I hesitate to say that "even in 10 years when the data is no longer needed". I wish the original benchmarks made by @schneems were committed or in a PR, so that I could actually validate the improvements against his original data.

> If you do want to keep the benchmark, keep just the benchmark and not the results/gemfiles. That way people can use and learn from the benchmark without taking up much space, and it still allows rechecking assumptions later. I would especially remove the gemfiles as we would want to test with with not just Falcon, but also Puma and Unicorn to judge the real world impact.

It sounds like you are bothered to include Falcon in a benchmark. I just used it because it's practical for the purpose of the benchmark. This benchmark is not affected by the server so much. We could use webrick or something else by default. But it didn't occur to me that someone would be unhappy about that, that I didn't use some other server or multiple servers. Making benchmarks reproducible is a good idea IMHO.

> Also, I think the benchmark should be made more realistic. You could consider worst case to be every middleware that Rack ships with that uses HeaderHash, and use that along with Rack::Response. 100 Rack::ContentLength middlewares doesn't tell us much about how this affects real world cases. I don't even think Rails uses that many middlewares. :)

I wish I could reproduce the original rails benchmark done by @schneems but that is lost to the sands of time because it was not committed or in any PR that I'm aware of. That's kind of the crux of the issue here and why I decided to include it. I believe any time we have a performance related PR, we should have a corresponding test or benchmark which is (reasonably) reproducible. This is pretty typical, e.g. CRuby.

ioquatix:
Just a few examples:

Async benchmarks to check assumptions, all fairly reproducible:

-  https://github.com/socketry/async/tree/master/benchmark

Useful for comparing assumptions across releases of Ruby, different implementations of Ruby.

Specs which include benchmarks:

https://github.com/socketry/async/blob/master/spec/async/performance_spec.rb

https://github.com/socketry/async/runs/420813752?check_suite_focus=true#step:5:66

There are others, but this is just one example which helps me catch regressions or identify performance issues over time.

It's also possible to do the specific regression testing on memory allocations in specs:

https://github.com/socketry/rspec-memory

honestly, something like the last one might be better than a benchmark, but these all work together to prevent and identify performance issues.

schneems:
I’m on mobile, but it looks like this used to be the behavior of HeaderHash.new, or at least it was when I made this PR https://github.com/rack/rack/pull/1204/files#diff-7c1a24d5b2fe58a6f925c7cacc6c55e7 I’m not able to see when or why it changed right now.


FWIW I extracted my benchmark code to derailed https://github.com/schneems/derailed_benchmarks#i-made-a-patch-to-to-rails-how-can-i-tell-if-it-made-my-rails-app-faster-and-test-for-statistical-significance you can also set the library to rack just like you can for Rails. It won’t show the change over time but will give an a/b perf comparison with some statistical analysis.

schneems:
> it looks like this used to be the behavior of HeaderHash.new, or at least it was when I made this PR

Looks like that logic was reverted by @jodosha in https://github.com/rack/rack/pull/1325

```
$ git log -S "HeaderHash ===" lib/rack/utils.rb
commit d671a0c3379bc05f17e3f670d943c6243c333df0
Author: Luca Guidi <me@lucaguidi.com>
Date:   Wed Jan 23 10:57:16 2019 +0100

    Ensure initialization of Utils::HeaderHash will dup given headers

commit e4c7c58597abaa3c0a8281df5f604da33d1304c3
Author: Luca Guidi <me@lucaguidi.com>
Date:   Fri Jan 4 16:13:24 2019 +0100

    Don't mutate given headers

commit cdf13618445fa7065c07d62b115907856d6d1095
Author: Eric Wong <normalperson@yhbt.net>
Date:   Thu Dec 10 21:33:53 2009 -0600

    HeaderHash.new avoids unnecessary object creation

    Creating a new HeaderHash is an O(n) operation in addition to
    the cost of allocating a new object.  When using multiple pieces
    of middleware, this can lead to unnecessary memory allocation
    and iteration overhead.   We now explicitly define the
    HeaderHash.new class method to return its original argument if
    it is already a HeaderHash to avoid repeating work.

    Signed-off-by: Joshua Peek <josh@joshpeek.com>
```

schneems:
Here's how I will test with Derailed and https://www.codetriage.com app (https://github.com/codetriage/codetriage)

```
$ git co schneems/rack-patch-speed-test
$ DERAILED_PATH_TO_LIBRARY="$(bundle show rack)" SHAS_TO_TEST="memoize-header-hash,59985351196ad348dbe53b799333ad24d8d64404" bundle exec derailed exec perf:library
```

The gist of how you perf test a library with derailed:

- Get an an app that uses the library (codetriage uses rack_
- Point the dependency to a new commit, I did this by modifying the gemfile in the `schneems/rack-patch-speed-test` branch of CodeTriage
- Tell derailed where the library you want to test lives `DERAILED_PATH_TO_LIBRARY="$(bundle show rack)" `
- Tell it which git refs you want to test against `SHAS_TO_TEST="memoize-header-hash,59985351196ad348dbe53b799333ad24d8d64404"` (this PR branch, and the commit of master)
- Then the actual command `bundle exec derailed exec perf:library`

Derailed will then boot the app in memory and hit it with 200 requests using rack-mock, this is done headlessly. It then toggles the git ref and hits that with 200 requests. It repeats this process 200 times. It will print out intermediate results along with statistical significance information as it runs. Ideally, I want histograms too, but that's not implemented yet https://github.com/schneems/derailed_benchmarks/pull/169 (and will require significant work).

I'm going to let this run a few times overnight and let you all know how it turns out.

ioquatix:
Thanks that sounds awesome looking forward to the results!

schneems:
**Big caveat** - I'm going to give the results of my tests, they are not THE results. They are one picture of results based off of how this patch behaves with CodeTriage. I've spent a lot of time trying to measure perf differences between code changes in my Rails patches. However, I do not have all the answers. Basically, use this data not as an answer, but rather as the basis for more questions.

**My results** I ran derailed against code triage 6 times last night and got statistically insignificant results for all six runs with a 95% confidence interval (derailed will report up to 99% significant if available). Some runs were faster some were slower. But we cannot say, with confidence, that the result was due to the change as opposed random chance on CodeTriage. Here's the results:

```
  FASTER 🚀🚀🚀 by:
    1.0096x [older/newer]
    0.9490% [(older - newer) / older * 100]

  FASTER 🚀🚀🚀 by:
    1.0093x [older/newer]
    0.9205% [(older - newer) / older * 100]

  SLOWER 🐢🐢🐢 by:
     0.9997x [older/newer]
    -0.0269% [(older - newer) / older * 100]

  FASTER 🚀🚀🚀 by:
    1.0212x [older/newer]
    2.0744% [(older - newer) / older * 100]

  SLOWER 🐢🐢🐢 by:
     0.9976x [older/newer]
    -0.2429% [(older - newer) / older * 100]

  SLOWER 🐢🐢🐢 by:
     0.9949x [older/newer]
    -0.5092% [(older - newer) / older * 100]
```

Again none of these results were significant, so they don't strictly _mean_ anything conclusive. I tried again this morning but so far over two runs, still no significant results. You cannot use these results to say "it is faster!" or "it is slower!" you cannot average or manipulate these results to mean anything, other than we cannot say with 95% confidence that the results are not the result of random chance.

That could mean that there is a change, but it's smaller than we are capable of measuring, or it could mean that there's effectively no measurable difference between the two branches (for this test). The app we're testing on for sure will make a difference. A "hello world" rack app would likely better be able to highlight a perf difference, but i'm not able to see one in CodeTriage.

I've attached the raw results in a zip file. If you want to chat more about the data, or derailed feel free to ping me. I'm in Nate's performance slack and the Rails basecamp. My email is visible on GitHub if you don't have access to either of those and want more synchronous communication.

[derailed-raw-results.zip](https://github.com/rack/rack/files/4154646/derailed-raw-results.zip)

ioquatix:
Nice work!

ioquatix:
I was reviewing the HeaderHash code, and realised that it's merging together headers using newlines. Which I subsequently have to revert in Falcon because no actual protocol supports this format:

https://github.com/socketry/falcon/blob/b8e9ca5aeaf748bec550a54a8bcdeb7f7c3d3fe7/lib/falcon/adapters/response.rb#L40-L50

For Rack 3.0, we could consider allowing headers to be an array of values and drop the newline format.

The practical benefits of this implementation seem limited, at least according to the data point provided by @schneems which I fully respect the statistical analysis he did, compared to the supplied micro benchmark.

That being said, in theory this does avoid memory allocations when there are lots of headers, so generally I think there are going to be use cases which benefit from this memoization.

Therefore, I'm going to tidy up this PR with some specs, remove the benchmarks, and merge. Then, with Rack 3.0 push, I will reintroduce benchmarks as part of the tests, and hopefully we can do so in other areas too.

test/spec_utils.rb
```diff
@@ -727,6 +727,35 @@ def initialize(*)
...
+      [200, headers, []]
+    end
+
+    app = Rack::ContentLength.new(app)
+
+    response = app.call(env)
+    assert_same response[1], headers
+  end
+
+	it "duplicates header hash" do
```
andrykonchin: Formatting issue?

ioquatix: Thanks, I'll fix it!

PR #1557, opened 2020-02-05
Prefer instance variables and use explicit returns.

ioquatix:


jeremyevans:
I'm OK with using instance variables instead of methods, but introducing explicit returns where they are not needed for early exit goes against standard ruby style.

jeremyevans:
My previous comment should probably have been phrased better.  I don't personally have a problem with explicit returns, even though I prefer implicit returns.  However, I think it makes little sense to have rubocop be nitpicky about whitespace and not explicit vs. implicit returns.  Thoughts on removing rubocop completely?  Alternatively, disabling the whitespace checks?

rafaelfranca:
I have the same opinion about explicit `returns`. They are not needed in Ruby. Most Ruby code doesn't us them, so I think it is better to keep following this.

> Thoughts on removing rubocop completely?

Yes. I'm against it very much. Consistency is good

> Alternatively, disabling the whitespace checks?

Why would you want to add extra whitespaces everywhere? If there is a tool to help us to not make something that we would consider a mistake, why should we not use it?

jeremyevans:
Why should we care about whitespace?  It causes no problems and I do not consider it a mistake, merely inconsequential.  Note that our rubocop configuration complains about extra whitespace at the end of lines, but requires unnecessary whitespace around braces and hash rockets.  So our rubocop configuration is inconsistent in how it treats whitespace depending on where the whitespace is.

I feel strongly that we should not care more about whitespace than about return style.  So we should either remove rubocop's complaining about whitespace, or we should have rubocop enforce return style (`Style/RedundantReturn`).

rafaelfranca:
> I feel strongly that we should not care more about whitespace than about return style.

We are in the same page then. I prefer to enforce return style.

ioquatix:
I'm fine with implicit return when the cyclomatic complexity of a function is 1, or the branch depth is 1. But when you start having several lines of code, the implicit return becomes a burden and a source of bugs. As in, was this an explicit part of what the function is supposed to do, or is it just a side effect of how the code was written?

Since it costs nothing, and you gain insight about what the function is supposed to do, I would argue pretty strongly about reverting the changes to this method.

jeremyevans:
I personally use implicit returns regardless of cyclomatic complexity, and explicit returns only for early exit.  I don't find it a burden or a source of bugs.  That being said, I'm also fine with explicit return style, even though I don't recommend it.  However, I'm not fine with enforcing nitpicky whitespace rules and not enforcing return style, since I consider the latter far more important.  If we don't want to enforce consistent return styles, we should not be enforcing nitpicky whitespace.

ioquatix:
> I personally use implicit returns regardless of cyclomatic complexity, and explicit returns only for early exit. I don't find it a burden or a source of bugs.

I've seen multiple bugs as a result of implicit return in non-trivial functions.

All it takes is someone to misunderstand a branch is part of an implicit return and add an extra statement. As a developer, you look at the code and wonder, is that implicit return a side effect or part of the contract of the function? It's not always obvious.

Because rack is so critical, explicit is definitely better than implicit unless it's so obvious that you could never make a mistake.

jeremyevans:
@ioquatix What are your thoughts on the current whitespace enforcement?

ioquatix:
- Consistency is good on a code base like this.
- Enforcement of consistency puts the burden on the person making the PR, not people trying to clean up later.
- Whitespace is kind of annoying but at least it's consistent.
- If you find whitespace rules annoying, maybe we can change them.

Personally, I think whitespace inside Hash literals to be pretty weird, e.g. `{ :foo => :bar }`. I prefer `{:foo => :bar}`.

So, I'd be okay with changing how we do it, but I still think we should probably enforce it if it saves time, w.r.t. keeping the code base consistent.

PR #1559, opened 2020-02-05
Remove `Rack::Files#response_body` as the implementation was broken.

ioquatix:
See https://github.com/rack/rack/pull/1153 for previous discussion.


chakrihacker:
@ioquatix when is the new release?

ioquatix:
This weekend hopefully.

PR #1561, opened 2020-02-05
Rework host/hostname/authority implementation.

### LINKED ISSUE #1560: IPv6 host normalization.
I reviewed the PR https://github.com/rack/rack/pull/1213 and it looks like we recently merged https://github.com/rack/rack/pull/1538 which prefered the non-square-brackets representation. However, this has not been released yet. So we could adjust it.

Based on the defintion of `URI#host`, I think we should probably prefer the square brackets representation... but I'm not sure what is canonical. I need to do some research.

_Originally posted by @ioquatix in https://github.com/rack/rack/pull/1213#issuecomment-582275811_

ioquatix:
Attempt to rework the host/hostname/authority methods to better follow the RFCs regarding host/authority, along with URI specifications for host/hostname. Fixes #1560.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file. For info on
 - `rack.session` request environment entry must respond to `to_hash` and return unfrozen Hash. ([@jeremyevans](https://github.com/jeremyevans))
 - Request environment cannot be frozen. ([@jeremyevans](https://github.com/jeremyevans))
 - CGI values in the request environment with non-ASCII characters must use ASCII-8BIT encoding. ([@jeremyevans](https://github.com/jeremyevans))
+- Improve SPEC/lint relating to SERVER_NAME, SERVER_PORT and HTTP_HOST. ([#1561](https://github.com/rack/rack/pull/1561), [@ioquatix](https://github.com/ioquatix))
 
 ### Added
 
@@ -37,7 +38,7 @@ All notable changes to this project will be documented in this file. For info on
 - `Rack::HeaderHash` is memoized by default. ([#1549](https://github.com/rack/rack/pull/1549), [@ioquatix](https://github.com/ioquatix))
 - `Rack::Directory` allow directory traversal inside root directory. ([#1417](https://github.com/rack/rack/pull/1417), [@ThomasSevestre](https://github.com/ThomasSevestre))
 - Sort encodings by server preference. ([#1184](https://github.com/rack/rack/pull/1184), [@ioquatix](https://github.com/ioquatix), [@wjordan](https://github.com/wjordan))
-- Rework host/hostname/authority implementation in `Rack::Request`. ([#1561](https://github.com/rack/rack/pull/1561), [@ioquatix](https://github.com/ioquatix))
+- Rework host/hostname/authority implementation in `Rack::Request`. `#host` and `#host_with_port` have been changed to correctly return IPv6 addresses formatted with square brackets, as defined by [RFC3986](https://tools.ietf.org/html/rfc3986#section-3.2.2). ([#1561](https://github.com/rack/rack/pull/1561), [@ioquatix](https://github.com/ioquatix))
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,6 +37,7 @@ All notable changes to this project will be documented in this file. For info on
 - `Rack::HeaderHash` is memoized by default. ([#1549](https://github.com/rack/rack/pull/1549), [@ioquatix](https://github.com/ioquatix))
 - `Rack::Directory` allow directory traversal inside root directory. ([#1417](https://github.com/rack/rack/pull/1417), [@ThomasSevestre](https://github.com/ThomasSevestre))
 - Sort encodings by server preference. ([#1184](https://github.com/rack/rack/pull/1184), [@ioquatix](https://github.com/ioquatix), [@wjordan](https://github.com/wjordan))
+- Rework host/hostname/authority implementation in `Rack::Request`. ([#1561](https://github.com/rack/rack/pull/1561), [@ioquatix](https://github.com/ioquatix))
 
 ### Removed
 
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -284,15 +284,13 @@ module Rack
       end
 
       ## The <tt>SERVER_NAME</tt> must be a valid authority as defined by RFC7540.
-      assert("env[SERVER_NAME] must be a valid host") do
-        server_name = env["SERVER_NAME"]
-        URI.parse("http://#{server_name}").host == server_name rescue false
+      assert("#{env[SERVER_NAME]} must be a valid authority") do
+        URI.parse("http://#{env[SERVER_NAME]}/") rescue false
       end
 
       ## The <tt>HTTP_HOST</tt> must be a valid authority as defined by RFC7540.
-      assert("env[HTTP_HOST] must be a valid host") do
-        http_host = env["HTTP_HOST"]
-        URI.parse("http://#{http_host}/").host == http_host rescue false
+      assert("#{env[HTTP_HOST]} must be a valid authority") do
+        URI.parse("http://#{env[HTTP_HOST]}/") rescue false
       end
 
       ## The environment must not contain the keys
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -271,12 +271,29 @@ module Rack
       ## accepted specifications and must not be used otherwise.
       ##
 
-      %w[REQUEST_METHOD SERVER_NAME SERVER_PORT
-         QUERY_STRING
+      %w[REQUEST_METHOD SERVER_NAME QUERY_STRING
          rack.version rack.input rack.errors
          rack.multithread rack.multiprocess rack.run_once].each { |header|
         assert("env missing required key #{header}") { env.include? header }
       }
+
+      ## The <tt>SERVER_PORT</tt> must be an integer if set.
+      assert("env[SERVER_PORT] is not an integer") do
+        server_port = env["SERVER_PORT"]
+        server_port.nil? || (Integer(server_port) rescue false)
+      end
+
+      ## The <tt>SERVER_NAME</tt> must be a valid authority as defined by RFC7540.
+      assert("env[SERVER_NAME] must be a valid host") do
+        server_name = env["SERVER_NAME"]
+        URI.parse("http://#{server_name}").host == server_name rescue false
+      end
+
+      ## The <tt>HTTP_HOST</tt> must be a valid authority as defined by RFC7540.
+      assert("env[HTTP_HOST] must be a valid host") do
+        http_host = env["HTTP_HOST"]
+        URI.parse("http://#{http_host}/").host == http_host rescue false
+      end
 
       ## The environment must not contain the keys
       ## <tt>HTTP_CONTENT_TYPE</tt> or <tt>HTTP_CONTENT_LENGTH</tt>
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -117,17 +117,19 @@ module Rack
       ##                         follows the <tt>?</tt>, if any. May be
       ##                         empty, but is always required!
 
-      ## <tt>SERVER_NAME</tt>, <tt>SERVER_PORT</tt>::
-      ##                        When combined with <tt>SCRIPT_NAME</tt> and
+      ## <tt>SERVER_NAME</tt>:: When combined with <tt>SCRIPT_NAME</tt> and
       ##                        <tt>PATH_INFO</tt>, these variables can be
       ##                        used to complete the URL. Note, however,
       ##                        that <tt>HTTP_HOST</tt>, if present,
       ##                        should be used in preference to
       ##                        <tt>SERVER_NAME</tt> for reconstructing
       ##                        the request URL.
-      ##                        <tt>SERVER_NAME</tt> and <tt>SERVER_PORT</tt>
-      ##                        can never be empty strings, and so
-      ##                        are always required.
+      ##                        <tt>SERVER_NAME</tt> can never be an empty
+      ##                        string, and so is always required.
+
+      ## <tt>SERVER_PORT</tt>:: An optional +Integer+ which is the port the
+      ##                        server is running on. Should be specified if
+      ##                        the server is running on a non-standard port.
 
       ## <tt>HTTP_</tt> Variables:: Variables corresponding to the
       ##                            client-supplied HTTP request
@@ -280,7 +282,7 @@ module Rack
       ## The <tt>SERVER_PORT</tt> must be an integer if set.
       assert("env[SERVER_PORT] is not an integer") do
         server_port = env["SERVER_PORT"]
-        server_port.nil? || (Integer(server_port) rescue false)
+        server_port.nil? || server_port.is_a?(Integer)
       end
 
       ## The <tt>SERVER_NAME</tt> must be a valid authority as defined by RFC7540.
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -279,10 +279,10 @@ module Rack
         assert("env missing required key #{header}") { env.include? header }
       }
 
-      ## The <tt>SERVER_PORT</tt> must be an integer if set.
-      assert("env[SERVER_PORT] is not an integer") do
+      ## The <tt>SERVER_PORT</tt> must be an Integer if set.
+      assert("env[SERVER_PORT] is not an Integer") do
         server_port = env["SERVER_PORT"]
-        server_port.nil? || server_port.is_a?(Integer)
+        server_port.nil? || (Integer(server_port) rescue false)
       end
 
       ## The <tt>SERVER_NAME</tt> must be a valid authority as defined by RFC7540.
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -219,7 +219,7 @@ module Rack
         end
       end
 
-      # The authority of the incoming reuqest as defined by RFC2976.
+      # The authority of the incoming request as defined by RFC3976.
       # https://tools.ietf.org/html/rfc3986#section-3.2
       #
       # In HTTP/1, this is the `host` header.
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -403,6 +403,7 @@ module Rack
       def form_data?
         type = media_type
         meth = get_header(RACK_METHODOVERRIDE_ORIGINAL_METHOD) || get_header(REQUEST_METHOD)
+
         (meth == POST && type.nil?) || FORM_DATA_MEDIA_TYPES.include?(type)
       end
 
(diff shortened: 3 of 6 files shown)


ioquatix:
This is still a draft/prototype.

ioquatix:
This is quite a complex change, so I welcome a little bit of scrutiny. That being said, the vast majority of the specs are passing without changes, so that's a good sign.

cc @AlexWayfer @tenderlove @jeremyevans

lib/rack/request.rb
```diff
@@ -232,52 +278,87 @@ def xhr?
...
+
+        return host
+      end
+
+      # Returns an address suitable for being used with `getaddrinfo`.
+      def hostname
+        host, address, port = split_authority(self.authority)
+
+        return address
       end
```
AlexWayfer: This looks like inconsistency: the method `hostname` returns `address`… why? I think we should name it either `hostname` or `address` ecerywhere.

ioquatix: It matches the implementation of `URI`.

AlexWayfer: I mean, if you even just rename local variable to `hostname` — it'd be better. I don't see `address` as URI method or in its sources.

ioquatix: Understood. It is actually confusing to me too in a way.

```
u = URI("http://[::1]/bar")
p u.hostname      #=> "::1"
p u.host          #=> "[::1]"
```

What does make sense to me is when you say "host_with_port". That means, the "escaped" host and `:port` appended. You can't have `hostname_with_port`. It's simply not possible because it becomes ambiguous to parse.

So, there's that logic... But yeah, I also see it's confusing.

ioquatix: Yes, we could rename it `hostname` everywhere. I see what you are saying, and I don't think it would be a problem. Except that technically, I don't even know if `192.168.1.1` is a hostname? Or `1::` is a hostname? To me, something like `www.google.com` is a hostname. Not sure what is the right word/definition here.

AlexWayfer: > Except that technically, I don't even know if `192.168.1.1` is a hostname? Or `1::` is a hostname? To me, something like `www.google.com` is a hostname. Not sure what is the right word/definition here.

I found [Wikipedia article](https://en.wikipedia.org/wiki/Hostname#Restrictions_on_valid_hostnames):

> Hostnames are composed of a series of labels concatenated with dots. For example, "en.wikipedia.org" is a hostname. Each label must be from 1 to 63 characters long, and the entire hostname including the delimiting dots, but not a trailing dot, has a maximum of 253 ASCII characters.

So… no, IP addresses are not hostnames. Maybe we should write better code, logical, instead of blind following of URI (and, probably, we should improve URI).

ioquatix: But "www.google.com" is also not an address.. is it?

ioquatix: One part that I wondered about... we are returning lots of bits... ports, hosts... when I think that most users probably just want a rich URI object. Then the interface for the specific methods is less important/out of our hands.

e.g.

```
class Request
  def uri
    URI.parse("#{scheme}://#{authority}/#{path}")
  end
end
```

AlexWayfer: > But "www.google.com" is also not an address.. is it?

I think, "address" is more common word, like address of a house. And in [Domain name article](https://en.wikipedia.org/wiki/Domain_name) I saw:

> Domain names are used in various networking contexts and for application-specific naming and **addressing** purposes.

But I also found [Network address article](https://en.wikipedia.org/wiki/Network_address), which is obviously more specific than common address, and there is no domain or hostname mentions inside, just numbers, IPs, MAC addresses…

So, that's not trivial question.

In such case we can follow stdlib, I think.

But, again, I've posted the first comment just about local variable name. :grimacing:

ioquatix: So, should we change `def hostname` to `def address`?

ioquatix: I know it's about local variable name but the entire thing is kind of confusing, so we should try to get it right for our users or we just perpetuate the chaos.

ioquatix: So we have several options:

1/ Rename local variable `address` -> `hostname` and `return address` to `return hostname`.
2/ Rename method `def hostname` to `def address`.
3/ Remove `def hostname`.
4/ Add `def uri` which returns an instance of URI which represents what the URI was used to make the request.

AlexWayfer: 1 is like compliance with URI from stdlib.
2 is like try to improve naming, but, as I see, still with logic violations.
3 is radical, and, I guess, we still need for a helper method. If not — the best code is no code.
4 seems interesting, and "URI" sounds common enough for IPs and domains, especially with `URI` instance inside.

So, I can't say before trying, but I'd try to make in this order: 3, 4, 1, 2.

ioquatix: I'll wait for feedback from @jeremyevans before I make any further changes.

ioquatix: Just for the record, there are several new methods, and that includes `def hostname`. It's not released in Rack 2.1.

lib/rack/request.rb
```diff
@@ -207,8 +219,42 @@ def scheme
...
+            return "#{host}:#{port}"
+          else
+            return host
+          end
+        end
+      end
+
+      def server_name
+        if name = get_header(SERVER_NAME)
+          return name
```
jeremyevans: In https://github.com/rack/rack/pull/1557#issuecomment-582611953, you stated `I'm fine with implicit return when the cyclomatic complexity of a function is 1, or the branch depth is 1.`  To keep at least a reasonably consistent style, can we avoid explicit returns in cases where the branch depth is <= 1 (I assume the branch depth is 1 here, and outside of the if the branch depth is 0).  Ditto for the rest of the file.

lib/rack/request.rb
```diff
@@ -232,52 +278,87 @@ def xhr?
...
-          "#{host}:#{port}"
+          return authority
         end
       end
 
+      # Returns a formatted host, suitable for being used in a URI.
       def host
-        # Remove port number.
-        strip_port hostname.to_s
+        host, address, port = split_authority(self.authority)
```
jeremyevans: Unused local variables, method body could probably just be `split_authority(authority)[0]`.  Likewise `hostname` could be `split_authority(authority)[1]`.  If you want really keep the existing style, use underscores for the unused local variables.  Similar issue exists in the `port` method below.

lib/rack/request.rb
```diff
@@ -232,52 +278,87 @@ def xhr?
...
           end
+        end
+
+        if forwarded_port = self.forwarded_port
+          return forwarded_port.first
+        end
 
-        result.to_i unless result.to_s.empty?
+        if scheme = self.scheme
+          if port = DEFAULT_PORTS[self.scheme]
```
jeremyevans: If you are going to set the `scheme` local variable, you should use it as the argument to `DEFAULT_PORTS.[]`.  Otherwise don't define the local variable.

lib/rack/request.rb
```diff
@@ -489,37 +584,24 @@ def parse_multipart
...
-
-      def hostname
-        if forwarded = get_header(HTTP_X_FORWARDED_HOST)
-          forwarded.split(/,\s?/).last
-        else
-          get_header(HTTP_HOST) ||
-            get_header(SERVER_NAME) ||
-            get_header(SERVER_ADDR)
-        end
+      def split_header(value)
```
jeremyevans: A generic name like `split_header` implies it will split all headers, which I don't think is possible as headers have different formats.  Maybe `split_ip_addresses` wasn't accurate, but I think `split_header` is more misleading as a name.

lib/rack/request.rb
```diff
@@ -489,37 +584,24 @@ def parse_multipart
...
-      def strip_port(ip_address)
-        # IPv6 format with optional port: "[2001:db8:cafe::17]:47011"
-        # returns: "2001:db8:cafe::17"
-        sep_start = ip_address.index('[')
-        sep_end = ip_address.index(']')
-        if (sep_start && sep_end)
-          return ip_address[sep_start + 1, sep_end - 1]
-        end
+      AUTHORITY = /(?<host>(\[(?<ip6>.*)\])|(?<ip4>[\d\.]+)|(?<name>[a-zA-Z0-9\.\-]+))(:(?<port>\d+))?/
+      private_constant :AUTHORITY
```
jeremyevans: This appears to be only used in one place.  I think it should be inlined into the method, and then you can use it with `=~` so that the named captures become local variables.

```ruby
/(?<host>(\[(?<ip6>.*)\])|(?<ip4>[\d\.]+)|(?<name>[a-zA-Z0-9\.\-]+))(:(?<port>\d+))?/ =~ authority
```

I would also use `ipv6` and `ipv4` as the capture names.

test/spec_request.rb
```diff
@@ -114,32 +114,38 @@ class RackRequestTest < Minitest::Spec
...
     req = make_request \
       Rack::MockRequest.env_for("/", "HTTP_HOST" => "localhost:81", "HTTP_X_FORWARDED_HOST" => "example.org:9292")
     req.host.must_equal "example.org"
+    req.hostname.must_equal "example.org"
 
     req = make_request \
       Rack::MockRequest.env_for("/", "HTTP_HOST" => "localhost:81", "HTTP_X_FORWARDED_HOST" => "[2001:db8:cafe::17]:47011")
-    req.host.must_equal "2001:db8:cafe::17"
+    req.host.must_equal "[2001:db8:cafe::17]"
+    req.hostname.must_equal "2001:db8:cafe::17"
```
jeremyevans: This is likely to be a significant breaking change for IPv6 users.  The CHANGELOG should specifically call this out as a breaking change.

test/spec_request.rb
```diff
@@ -114,32 +114,38 @@ class RackRequestTest < Minitest::Spec
...
     env.delete("SERVER_NAME")
     req = make_request(env)
     req.host.must_equal "192.168.1.1"
+    req.hostname.must_equal "192.168.1.1"
 
     env = Rack::MockRequest.env_for("/")
     env.delete("SERVER_NAME")
     req = make_request(env)
-    req.host.must_equal ""
+    req.host.must_be_nil
```
jeremyevans: I think removing this test completely is better.  `SERVER_NAME` is required by SPEC.  A Request without `SERVER_NAME` should be undefined behavior.

test/spec_request.rb
```diff
@@ -214,7 +220,7 @@ class RackRequestTest < Minitest::Spec
     req.host_with_port.must_equal "example.org:9292"
 
     req = make_request \
-      Rack::MockRequest.env_for("/", "SERVER_NAME" => "example.org", "SERVER_PORT" => "")
```
jeremyevans: Likewise, I think we can remove this test.  `SERVER_PORT` is already required by SPEC to be present and not empty, and we don't need to test undefined behavior.

test/spec_request.rb
```diff
@@ -227,7 +233,7 @@ class RackRequestTest < Minitest::Spec
 
     req = make_request \
       Rack::MockRequest.env_for("/", "HTTP_HOST" => "localhost:81", "HTTP_X_FORWARDED_HOST" => "2001:db8:cafe::17")
-    req.host_with_port.must_equal "2001:db8:cafe::17"
+    req.host_with_port.must_equal "[2001:db8:cafe::17]"
```
jeremyevans: As with `host`, `host_with_port` now using surrounding brackets for IPv6 should be specifically mentioned as breaking the CHANGELOG.

CHANGELOG.md
```diff
@@ -37,6 +37,7 @@ All notable changes to this project will be documented in this file. For info on
 - `Rack::HeaderHash` is memoized by default. ([#1549](https://github.com/rack/rack/pull/1549), [@ioquatix](https://github.com/ioquatix))
 - `Rack::Directory` allow directory traversal inside root directory. ([#1417](https://github.com/rack/rack/pull/1417), [@ThomasSevestre](https://github.com/ThomasSevestre))
 - Sort encodings by server preference. ([#1184](https://github.com/rack/rack/pull/1184), [@ioquatix](https://github.com/ioquatix), [@wjordan](https://github.com/wjordan))
+- Rework host/hostname/authority implementation in `Rack::Request`. ([#1561](https://github.com/rack/rack/pull/1561), [@ioquatix](https://github.com/ioquatix))
```
rafaelfranca: I'd like to see more context on this CHANGELOG entry. What rework means? Is any behavior going to change? Are users expected to do anything differently? Is it just trying to fix some bug?

lib/rack/lint.rb
```diff
@@ -271,13 +271,28 @@ def check_env(env)
...
 
-      %w[REQUEST_METHOD SERVER_NAME SERVER_PORT
-         QUERY_STRING
+      %w[REQUEST_METHOD SERVER_NAME QUERY_STRING
          rack.version rack.input rack.errors
          rack.multithread rack.multiprocess rack.run_once].each { |header|
         assert("env missing required key #{header}") { env.include? header }
       }
 
+      ## The <tt>SERVER_PORT</tt> must be an integer if set.
```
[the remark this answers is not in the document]
ioquatix: I found that some servers just set it to an empty string. There is a proviso it should be nil if empty somewhere in the spec. So... I don't know. I'd like to think if it's present, it's a valid integer. Otherwise, it's nil.

lib/rack/lint.rb
```diff
@@ -271,13 +271,28 @@ def check_env(env)
...
+        server_port = env["SERVER_PORT"]
+        server_port.nil? || (Integer(server_port) rescue false)
+      end
+
+      ## The <tt>SERVER_NAME</tt> must be a valid authority as defined by RFC7540.
+      assert("#{env[SERVER_NAME]} must be a valid authority") do
+        URI.parse("http://#{env[SERVER_NAME]}/") rescue false
+      end
+
+      ## The <tt>HTTP_HOST</tt> must be a valid authority as defined by RFC7540.
```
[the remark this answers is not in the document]
ioquatix: The authority header is well defined in RFC7540. If we don't require it to meet a certain standard here, we have to ensure all middleware can handle whatever junk is in this field.

lib/rack/request.rb
```diff
@@ -207,8 +219,42 @@ def scheme
         end
       end
 
+      # The authority of the incoming reuqest as defined by RFC2976.
```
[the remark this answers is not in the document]
ioquatix: Thanks!

lib/rack/request.rb
```diff
@@ -207,8 +219,42 @@ def scheme
...
+      # In HTTP/2, this is the `:authority` pseudo-header.
       def authority
-        get_header(SERVER_NAME) + ':' + get_header(SERVER_PORT)
+        forwarded_authority || host_authority || server_authority
+      end
+
+      # The authority as defined by the `SERVER_NAME`/`SERVER_ADDR` and
+      # `SERVER_PORT` variables.
+      def server_authority
+        host = self.server_name
```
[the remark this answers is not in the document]
ioquatix: Servers are sending through empty strings?

lib/rack/request.rb
```diff
@@ -207,8 +219,42 @@ def scheme
...
+          else
+            return host
+          end
+        end
+      end
+
+      def server_name
+        if name = get_header(SERVER_NAME)
+          return name
+        elsif address = get_header(SERVER_ADDR)
```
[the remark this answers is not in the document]
ioquatix: This was the previous logic. It's required for specs to pass.

lib/rack/request.rb
ioquatix: If you use it in the method, doesn't the regexp get compiled multiple times?

ioquatix:
Thanks @jeremyevans I will update the specs according to your feedback.

lib/rack/request.rb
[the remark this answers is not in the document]
ioquatix: I guess the confusion here is that some servers are running on unix domain sockets or other things which don't actually have a name or a port. They might not even have an address. Probably we should relax the spec to allow SERVER_PORT to be nil, or maybe that's implicit based on 

```
      ## The environment is required to include these variables
      ## (adopted from PEP333), except when they'd be empty, but see
      ## below.
```

lib/rack/lint.rb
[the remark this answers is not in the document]
ioquatix: (Then all bets are off regarding the behaviour of `Rack::Request`).

lib/rack/lint.rb
```diff
@@ -271,13 +271,28 @@ def check_env(env)
...
+      %w[REQUEST_METHOD SERVER_NAME QUERY_STRING
          rack.version rack.input rack.errors
          rack.multithread rack.multiprocess rack.run_once].each { |header|
         assert("env missing required key #{header}") { env.include? header }
       }
 
+      ## The <tt>SERVER_PORT</tt> must be an integer if set.
+      assert("env[SERVER_PORT] is not an integer") do
+        server_port = env["SERVER_PORT"]
+        server_port.nil? || (Integer(server_port) rescue false)
```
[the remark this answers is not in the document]
ioquatix: Okay, makes sense, I'll update the SPEC.

lib/rack/request.rb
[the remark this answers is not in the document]
ioquatix: `SERVER_ADDR` comes from https://github.com/rack/rack/commit/fb32ff74f1e68650b7b4fa9889d277ef1fd30380

lib/rack/lint.rb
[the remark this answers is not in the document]
ioquatix: I will implement the current behaviour in this PR, and created an issue to discuss further.

lib/rack/request.rb
jeremyevans: No, regexps without interpolation are static.

lib/rack/request.rb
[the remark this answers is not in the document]
jeremyevans: It looks like when it was added doesn't even mention a deliberate decision to include it.  SPEC says nothing about SERVER_ADDR. I think we should drop support for it in Request and mention the dropping in CHANGELOG.  However, if you think it is important to keep, I do not object.

lib/rack/request.rb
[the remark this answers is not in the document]
jeremyevans: Agreed for allowing nil SERVER_PORT to support unix domain sockets.

lib/rack/lint.rb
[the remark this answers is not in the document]
jeremyevans: As we discussed, OK to use HTTP_HOST in preference to SERVER_NAME if it is present in env.

lib/rack/request.rb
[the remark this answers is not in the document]
ioquatix: I am happy to drop it. But let's do it in a separate PR.

lib/rack/request.rb
ioquatix: Okay, I'll try to rework it.

test/spec_request.rb
ioquatix: Done.

test/spec_request.rb
ioquatix: Test was removed.

test/spec_request.rb
ioquatix: It's an existing spec. We could remove it. I wonder why lint doesn't fail?

test/spec_request.rb
ioquatix: Done.

CHANGELOG.md
ioquatix: Done.

lib/rack/request.rb
ioquatix: It's quite a big regex so I added documentation and made it `/x`.

PR #1565, opened 2020-02-05
Enable `Layout/Tab` to avoid hard tab indentation in the future

kamipo:
Ref https://github.com/rack/rack/pull/1549#discussion_r375470299, 1f89eaafd39040707bf2a76cab134dd7a78b1441.


ioquatix:
I like my tabs hard, and my beers cold.

PR #1566, opened 2020-02-05
Refactor and document Directory

jeremyevans:
The previous implementation of directory listings was inefficient for large
directories.  First it built a bunch of file entry strings, then it joined
them into a large string containing all entries, then that was interpolated
into the header and footer creating another large string.  Then each_line was
called on the large string, creating one string per line.

This switches to a streaming approach for directory listings.  We yield the
header, then one string for each listing, then the footer.

Testing with puma with a directory with 10,000 files with 100 byte file
names, directory listing before took 3.5 seconds, with 900ms before
first byte.  After this commit, it takes 2.1 seconds, with 10ms before
the first byte.  Note that total time is not hugely different, because
most of the time is probably taken in stat(2) calls.  Time to first byte is
greatly improved, though.

Remove Directory#path.  path is used as a local variable, but the instance
variable is never set, so the value would always be nil.

The previous approach used a glob for getting files.  That resulted in
broken behavior when the path being served includes glob metacharacters.
Switch to a `Dir.foreach` approach to avoid that issue.  Because the glob
skips hidden files, continue to skip files starting with a period.  Note
that Directory will still serve hidden files and display directory listings
for hidden directories, so this should not be considered a security feature.

As a small behavior change, do not show a Parent directory link for the
root directory, since it takes you to the same page.

Switch list_path to not use exceptions for flow control.  Remove all
exception handling, and switch to using Directory#stat, which already handles
the same exceptions.  This has the advantage so that if a non-file,
non-directory is requested, you get a valid 404 rack response instead of
nil.

### DIFF
--- a/lib/rack/directory.rb
+++ b/lib/rack/directory.rb
@@ -11,8 +11,8 @@ module Rack
   # If +app+ is not specified, a Rack::Files of the same +root+ will be used.
 
   class Directory
-    DIR_FILE = "<tr><td class='name'><a href='%s'>%s</a></td><td class='size'>%s</td><td class='type'>%s</td><td class='mtime'>%s</td></tr>"
-    DIR_PAGE = <<-PAGE
+    DIR_FILE = "<tr><td class='name'><a href='%s'>%s</a></td><td class='size'>%s</td><td class='type'>%s</td><td class='mtime'>%s</td></tr>\n"
+    DIR_PAGE_HEADER = <<-PAGE
 <html><head>
   <title>%s</title>
   <meta http-equiv="content-type" content="text/html; charset=utf-8" />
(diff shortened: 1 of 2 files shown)


jeremyevans:
Due to the extent of changes, I would like another core team member to review this before merging.

lib/rack/directory.rb
```diff
@@ -33,141 +33,161 @@ class Directory
...
     end
 
-    def stat(node)
-      ::File.stat(node)
+    # File::Stat for the given path, but return nil for missing/bad entries.
+    def stat(path)
+      ::File.stat(path)
     rescue Errno::ENOENT, Errno::ELOOP
-      return nil
+      # nil
```
ioquatix: I think `return nil` is better. Why would you change that to a comment? Same with `list_path` below.

lib/rack/directory.rb
```diff
@@ -33,141 +33,161 @@ class Directory
...
+          yield(DIR_FILE % DIR_FILE_escape(f))
+        end
+
+        yield(DIR_PAGE_FOOTER)
       end
 
       private
+
+      # Escape each element in the array of html strings.
       def DIR_FILE_escape(htmls)
```
ioquatix: I don't understand the casing of this method. Why not `dir_file_escape`?

ioquatix:
It looks good to me. It's probably not used much, so I don't feel too concerned about any latent issues. I assume there are some tests for it...

lib/rack/directory.rb
jeremyevans: If you look at the generated instruction sequences, the explicit `return nil` requires more instructions, and the behavior is the same. I was nice enough to leave a comment indicating that we expect nil to be return if the rescue block is hit. :) 

That said, I don't have a problem changing these back, so I'll do so before committing.

lib/rack/directory.rb
jeremyevans: The escaped values are interpolated into a string constant called DIR_FILE.

jeremyevans:
There are tests for it (don't forget, we recently got to 100% line coverage).  I'll recheck that everything is covered.

PR #1567, opened 2020-02-06
Initial conversion and formatting.

ioquatix:


jeremyevans:
I'm against this change. Also, note that SPEC is generated from comments in Lint. so you would need to port all changes to Lint.  You would then need to update the Rakefile for this change, then regenerate using `rake SPEC.md`.

ioquatix:
Understood. Maybe we can just automate the spec conversion to markdown (so we have both rdoc and markdown).

jeremyevans:
I'm fine with having both as long as the content is the same and the conversion process only requires what is included with Ruby (no dependency on a third party document converter or gem).

ioquatix:
Apparently:

```
require 'rdoc'
converter = RDoc::Markup::ToMarkdown.new
rdoc = File.read(ARGV[0] || 'README.rdoc')
puts converter.convert(rdoc)
```

I'm going to close this PR.

If anyone wants to have a go at making a PR, it would be great. But we also need to fix the source documentation.

jeremyevans:
Before anyone spends the time to convert, we should discuss if we want it converted.  You want it converted, I don't.  I don't understand what the benefits are.  Is it just Markdown=good, RDoc=bad, or is there a technical reason to prefer Markdown?

ioquatix:
RDoc is Ruby specific whereas Markdown is a general text format which is more universally recognised and used.

To be honest, now that I understand that it is automatically generated, I'm less interested in doing this.

But there do appear to be lots of opportunities for improving the RDoc formatting. The markdown conversion I did, I did add better formatting in certain areas. So, if we close this, which I'm fine with, I think we need to them improve `rack/lint.rb`.

ioquatix:
So, in conclusion, I won't bother finishing this or merging it, the use of markdown from the original issue could be implemented using a conversion tool, so feel free to bring that PR to the table, and as a consequence of my better understanding, let's focus on improving the docs in `lint.rb`.

PR #1572, opened 2020-02-07
Make Request#params not rescue EOFError

### LINKED ISSUE #761: params is empty if multipart request boundary is incorrect
params in request.rb is empty if a multipart request ends with an incorrect boundary e.g. with `-9e923a4c2c2cbc9dd5f104b02ec111df-` and the header contains `Content-Type: multipart/form-data; =-6f88629f6762cfcdcdbe4b9c0156a786`. This is caused by the fact that params catches EOFError. Is there a specific reason that the exception is not propagated?

I'm using 1.6.0.beta.

jeremyevans:
There was no indication why this was done originally, and it is
possible that it will rely in silent data loss.

Fixes #761

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,8 +24,9 @@ All notable changes to this project will be documented in this file. For info on
 
 ### Changed
 
+- `Request#params` no longer rescues EOFError. ([@jeremyevans](https://github.com/jeremyevans))
 - `Directory` uses a streaming approach, significantly improving time to first byte for large directories. ([@jeremyevans](https://github.com/jeremyevans))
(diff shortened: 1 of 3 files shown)


ioquatix:
Looks good to me. Seems CHANGELOG includes unrelated changes?

jeremyevans:
Yes, I was fixing the wording of a previous CHANGELOG entry.

dblock:
This caused https://github.com/ruby-grape/grape/pull/2161, was this the right decision to make this change here?

jeremyevans:
I think so.  Silently ignoring errors is a bad idea.  Looks like @ioquatix agreed.  Not sure how other committers feel about it.

PR #1575, opened 2020-02-08
Tidy up gemspec.

ioquatix:


ioquatix:
@asppsa you asked about dropping 2.2 support and I think we will do that in the next release.

ioquatix:
@jeremyevans if you are happy with this can you merge before doing the 2.2 release?

ioquatix:
Also I think we should shut down the mailing list. People can use github issues.

rack.gemspec
```diff
@@ -5,37 +5,37 @@ require_relative 'lib/rack/version'
...
+  s.description = <<~EOF
+    Rack provides a minimal, modular and adaptable interface for developing
+    web applications in Ruby. By wrapping HTTP requests and responses in
+    the simplest way possible, it unifies and distills the API for web
+    servers, web frameworks, and software in between (the so-called
+    middleware) into a single method call.
+  EOF
+
+  s.files = Dir['{bin/*,contrib/*,example/*,lib/**/*}'] +
+    %w(MIT-LICENSE rack.gemspec Rakefile README.rdoc SPEC)
```
jeremyevans: This should be `SPEC.rdoc` now. I missed this when I changed the name.

jeremyevans:
Other than the SPEC -> SPEC.rdoc change, this looks fine.  I prefer Google Groups to GitHub Issues for non-bug reports, but I don't feel strongly about it.

jeremyevans:
I fixed the SPEC filename after merging this.

ioquatix:
Thanks. I was 💤.

PR #1578, opened 2020-02-09
External gems

ioquatix:
@jeremyevans what do you think about doing something like this?

I suggest we try to pull in a few more external gems, and then add that in the GitHub Actions tests, we can make a workflow just for external dependencies. It's a nice way add some higher level/extended extensive acceptance tests.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -128,3 +128,25 @@ task pushsite: :pushdoc do
   sh "rsync -avz site/ rack.rubyforge.org:/var/www/gforge-projects/rack/"
   sh "cd site && git push"
 end
+
+def clone_and_test(url, name, command)
+  path = "external/#{name}"
+  FileUtils.rm_rf path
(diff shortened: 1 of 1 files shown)


jeremyevans:
I guess I'm OK with this idea.  However, if we run into issues where the external gems tests fail due to non-rack issues, we should probably be aggressive about dropping them.

I think the most important tests would be framework tests, so rails and sinatra at least and maybe grape and roda if we want to be more thorough.  In terms of webservers, falcon makes sense.  Puma, unicorn, iodine, and agoo don't depend on rack, so no point in testing them.  Passenger depends on rack, but I'm against adding it due to build time.  In terms of middleware, maybe warden?

ioquatix:
I’m okay just adding all the above, and pruning them if they are unhelpful. Thanks for all the suggestions!

PR #1580, opened 2020-02-09
Add external tests.

ioquatix:


ioquatix:
@jeremyevans I've added falcon and rack-attack.

There should be more. Maybe rack-cache is another good one. And Roda.

Can I invite you at your leisure to add a few more?

PR #1585, opened 2020-02-09
Extend Rack::Builder with support for Rack::Environment

ioquatix:


jeremyevans:
Looks like you already switched the warn to fail in b95cb72463bac5a087294c7ecf94674db7aa41fe, so I think this can now be closed.

PR #1587, opened 2020-02-10
Fix to handle same_site option for session pool

kamipo:
Follow up of #1543.

### DIFF
--- a/lib/rack/session/abstract/id.rb
+++ b/lib/rack/session/abstract/id.rb
@@ -252,6 +252,7 @@ module Rack
           @default_options = self.class::DEFAULT_OPTIONS.merge(options)
           @key = @default_options.delete(:key)
           @cookie_only = @default_options.delete(:cookie_only)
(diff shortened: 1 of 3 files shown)


ioquatix:
This seems reasonable to me. @jeremyevans WDYT?

kamipo:
I've found this issue at https://buildkite.com/rails/rails/builds/66924#e59fd8a6-3524-46b4-98ae-25369c786941, using subclasses of `Rack::Session::Abstract::Persisted` except `Rack::Session::Cookie` will raise "instance variable @same_site not initialized" warnings.

ioquatix:
Can you confirm this is just to fix warnings, right?

ioquatix:
Sorry, I misunderstood the PR. I've merged it. I will release as part of v2.2.2 tomorrow after I wake up. So, if there are any other issues, please commit it in the next 12 hours and assign it milestone v2.2.2

PR #1589, opened 2020-02-10
Double assignment is still needed to prevent an "unused variable" warning

kamipo:
That warning is still raised on latest Ruby.

```
% bundle exec ruby -w -Itest test/spec_show_exceptions.rb test/spec_show_status.rb
/Users/kamipo/src/github.com/rack/rack/lib/rack/show_exceptions.rb:66: warning: assigned but unused variable - path
/Users/kamipo/src/github.com/rack/rack/lib/rack/show_exceptions.rb:68: warning: assigned but unused variable - frames
Run options: --seed 23328

# Running:

........

Finished in 0.240203s, 33.3052 runs/s, 374.6831 assertions/s.

8 runs, 90 assertions, 0 failures, 0 errors, 0 skips

% bundle exec ruby -w -Itest test/spec_show_status.rb
/Users/kamipo/src/github.com/rack/rack/lib/rack/show_status.rb:26: warning: assigned but unused variable - req
/Users/kamipo/src/github.com/rack/rack/lib/rack/show_status.rb:30: warning: assigned but unused variable - detail
Run options: --seed 41036

# Running:

.......

Finished in 0.061453s, 113.9082 runs/s, 829.9025 assertions/s.

7 runs, 51 assertions, 0 failures, 0 errors, 0 skips
```

### DIFF
--- a/lib/rack/show_exceptions.rb
+++ b/lib/rack/show_exceptions.rb
@@ -63,9 +63,13 @@ module Rack
     def pretty(env, exception)
       req = Rack::Request.new(env)
 
-      path = (req.script_name + req.path_info).squeeze("/")
+      # This double assignment is to prevent an "unused variable" warning.
+      # Yes, it is dumb, but I don't like Ruby yelling at me.
+      path = path = (req.script_name + req.path_info).squeeze("/")
 
-      frames = exception.backtrace.map { |line|
+      # This double assignment is to prevent an "unused variable" warning.
+      # Yes, it is dumb, but I don't like Ruby yelling at me.
+      frames = frames = exception.backtrace.map { |line|
         frame = OpenStruct.new
         if line =~ /(.*?):(\d+)(:in `(.*)')?/
           frame.filename = $1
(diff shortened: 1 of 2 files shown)


ioquatix:
Thanks for this I will check it.

ioquatix:
This is because of ERB and using binding.

Maybe another solution is to pass a hash of the relevant variables into ERB.

I don't know a lot about ERB, but this is the approach I use in Trenni (similar to ERB) in my own projects.

kamipo:
I already tried that, `ERB#result_with_hash` which is introduced at Ruby 2.5 uses `TOPLEVEL_BINDING`, but in this case we need to use local binding object to use `h` helper in the ERB template.

ioquatix:
@mame do you have any suggestions about how we can improve this?

ioquatix:
I've merged this because I agree we don't want warnings. But maybe there is a better way.

mame:
:thinking:

@nobu What do you think?

Issue #1592, opened 2020-02-10
Make response headers a hash and enforce lower-case keys.


ioquatix:
HTTP/2 makes it clear that headers should be lower-case (https://http2.github.io/http2-spec/#rfc.section.8.1.2):

> Just as in HTTP/1.x, header field names are strings of ASCII characters that are compared 
> in a case-insensitive fashion. However, header field names MUST be converted to lowercase
> prior to their encoding in HTTP/2. A request or response containing uppercase header field
> names MUST be treated as malformed (Section 8.1.2.6).

HTTP/1 was often case-preserving/case-insensitive. However, sometimes this behaviour caused bugs when different implementations worked together. So, it's quite a risky default.

I suggest with Rack 3.0 we take the opportunity to enforce the HTTP/2 semantics. It also simplifies the requirements around `headers` being a hash, to the point where I think it would be acceptable for Rack 3.0.

In terms of implementation it would require:

- `Rack::Lint` validates response headers should be lower case.
- All header key string constants are changed to lower case.
- Rework/remove `Rack::HeaderHash`.

ioquatix:
Here is a proposed implementation: https://github.com/rack/rack/pull/1485 but I think we can lock it down even further, and we have justification to do so based on the HTTP/2 RFCs.

ioquatix:
This has also been discussed in https://github.com/rack/rack/issues/1222.

ioquatix:
Node.js does normalisation like this too apparently: https://nodejs.org/en/docs/guides/anatomy-of-an-http-transaction/#method-url-and-headers

jeremyevans:
I think requiring lower case headers has the potential to break lots of rack applications, frameworks, and middleware.  However, it would also simplify a lot of code in the long run.  Since HTTP/2 requires it and HTTP/1 works fine with it, it seems a reasonable change for SPEC.  I definitely prefer this approach over requiring that response headers be case insensitive.

I think if we take this approach, we should ship with a `Rack::LowerCaseResponseHeaders` or some such middleware that automatically downcases all response header keys. Basically, it should take any response headers valid by the current SPEC (enumerable of string keys and values) and turn it into response headers valid by the new SPEC.

ioquatix:
> I think requiring lower case headers has the potential to break lots of rack applications, frameworks, and middleware.

Agreed. I even know for a fact this will break `rack-test` which is why I am keen to get that repo here so we can coordinate the releases.

> However, it would also simplify a lot of code in the long run.

[...]

PR #1597, opened 2020-02-11
Require the response headers be an unfrozen hash in SPEC

### LINKED ISSUE #1222: Headers specs are not clear enough

ioquatix:
This is stricter than what was previously required.  However,
non-hash response headers would break most of the middleware
that accesses response headers.

Middleware in many cases adds or removes headers, so require
the hash not be frozen, so that this can be done efficiently.

Fixes #1222

### DIFF
--- a/SPEC.rdoc
+++ b/SPEC.rdoc
@@ -268,7 +268,7 @@ This is an HTTP status. It must be an Integer greater than or equal to
 
 === The Headers
 
-The header must respond to +each+, and yield values of key and value.
+The headers must be a unfrozen Hash.
 The header keys must be Strings.
 Special headers starting "rack." are for communicating with the
 server, and must not be sent back to the client.
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -676,13 +676,17 @@ module Rack
       ##
       ## === The Headers
       ##
-      def check_headers(header)
-        ## The header must respond to +each+, and yield values of key and value.
-        unless header.respond_to? :each
-          raise LintError, "headers object should respond to #each, but doesn't (got #{header.class} as headers)"
+      def check_headers(headers)
+        ## The headers must be a unfrozen Hash.
+        unless headers.kind_of?(Hash)
+          raise LintError, "headers object should be a hash, but isn't (got #{headers.class} as headers)"
+        end
+        
+        if headers.frozen?
+          raise LintError, "headers object should not be frozen, but is"
         end
 
-        header.each { |key, value|
+        headers.each { |key, value|
           ## The header keys must be Strings.
           unless key.kind_of? String
             raise LintError, "header key must be a string, was #{key.class}"
--- a/test/spec_lint.rb
+++ b/test/spec_lint.rb
@@ -284,15 +284,24 @@ describe Rack::Lint do
   end
 
   it "notice header errors" do
+    obj = Object.new
+    def obj.each; end
     lambda {
       io = StringIO.new('a')
       io.binmode
       Rack::Lint.new(lambda { |env|
                        env['rack.input'].each{ |x| }
-                       [200, Object.new, []]
+                       [200, obj, []]
                      }).call(env({ "rack.input" => io }))
     }.must_raise(Rack::Lint::LintError).
-      message.must_equal "headers object should respond to #each, but doesn't (got Object as headers)"
+      message.must_equal "headers object should be a hash, but isn't (got Object as headers)"
+    lambda {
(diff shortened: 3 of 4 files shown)


ioquatix:
I'm happy with this as a starting point for https://github.com/rack/rack/issues/1592

ioquatix:
@ bryanmacfarlane sorry to ping you.... but I just don't understand why this is going wrong. We have a workflow set up for push and pull_request, and now we seem to get 2x the jobs executing, i.e. for both push and pull_request. What are we doing wrong here? Thanks and sorry to bother you.

ioquatix:
@bryanmacfarlane sorry to ping you directly.... but I just don't understand why this is going wrong. We have a workflow set up for push and pull_request, and now we seem to get 2x the jobs executing, i.e. for both push and pull_request. What are we doing wrong here? Thanks and sorry to bother you.

ioquatix:
These changes ensure that if middleware uses `Rack::Response` for manipulating headers, that it will work as expected, even if the header case is not lower case. But should we also issue a warning?

e.g.

```
set_header('Foo', 'bar')
Header key "Foo" is not lower case, this behaviour will be deprecated.
```

mpalmer:
If you only state that "headers must be a hash", then adding another value to the hash when there's already a value there is... fraught with peril.  If "headers must be a hash" is where SPEC goes, I'd suggest going a bit further and saying "headers must be a `Hash.new { |h, k| h[k] = [] }`" (and "all values must be arrays of strings").  Then `set_header` is `[]=` and `add_header` is `<<`.

ioquatix:
@mpalmer I actually agree with you but I received a lot of push back on this point. https://github.com/rack/rack/issues/1598

However, I still believe it's the right approach to use `Array`, at least in cases like this.

mpalmer:
There's a lot going on in that linked issue, and you're right that there's some opinion against the proposal you made there.  However, it's discussing a *slightly* different point to what we're talking about in this issue.  Over in #1597, you're proposing that arrays be *allowed* header values, but not requiring them, within the context of not otherwise changing the format of `headers` in SPEC.  I agree with Jeremy, et al, that just making that change isn't worthwhile -- it's just complicating all code that has to handle headers, without really providing any benefit.

This issue, though, is about mandating a more restrictive interface for `headers`, at which case it's worth doing the job properly and embracing the array of values.  I don't think you can sanely expect everything that interacts with `headers`, a hash in which some keys could legitimately have multiple values -- and for which the "join" semantics differ between keys -- to implement the correct "join with a comma, except if you don't, but only if there's already a value" logic.

Yes, everyone *could* just use `HeaderHash`, but if you specify an interface that essentially requires `HeaderHash`, without mandating that `HeaderHash` is necessary, you're *de facto* mandating using `HeaderHash` anyway, at which point you may as well just say "you must use `HeaderHash`" and be done with it.  But I agree with [what Bo said](https://github.com/rack/rack/issues/1598#issuecomment-593468208), and I too would prefer it if the Rack specification didn't *require* me to use the `rack` gem, so sticking to simple data structures with well-described semantics for usage would be my preference.

ioquatix:
I've reverted this back to the original commit which was co-authored by @jeremyevans and I believe we agree on, so I'll push forward with this in the first instance.

PR #1606, opened 2020-02-16
Relax parsing restrictions around host and hostname

### LINKED ISSUE #1604: Rack::Request#host returns truncated data for domain names containing underscores, other characters
In Rack 2.1.2, `Rack::Request#host` did no hostname validation, returning everything from `#host_with_port` up to the colon.

``` ruby
      def host
        # Remove port number.
        h = host_with_port
        if colon_index = h.index(":")
          h[0, colon_index]
        else
          h
        end
      end
```

This did not support IPv6 well, so #1561 was implemented to solve that.  This changed the implementation of `#host` to one that relied on regular expressions.  This was released in Rack 2.2.0.

``` ruby
      def host
        split_authority(self.authority)[0]
      end

      def authority
        forwarded_authority || host_authority || server_authority
      end

      def forwarded_authority
        if value = get_header(HTTP_X_FORWARDED_HOST)
          wrap_ipv6(split_header(value).first)
        end
      end

      def host_authority
        get_header(HTTP_HOST)
      end

      def server_authority
        host = self.server_name
        port = self.server_port

        if host
          if port
            "#{host}:#{port}"
          else
            host
          end
        end
      end

      def split_authority(authority)
        if match = AUTHORITY.match(authority)
          if address = match[:ip6]
            return match[:host], address, match[:port]&.to_i
          else
            return match[:host], match[:host], match[:port]&.to_i
          end
        end

        # Give up!
        return authority, authority, nil
      end

      AUTHORITY = /
        # The host:
        (?<host>
          # An IPv6 address:
          (\[(?<ip6>.*)\])
          |
          # An IPv4 address:
          (?<ip4>[\d\.]+)
          |
          # A hostname:
          (?<name>[a-zA-Z0-9\.\-]+)
        )
        # The optional port:
        (:(?<port>\d+))?
      /x
```

This provided not only better IPv6 support, but also did some validation of hostnames, as per RFC 952.

It was soon discovered that this was improperly parsing hostnames that were valid under the more lenient RFC 1123 (permitting hostnames to begin with digits and hyphens) as IP addresses (#1590), so in #1591, the regular expression was updated to match [an entire line of] the authority.  This was released as Rack 2.2.2.

``` ruby
      AUTHORITY = /^
        # The host:
        (?<host>
          # An IPv6 address:
          (\[(?<ip6>.*)\])
          |
          # An IPv4 address:
          (?<ip4>[\d\.]+)
          |
          # A hostname:
          (?<name>[a-zA-Z0-9\.\-]+)
        )
        # The optional port:
        (:(?<port>\d+))?
      $/x
```

---

### The Issue

While the match being performed in the `AUTHORITY` regex is technically correct for _hostnames_, it is not correct for _DNS_ names (as per [RFC 2181, Section 11](https://tools.ietf.org/html/rfc2181#section-11)).  This also causes issues in environments like Docker, where some tools (e.g. Docker Compose) implicitly create hostnames containing underscores.  These hostnames are being truncated in Rack 2.2.0, and being left unprocessed in Rack 2.2.2.

### Suggestions

Given RFC 2181 relaxes all restrictions on DNS names (except segment and total length), it seems unnecessarily risky to use anything but a wildcard match in that position.  IP addresses, however, have a very predictable form, and regular expressions for matching them correctly are already a part of the Ruby standard library (see [`Resolv::IPv4::Regex`](https://github.com/ruby/ruby/blob/v2_3_8/lib/resolv.rb#L2362-L2366) and [`Resolv::IPv6::Regex`](https://github.com/ruby/ruby/blob/v2_3_8/lib/resolv.rb#L2438-L2477)) in every supported version of Ruby.

pvande:
This approach allows us to match and process any potential input and still meet the contractual requirements of `Request#split_authority` _without_ requiring the input to be well-formed or RFC-compliant.  In turn, this eliminates unexpected failures on invalid-but-functional inputs.

If we deemed it necessary to do validation (particularly on IP addresses), the `AUTHORITY` regex still provides enough context on which validations should apply, though doing so would require some consideration on what should be returned upon validation failure.

If validations can be ruled unnecessary – I suspect this should be the case – the implementation could be replaced by the semantically identical:

``` ruby
def split_authority(authority)
  /\A(?<host>\[\g<addr>\]|(?<addr>.*?))(:(?<port>\d+))?\Z/ =~ authority
  return host, addr, port&.to_i
end
```

This gives up differentiation between IPv6, IPv4, and DNS addresses, but is arguably simpler.

(Resolves #1604)

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -598,35 +598,29 @@ module Rack
         value ? value.strip.split(/[,\s]+/) : []
       end
 
-      AUTHORITY = /^
-        # The host:
+      AUTHORITY = /
+        \A
         (?<host>
-          # An IPv6 address:
-          (\[(?<ip6>.*)\])
+          # Match IPv6 as a string of hex digits and colons in square brackets
+          \[(?<address>(?<ipv6>.*))\]
           |
-          # An IPv4 address:
-          (?<ip4>[\d\.]+)
+          # Match IPv4 as four dot-separated groups of digits
+          (?<address>(?<ipv4>[\d.]{7,}))
           |
-          # A hostname:
-          (?<name>[a-zA-Z0-9\.\-]+)
+          # Match any other string as a hostname
+          (?<address>(?<hostname>.*?))
         )
-        # The optional port:
         (:(?<port>\d+))?
-      $/x
+        \Z
+      /x
 
       private_constant :AUTHORITY
 
       def split_authority(authority)
-        if match = AUTHORITY.match(authority)
-          if address = match[:ip6]
-            return match[:host], address, match[:port]&.to_i
-          else
-            return match[:host], match[:host], match[:port]&.to_i
-          end
-        end
+        return [] if authority.nil?
 
-        # Give up!
-        return authority, authority, nil
+        match = AUTHORITY.match(authority)
+        return match[:host], match[:address], match[:port]&.to_i
       end
 
       def reject_trusted_ip_addresses(ip_addresses)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file. For info on
 
 ## [2.3.0] - Unreleased
 
+- Relax validations around `Rack::Request#host` and `Rack::Request#hostname`. ([#1606](https://github.com/rack/rack/issues/1606), [@pvande](https://github.com/pvande))
+
 ## [2.2.2] - 2020-02-11
 
 ### Fixed
--- a/test/spec_request.rb
+++ b/test/spec_request.rb
@@ -122,6 +122,36 @@ class RackRequestTest < Minitest::Spec
     req.hostname.must_equal "123foo.example.com"
 
     req = make_request \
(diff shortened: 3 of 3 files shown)


ioquatix:
This looks good to me, you took what I did and made it way better. Thanks!

ioquatix:
I’ll backport to 2-2-stable as I consider this a bug fix.

lib/rack/request.rb
```diff
@@ -598,35 +598,29 @@ def split_header(value)
         value ? value.strip.split(/[,\s]+/) : []
       end
 
-      AUTHORITY = /^
-        # The host:
+      AUTHORITY = /
+        \A
```
ioquatix: @pvande what is the difference between `\A` and `\Z` vs `^` and `$`?

jeremyevans: `\A` matches the beginning of the string.  `^` matches any beginning of line in the string. `\Z` matches the end of string, allowing possible newline at the end of the string. `$` matches the end of any line in the string.

In general, you are much more likely to want `\A` and `\Z` (or `\z`, which matches the end of string) than `^` and `$`.  Using `^` and `$` usually leads to bugs, in my experience.

PR #1607, opened 2020-02-16
Ensure that multiline input doesn't cause Rack::Request to break

pvande:
As written, the `AUTHORITY` regex intends to match the entire input string, and the sole usage of it expects that it will always return a match.  After some consideration, there *is* a way to feed it input that will not match, causing errors on the subsequent line.

The issue at hand is that `.` in a regular expression matches everything *except newlines* by default, meaning that even the most generous regular expression with start- and end-of-string anchors (`/\A.*\z/`) cannot match a multiline string.  The solution, then, is to use the `m` (multiline) flag on the regex, which allows `.` to match newlines as well.

This change also adds a test for the behavior of the `\Z` anchor (end-of-chomped-string) vs the `\z` anchor (actual end-of-string).

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -612,7 +612,7 @@ module Rack
         )
         (:(?<port>\d+))?
         \Z
-      /x
+      /mx
 
       private_constant :AUTHORITY
 
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -603,6 +603,9 @@ module Rack
       AUTHORITY = /
         \A
         (?<host>
+          # Match IPvFuture as per RFC 3986
+          \[(?<address>(?<ipvfuture>v\h+ [[:alnum:]._~\-!$&'()*+,;=:]+))\]
+          |
           # Match IPv6 as a string of non-blank characters in square brackets
           \[(?<address>(?<ipv6>[[:graph:]]*))\]
           |
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'resolv'
+
 module Rack
   # Rack::Request provides a convenient interface to a Rack
   # environment.  It is stateless, the environment +env+ passed to the
@@ -601,25 +603,34 @@ module Rack
       AUTHORITY = /
         \A
         (?<host>
-          # Match IPv6 as a string of hex digits and colons in square brackets
-          \[(?<address>(?<ipv6>.*))\]
+          # Match IPv6 as a string of non-blank characters in square brackets
+          \[(?<address>(?<ipv6>[[:graph:]]*))\]
           |
           # Match IPv4 as four dot-separated groups of digits
-          (?<address>(?<ipv4>[\d.]{7,}))
+          (?<address>(?<ipv4>[\d.]*))
           |
-          # Match any other string as a hostname
-          (?<address>(?<hostname>.*?))
+          # Match any other string of non-blank characters as a hostname
+          (?<address>(?<hostname>[[:graph:]]*?))
         )
         (:(?<port>\d+))?
         \Z
-      /mx
+      /x
 
       private_constant :AUTHORITY
 
       def split_authority(authority)
-        return [] if authority.nil?
-
         match = AUTHORITY.match(authority)
+
+        if match.nil?
+          return nil, nil, nil
+        elsif match[:ipv6]
+          # If it matched as IPv6, it should look like a reasonable IPv6 address.
+          return nil, nil, nil unless Resolv::IPv6::Regex.match?(match[:ipv6])
+        elsif match[:ipv4]
+          # If it matched as IPv4, it should look like a reasonable IPv4 address.
+          return nil, nil, nil unless Resolv::IPv4::Regex.match?(match[:ipv4])
+        end
+
         return match[:host], match[:address], match[:port]&.to_i
       end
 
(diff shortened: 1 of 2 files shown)


jeremyevans:
I'm not sure if we want to accept hosts that are definitely incorrect.  More relaxing parsing than the RFC requires may be fine, but accepting newlines may be going a bit too far.  Are there clients are actually sending such hosts?

Is the current behavior when such hosts are given an exception?  If so, that sounds wrong, we should probably just skip that authority in such cases and use the next authority.

pvande:
I have no idea whether there are clients sending such hosts, and I would consider them broken if there were.  (You *might* be able to fudge it through `curl`.)  It *can* absolutely happen through the Ruby API, though.

The I took this approach is because the current implementation assumes no newlines (and throws exceptions otherwise), and it’s backwards-compatible with the previous implementation.

I think that using fallback authorities sounds great in principle, but would break backwards compatibility in subtle ways.  As someone who lost over half a day to debugging why a trivial Rails upgrade was causing asset serving to fail (tl;dr: Webpacker depends on Rack::Proxy (neither updated), which uses Rack::Request#host to determine which server to proxy to, and the Rails version bump also bumped Rack, returning different results in my environment), I am personally disinterested in making changes that will cause similar problems for others.

ioquatix:
How would you send a host header including newlines with HTTP/1? I think it’s invalid.

pvande:
@ioquatix There's no question that it's invalid, and you're very likely correct that it **cannot** be sent over HTTP/1.  `Rack::Request` *does* get used sometimes with data that doesn't come directly from an HTTP request, however, and retaining newlines is important for maintaining compatibility for that case, despite the absurdity.

To be clear, I really only see this as an issue for the 2.x line of releases – it seems completely reasonable to do more/better validations in a new major version, including:
* IPv6 address contains non-hex characters
* IPv6 address contains more than one elision (`::`)
* IPv6 address segment contains more that four digits
* IPv6 address contains more than eight segments
* IPv6 address contains invalid IPv4 address
* IPv4 address does not contain four segments
* IPv4 segment contains a number greater than 255
* DNS name contains a newline
* DNS name contains a colon

ioquatix:
Fair enough I get your point. Previously the parser was very loose. But now it’s a bit more strict. The key point is do we have some definition of what a host is, as in `Request#host`. My internal definition is a strong that fits the authority/host part of an absolute URI. I don’t think it was documented or specced clearly but under this definition we shouldn’t accept newline characters (or any non-printing characters actually). Because `#host` should return a string suitable for making a URI, as in other methods of `Request`.

pvande:
> #host should return a string suitable for making a URI…

That’s also a completely fair point, and I suspect that’s also a fairly
common expectation of the contract (if not its use in practice).  Given
that, the previous behavior is obviously a bug, and anyone who depended on
that behavior would most likely have ended up with a non-functional URI.

I’ll update the PR accordingly as soon as my hands are free.  👍

pvande:
The latest commit narrows acceptable inputs somewhat, and performs some additional validations on the machine-facing classes.

* Any address containing blank characters is rejected.
* Any address wrapped in square brackets is captured as an IPv6 address.
* Any address comprised entirely of decimal digits and dots is captured as an IPv4 address.
* Any other address is captured as a hostname.
* IPv6 and IPv4 captures are tested against the appropriate regular expression from `resolv` ([IPv6](https://github.com/ruby/ruby/blob/master/lib/resolv.rb#L2424-L2468), [IPv4](https://github.com/ruby/ruby/blob/master/lib/resolv.rb#L2350-L2357)), and are rejected if they do not match.
  * This seemed appropriate given the machine-facing nature of these addresses, the relaxed constraints implied by the regular expressions, and the goal of ultimately producing functional URIs.

ioquatix:
The URI RFCs specifically allow future versions of ip addresses in square brackets. Do we handle this correctly?

pvande:
Great question!

>  IP-literal = "[" ( IPv6address / IPvFuture ) "]"

> IPvFuture  = "v" 1*HEXDIG "." 1*( unreserved / sub-delims / ":" )


> If a URI containing an IP-literal that starts with "v"
(case-insensitive), indicating that the version flag is present, is
dereferenced by an application that does not know the meaning of that
version flag, then the application should return an appropriate error for
"address mechanism not supported".

This PR captures anything wrapped in square brackets as IPv6, but it
matches *anything* (non-blank).  Addresses captured as IPv6 are validated
as IPv6, which will fail for IPvFuture and return nil.  As we cannot
currently purport to understand or validate a future version of the spec,
this seems to be a credible interpretation of the RFC.

In the future, when we do need to add compatibility with an iteration of
IPvFuture, we can prepend a clause specifically for that `v` prefix, and
validate it independently.

ioquatix:
This in theory looks good to me but I need to do a closer review.  Can you please rebase on master?

ioquatix:
I agree with Jeremy on this point, and also just for the sake of efficiency, a single Regexp should be enough.

jeremyevans:
I did another review of this and I don't think all of this complexity is needed.  Handling IPvFuture seems unnecessary.  The odds are low that rack will have to deal with anything beyond IPv6.  I don't think it makes sense to allow trailing newlines in hostnames. We don't actually use the IPv4 vs. hostname distinction, so that can be removed.  We can tighten the IPv6 parsing by inlining the resolv values, so this doesn't need to depend on resolv at runtime. I submitted a pull request for my recommended approach: #1835

PR #1609, opened 2020-02-18
README: Link to the specification

XrXr:
We could alternatively use `rdoc-ref` but then Github doesn't render it
as a link. I think it's important that people have a link to the
specification when viewing the README on Github as that is the de-facto
front page of the project.


ioquatix:
Thanks great idea.

PR #1614, opened 2020-03-03
Fully qualify ::Rack::Utils usages

stevenharman:
As of ab41dccfe287b7d2589778308cb297eb039e88c6 it's expected that consumers of Rack require the full library so that `lib/rack.rb` can properly setup autoloading of constants. However, there exist apps and other libraries with dependencies on Rack which do not do that. They instead require in only the bits they think they need, e.g., `require "rack/response"`. Those consumers need to update their usage so that autoloading works as expected. But in the meantime, they can raise confusing errors like `uninitialized constant Rack::Response::Utils` because we're not fully-qualifying the usage of the `Utils` constant.

Changing these instances to also fully qualify means downstream consumer will at least get a more meaningful error: `uninitialized constant Rack::Utils`. They still need to make a change, but hopefully with a little less confusion as to why they're seeing the message in the first place.

_NOTE:_ There are a number of non-qualified usages of the `Utils` constant throughout `lib/`. I've only fully-qualified the usages in `lib/rack/request.rb` and `lib/rack/response.rb` because those are two specific modules that I've seen directly required into various apps and libraries. Perhaps we should update all usages of the `Utils` constant? Or... perhaps this entire idea is not desired and we should remove existing fully-qualified usages? I'm open to either, but we should probably shoot for consistency, which ever route we choose.

See also: https://github.com/cvonkleist/encrypted_cookie/pull/16

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -261,7 +261,7 @@ module Rack
         string = get_header(HTTP_COOKIE)
 
         unless string == get_header(RACK_REQUEST_COOKIE_STRING)
-          hash.replace Utils.parse_cookies_header(string)
+          hash.replace ::Rack::Utils.parse_cookies_header(string)
           set_header(RACK_REQUEST_COOKIE_STRING, string)
         end
 
@@ -583,7 +583,7 @@ module Rack
       end
 
       def query_parser
-        Utils.default_query_parser
+        ::Rack::Utils.default_query_parser
       end
 
       def parse_query(qs, d = '&')
--- a/lib/rack/response.rb
+++ b/lib/rack/response.rb
@@ -21,7 +21,7 @@ module Rack
     end
 
     CHUNKED = 'chunked'
-    STATUS_WITH_NO_ENTITY_BODY = Utils::STATUS_WITH_NO_ENTITY_BODY
+    STATUS_WITH_NO_ENTITY_BODY = ::Rack::Utils::STATUS_WITH_NO_ENTITY_BODY
 
     attr_accessor :length, :status, :body
     attr_reader :headers
@@ -41,7 +41,7 @@ module Rack
     # Providing a body which responds to #to_str is legacy behaviour.
     def initialize(body = nil, status = 200, headers = {})
       @status = status.to_i
-      @headers = Utils::HeaderHash[headers]
+      @headers = ::Rack::Utils::HeaderHash[headers]
 
       @writer = self.method(:append)
 


jeremyevans:
I think I would rather go the direction of removing the explicit `::Rack::` and relying on normal Ruby constant lookup.  But I don't feel especially strongly about it.  Let's see what other maintainers think.

ioquatix:
I think we should just use `require_relative` to ensure loading specific files works as expected.

jeremyevans:
@ioquatix You approved removing the related requires in #1529.  Are you sure you want to add them back?

ioquatix:
Here is my opinion:

- If autoloading worked perfectly, it would be great.
- Ruby's "autoload" feature is tricky to implement and tricky to use. Thread race conditions, etc.
- I prefer simple, predictable code, so I like `require_relative` since to me it's an expression of a dependency in it's simplest form.
- I think if you require a file, the code defined in that file should work. i.e. it should require all dependent code.

Now, the caveat to all that, is I did try `zeitwerk` and thought that it's a great idea. It automates the convention of loading constants and establishes rules for file names, constant names, etc. It's also compatible with `require_relative`.

So, thinking on this issue, I guess:

- We should fix files so that they work as per my initial outline above (e.g. add `require_relative`).
- I think as far as users are concerned, we are stuck with `autoload` in `lib/rack.rb`. But it's an optimisation, in theory we can replace it by `require_relative`.
- We should make sure our design is compatible with zeitwerk and other code loaders.

stevenharman:
@ioquatix I tend to agree. Specifically:

1. We should allow components to be required independently. e.g., a consumer should be able to `require "rack/response"`.
2. `lib/rack.rb` setting up `autoload` is an optimization to avoid loading things that are not going to be used at runtime. It also has a side effect of (maybe) making constant resolution _easier_ for developers. But that's not incompatible with the first point.

Perhaps the sweet spot/compromise we're looking for is to set up `autoload` for _components_ (where "component" is hand-wavy and undefined as of now), while each component manages its own dependencies directly via `require_relative`. The real trick there is figure out which objects and files comprise which components.

Alternately, we could do both - keep the `autoload`s as they are now in `lib/rack.rb`, while also having _components_ explicitly define their dependencies via `require_relative`. As a bonus, doing this doesn't prevent us from pairing down the `autoload` configuration later.

As far zeitwerk goes... I love it as a constant resolution, loading, and reloading tool at the application/framework level. And I'm plenty happy it's been rolled into Rails. But I'm not sure how that's all going to shake out for individual libraries. That said, I don't know that either `autoload` or `require_relative` necessarily break zeitwerk, outside of the auto-reloading maybe not working. But that feature seems more targeted at auto-reloading the application's code than that of a library.

I suppose all of this is to say, I'm a 👍 for bringing back `require_relative` from w/in the individual bits and pieces of Rack, thus enabling a consumer to `require "rack/response"` w/o having to pull in all of Rack via `require "rack"`. And I'm happy to do that work with y'all's guidance. I'm equally happy to go all in on `autoload`, but make it clear (via docs, I suppose) that requiring bits and pieces of Rack in isolation is explicitly NOT supported.

Thank you all for putting in the effort, time, and brain power to keep Rack moving forward.

ioquatix:
> I'm a 👍 for bringing back require_relative from w/in the individual bits and pieces of Rack, thus enabling a consumer to require "rack/response" w/o having to pull in all of Rack via require "rack"

I would accept a PR for this.

jeremyevans:
A PR for this would mostly be a revert of the removal of requires in 0d14bf5797ebe9ef51f55102e27b9c147c1dc283, followed by switching from `require` to `require_relative`.

ioquatix:
After reviewing this issue, I don't think we should introduce full qualification of constants. However, I'm 👍🏼 for more use of `require_relative`.

Issue #1617, opened 2020-03-10
Document thread safety


ioquatix:
I would like to start a discussion about thread safety of rack applications and defining a thread safety specification for rack 3.

Personally, I think the safest option is to define rack apps as unsafe to invoke from multiple threads. I can elaborate, but I'd be interested to hear other opinions first.

jeremyevans:
I don't think the rack specification should concern itself with thread safety.  The rack API itself is functional (does not require side effects).  Why would `proc{|env| [200, {}, []]}` be unsafe to invoke from multiple threads?  Defining all rack apps as thread-unsafe just because some rack apps are thread-unsafe seems wrong to me.

ioquatix:
> I don't think the rack specification should concern itself with thread safety.

I believe that too.

But the default assumption should be that methods are thread unsafe.

It's trivial to make a middleware that is not thread safe.

> Defining all rack apps as thread-unsafe just because some rack apps are thread-unsafe seems wrong to me.

"Defining all rack apps as thread-safe just because some rack apps are thread-safe seems wrong to me".

Do you believe a method is thread-safe or thread-unsafe by default? Because in my experience there is plenty of code which is the latter, and even some code which tries to be the former fails.

jeremyevans:
Whether a rack app is thread safe or not is a property of the app, it has nothing to do with the rack specification or library, IMO.  I don't think rack should document whether rack apps are thread-safe or not, since rack has no control over rack apps.  I suppose the furthest I would be willing to go would be to say that "rack apps may not be thread-safe", or "whether a rack app is thread-safe depends on the app".  However, while true, I don't see the point in documenting that.

It is true that it is trivial to make a thread-unsafe middleware.  However, I would respond that it is even more trivial to write a thread-safe middleware: `class M; def initialize(a) @a = a end; def call(h) @a.call(h) end; end`

ioquatix:
> I would respond that it is even more trivial to write a thread-safe middleware.

Thread safety is non-trivial when considering actual programs. A nice summary of different issues you might encounter: https://bearmetal.eu/theden/how-do-i-know-whether-my-rails-app-is-thread-safe-or-not/

> Whether a rack app is thread safe or not is a property of the app, it has nothing to do with the rack specification or library, IMO

Reentrancy is an important quality which should be documented. Either we expect rack apps to be thread safe, or we don't, in which case we can't assume any rack apps are thread safe in general.

The point of rack is to bridge web applications and servers. Either:

[...]

PR #1618, opened 2020-03-10
Remove `rack.multithread`/`rack.multiprocess`/`rack.run_once`.

ioquatix:
These variables generally come too late to be useful. Removed `Rack::Lock`
which depends on these variables.


ioquatix:
Relates to https://github.com/rack/rack/issues/1617

nateberkopec:
I just realized Puma actually _never_ set these variables correctly (we were always multithread -> true and multiprocess -> false), so I guess we can safely say no one ever paid attention, at least when running apps in Puma! 😅

ioquatix:
> I just realized Puma actually never set these variables correctly (we were always multithread -> true and multiprocess -> false), so I guess we can safely say no one ever paid attention, at least when running apps in Puma!

Setting `multithread` to true (actually should be `reentrant`), already gets you the most restrictive behaviour. So, you should be "fine" in that as long as everything is thread safe...

jeremyevans:
I'm OK with removing these, but as it is a SPEC change, we should probably get input from another maintainer.

ioquatix:
Consider this PR a draft. Let's take it as a starting point for a discussion about how to handle thread safety within Rack. Maybe we merge it, maybe we don't.

tenderlove:
I don't think I'm OK with this.  Removing `Rack::Lock` is good, but I think it's generally a good thing that an app server or middleware can know the concurrency model under which it's being executed.

tenderlove:
> Always use mutex. Uncontested mutex is not slow.

Locking an uncontested mutex is slower than doing nothing.  If the middleware knew at boot the concurrency model, it could configure a object that uses a mutex or not.  Maybe we should provide this information earlier?

jeremyevans:
@tenderlove Are you OK with removing these env variables if we provide a replacement via config.ru that applications can use to inform them of whether they are being run in multiple processes/threads/fibers/etc.  Example:

```ruby
# config.ru

require './app'

if multiprocess?
  # ...
end

unless multithread?
  App::Foo.single_threaded = true
end

run App
```

tenderlove:
@jeremyevans ya, I think so.  Maybe we could provide it to the middleware builder too?  I don't think that middleware know about the builder object, but it seems like it would be nice for middleware to know the context in which they are being built.

Maybe something like this:

```diff
diff --git a/lib/rack/builder.rb b/lib/rack/builder.rb
index 816ecf6..a37a804 100644
--- a/lib/rack/builder.rb
+++ b/lib/rack/builder.rb
@@ -151,6 +151,8 @@ module Rack
     # The +call+ method in this example sets an additional environment key which then can be
     # referenced in the application if required.
     def use(middleware, *args, &block)
+      some_context_object = self # Just as an example
+      middleware = middleware.call(some_context_object) if middleware.respond_to?(:call)
       if @map
         mapping, @map = @map, nil
         @use << proc { |app| generate_map(app, mapping) }
```

I just used `self` as an example.  It might be nice to have a dedicated object that represents the build context.

jeremyevans:
I'm OK with that, with the exception that we cannot use `call` for it.  There's probably way to much middleware that already uses `call` for other purposes, we should be using a rack specific method name for this.

tenderlove:
> There's probably way to much middleware that already uses `call` for other purposes, we should be using a rack specific method name for this.

Ya, definitely.

The reason I want this in the builder is that some apps (Rails) build middleware via the builder, not necessarily via `config.ru`.  Also, `config.ru` is where user land code goes.  I don't expect users to know that `Rack::Lock` should essentially be a noop on Unicorn, rather it would make more sense for `Rack::Lock` to remove itself from the middleware stack if the concurrency model doesn't apply.

IOW `config.ru` should be reasonably portable between webservers with minimal interaction from the user (as it is today).

ioquatix:
I can update the PR based on the discussion here.

jeremyevans:
Sorry, when I was referring to `config.ru`, I was really talking about Builder (`config.ru` is basically just an instance_eval on an Builder instance).  I agree with @tenderlove regarding `Rack::Lock`.

matthewd:
I feel somewhat attached to Rack::Lock, just because it is the answer to "I don't know how to do multithreading, but the server has just informed me [via whatever protocol we define] that it is going to do that". 

It's not terribly useful in a root application (just configure your server to use a single thread instead), but might still make sense as a solution when the lock-needing app is being mounted under an infrequently-accessed URL.

ioquatix:
```ruby
use Rack::Lock # -> Rack::Lock.call(builder)

class Rack::Lock
  def self.call(builder)
    builder.add(Rack::Lock) if builder.reentrant?
  end
end
```

matthewd:
@ioquatix perhaps, yeah (though as @jeremyevans noted, maybe not `call`). That sort of involvement during the building process feels like it'd be useful for other things too -- if only from the fact it means you get invoked, at boot time, with the context of the process/server you're going to be serving.

Or, even without the conditional, it still seems useful for us to ship it as an adapter that a user can manually reach for in order to run a non-threadsafe workload under a multi-threaded server.

ioquatix:
> though as @jeremyevans noted, maybe not call

What do you suggest?

matthewd:
Maybe something like `rack_build`, or something even longer?

It's low-level protocol, so seems safest to go for an overly-specific name. Then the `use` protocol could perhaps be something like:

```ruby
      if @map
        mapping, @map = @map, nil
        @use << proc { |app| generate_map(app, mapping) }
      end

      if middleware.respond_to?(:rack_build)
        @use << middleware.rack_build(self, *args, &block)
      else
      end
    end
```

Or maybe it's:

```ruby
        @use << proc { |app| middleware.rack_build(self, app, *args, &block) }
```

Both of those forms deprive the callee of the actual rack builder interface, though, so I guess the closer to your suggested invocation would be:

```ruby
      if middleware.respond_to?(:rack_build)
        middleware.rack_build(self, *args, &block)
      else
      end
```

(though that version might be harder to support a smooth compatible transition on)

jeremyevans:
> > though as @jeremyevans noted, maybe not call
> 
> What do you suggest?

Naming is hard. Possible method names:

* `rackup`
* `rack_configure`
* `rack_build` (as mentioned by @matthewd )

I would prefer if the API didn't require the middleware to mutate the builder object.  Also, a middleware should be able to indicate that it isn't needed in certain builder configurations.  So an API like:

```ruby
      if middleware.respond_to?(:rack_build)
        if mid = middleware.rack_build(self, *args, &block)
          @use << mid
        end
      else
      end
    end
```

The `if` part here leaving a way for the middleware to indicate it isn't needed based on the builder by returning nil/false.

ioquatix:
> I would prefer if the API didn't require the middleware to mutate the builder object.

That's a reasonable request. @tenderlove how would that fit into Rails?

@jeremyevans how would it work if you wanted to add more than one middleware, or do something more advanced like `Builder#map`, etc

matthewd:
Ah, yeah, I had thought that allowing it to return nil or an array would be a nice and fairly simple extension, then forgot to mention it. (The caller of this method is library-internal, so allowing multiple returns just adds a conditional to our code, it doesn't force implementers to handle both in the way it would if the method was SPEC-level.)

We do our own builder-like thing in https://github.com/rails/rails/blob/master/actionpack/lib/action_dispatch/middleware/stack.rb, so without actually working it through, I think the return-a-proc-that-builds-a-wrapped-app would be easier for us to work with... either passing along the real rack builder as the argument, or possibly our own extended proxy (so we could expose other Rails-specific configuration context, as well as passing along e.g. concurrency details from the underlying server).

jeremyevans:
It's possible to build a middleware that runs multiple middleware internally or does internal URL mapping, but it's not something I think should we should optimize for.  I think the purpose of middleware-configured-based-on-builder is to customize the middleware based on the builder, not do potentially arbitrary actions such as adding multiple middleware or setting up URL maps.  That's really getting into more plugin/extension territory, and I don't think we should go that far.

tenderlove:
> Both of those forms deprive the callee of the actual rack builder interface

> I would prefer if the API didn't require the middleware to mutate the builder object.

+1 to both of these.  I don't think we should give the builder object to the middleware at all.  It makes the API footprint too broad.  Passing a type of "build environment" object in is more what I have in mind.

`rack_build` seems good IMO (but I think any of those 3 method names is fine).

ioquatix:
I played around with this. It turns out unless we cannot modify the source code, such a `rack_build` method is not required. Once can just do something like `MyMiddleware.rack_build(self)`.

The value of adding some kind of magic lookup would be to make this transparent.

I've made a number of commits here to tidy up the existing `Rack::Builder` implementation and expose some predicates. It's by no means a final design but maybe time for some more discussion about the direction could be good.

The way I see this working for servers, is that they'd make a server-specific subclass of `Rack::Builder` and expose relevant predicates.

Or, they could use `Rack::Builder` and provide options to control the default predicates.

Or a combination of the two.

Thoughts and feedback welcome.

lib/rack/builder.rb
```diff
@@ -35,39 +35,6 @@ class Builder
...
-    #   Rack::Builder.parse_file('app.rb')
-    #   # requires app.rb, which can be anywhere in Ruby's
-    #   # load path. After requiring, assumes App constant
-    #   # contains Rack application
-    #
-    #   Rack::Builder.parse_file('./my_app.rb')
-    #   # requires ./my_app.rb, which should be in the
-    #   # process's current directory.  After requiring,
-    #   # assumes MyApp constant contains Rack application
-    def self.parse_file(path)
```
jeremyevans, line 62: This removes the ability for rackup to take an `.rb` file instead of an `.ru`, which breaks backwards compatibility.  It seems unrelated to `rack.multithread`/`rack.multiprocess`/`rack.run_once`.  The entire commit that removes this method seems unrelated to `rack.multithread`/`rack.multiprocess`/`rack.run_once`, and I don't think should be considered in this pull request.

lib/rack/builder.rb
```diff
@@ -93,28 +60,50 @@ def self.load_file(path)
...
+      @freeze_app = false
+
+      @options = options
+
       instance_eval(&block) if block_given?
     end
 
+    attr :options
+
+    # Whether the application server will invoke the application from multiple threads. Implies {reentrant?}.
```
jeremyevans, line 96: `multithread?` and `reentrant?` are not strictly related to the removal of `rack.multithread`/`rack.multiprocess`/`rack.run_once` and I also think they should not be part of the same pull request.

ioquatix: They are related in the sense that we need to replace with something else but sure it can be a separate PR. I'll split it up.

lib/rack/builder.rb
ioquatix: It's related in the sense that the replacement (Builder#options) cannot possibly make sense with a static middleware loaded from a `.rb` file. If we are going to continue to support this, it would make more sense for us to invoke `#call` on the class and pass in a builder instance. But we can have this in another PR, it just means there will be a point where the master branch no longer exposes any concurrency/parallelism metadata. This PR was to remove it and also propose a replacement.

ioquatix:
Okay, I'll close this PR and open two new ones.

ioquatix:
Follow up in https://github.com/rack/rack/pull/1718

PR #1620, opened 2020-03-11
Fix possible thread safety issue in Rack::Session::Pool

jeremyevans:
Internal use in Rack::Session::Pool is already thread-safe.
However, this can be called from commit_session and is a public
method, so make those calls thread-safe.

Note that Rack::Session::Pool#generate_sid did no support taking
the same arguments as the superclass #generate_sid did.  This
overrides the method to support the same arguments, but adds a
keyword argument for whether to use a mutex, which defaults to
true.

### DIFF
--- a/lib/rack/session/pool.rb
+++ b/lib/rack/session/pool.rb
@@ -36,17 +36,17 @@ module Rack
         @mutex = Mutex.new
       end
 
-      def generate_sid
+      def generate_sid(*args, use_mutex: true)
         loop do
-          sid = super
-          break sid unless @pool.key? sid.private_id
+          sid = super(*args)
+          break sid unless use_mutex ? @mutex.synchronize { @pool.key? sid.private_id } : @pool.key?(sid.private_id)
         end
       end
 
       def find_session(req, sid)
         with_lock(req) do
           unless sid and session = get_session_with_fallback(sid)
-            sid, session = generate_sid, {}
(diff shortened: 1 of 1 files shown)


ioquatix:
One other option to consider is making the private method without a lock and the public method with a lock. I don't think whether a lock is used or not should be an option.

A `Monitor` is also an option, but they are slightly slower.

jeremyevans:
I'm open to other options.  Using separate methods is tricky due to the use of super.  In general I consider Monitor a smell as it means you don't know whether you should be locking or not.  However, I'm fine with that approach if you feel better about it.  In either case, can you please push a separate PR for the approach you want and I can close this one?

ioquatix:
Agreed the use of `super` makes it tricky, and agreed `Monitor` is usually a code smell... haha

jeremyevans:
@ioquatix Are you OK with this going in?  You can switch it to a different approach later if you want.

ioquatix:
I believe we need to store the generated id into the pool before releasing the mutex otherwise we still have a race, no?

jeremyevans:
That already happens if you call `find_session` (see the `with_lock` call).  This only makes it so calling `generate_sid` manually will use the mutex, so it won't generate two separate sids.

ioquatix (approved):
I'm a little concerned about the additional complexity this introduces, but it basically seems like a net positive.

PR #1621, opened 2020-03-11
Enable Directly requiring Rack Components

stevenharman:
## Primary Change

[As discussed in #1614](https://github.com/rack/rack/pull/1614#issuecomment-594296721), the goal here is to enable a consumer of Rack to load only a subset of Rack (e.g., `require "rack/request"`) without having to pull in all of Rack via `require "rack"`. This is a common pattern for tools built to work atop or next to Rack. A prior change (0d14bf5) was made to lean more heavily on Ruby's `autoload` feature, breaking this usage pattern. This commit aims to re-introduce the `require_relative`s necessary to allow any subset of Rack to be directly loaded.

In some cases that means a file must also effectively `require_relative 'lib/rack'` because the file has dependencies on constants defined in `lib/rack.rb`. The net effect is that no other `require_relatives` would then be needed in such a file because the `autoload` configuration of `lib/rack.rb` would be in play. We've chosen to still `require_relative` those dependencies as it helps to highlight the dependencies used within the file.

## Secondary Change

A secondary change (which perhaps we want to drop here and address in another PR) is to normalize how we reference Racks constants from within Rack itself. That is, rather than fully-qualifying a dependency (e.g., `Rack::Utils` or `::Rack::Utils`), we can leverage Ruby's constant resolution strategies to use a module-relative name. For example, from anywhere with the `Rack` module, we can say `Utils`, or `Multipart::UploadedFile`.

One downside of module-relative naming is that an unloaded constant can give a confusing error, like `uninitialized constant Rack::Response::Utils`. Whereas if that constant were fully qualified the error would be `uninitialized constant Rack::Utils`, which gives the developer a better pointer to the problem.

This secondary change is WIP; I've only done it in two spots to show what it could look like. We should either adopt this across the code base, or perhaps go the other way and fully-qualify all usages by way of `::Rack::...`. I'll adjust this commit, or remove it entirely, pending further discussion.

### DIFF
--- a/lib/rack/auth/digest/md5.rb
+++ b/lib/rack/auth/digest/md5.rb
@@ -1,5 +1,6 @@
 # frozen_string_literal: true
 
+require_relative '../../utils'
 require_relative '../abstract/handler'
 require_relative 'request'
 require_relative 'params'
--- a/lib/rack/auth/digest/md5.rb
+++ b/lib/rack/auth/digest/md5.rb
@@ -99,7 +99,7 @@ module Rack
 
         def valid_digest?(auth)
           pw = @authenticator.call(auth.username)
-          pw && Rack::Utils.secure_compare(digest(auth, pw), auth.response)
+          pw && Utils.secure_compare(digest(auth, pw), auth.response)
         end
 
         def md5(data)
--- a/lib/rack/directory.rb
+++ b/lib/rack/directory.rb
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
 require 'time'
+require_relative '../rack'
+require_relative 'files'
+require_relative 'head'
+require_relative 'mime'
+require_relative 'utils'
 
 module Rack
   # Rack::Directory serves entries below the +root+ given, according to the
--- a/lib/rack/auth/abstract/request.rb
+++ b/lib/rack/auth/abstract/request.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative '../../request'
+
 module Rack
   module Auth
     class AbstractRequest
--- a/lib/rack/builder.rb
+++ b/lib/rack/builder.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+require_relative 'server'
+require_relative 'urlmap'
+
 module Rack
   # Rack::Builder implements a small DSL to iteratively construct Rack
   # applications.
--- a/lib/rack/chunked.rb
+++ b/lib/rack/chunked.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+require_relative '../rack'
+require_relative 'utils'
+
 module Rack
 
   # Middleware that applies chunked transfer encoding to response bodies
--- a/lib/rack/common_logger.rb
+++ b/lib/rack/common_logger.rb
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
+require_relative '../rack'
+require_relative 'body_proxy'
+require_relative 'utils'
+
 module Rack
   # Rack::CommonLogger forwards every request to the given +app+, and
   # logs a line in the
--- a/lib/rack/conditional_get.rb
+++ b/lib/rack/conditional_get.rb
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
+require_relative '../rack'
+require_relative 'body_proxy'
+require_relative 'utils'
+
 module Rack
 
   # Middleware that enables conditional GET using If-None-Match and
--- a/lib/rack/content_length.rb
+++ b/lib/rack/content_length.rb
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
+require_relative '../rack'
+require_relative 'body_proxy'
+require_relative 'utils'
+
 module Rack
 
   # Sets the Content-Length header on responses that do not specify
--- a/lib/rack/content_type.rb
+++ b/lib/rack/content_type.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+require_relative '../rack'
+require_relative 'utils'
+
 module Rack
 
   # Sets the Content-Type header on responses which don't have one.
--- a/lib/rack/deflater.rb
+++ b/lib/rack/deflater.rb
@@ -2,6 +2,10 @@
 
 require "zlib"
 require "time"  # for Time.httpdate
+require_relative '../rack'
+require_relative 'body_proxy'
+require_relative 'request'
+require_relative 'utils'
 
 module Rack
   # This middleware enables content encoding of http responses,
--- a/lib/rack/events.rb
+++ b/lib/rack/events.rb
@@ -1,5 +1,9 @@
 # frozen_string_literal: true
 
+require_relative 'body_proxy'
+require_relative 'request'
+require_relative 'response'
+
 module Rack
   ### This middleware provides hooks to certain places in the request /
   # response lifecycle.  This is so that middleware that don't need to filter
--- a/lib/rack/files.rb
+++ b/lib/rack/files.rb
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
 require 'time'
+require_relative '../rack'
+require_relative 'head'
+require_relative 'mime'
+require_relative 'request'
+require_relative 'utils'
 
 module Rack
   # Rack::Files serves files below the +root+ directory given, according to the
--- a/lib/rack/handler/cgi.rb
+++ b/lib/rack/handler/cgi.rb
@@ -1,5 +1,8 @@
 # frozen_string_literal: true
 
+require_relative '../rewindable_input'
+require_relative '../version'
+
 module Rack
   module Handler
     class CGI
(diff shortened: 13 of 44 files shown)


lib/rack/auth/digest/md5.rb
```diff
@@ -98,7 +99,7 @@ def valid_nonce?(auth)
 
         def valid_digest?(auth)
           pw = @authenticator.call(auth.username)
-          pw && Rack::Utils.secure_compare(digest(auth, pw), auth.response)
+          pw && Utils.secure_compare(digest(auth, pw), auth.response)
```
stevenharman: This is an example of the mentioned **Secondary Change**. Even in this one file we reference the `Utils` constant inconsistently. I can be convinced that either fully qualified or module-relative is the way to go for Rack. But we should be consistent, whichever way we go.

lib/rack/directory.rb
```diff
@@ -1,6 +1,11 @@
 # frozen_string_literal: true
 
 require 'time'
+require_relative '../rack'
+require_relative 'files'
+require_relative 'head'
+require_relative 'mime'
+require_relative 'utils'
```
stevenharman: Here is an example of the slight awkwardness mentioned in **Primary Change** in the PR description. Technically the `require_relative '../rack'` is sufficient and the other 4 `require_relative`s are unnecessary. But personally, I kind of like seeing that `Rack::Directory` has a dependency on several other parts of Rack.

What do y'all think?

jeremyevans:
I think if we are going to support this, we should make modifications to the tests so that running a single test file requires just the rack component being tested (e.g. remove the `require_relative '../lib/rack'` from `test/helper.rb`).  Then we should add a test suite that tests each test file in a separate process. Otherwise we will probably end up with a situation where one component implicitly depends on another instead of explicitly.

I'm fine with switching all Rack constant to relative requires (e.g. `Utils`, not `Rack::Utils` or `::Rack::Utils`).

ioquatix:
All looks good to me.

tenderlove:
I think this is fine, but I'm not excited about supporting two different ways for accessing Rack constants.  As @jeremyevans says, we'll have to add a test suite that makes sure we don't have implicit dependencies. 🤷‍♀️

ioquatix:
> I think if we are going to support this, we should make modifications to the tests so that running a single test file requires just the rack component being tested

This is also my preferred approach, it makes running individual tests a bit faster and isolated.

stevenharman:
Agree 💯 on updating the test suite to be isolated. Will do.

> Then we should add a test suite that tests each test file in a separate process. 

I'm game to give this a whack, if anyone has an idea for an implementation.

> I'm not excited about supporting two different ways for accessing Rack constants

I have the same instinct. I've always viewed Ruby's `autoload` as an optimization to avoid loading things into memory unless they're actually needed, and doing so on-demand. Perhaps that's an incorrect understanding or expectation? This fits pretty well with a lot of Rack and how it's typically used - via a `*.ru` or as middleware, etc...

I wonder if some of the current `autoload` config for constants that are more low-level abstractions, rather than [middleware](https://github.com/rack/rack#available-middleware-shipped-with-rack-) or [convenience helpers](https://github.com/rack/rack#convenience-). e.g., `Rack::Auth::AbstractHandler` is something you'd build atop of, rather than a module you'd plug together to configure a Rack app.

Though perhaps this is too nuanced a point and will only lead to confusion?

tenderlove:
> I wonder if we could remove the `autoload` config for constants that are more low-level abstractions, rather than [middleware](https://github.com/rack/rack#available-middleware-shipped-with-rack-) or [convenience helpers](https://github.com/rack/rack#convenience-)? e.g., `Rack::Auth::AbstractHandler` is something you'd build atop of, rather than a module you'd plug together to configure a Rack app. As such, I can see needing/expecting to directly `require "rack/auth/abstract/handler"` in custom code that builds atop it.
> 
> Though perhaps this is too nuanced a point and will only lead to confusion?

It makes sense.  The thing I'm unsure about is being the arbiter of what is "low level" and what is not.  As a maintainer I don't mind playing that role, but having to justify the strategy for loading any particular constant sounds exhausting.  It would be nicer to have one particular rule.

Are we re-introducing the ability to `require` a particular constant because of backwards compatibility?  Or is this a feature we genuinely need?

stevenharman:
> the ability to `require` a particular constant because of backwards compatibility?

Sure. Maybe? Schmaybe?

Prior to v2.2.0 (ab41dccfe287b7d2589778308cb297eb039e88c6) it was possible to directly require constants such as `Rack::Request` or `Rack::Utils` without first doing a `require "rack"`. Other libraries (like test helpers, middleware, etc...) would do this to keep the surface are of their dependency on Rack minimal. I don't know if we ever explicitly supported doing so, but it was done, and it worked. Hence... schmaybe?

It's a fair question wether or not we want to support that usage pattern. If we do, and we also want to lazily load constants via `autoload`, I'm not sure how to avoid needing an arbiter. Though perhaps some explicit expectations could help? e.g. everything should always be `require "rack/..."` capable. But middleware and [some name for the helpers] can be lazily loaded via `autoload`.

🤷‍♂

jeremy:
Feeling a maintenance headache brewing 😆

> > I wonder if we could remove the autoload config for constants that are more low-level abstractions, rather than middleware or convenience helpers?
> > Though perhaps this is too nuanced a point and will only lead to confusion?
> It makes sense.

Yeah, let's draw some lines! What's "internal"?

> file must also effectively require_relative 'lib/rack' because the file has dependencies on constants defined in lib/rack.rb. The net effect is that no other require_relatives would then be needed in such a file because the autoload configuration of lib/rack.rb would be in play. We've chosen to still require_relative those dependencies as it helps to highlight the dependencies used within the file.

Should we only use explicit deps within Rack if we go this route. Walking an error-prone path.

> A secondary change (which perhaps we want to drop here and address in another PR) is to normalize how we reference Racks constants from within Rack itself.

I'm not following the motivation for this secondary change to constant resolution. Could you expand?

lib/rack/auth/digest/md5.rb
ioquatix: I like this change.

lib/rack/directory.rb
ioquatix: I think being explicit is okay, but we should also avoid being unnecessarily explicit if it creates performance issues (i.e. lots of stats/hash lookups). Rack is loaded a lot.

ioquatix:
@stevenharman are you able to rebase this on master, as it stands I'd accept this, and based on the follow up discussion, there are some other parts we can improve `autoload` etc. However we can do that in a separate PR.

stevenharman:
@ioquatix i will try to get it rebased in the next couple of days and see how it looks.

lib/rack/directory.rb
stevenharman: Totally fair!

Please double-check my understanding: wouldn't there still only be a one-time hit for each of these `require_relative`s? For example, once `lib/rack/directory.rb` was loaded - either via the `autoload` mechanism, the `require "rack/directory"` from a consuming app, or `require_relative` from w/in another part of Rack, any further attempts to require the file be a no-op since it's already loaded? Meaning these `require_relatives` wouldn't be invoked again, thus wouldn't cause extraneous hash lookups.

stevenharman:
I just re-read the comments to load this context back into my brain. I think there are still a couple of pending questions that we might want to address before merging. Or at least have a plan for. Of note:

1. It sounds like we're in favor of removing the fully-qualified references to other `Rack` constants (e.g., we want to normalize on a module-relative `Utils` rather than `::Rack::Utils`). Shall I make that consistent across the code base with this PR? As it stands I've only addressed this inconsistency in two files where I knew it to be a problem due to common usage patterns.
2. Should we roll-back some of the explicit-but-not-necessary `require_relative` usages w/in Rack? For example: https://github.com/rack/rack/pull/1621#discussion_r791413304
3. What about the process-per-rack-file testing strategy to ensure we don't have implicit dependencies?

ioquatix:
1. Yes
2. Sure.
3. Don't worry about it for now.

jeremyevans:
> 1. It sounds like we're in favor of removing the fully-qualified references to other `Rack` constants (e.g., we want to normalize on a module-relative `Utils` rather than `::Rack::Utils`). Shall I make that consistent across the code base with this PR? As it stands I've only addressed this inconsistency in two files where I knew it to be a problem due to common usage patterns.

This is a nice to have, but not a requirement.  After all, fully-qualified references still work.

> 2. Should we roll-back some of the explicit-but-not-necessary `require_relative` usages w/in Rack? For example: [Enable Directly requiring Rack Components #1621 (comment)](https://github.com/rack/rack/pull/1621#discussion_r791413304)

Eliminating unnecessary relative requires is also a nice to have.  We need to make sure that they are unnecessary, which leads to the next point.

> 3. What about the process-per-rack-file testing strategy to ensure we don't have implicit dependencies?

We 100% need to add separate testing per file if we are going to officially support requiring individual parts.  Otherwise, there is no ability to ensure that requiring individual parts works correctly.

jeremyevans:
I've added #1801 to implement this, including testing that each part required separately still works.

We may want to keep this issue open for the switch from absolute constant references to relative constant references, though that is orthogonal to enabling directly requiring components.

ioquatix:
Awesome!

jeremyevans:
Closing as this was handled by #1801.

PR #1626, opened 2020-03-23
Don't ary.inspect in the lint assertions

### LINKED ISSUE #1625: 2.2.x Rack::Lint severe performance regression
In version 2.2.x, Rack::Lint does this check in `_call`:

```ruby
      assert("response #{ary.inspect} is not an Array , but #{ary.class}") {
        ary.kind_of? Array
      }
      assert("response array #{ary.inspect} has #{ary.size} elements instead of 3") {
        ary.size == 3
      }
```

_Even if_ the assertion is successful, the result of `@app.call(env)` is inspected, allocated, and dropped into a string that will never be used. Twice. For every middleware being linted. `@app.call(env)` returns an array of `status` (no big deal), `headers` (this is rather large once you've got stuff like Devise loaded), and `body` (hundreds to thousands of kilobytes!). So, Rack is allocating RAM for the response body at least three times, but probably dozens of times.

This resulted in this performance regression of our large production application (purple is "Middleware"):

image

I ran a transaction trace and found that each of the ~15 middlewares eats up about 1.2% of the entire request time in `#inspect`, for a total of 16%.

We've downgraded to 2.1.2 for the time being. I'm not sure if you actually need to inspect the array, but if you do, I'd suggest only doing it in development.

ziggythehamster:
Fixes #1625

### DIFF
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -48,10 +48,10 @@ module Rack
 
       ## and returns an Array of exactly three values:
       ary = @app.call(env)
-      assert("response #{ary.inspect} is not an Array , but #{ary.class}") {
+      assert("response is not an Array, but #{ary.class}") {
         ary.kind_of? Array
       }
-      assert("response array #{ary.inspect} has #{ary.size} elements instead of 3") {
+      assert("response array has #{ary.size} elements instead of 3") {
         ary.size == 3
       }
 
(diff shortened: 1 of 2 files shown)


ziggythehamster:
I'm not sure what the deal is with circleci

bpinto:
I've notice a new release for 2.x is out but it does not include this patch. Would it be possible to back port this commit into 2.x branch?

ioquatix:
There is no reason why it couldn't be back ported. I would say that this was a bug fix more than a feature/enhancement so it is a valid candidate for a 2.2.x update.

Issue #1598, opened 2020-02-11
Allow response headers to contain array of values


ioquatix:
A while ago I discussed with @tenderlove - we currently use `\n` characters to separate multiple headers. e.g.

```ruby
{"set-cookie" => "foo=x\nbar=y"}
```

Internally `HeaderHash` uses an Array I believe, e.g. something like:

```ruby
{"set-cookie" => ["foo=x", "bar=y"]}
```

Personally I like this formulation better, it's more efficient for the server implementation and it's easier from the user POV to deal with too. It seems like less of a hack, and it should minimise

So, I vote to deprecate/remove the "\n" representation and prefer the Array representation.

The only limitation is going to be if people start using it everywhere, e.g..

```ruby
{"content-type" => ["text/html"]}
```

There are different solutions to this problem: e.g., we could define only certain headers can work like this (notably `set-cookie` would make sense) or we could mark certain headers as NOT being allowed to be an array (e.g. `content-length`).

(Epic https://github.com/rack/rack/issues/1593)

boazsegev:
> I vote to deprecate/remove the "\n" representation and prefer the Array representation.

What an amazing coincidence, I almost posted an issue about this a few weeks ago.

However, after writing it down and adding benchmarks to show why it's faster (by ~40%), I realized that servers will have to support both variants (the `\n` and the Array value) due to backwards compatibility.

In the end, the positive change in the SPECs might not translate to a positive change in practice.

For me (iodine), the `\n` promises cache locality when parsing the response. However, the fact that the application has to concat strings, which requires memory reallocation and data copying... switching to an Array is much better.

> The only limitation is going to be if people start using it everywhere

This will be the same as current use case for `\n`, which can be used anywhere (including invalid places such as `content-length`).

I think this part of header "correctness" could be left to common sense.

ioquatix:
*most* headers can be split by comma. The only one which does not follow this rule that I'm aware of is `set-cookie`. In `Protocol::HTTP::Headers` this is encoded specifically in the representation. I followed this model from the Mozilla implementation: https://github.com/socketry/protocol-http/blob/master/lib/protocol/http/headers.rb#L159-L161

One way to deal with this is to simply allow `set-cookie` to be the only array, i.e. if defined it must be an array of zero or more values which get written out as individual headers.

Even thought it's kind of ugly to specify custom behaviour, at least it would make it:

- Easy for servers to do the right thing
[...]

PR #1630, opened 2020-03-30
Bring Rack::Mock cookie parsing into SPEC (fixes #1629)

### LINKED ISSUE #1629: Rack::Mock#parse_cookies_from_header does not follow SPEC
From SPEC:

> The header must respond to each, and yield values of key and value.

However, [Rack::Mock#parse_cookies_from_header](https://github.com/rack/rack/blob/master/lib/rack/mock.rb#L234) calls `has_key?` on the headers.  As such, this breaks any Rack application which takes the spec at its word, and provides another object which responds to `each` and yields values of key and value (such as, for instance, an array of two-element arrays representing header names and values).

mpalmer:
#has_key? isn't required to be supported by the object in the
`headers` element of the response coming back out of the app.
This replaces that call with a transmogrification into a
HeaderHash so we can use an index as normal.  The odd
`hash[] || ""` construction is necessary because HeaderHash#fetch
doesn't respect header name case insensitivity.

### DIFF
--- a/lib/rack/mock.rb
+++ b/lib/rack/mock.rb
@@ -231,21 +231,19 @@ module Rack
 
     def parse_cookies_from_header
       cookies = Hash.new
-      if original_headers.has_key? 'Set-Cookie'
-        set_cookie_header = original_headers.fetch('Set-Cookie')
-        set_cookie_header.split("\n").each do |cookie|
-          cookie_name, cookie_filling = cookie.split('=', 2)
-          cookie_attributes = identify_cookie_attributes cookie_filling
-          parsed_cookie = CGI::Cookie.new(
-            'name' => cookie_name.strip,
-            'value' => cookie_attributes.fetch('value'),
-            'path' => cookie_attributes.fetch('path', nil),
-            'domain' => cookie_attributes.fetch('domain', nil),
-            'expires' => cookie_attributes.fetch('expires', nil),
-            'secure' => cookie_attributes.fetch('secure', false)
-          )
-          cookies.store(cookie_name, parsed_cookie)
-        end
+      hashed_headers = Rack::Utils::HeaderHash[original_headers]
+      [hashed_headers.fetch("Set-Cookie", "")].flatten.join("\n").split("\n").each do |cookie|
+        cookie_name, cookie_filling = cookie.split('=', 2)
+        cookie_attributes = identify_cookie_attributes cookie_filling
+        parsed_cookie = CGI::Cookie.new(
+          'name' => cookie_name.strip,
+          'value' => cookie_attributes.fetch('value'),
+          'path' => cookie_attributes.fetch('path', nil),
+          'domain' => cookie_attributes.fetch('domain', nil),
+          'expires' => cookie_attributes.fetch('expires', nil),
+          'secure' => cookie_attributes.fetch('secure', false)
+        )
+        cookies.store(cookie_name, parsed_cookie)
       end
       cookies
     end
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -474,6 +474,14 @@ module Rack
       alias_method :member?, :include?
       alias_method :key?, :include?
 
+      def fetch(*args)
+        begin
+          super(args.first)
+        rescue KeyError
+          super(@names[args.first.downcase], *args[1..-1])
+        end
+      end
+
       def merge!(other)
         other.each { |k, v| self[k] = v }
         self
--- a/test/spec_mock.rb
+++ b/test/spec_mock.rb
@@ -324,6 +324,21 @@ describe Rack::MockResponse do
     res.cookie("i_dont_exist").must_be_nil
   end
 
+  it "parses cookie headers provided as an array" do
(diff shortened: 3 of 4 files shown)


mpalmer:
I'm inclined to think that the circleci error is more that someone hasn't finished configuring that job, rather than the code being broken.

ioquatix:
@tenderlove can you please disable CircleCI it's irrelevant now.

ioquatix:
`HeaderHash#fetch` in theory is an internal API.

I am happy for you to include in this PR any changes you think make sense regarding that interface in terms of making the main change more straight forward.

Maybe `HeaderHash#each_value(header_key)` which yields items from the array or tries to split it into an array would make sense. I'm open to ideas.

lib/rack/mock.rb
```diff
@@ -231,9 +231,11 @@ def cookie(name)
 
     def parse_cookies_from_header
       cookies = Hash.new
-      if original_headers.has_key? 'Set-Cookie'
-        set_cookie_header = original_headers.fetch('Set-Cookie')
-        set_cookie_header.split("\n").each do |cookie|
+      hashed_headers = Rack::Utils::HeaderHash[original_headers]
+      if hashed_headers.has_key?("Set-Cookie")
+        values = (v = hashed_headers["Set-Cookie"]).respond_to?(:to_ary) ? v.to_ary : v.split("\n")
```
ioquatix: I see, the behaviour here is a bit... messed up.

mpalmer: Yes, `HeaderHash` is a bit... under-specified, compared to the reality of headers and their oddities, as I notice has been [discussed at some length](https://github.com/rack/rack/issues/1598).  I am in violent agreement that "considering HTTP headers as general metadata is generally a mistake, in my experience. There are too many things which have specific semantic behaviour which requires specific things to handle."

I'm a bit down on `HeaderHash` overall, after discovering it doesn't handle `[["set-cookie", "foo=bar"], ["Set-Cookie", "baz=wombat"]]` correctly (though that's something for another PR).  Looking at its history, the joining of arrays by newlines has always been on a bit of an "as-needed" basis, and I'm loathe to change the behaviour in `#fetch`, because it appears to be used heavily and garbage collection, etc.

I've just pushed a new version of the PR, with an extra commit to make `HeaderHash#fetch` case-insensitive.  I also tried a different approach to handling "could be a string, could be an array of strings, who knows?" in the cookie parsing.  It's at least *differently* ugly this time.  Let me know which approach you prefer; I'm equally unhappy with both of them.

lukaso:
If you want to work around the case insensitivity error seen here plus a strange base64 encoded error is saw in a cookie, add this to our capybara configuration file:

```ruby
module Rack
  module Utils
    # Makes fetch case insensitive
    class HeaderHash
      def fetch(*args)
        patch_bug(super(args.first))
      rescue KeyError
        patch_bug(super(@names[args.first.downcase], *args[1..-1]))
      end

      private

      # this cookie from cloudflare with base64 encoded value that ends in
      # '=', kills cookie handling
      # Not patch received for this part yet.
      # '__cf_bm=somebase64encodedstringwithequalsatthened='
      def patch_bug(string)
        string.gsub(/=; /, '= ; ')
      end
    end
  end
end
```

The fix to the second error is to be added to this line:

https://github.com/rack/rack/blob/master/lib/rack/mock.rb#L260

Where `attribute_value.strip` but `attribute_value` will be `nil`.

tenderlove:
I've been trying to figure out how to remove Circle for the past 20 min.  I am going to stop now.  I'll try again later.  If anyone has ideas how to remove it I'm all ears.  I couldn't find any settings in the Circle UI, and I don't see any integrations in GitHub settings so I'm kind of at a loss right now.

ioquatix:
@tenderlove that's weird!? I took a look and can't see anything either. Maybe a webhook? Or some other integration?

olleolleolle:
@tenderlove Hope this helps:

Settings "Branches" - enter the `master` branch settings.

There will be named checkboxes there.

tenderlove:
@olleolleolle is that on the circle ci side?  Because I don't see that on the GitHub side.  Here is the GitHub side:

Branches 2020-05-21 17-10-18

CircleCI seems to think that `rack/rack` is untested:

Org settings - rack - CircleCI 2020-05-21 17-07-15

There's nothing in Integrations:

Apps and integrations 2020-05-21 17-11-29

Nothing in Webhooks:

Webhooks 2020-05-21 17-11-52

Nothing in the org level web hooks:

Webhooks 2020-05-21 17-12-41

I have no idea why this is happening.

olleolleolle:
The master branch has a button Edit which opens a list of "required CI jobs that must pass" - extremely unintuitive. @tenderlove I understand that you missed it. Thanks for the thorough look.

tenderlove:
@olleolleolle I actually clicked that edit button once and didn't see anything relevant so I didn't post a screenshot.  Here it is though:

Branch protection rule 2020-05-22 08-53-38

FWIW, newer pull requests aren't running the CircleCI build afaict.  For example [this one](https://github.com/rack/rack/pull/1648).

PR #1631, opened 2020-04-01
Request#port: Avoid warning about unused variable

olleolleolle:
This PR avoids an emitted Ruby warning.

```
.../rack-2.2.2/lib/rack/request.rb:309: warning: assigned but unused variable - authority
.../rack-2.2.2/lib/rack/request.rb:321: warning: assigned but unused variable - scheme
```

The solution here **uses the value** of two assigned variables instead of calling its getters again.

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -307,7 +307,7 @@ module Rack
 
       def port
         if authority = self.authority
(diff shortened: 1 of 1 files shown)


ioquatix:
Ah, that's probably what it should have been originally.

PR #1643, opened 2020-05-02
Add missing Logger methods to Rack::NullLogger

andrykonchin:
Despite `Rack::NullLogger` should implement the interface of `rack.logger` only currently it implements the full `Logger` interface (in was introduced [here](https://github.com/rack/rack/pull/717)).

The [interface](https://ruby-doc.org/stdlib-2.4.0/libdoc/logger/rdoc/Logger.html) of `Logger` was changed since 2014 so it makes sense to update `Rack::NullLogger` as well.

Added following empty methods to the Rack::NullLogger:
- debug!
- error!
- fatal
- info!
- warn!
- log
- reopen

### DIFF
--- a/lib/rack/null_logger.rb
+++ b/lib/rack/null_logger.rb
@@ -22,6 +22,11 @@ module Rack
     def warn? ;  end
     def error? ; end
     def fatal? ; end
+    def debug! ; end
+    def error! ; end
+    def fatal ; end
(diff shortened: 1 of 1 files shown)


ioquatix:
Thanks for this. It's my fault, I suggested some changes to `Logger` and they were merged.

PR #1644, opened 2020-05-04
Fix response status usage - convert it with to_i method

andrykonchin:
Some middleware treat response status as Integer but according to the specification it's just an object with `to_i` method which returns Integer.

Changes:
- fixed Rack::ConditionalGet
- fixed Rack::ETag

### DIFF
--- a/lib/rack/conditional_get.rb
+++ b/lib/rack/conditional_get.rb
@@ -26,7 +26,7 @@ module Rack
       when "GET", "HEAD"
         status, headers, body = @app.call(env)
         headers = Utils::HeaderHash[headers]
-        if status == 200 && fresh?(env, headers)
+        if status.to_i == 200 && fresh?(env, headers)
           status = 304
           headers.delete(CONTENT_TYPE)
           headers.delete(CONTENT_LENGTH)
--- a/lib/rack/etag.rb
+++ b/lib/rack/etag.rb
@@ -50,7 +50,7 @@ module Rack
     private
 
       def etag_status?(status)
-        status == 200 || status == 201
+        [200, 201].include?(status.to_i)
       end
 
       def etag_body?(body)


andrykonchin:
Actually I am thinking about general approach *in tests* to ensure middleware follow the specification and use only declared interface of response status, headers and body.

For instance in the tests all the applications can wrap a response with some helper and convert status, headers and body to objects which respond to declared method e.g. 'to_i' and `each`:

```ruby
app = lambda { |env| Rack::Lint::Response.wrap(201, { 'Content-Type' => 'text/plain' }, ["Hello, World!"]) }

module Rack
  class Lint
    module Response
      def self.wrap(status, headers, body)
        [status.to_s, headers.to_enum, body.to_enum]
      end
    end
  end
end
```

This is the simple implementation and it isn't strict as it could be but it's enough to catch issues with treating status as Integer, headers as Hash and body as Array.

If it's reasonable approach I will be happy to create a PR and update the specs.

jeremyevans:
I'm not sure how other maintainers feel about this, but I think in Rack 3 we should change SPEC to require the status to be an Integer instead of requiring that it can be converted to an integer.  Having all middleware perform conversion to get the status is inefficient, and I can't see a significant advantage in supporting non-Integer status values.

We already have a similar pull request to make the response headers be a hash (#1597).

ioquatix:
I agree status should be an integer, I don't know any use case that benefits from this not being the case. Any counter arguments?

ioquatix:
@tenderlove can we get your historical context on this?

tenderlove:
> @tenderlove can we get your historical context on this?

Honestly I think it's legacy.  I agree it should just be an integer and we should fix SPEC

ioquatix:
@olleolleolle if you are interested to have a go at helping, this one is pretty low hanging fruit (updating the SPEC).

ioquatix:
Fixed by https://github.com/rack/rack/pull/1662

PR #1645, opened 2020-05-05
Ensure some cookies don't cause failure

lukaso:
Cookies can on occasion have base 64 encoded strings as their value. Base 64 encoded strings tend to end in '=' as this is a filler character. Cookies of this form cause rack to fail.

An example is `__cf_bm=_somebase64encodedstringwithequalsatthened=; array=awesome`.

This fixes the bug.

Alternative approach would be that in the function I've patched, to only parse the headers required  in the enclosing function.

### DIFF
--- a/lib/rack/mock.rb
+++ b/lib/rack/mock.rb
@@ -254,7 +254,7 @@ module Rack
       cookie_bits = cookie_filling.split(';')
       cookie_attributes = Hash.new
       cookie_attributes.store('value', cookie_bits[0].strip)
-      cookie_bits[1..].each do |bit|
+      cookie_bits.drop(1).each do |bit|
         if bit.include? '='
           cookie_attribute, attribute_value = bit.split('=',2)
           cookie_attributes.store(cookie_attribute.strip, attribute_value.strip)
--- a/lib/rack/mock.rb
+++ b/lib/rack/mock.rb
@@ -254,10 +254,10 @@ module Rack
       cookie_bits = cookie_filling.split(';')
       cookie_attributes = Hash.new
       cookie_attributes.store('value', cookie_bits[0].strip)
-      cookie_bits.each do |bit|
+      cookie_bits[1..].each do |bit|
         if bit.include? '='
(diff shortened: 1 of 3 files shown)


lib/rack/mock.rb
```diff
@@ -257,7 +257,7 @@ def identify_cookie_attributes(cookie_filling)
       cookie_bits.each do |bit|
         if bit.include? '='
           cookie_attribute, attribute_value = bit.split('=')
```
ioquatix, line 259: Would it make mores sense to use `split('=', 2)`?

lukaso: Yes, also works. Either way, the results of this function are discarded other than for these areas:
https://github.com/rack/rack/pull/1645/files#diff-4fc9bc1f7d91630f4f9f47fc6663f3f7L240-L245

It is probably incorrect to assume '=' is a divider.

A third option is not to process the first element of the `cookie_bits` array at all, since this is the value. So `cookie_bits[1..].each do |bit|`.

Fourth option, do both since the value could possibly also contain '='.

lukaso: Done both in f48f0b83

lukaso:
BTW: the build failures appear to be due to this: https://github.com/rubygems/rubygems/issues/3570

lib/rack/mock.rb
```diff
@@ -254,9 +254,9 @@ def identify_cookie_attributes(cookie_filling)
       cookie_bits = cookie_filling.split(';')
       cookie_attributes = Hash.new
       cookie_attributes.store('value', cookie_bits[0].strip)
-      cookie_bits.each do |bit|
+      cookie_bits[1..].each do |bit|
```
tenderlove, line 257:
```suggestion
      cookie_bits.drop(1).each do |bit|
```

I don't think endless ranges work on Ruby 2.5, but `drop(1)` should do the same thing.

lukaso:
> This seems good, but we need to make sure the build passes. It looks like there is a legit failure in 2.5

Done in c17569c

PR #1648, opened 2020-05-11
Refactor Multipart::Parser parse loop logic

fatkodima:
During refactoring, I also removed `#on_read` method. 
I'm wondering, am I allowed to do that? It seems for internal usage only.


tenderlove:
@fatkodima ya, it's fine

PR #1658, opened 2020-05-24
Remove servers

ioquatix:


ioquatix:
@tenderlove I could not remove the CGI handler without breaking a ton of specs, so I have left it in place for now.

However, one thing it did bring attention to is how ancient and "specific" the `Rack::Server` and `Rack::Handler` code is. For example, some of the specs are testing webrick by sending interrupts to the current process, while running webrick in a thread. It's not particularly elegant.

I have a moderately strong feeling to remove rackup to it's own gem, along with the remaining handlers and tests.

Good because:

- Smaller surface area for core rack (less potential for CVEs)
- Easier to maintain.

Bad because:

- Users cannot use rackup (does anyone use it much?) without installing separate gem.

Thoughts?

jeremyevans:
I'm against moving rackup to another gem.  I think you would be surprised how many users use rackup.

Have you setup a gem for the removed servers?  If not, I expect we'll receive pushback when Rack 3.0 is released.

ioquatix:
> Have you setup a gem for the removed servers? If not, I expect we'll receive pushback when Rack 3.0 is released.

We discussed this and you said not to bother.

> I don't think there is need to do that. We may want to reach out to the gem authors of the handlers and ask them to include the rack handler in their library if they want to support it. If not, users that want to continue using the handlers can copy the old version and maintain their own gem.

jeremyevans:
Ah, OK. :)

ioquatix:
I guess we should let @macournoyer know this has finally landed.

I was added as a maintainer of thin but I have little time to support it. That being said, I think there are a few options to consider:

- Someone can invest time to support and maintain thin, and copy the (short) handler code: https://github.com/rack/rack/blob/19149461e480e509b7deaff4f1574f18e3402bdf/lib/rack/handler/thin.rb
- Archive `thin`. It was a great project and it has lived it's life well.

While we would try to minimise breaking changes to server code in Rack 3 (nightmare for server maintainers to support Rack 1 - 3 if there are breaking changes), it's likely **some** changes will be required. So, if we can't find maintainer/support for thin, I recommend archive it to avoid users getting frustrated.

macournoyer:
No worries, I’ll take care of update Thin. Thanks for the heads up!

PR #1662, opened 2020-05-24
SPEC: Require "status" to be Integer and >= 100 (BREAKING CHANGE)

olleolleolle:
This PR adds a SPEC change: "require Status to be an Integer >100".

The change is: check `status.is_a?(Integer)` before trying the >=.

- Tests which tested the previous, lenient `#to_i`-accepting behavior: dropped

See comment in #1644

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file. For info on
 
 ### Changed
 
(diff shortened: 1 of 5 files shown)


ioquatix:
Great work, can you please add an entry to the changelog as this is a breaking change.

olleolleolle:
@ioquatix Not many BREAKING CHANGE markers in there, I imitated one I found.

ioquatix:
Thanks for your effort.

PR #1663, opened 2020-05-24
Remove options from `Builder#parse_file` & `Builder#load_file`.

ioquatix:


ioquatix:
There are lots of tests that write `app, _ = Builder.parse_file`. I've left them. The `_` now becomes `nil` which I think is acceptable for backwards compatibility.

Can you see anything else we need to do?

cc @tenderlove

PR #1664, opened 2020-05-24
Update to RuboCop 0.84.0: Use new name for a rule; auto-correct issues

olleolleolle:
This PR makes RuboCop 0.84.0 pass on master.

## Background

I ran `bundle install`, then looked in the Gemfile, saw RuboCop there, and tried to run it.

- Use a new name for one of the rules
- Run `rubocop -a` to auto-fix outstanding issues

Why this didn't work at first: external tests - the external projects may run a `rubocop` Rake task, which looks for additional .rubocop.yml files in parent directories.

**Workaround:** To work around that, I added a commit which renames Rack's `.rubocop.yml` (for the duration of the external tests).

## Notes

- 🔧  **2.3 syntax** is **no longer supported** by RuboCop. Perhaps hastily, I picked 2.4, to be able to run the check with that latest RuboCop release

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,5 +1,5 @@
 AllCops:
-  TargetRubyVersion: 2.3
+  TargetRubyVersion: 2.4
   DisabledByDefault: true
   Exclude:
     - '**/vendor/**/*'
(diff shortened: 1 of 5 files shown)


ioquatix:
Thanks!

PR #1667, opened 2020-05-28
Fix using Rack::Session::Cookie with coder: Rack::Session::Cookie::Base64::{JSON,Zip}

### LINKED ISSUE #1666: Rack::Session::Cookie no longer round-trips with JSON coder
I'm upgrading an application currently using Rack 1.6.11 to the latest Rack 2.2.2. I'm running in to an issue with `Rack::Session::Cookie` configured with `coder: Rack::Session::Cookie::Base64::JSON.new`:
```
NoMethodError:
       undefined method `public_id' for #<String:0x00007f868b8594e0>
       Did you mean?  public_send
# ./.bundle/gems/rack-2.2.2/lib/rack/session/abstract/id.rb:458:in `[]'
```

I've reduced this down to:
``` ruby
require "rack"
# Rack::Session::Cookie needs this but doesn't require it?
require "delegate"

# our app
app = Proc.new do |env|
  # write something to the session to create the cookie on first call
  env["rack.session"]["foo"] = "bar"
  # read session ID to trigger bug on second call
  env["rack.session"]["session_id"]
  [200, {}, ["ok"]]
end

# wrap app in cookie session middleware, with json coder
wrapped = Rack::Session::Cookie.new(app,
  secret: "foo",
  coder: Rack::Session::Cookie::Base64::JSON.new)

# first call to get a cookie
env = {}
status, headers, body = wrapped.call(env)
cookie = headers["Set-Cookie"].split(";").first

# trigger the bug
env = {"HTTP_COOKIE" => cookie}
wrapped.call(env)
```

This worked fine with 1.6.11 (and presumably before, although I haven't tested), and with 2.0.1 through 2.0.7.

1.6.12 and 1.6.13 fail because the json library used (`OkJson`) refuses to encode `Rack::Session::SessionId`.
2.0.8 fails with `RuntimeError` and no message.
2.0.9 is the start of the `NoMethodError: undefined method 'public_id'` error.

jeremyevans:
This is an issue because SessionId doesn't round trip through JSON.
However, it probably has the same security issue as before SessionId
was introduced.

It may be better to eliminate the session id completely for cookie
sessions, since there is no reason cookie sessions need an id (an
id is only needed for memcache/memory/database sessions).

Fixes #1666

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file. For info on
 
 ### Fixed
 
+- Fix using Rack::Session::Cookie with coder: Rack::Session::Cookie::Base64::{JSON,Zip}. ([#1666](https://github.com/rack/rack/issues/1666), [@jeremyevans](https://github.com/jeremyevans))
 - Avoid NoMethodError when accessing Rack::Session::Cookie without requiring delegate first. ([#1610](https://github.com/rack/rack/issues/1610), [@onigra](https://github.com/onigra))
 - Handle cookies with values that end in '=' ([#1645](https://github.com/rack/rack/pull/1645), [@lukaso](https://github.com/lukaso))
 
--- a/lib/rack/session/abstract/id.rb
+++ b/lib/rack/session/abstract/id.rb
@@ -455,7 +455,12 @@ module Rack
           def [](key)
             if key == "session_id"
               load_for_read!
-              id.public_id if id
+              case id
+              when SessionId
+                id.public_id
+              else
+                id
+              end
             else
               super
             end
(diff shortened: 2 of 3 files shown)


ioquatix:
`case` statements around what should probably just be an inheritance/method override seems like a fragile design.

We fundamentally have two different kinds of sessions - a session where all data is stored on the client, and a session where the client only stores an id and we have an internal database of session data. Do we model this correctly? Is this problem caused by modelling it incorrectly?

jeremyevans:
I agree that this design is suboptimal.  There should no reason to force use of a session id for cookie sessions.  Roda's sessions plugin, which also uses cookies, does not use session ids.

However, in terms of fixing #1666, this appears to me to be the simplest approach.  We could drop the session id for cookie sessions, but I'm guessing people are using it even though it is not strictly needed.

I'm not sure there is a security issue with session ids when using cookie sessions, since the session id isn't actually used, unlike for memcache/memory/database sessions.

PR #1672, opened 2020-06-11
Make Rack::ETag not set ETag for no-cache responses by default

### LINKED ISSUE #1619: Version 2.2.x breaks Rails Actioncontroller:Live streaming
When you upgrade rack from 2.1.x to 2.2.x the streaming functionality in rails stops working.
A simple controller like this:

```ruby
class MyController < ActionController::Base
  include ActionController::Live

  def index

    response.headers["X-Accel-Buffering"] = "no"
    response.headers['Content-Type'] = 'text/event-stream'

    5.times do |i|
      logger.info "stream"

      response.stream.write "hello #{i}\n"
      sleep 1
    end
  rescue ActionController::Live::ClientDisconnected => e
    # ignore client disconnections
    nil
  ensure
    response.stream.close
  end
  
end
```
will not stream anymore and will send all the data at the end of the cicle.
After many attempts I've pinned down that the culprit is Rack 2.2.x.

with version 2.2.0 if you invoke that controller will raise an exception:

```
#<NoMethodError: undefined method `reject' for nil:NilClass>
ruby-2.6.5@mod/gems/rack-2.2.0/lib/rack/request.rb:623:in `reject_trusted_ip_addresses'
ruby-2.6.5@mod/gems/rack-2.2.0/lib/rack/request.rb:363:in `ip'
ruby-2.6.5@mod/gems/actionpack-5.1.2/lib/action_dispatch/http/request.rb:250:in `ip'
ruby-2.6.5@mod/gems/railties-5.1.2/lib/rails/rack/logger.rb:51:in `started_request_message'
ruby-2.6.5@mod/gems/railties-5.1.2/lib/rails/rack/logger.rb:35:in `block in call_app'
ruby-2.6.5/lib/ruby/2.6.0/logger.rb:465:in `add'
```

with version 2.2.1 no exception is raised but the streaming does not work. same on version 2.2.2

I've tested it with Rails 5.1.x and 6.0.2.1 (the latest). Same behaviour.

jeremyevans:
This reverts commit 0371c69a0850e1b21448df96698e2926359f17fe,
"Remove check for Cache-Control: no-cache when generating etag".

It makes the previous behavior opt-in using a skip_etag_if_no_cache: false
keyword argument to Rack::ETag, and defaults to the historical behavior of
not setting the ETag if Cache-Control contians no-cache.

Fixes #1619

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file. For info on
 - Relax validations around `Rack::Request#host` and `Rack::Request#hostname`. ([#1606](https://github.com/rack/rack/issues/1606), [@pvande](https://github.com/pvande))
 - Removed antiquated handlers: FCGI, LSWS, SCGI, Thin. ([#1658](https://github.com/rack/rack/pull/1658), [@ioquatix](https://github.com/ioquatix))
 - Removed options from `Rack::Builder.parse_file` and `Rack::Builder.load_file`. ([#1663](https://github.com/rack/rack/pull/1663), [@ioquatix](https://github.com/ioquatix))
+- Rack::ETag will no longer set ETag for no-cache responses, unless skip_etag_if_no_cache: false keyword is used. ([#1619](https://github.com/rack/rack/issues/1619), [@jeremyevans](https://github.com/jeremyevans))
 
 ### Fixed
 
--- a/lib/rack/etag.rb
+++ b/lib/rack/etag.rb
@@ -10,17 +10,25 @@ module Rack
   # a sendfile body (body.responds_to :to_path) is given (since such cases
   # should be handled by apache/nginx).
   #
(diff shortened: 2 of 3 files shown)


ioquatix:
I want to check this before we merge it, because while I respect that it's broken, and we probably don't want to break Rails, I also want to know how we move forward with this in Rack 3.0 since ultimately I don't think `Rack::ETag` is doing the wrong thing here.

jeremyevans:
I'm not in favor of merging this.  I merely added it so that if we want to revert to the historical behavior for compatibility, we could use this approach to allow for the more correct current behavior.

ioquatix:
I would be okay considering this as a fix for Rack 2.x but I think we need a better model for describing streaming responses because this won't be the only issue with streaming responses in Rack.

Pencab:
https://github.com/rack/rack/pull/1672#issuecomment-643295672

PR #1687, opened 2020-07-14
Make Rack::MockResponse handle non-hash headers

### LINKED ISSUE #1629: Rack::Mock#parse_cookies_from_header does not follow SPEC

jeremyevans:
Rack::MockResponse inherits from Rack::Response, which already
uses a HeaderHash for the headers.  The original_headers were
only used for cookie parsing, which for some reason was happening
before the call to super in initialize (so the headers weren't
available yet). There seems to be no reason why the cookie parsing
can't happen after the call to super, in which case we can use the
headers directly.

Fixes #1629
Fixes #1630

Co-authored-by: Matt Palmer <mpalmer@hezmatt.org>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file. For info on
 
 ### Fixed
 
+- Make Rack::MockResponse handle non-hash headers. ([#1629](https://github.com/rack/rack/issues/1629), [@jeremyevans](https://github.com/jeremyevans))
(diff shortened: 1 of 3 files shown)


vitoravelino:
It would be nice to have either this one or #1630 merged to fix the issues.

jeremyevans:
@vitoravelino Thanks for the reminder.  I'll merge this now.

PR #1690, opened 2020-07-14
Make Rack::Request#{[],[]=} warn even in non-verbose mode

### LINKED ISSUE #1277: Clarification regarding `Rack::Request#[]`
This method issues a warning that it will be deprecated.

But it's used in a lot of code.

What is the reason for it's deprecation and what is the timeline for removal?

I couldn't find any documentation regarding this.

It's a convenient shortcut.

jeremyevans:
Additionally, use the :uplevel keyword to warn, so it includes
the caller information in the warning output (similar to rb_warn).
This only works on Ruby 2.5+, so output will look a little odd on
Ruby 2.3/2.4, but I don't think it warrants separate handling in
older versions.

Fixes #1277


ioquatix:
Would it make more sense to remove these methods for Rack 3.0? Rack 4.0?

jeremyevans:
They were only warned in non-verbose mode before, so going directly to removed is likely to break things.  We can probably remove them in Rack 3.1 or 4.0.

PR #1691, opened 2020-07-14
Remove Rack::HTTP_VERSION and don't set HTTP_VERSION in env in cgi and webrick

### LINKED ISSUE #970: Stop setting HTTP_VERSION to SERVER_PROTOCOL
I'm opening this to discuss the merits and reasonings for setting HTTP_VERSION inside of the different rack handlers.

See my "Notes" under #969 for a little bit of context.

tl;dr is in all of the handlers, rack does this little bit:

``` ruby
env[HTTP_VERSION] ||= env[SERVER_PROTOCOL]
```

This causes an environment variable to get set to for `HTTP_VERSION`, when according to the CGI spec, `HTTP_*` is explicitly reserved for client headers. This makes it confusing if something downstream is attempting to collect all client headers, this becomes ambiguous. Did the client sent a `Version` header? Or is this the `Version` header that Rack is declaring?

This behavior is wrong, but I'm not sure of the ramification of simply... stopping doing this. I can't seem to find things that reference this, except the internals that were fixed up in #969, but no idea how common this is downstream to actually use this value.

jeremyevans:
Fixes #970


ioquatix:
Excellent.

Issue #1692, opened 2020-07-16
Add spec for early hints


casperisfine:
Both `puma` and `falcon` now expose `rack.early_hints`, and I'm working on adding support for it in `unicorn`.

However I can't see it in the rack spec.

jeremyevans:
That's because it isn't currently an official rack spec.  However, considering the support in falcon, puma, and unicorn on the webserver end and rails, roda, and hanami on the web framework end, it seems reasonable to add to the spec.  Can you submit a pull request to update lib/rack/lint.rb appropriately for early hints (SPEC.rdoc is generated from lib/rack/lint.rb, see rake spec task).

casperisfine:
> Can you submit a pull request

I'll try my hand at it early next week.

ioquatix:
I have some issues with how early hints is currently implemented. So, I want to ensure that this is implemented in a generic and useful way. Please make sure there is time for me to review before merging.

ioquatix:
When I implemented this in falcon, I found the interface defined by Puma somewhat HTTP/1 specific.

https://www.codeotaku.com/journal/2019-02/falcon-early-hints/index

Can you check what I wrote?

byroot:
> Can you check what I wrote?

I did when I added Early Hints support to Unicorn.

Quick answer because I'm on mobile, I can elaborate more tomorrow if needed.

I think your blog post assume that 103 Early Hints sole goal is to be translated into a push promise by the reverse proxy (h2o most likely).

It is true that today that's pretty much all you can do with it.

However since HTTP/2 server push has been implemented, there has been a lot of experimentations that showed that because the server doesn't know the state of the browser cache, often pushing the assets is slower than letting the browser request it as the browser can't cancel the push if it's useless.

So I believe that when browsers will start to support 103 early hints, these will likely exhibit better performance than push promises.

So maybe what you are looking for is a distinct API specifically for pushes.

ioquatix:
That all makes sense.

So with that in mind, what I'm saying is, we should not make an API which is tied to the format of the HTTP/1 headers.

I'm less concerned about functionality. Yes great to support both push and 1xx informal responses. That being said, I'd need to actually implement it in HTTP/2 to provide feedback. Do any clients actually support it?

https://bugs.chromium.org/p/chromium/issues/detail?id=671310

[...]

PR #1695, opened 2020-07-20
Add a linter for rack.early_hints

casperisfine:
Fix: https://github.com/rack/rack/issues/1692

This is kind of a first stab, I'm not familiar enough with `Rack::Lint` to understand exactly what's needed. Please let me know if there are specifications I missed, or if the main description should be improved.

cc @jeremyevans

### DIFF
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -330,6 +330,8 @@ module Rack
         check_error env[RACK_ERRORS]
         ## * There may be a valid hijack stream in <tt>rack.hijack_io</tt>
         check_hijack env
+        ## * There may be a valid early hints callback in <tt>rack.early_hints</tt>
+        check_early_hints env
 
         ## * The <tt>REQUEST_METHOD</tt> must be a valid token.
         unless env[REQUEST_METHOD] =~ /\A[0-9A-Za-z!\#$%&'*+.^_`|~-]+\z/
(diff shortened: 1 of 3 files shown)


lib/rack/lint.rb
```diff
@@ -655,6 +657,29 @@ def check_hijack_response(headers, env)
...
+    ##
+    ## The application or any middleware may call the <tt>rack.early_hints</tt>
+    ## with pairs of headers that will be sent as a 103 Early Hints provisional
+    ## response.
+    def check_early_hints(env)
+      if env[RACK_EARLY_HINTS]
+        ##
+        ## If rack.early_hints is present it must respond to #call.
+        assert("rack.early_hints must respond to call") { env[RACK_EARLY_HINTS].respond_to?(:call) }
+        original_callback = env[RACK_EARLY_HINTS]
```
jeremyevans, line 670: This local variable is not used.  It looks like other than checking `env['rack.early_hints']` responds to call, the value is ignored.  Shouldn't `original_callback` be called with the headers inside the lambda below?  Otherwise `Rack::Lint` breaks applications using early hints?

lib/rack/lint.rb
```diff
@@ -655,6 +657,29 @@ def check_hijack_response(headers, env)
...
+        env[RACK_EARLY_HINTS] = lambda do |headers|
+          ## The headers must respond to #each_pair
+          assert("rack.early_hints headers must respond to each_pair") { headers.respond_to?(:each_pair) }
+          headers.each_pair do |key, value|
+            ## The headers keys and values must respond to #to_s
```
jeremyevans, line 675: Period at the end of sentences in Lint comments, since they are combined in SPEC.  Otherwise I think RDoc makes them appear as a single sentence.

It would also be better to be more descriptive. `The rack.early_hints argument must respond to #each_pair and yield a key string and a value string to the provided block for each header.`  This would change the specification slightly (`is_a?(String)` instead of `respond_to?(:to_s)`).

ioquatix:
I'm okay with this, but I think we should defer to the response headers rather than re-specifying it to avoid duplicating the specifications.

jeremyevans:
With the changes to SPEC since this was originally submitted, I think this will require significant changes before merging.  `rack.early_hints` should be called with an argument that would be valid as response headers.  The Lint implementation can call `check_headers` with the value.

@casperisfine Unless you want to work on this, I can see if I can update it tomorrow.

casperisfine:
I don't absolutely want to work on it, feel free to take over if you wish, but otherwise I can probably do that in a day or two.

casperisfine:
Sorry for the delay, I rebased and replaced the content of `check_early_hints` by `check_headers`.

It's been a while, so I hope the result is decent. Again feel free to directly take over if you'd rather quickly improve it rather than go through review rounds.

casperisfine:
Ah that great!

PR #1697, opened 2020-07-21
Parsing behavior change from foo[] to "foo" => []

### LINKED ISSUE #1696: Change nested query parsing behavior of "foo[]" from {"foo" => [nil]} to {"foo" => []}
Hello lovely folks,

first and foremost thanks for your excellent work on this project practically the entire ruby webdev eco system relies on :green_heart: :tada: :rocket: 

## Problem

I've come here today to propose a change in the parsing behavior of the query string `param[]`.

There is a spec asserting the current behavior so I assume it's intentional
https://github.com/rack/rack/blob/649c72bab9e7b50d657b5b432d0c205c95c2be07/test/spec_utils.rb#L164-L165

I believe/hope the more correct test would be:

```ruby
    Rack::Utils.parse_nested_query("foo[]").
      must_equal "foo" => []
```

I base this on three things:

* we otherwise have no way to indicate an empty array and an array with `nil` doesn't seem particularly useful (or intuitive) vs. an empty array
* the introduction was 12 years ago (  6e330bdda5b003f4c2ed1b20af47c05c544 ) and might be a side effect of the implementation due to the splitting behavior exhibited here: https://github.com/rack/rack/blob/649c72bab9e7b50d657b5b432d0c205c95c2be07/lib/rack/query_parser.rb#L67-L72
* the popular HTTP client library faraday (which `her` and many projects rely upon) changed it's way of serializing empty arrays to the query string above https://github.com/lostisland/faraday/pull/801

Especially due to the last point this creates a sort of dissonance in the Ruby eco system where I send an empty array from the one side and it arrives as `[nil]` on the server side. Of course, the fix might also be on the faraday side.

I have not yet checked the HTTP specs or anything about this to see what's "officially correct"

To have complete traceability, I came here via https://github.com/ruby-grape/grape/issues/2079 :) 

## Implementation

I'm not sure if it's the complete implementation but it seems to be easily fixable in https://github.com/rack/rack/blob/649c72bab9e7b50d657b5b432d0c205c95c2be07/lib/rack/query_parser.rb#L102-L105

all it needs is `unless v.nil?`:

```ruby
      elsif after == "[]"
        params[k] ||= []
        params[k] << v unless v.nil?
```

All tests (with exception of the one mentioned above) pass with this change.

(I have the code, I found the old spec while implementing this/my spec was elsewhere)

Happy to make a PR should it be wanted!

Thanks for all your work! :tada: :rocket:

PragTob:
Pending discussion in #1696 and if merged also fixes #1696

### DIFF
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -102,7 +102,7 @@ module Rack
       elsif after == "[]"
         params[k] ||= []
         raise ParameterTypeError, "expected Array (got #{params[k].class.name}) for param `#{k}'" unless params[k].is_a?(Array)
-        params[k] << v
+        params[k] << v unless v.nil?
       elsif after =~ %r(^\[\]\[([^\[\]]+)\]$) || after =~ %r(^\[\](.+)$)
         child_key = $1
         params[k] ||= []
--- a/test/spec_utils.rb
+++ b/test/spec_utils.rb
@@ -162,7 +162,7 @@ describe Rack::Utils do
       must_equal "pid=1234" => "1023", "a" => "b"
 
     Rack::Utils.parse_nested_query("foo[]").
-      must_equal "foo" => [nil]
+      must_equal "foo" => []
     Rack::Utils.parse_nested_query("foo[]=").
       must_equal "foo" => [""]
     Rack::Utils.parse_nested_query("foo[]=bar").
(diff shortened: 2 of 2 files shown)


ioquatix:
Closing in favour of #1699. Thanks for your efforts.

PragTob:
totes, thanks a lot for getting back to this!

PR #1699, opened 2020-07-21
Make query string parsing conform to URL spec

### LINKED ISSUE #1696: Change nested query parsing behavior of "foo[]" from {"foo" => [nil]} to {"foo" => []}

jeremyevans:
The URL spec section 5.1.3.3 specifies that if = is not present
in the byte sequence, it should be treated as if the byte sequence
is the name of the tuple and the value is the empty string.

This affects all parameters without =, not just arrays:

```ruby
Rack::Utils.parse_nested_query("foo[bar]&baz[]&quux")
{"foo"=>{"bar"=>nil}, "baz"=>[nil], "quux"=>nil} # Before
{"foo"=>{"bar"=>""}, "baz"=>[""], "quux"=>""}    # After
```

Fixes #1696

### DIFF
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -83,6 +83,7 @@ module Rack
       name =~ %r(\A[\[\]]*([^\[\]]+)\]*)
       k = $1 || ''
       after = $' || ''
+      v ||= String.new
 
       if k.empty?
         if !v.nil? && name == "[]"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ All notable changes to this project will be documented in this file. For info on
 ### Changed
 
 - BREAKING CHANGE: Require `status` to be an Integer. ([#1662](https://github.com/rack/rack/pull/1662), [@olleolleolle](https://github.com/olleolleolle))
+- BREAKING CHANGE: Query parsing now treats parameters without `=` as having the empty string value instead of nil value, to conform to the URL spec. ([#1696](https://github.com/rack/rack/issues/1696), [@jeremyevans](https://github.com/jeremyevans))
 - Relax validations around `Rack::Request#host` and `Rack::Request#hostname`. ([#1606](https://github.com/rack/rack/issues/1606), [@pvande](https://github.com/pvande))
 - Removed antiquated handlers: FCGI, LSWS, SCGI, Thin. ([#1658](https://github.com/rack/rack/pull/1658), [@ioquatix](https://github.com/ioquatix))
 - Removed options from `Rack::Builder.parse_file` and `Rack::Builder.load_file`. ([#1663](https://github.com/rack/rack/pull/1663), [@ioquatix](https://github.com/ioquatix))
--- a/test/spec_utils.rb
+++ b/test/spec_utils.rb
@@ -135,7 +135,7 @@ describe Rack::Utils do
 
   it "parse nested query strings correctly" do
     Rack::Utils.parse_nested_query("foo").
-      must_equal "foo" => nil
+      must_equal "foo" => ""
     Rack::Utils.parse_nested_query("foo=").
       must_equal "foo" => ""
     Rack::Utils.parse_nested_query("foo=bar").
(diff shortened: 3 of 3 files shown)


ioquatix:
The implications of this change are:

1. Can no longer specify nil values in any way (maybe a good thing?)
2. Existing code that depends on nil may be broken in subtle ways.

jeremyevans:
You are correct about the implications. In general, you should omit the parameter if it should have a nil value.  That's more of an implicit nil, this would remove the explicit nil.

Correct that this will change behavior and can break things that depend on nil values.  That's why I labeled it a breaking change in the CHANGELOG.

Looks like the result of `Rack::Utils.build_query({x: [nil, nil, nil]})` is bad.  It should be `x[]=&x[]=&x[]=` or `x[]&x[]&x[]` or the empty string (if we drop elements with nil values).  That's not strictly related to parsing, so it should probably be fixed in a separate pull request.  nil won't roundtrip through query parsing, but neither do empty arrays, empty hashes, numbers, booleans, etc.  Anything that needs more than what query parsing provides should probably use JSON.

A somewhat informal explanation of query parsing types using an algebraic data types:

A QueryParseResult (result of `Rack::Utils.parse_nested_query`) is a hash with String keys and QueryParseValue values.

A QueryParseValue is either a:

* String
* QueryParseResult
* Array of String|QueryParseResult

ioquatix:
I basically agree with this but I'm concerned about the blast radius of this change.

lib/rack/query_parser.rb
```diff
@@ -86,6 +86,7 @@ def normalize_params(params, name, v, depth)
       name =~ %r(\A[\[\]]*([^\[\]]+)\]*)
```
ioquatix, line 86: Do you think you can tidy up this parsing logic a bit? Using `$` variables concerns me a little.

ioquatix (approved):

lib/rack/query_parser.rb
jeremyevans: I can, but I don't see the reason to do so in this PR.  It's better to keep the implementation the same as before.  After all, there are other `$` variables used elsewhere in the code, including later in this method.  If you would like to remove `$` variable usage, we should probably do that globally in a separate PR.

PR #1702, opened 2020-08-06
Fix MockResponse#body encoding

Quiwin:
Hello,

This pr fix the following issue:

Since 2.1.0, `MockResponse#body` ignores the original encoding of the body and always returns with `ASCII-8BIT` encoding.
```ruby
body = ["pong"]
body.first.encoding
=> #<Encoding:UTF-8>
Rack::MockResponse.new(200, {}, body).body.encoding
=> #<Encoding:ASCII-8BIT>
```
This issue stems from the use of `String.new` when joining the body, which by default produces a String encoded in `ASCII-8BIT`.

https://github.com/rack/rack/blob/649c72bab9e7b50d657b5b432d0c205c95c2be07/lib/rack/mock.rb#L213

### DIFF
--- a/lib/rack/mock.rb
+++ b/lib/rack/mock.rb
@@ -210,9 +210,10 @@ module Rack
       #     ...
       #     res.body.should == "foo!"
       #   end
-      buffer = String.new
+      enum = super.each
+      buffer = String.new(encoding: enum.first&.encoding)
 
-      super.each do |chunk|
+      enum.each do |chunk|
         buffer << chunk
       end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file. For info on
 
 ### Fixed
 
+- Fix encoding of `MockResponse#body` to match input body encoding.([#1702](https://github.com/rack/rack/pull/1702), [@Quiwin](https://github.com/quiwin))
 - TempfileReaper now deletes temp files if application raises an exception. ([#1679](https://github.com/rack/rack/issues/1679), [@jeremyevans](https://github.com/jeremyevans))
 - Fix using Rack::Session::Cookie with coder: Rack::Session::Cookie::Base64::{JSON,Zip}. ([#1666](https://github.com/rack/rack/issues/1666), [@jeremyevans](https://github.com/jeremyevans))
 - Avoid NoMethodError when accessing Rack::Session::Cookie without requiring delegate first. ([#1610](https://github.com/rack/rack/issues/1610), [@onigra](https://github.com/onigra))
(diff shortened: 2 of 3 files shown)


lib/rack/mock.rb
```diff
@@ -210,9 +210,10 @@ def body
       #     ...
       #     res.body.should == "foo!"
       #   end
-      buffer = String.new
+      enum = super.each
```
ioquatix, line 213: This is not guaranteed to work. The only way you could do this is by initializing the buffer during `each`:

``` ruby
body.each do |chunk|
  if buffer.nil?
    buffer = chunk.dup
  else
    buffer << chunk
  end
end
```
```

Comment (Quiwin) [reply lib/rack/mock.rb:213]: @ioquatix -- Thanks for your swiftness!
Before I change it, could you point in which case it fail?

**Comment by BuonOmo** [comment]
Hi!

@ioquatix, do you know when this one will be merged ? It prevents us from upgrading our rack version :/

Cheers

**Comment by jeremyevans** [comment]
Can you explain why the current behavior is a problem?  An empty string with ASCII-8BIT encoding has its encoding changed to UTF-8 if you append a UTF-8 string to it with non-ASCII characters.

```ruby
body = ["pong\u1234"]
# => ["pong\u1234"]
body.first.encoding
# => #<Encoding:UTF-8>
Rack::MockResponse.new(200, {}, body).body.encoding
# => #<Encoding:UTF-8>
```

BuonOmo:
@jeremyevans we want in our tests to ensure that we always send UTF-8, which we cannot do with the actual implementation. However, we would be able to do so with this implementation. And I don't see the downside of it anyway.

jeremyevans:
If you want to be sure you are sending UTF-8, you can't just check the string encoding.  You should check the data is valid for the encoding you want:

```ruby
assert string.force_encoding('UTF-8').valid_encoding?
```

You'll notice that this returns true for an ASCII-8BIT string if all data in it is valid UTF-8, such as `pong`.

BuonOmo:
@jeremyevans that's a good point thanks ! And I think we'll add this bit to our test suite.

However, our mobile apps want us to always send the same (UTF-8) encoding. Hence this verification tells that the string is UTF-8_ ready, but not UTF-8... Do I miss something here?

jeremyevans:
What I'm saying is you shouldn't be concerned that the Ruby string has UTF-8 encoding, you should be concerned that the data contained in the string is valid UTF-8.  Your mobile app doesn't care about the encoding of the Ruby string, it cares that it receives correctly encoded UTF-8 data.  Testing the encoding of the Ruby string is testing at the wrong level, because there is no guarantee that a Ruby string's encoding matches the encoding of the data contained in it, hence the need for the `String#valid_encoding?` method.

BuonOmo:
Ok it makes sense indeed. Thank you very much for your time, we'll update this in our tests. 

As for this Pull Request, I guess it is not mandatory for us anymore and could be closed. Or merged, depending on other side effects of encodings I don't know about...

ioquatix:
Thanks @jeremyevans

Issue #1668, opened 2020-06-01
Proxy all (or at least more) methods on Rack::RewindableInput?


doriantaylor:
Running up to an issue with `Rack::RewindableInput` where I have a construct that looks like:

```ruby
    content = case content
              when nil          then self.content
              when IO, StringIO then content
              when String       then StringIO.new content
              when Pathname     then content.open('rb')
              when Proc         then content.call
              when -> x { %i[read seek pos].all? { |m| x.respond_to? m } }
                content
              else
                raise ArgumentError,
                  "Cannot scan content of type #{content.class}"
              end
```
What this thing does is effectively coalesce the input into an `IO`-like object (or at least something that quacks like one). Problem is `Rack::RewindableInput` does not implement `seek` or `pos`/`tell` despite the proxied `Tempfile` class implementing them, nor does it inherit `IO` (although neither does `Tempfile` apparently). As such, I found myself writing the following monkeypatch:

```ruby
# Rack::RewindableInput doesn't provide the full complement of methods
# of the object it proxies
class Rack::RewindableInput

  # not sure if this is a good idea or a terrible one
  def respond_to? sym
    # of course we can't query the proxy object until we bootstrap it
    make_rewindable unless @rewindable_io
    @rewindable_io.respond_to?(sym) || super(sym)
  end

  # make sure we don't redefine method_missing if it already exists
  unless self.instance_methods.include? :method_missing
    def method_missing which, *args
      # do this now
      make_rewindable unless @rewindable_io

      # toss if the proxy object doesn't respond (& mimic original error)
      raise NoMethodError,
        "undefined method #{which} for #{self.inspect}:#{self.class}" unless
        @rewindable_io.respond_to? which

      # inline the method so this doesn't get called again
      self.class.define_method which do |*args|
        # do it here too for parity
        make_rewindable unless @rewindable_io
        @rewindable_io.__send__(which, *args)
      end

      # don't forget to actually invoke it this time around too
      @rewindable_io.__send__(which, *args)
    end
  end
end
```
I can't tell if this is a great idea or a terrible one, but it seems like `Rack::RewindableInput` could use the full complement of methods available to its proxy object so it can at least be duck-typed.

Thoughts?

ioquatix:
RewindableInput will be getting removed. Does that impact your proposal?

doriantaylor:
I'm not committed to it, as long as I can have a seekable IO-looking thing for request bodies when running under FastCGI. What's it being replaced with?

[...]

PR #1703, opened 2020-08-24
add size to Rack::RewindableInput

ahorek:
to improve compatibility with fcgi

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file. For info on
 
 ## [3.0.0] - Unreleased
 
+### Added
+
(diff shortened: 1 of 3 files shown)


jeremyevans:
This seems OK to me.  Note that we removed the fcgi handler already, so you would have to use an external fcgi handler now.  Assuming no other committer objects, I'll merge this in a couple days.

ahorek:
I was asked to keep up the support in Redmine
see https://www.redmine.org/issues/10832 and https://www.redmine.org/issues/33752
that's the only reason

but the original issue is very old, so I'm not sure if anyone really cares about fcgi anymore.

there's also a discussion about the complete removal of Rack::RewindableInput here https://github.com/rack/rack/issues/1668

PR #1705, opened 2020-09-05
Fix RuboCop offenses

utkarsh2102:
Hey,

Thanks all for your work in maintaining rack in such a great way! ❤️ 

However, whilst maintaining this in Debian, I found a couple of remaining RuboCop offenses that were left unfixed.
This PR fixes the same and makes the `Layout/SpaceInsideHashLiteralBraces` cop happy.

Signed-off-by: Utkarsh Gupta <<utkarsh@debian.org>>

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,3 +1,6 @@
+require:
+  - rubocop-packaging
+
 AllCops:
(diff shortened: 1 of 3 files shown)


ioquatix:
Thanks.

PR #1706, opened 2020-09-16
Add :same_site cookie option to the param docs.

alecclarke:
This change adds the `:same_site` cookie option to the Rack::Session::Abstract::Persisted parameter documentation as it's a valid option, but isn't documented.


jeremyevans:
Thanks for the patch!

PR #1710, opened 2020-09-25
use @status instead of status.to_i

ceclinux:
`status.to_i` looks strange in `Rack::Response`. Changed to `@status`


jeremyevans:
Looks good, thanks!

Issue #1712, opened 2020-10-15
ETag middleware no longer handles a Rack::Response body of [nil] 


stanhu:
Prior to Rack v2.1.0, a `Rack::Response` initialized with `[nil]` would be handled gracefully because `to_s` would be run on each element: https://github.com/rack/rack/blob/85684323f8f58409e717af91e446d257d496f8b8/lib/rack/response.rb#L40-L43

Now that https://github.com/rack/rack/pull/1434/files has merged, if `[nil]` is used, the ETag calculations fails with:

```
NoMethodError (undefined method `empty?' for nil:NilClass):
rack (2.1.4) lib/rack/etag.rb:70:in `block in digest_body'
rack (2.1.4) lib/rack/body_proxy.rb:34:in `block in each'
rack (2.1.4) lib/rack/body_proxy.rb:34:in `each'
rack (2.1.4) lib/rack/body_proxy.rb:34:in `each'
rack (2.1.4) lib/rack/etag.rb:68:in `digest_body'
rack (2.1.4) lib/rack/etag.rb:31:in `call'
rack (2.1.4) lib/rack/conditional_get.rb:40:in `call'
rack (2.1.4) lib/rack/head.rb:14:in `call'
actionpack (6.0.3.3) lib/action_dispatch/http/content_security_policy.rb:18:in `call'
```

Is `[nil]` a valid input? If so, does it make sense to fix this in `Rack::Response` so ETag calculations use `''`, or fix in the ETag handler?

Related issue in Grape: https://github.com/ruby-grape/grape/issues/2118

stanhu:
Hmm, this might have been fixed by 549412368.

stanhu:
Ok, I think I finally understand the problem:

1. Grape's own Rack middleware (`base.rb`) sends essentially sends something like `Rack::Response.new([nil], 200, {})`: https://github.com/ruby-grape/grape/blob/0e0c1550999bd1d80618e534880391796f2d9a98/lib/grape/middleware/base.rb#L62
1. `response.body` gets passed to `Rack::Etag`: https://github.com/rack/rack/blob/5ce5b2ccb151c62652e25b4711218ede11497cc3/lib/rack/etag.rb#L27
1. `Rack::Etag` iterates on the `Array` and calls `String#empty?` on all elements: https://github.com/rack/rack/blob/5ce5b2ccb151c62652e25b4711218ede11497cc3/lib/rack/etag.rb#L70

As a result, with Rack v2.1.4, we get:

```
[...]

PR #1713, opened 2020-10-15
Gracefully handle nil values in ETag middleware

stanhu:
Prior to Rack v2.1.0, a `Rack::Response` initialized with `[nil]` would
be handled gracefully because `to_s` would be run on each element:
https://github.com/rack/rack/blob/85684323f8f58409e717af91e446d257d496f8b8/lib/rack/response.rb#L40-L43

Now that https://github.com/rack/rack/pull/1434/files has been merged,
if `[nil]` is used, the ETag calculations fails with:

```
NoMethodError (undefined method `empty?' for nil:NilClass):
rack (2.1.4) lib/rack/etag.rb:70:in `block in digest_body'
rack (2.1.4) lib/rack/body_proxy.rb:34:in `block in each'
rack (2.1.4) lib/rack/body_proxy.rb:34:in `each'
rack (2.1.4) lib/rack/body_proxy.rb:34:in `each'
rack (2.1.4) lib/rack/etag.rb:68:in `digest_body'
rack (2.1.4) lib/rack/etag.rb:31:in `call'
rack (2.1.4) lib/rack/conditional_get.rb:40:in `call'
rack (2.1.4) lib/rack/head.rb:14:in `call'
actionpack (6.0.3.3) lib/action_dispatch/http/content_security_policy.rb:18:in `call'
```

We restore the original behavior by calling `to_str` on each body part.

Fixes https://github.com/rack/rack/issues/1712

### DIFF
--- a/lib/rack/etag.rb
+++ b/lib/rack/etag.rb
@@ -66,6 +66,7 @@ module Rack
         digest = nil
 
         body.each do |part|
+          part = part.to_s unless part.is_a?(String)
           parts << part
           (digest ||= Digest::SHA256.new) << part unless part.empty?
         end
--- a/test/spec_etag.rb
+++ b/test/spec_etag.rb
@@ -72,6 +72,12 @@ describe Rack::ETag do
     response[1]['ETag'].must_be_nil
   end
 
+  it "not set ETag if body has nil values" do
+    app = lambda { |env| [200, { 'Content-Type' => 'text/plain' }, [nil]] }
+    response = etag(app).call(request)
+    response[1]['ETag'].must_be_nil
+  end
+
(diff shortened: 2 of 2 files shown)


stanhu:
I believe this PR also fixes the case where an Integer is passed to the body (e.g. `[10]`).

jeremyevans:
Both `[nil]` and `[10]` are invalid bodies according to the rack SPEC.  From SPEC:

```
The Body must respond to +each+
and must only yield String values
```

I don't think the ETag middleware should attempt to handle invalid bodies.  Callers should be fixed to use valid bodies.  Before Rack 2.1.0, Rack::Response converted some invalid bodies to valid bodies, but that was an implementation detail, not by design.

PR #1715, opened 2020-10-21
Remove thin spec

### LINKED ISSUE #1714: Move Thin handler test into Thin?
The Thin dependency was removed by 98d9cf5834d4e27e34bbaa017cdfc68795763b55 and moved to [Thin](https://github.com/macournoyer/thin/commit/b201525f0db7d7e9faf26681b68646e0be9f96ec). Shouldn't the tests, i.e. the `test/spec_thin.rb` follow the suite? I don't think the test are executed anyway, because nothing pulls Thin in since 20d248e9ad32515bb4f12f4f7211a6c89441b93f.

jeremyevans:
Support for thin was removed earlier, so no reason to keep this.

Fixes #1714


jeremyevans:
All failures appear to be due to an old version of rubocop vendored by rack-attack.  Maybe someone with knowledge of GitHub Action configuration can fix that?

simi:
@jeremyevans I did a quick check and actually the same error is on rack-attack CI side. It is not related to any change or configuration in rack repo IMHO.

https://travis-ci.org/github/rack/rack-attack/jobs/736690080

jeremyevans:
Thanks for checking.  Maybe we can make it so that rack commits/PRs don't run the rack-attack CI, at least until rack-attack is fixed?

simi:
I have submitted PR to fix it for now https://github.com/rack/rack-attack/pull/502.

PR #1719, opened 2020-11-14
Skip rubocop for external tests

tlatsas:
Rubocop [fails](https://travis-ci.org/github/rack/rack-attack/builds) on rack/rack-attack repo. This commit explicitly runs only the tests of the rack-attack repo which is what rack is concerned about.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -149,7 +149,7 @@ task :external do
   FileUtils.mv ".rubocop.yml", ".rack.rubocop.yml.disabled"
 
   Bundler.with_clean_env do
-    clone_and_test("https://github.com/kickstarter/rack-attack", "rack-attack", "bundle exec rake")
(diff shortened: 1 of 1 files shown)


Rakefile
```diff
@@ -149,7 +149,7 @@ task :external do
   FileUtils.mv ".rubocop.yml", ".rack.rubocop.yml.disabled"
```
tlatsas, line 149: Should this be removed? 🤔

ioquatix:
I think this might be a good idea, but I've got a fix for rack-attack which also fixes the issues. Maybe we should just opt for that (although I also agree running tests should not invoke RuboCop).

Rakefile
ioquatix: Yes, it could be.

simi:
@ioquatix My fix was already submitted at https://github.com/rack/rack-attack/pull/502.

ioquatix:
Awesome, thanks for your work fixing this.

PR #1720, opened 2020-11-14
Define Rack::Builder::Config, and support middleware.rackup configuring middleware based on server configuration

jeremyevans:
This is an alternative approach to #1718.  Instead of creating a new
builder for every middleware that responds to `rackup`, we have a
single configuration object that stores the server's configuration,
such as whether it is multithreaded.  If middleware.rackup is defined,
we call it with the configuration object as the first object, and the
remaining arguments and block that would have been passed to new.
The middleware.rackup method can return the app itself if it not
needed, call new with the remaining arguments if it is needed, or
potentially other actions for more complex cases.

To handle the very rare case where a middleware would want to
delegate to other middleware in certain server configurations, and
doesn't know whether the other middleware supports the rackup
method or not, The configuration object supports a rackup method,
which will call rackup on the middleware if defined, or new
otherwise.  So if middleware A wants to use external middleware B
in a certain server configuration, middleware A's rackup method
could be something like:

```ruby
  def self.rackup(config, app)
    if config.multithread?
      config.rackup(MiddlewareB, app)
    else
      new(app)
    end
  end
```

The configuration rackup method is also used internally to implement
the builder.

This readds Rack::Lock, using the rackup method, which only uses
the Rack::Lock middleware if the configuration indicates the
server is multithreaded.

The advantage of this approach is that it doesn't require exposing
the entire builder API to the middleware, it only exposes the
server configuration, which is all the middleware should need to
appropriately configure itself.


newblaque (approved):

ioquatix:
I generally like this idea, I'm not strongly in favour of any particular approach, but I would like to explore a bit more about how this works in practice. The idea of a config object is nice since it isolates the responsibilities, but it also introduces additional complexities. However, I think the net gain for users is that it simplifies more complex construction of the application and provides a very well defined interface for the kinds of things people should depend on.

The only thing I wonder about is whether we should allow config to contain user specific keys/values. I imagine that such a model could be really useful but at the expense of opening up the design to more complexity. However, the advantage would be that users could depend on a configuration context for their own usage rather than having multiple sources of truth for configuration.

ioquatix:
@jeremyevans do you mind rebasing this on master?

lib/rack/lock.rb
```diff
@@ -1,32 +1,27 @@
...
+        @app.call(env)
       end
     end
 
-    private
-
-    def unlock
-      @mutex.unlock
-      @env[RACK_MULTITHREAD] = @old_rack_multithread
+    def self.rackup(config, app)
```
ioquatix, line 19: I feel like this should be `app, config` simply because config is similar to `**options`. WDYT?

ioquatix (changes requested):
Looking great. Just some points for us to discuss and iron out.

CHANGELOG.md
```diff
@@ -34,6 +34,10 @@ All notable changes to this project will be documented in this file. For info on
 
 - [[CVE-2020-8184](https://nvd.nist.gov/vuln/detail/CVE-2020-8184)] Do not allow percent-encoded cookie name to override existing cookie names. BREAKING CHANGE: Accessing cookie names that require URL encoding with decoded name no longer works. ([@fletchto99](https://github.com/fletchto99))
 
+### Removed
```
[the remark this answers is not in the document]
jeremyevans: Sure, I can do that.  Note that most users won't be using it, only middleware/server authors.

README.rdoc
```diff
@@ -77,7 +77,6 @@ middleware:
 * Rack::Files, for serving static files.
 * Rack::Head, for returning an empty body for HEAD requests.
 * Rack::Lint, for checking conformance to the \Rack API.
-* Rack::Lock, for serializing requests using a mutex.
```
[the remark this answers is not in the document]
jeremyevans: Correct, I'll restore it.

lib/rack/builder.rb
```diff
@@ -31,6 +31,30 @@ module Rack
   # You can use +map+ to construct a Rack::URLMap in a convenient way.
 
   class Builder
+    # Config stores settings on what the server supports, such as whether it
+    # is multithreaded.
+    class Config
+      def initialize(multithread: true, reentrant: multithread)
```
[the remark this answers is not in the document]
jeremyevans: My idea here is that server authors can subclass this class to provide a richer API if they need to. Currently, the only use for the configuration is for concurrency. The issue with `**options` is that it turns typos into silent failures.

lib/rack/builder.rb
```diff
@@ -31,6 +31,30 @@ module Rack
...
+        @multithread = multithread
+        @reentrant = reentrant
+      end
+
+      # Whether the application server will invoke the application from multiple threads. Implies {reentrant?}.
+      def multithread?; @multithread; end
+
+      # Re-entrancy is a feature of event-driven servers which may perform non-blocking operations. When
+      # an operation blocks, that particular request may yield and another request may enter the application stack.
+      def reentrant?; @reentrant; end
```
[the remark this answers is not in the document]
jeremyevans: If `multithread?` is an implementation detail, so is `multifiber?`.  `parallel?` is not accurate on CRuby, since Ruby code does not execute in parallel on CRuby (though that is also an implementation detail).  Replacing `reentrant?` with `concurrent?` seems like a good idea to me, so I'll make that change.

I would like to get rid of `multithread?`, but I think it it is necessary until we drop Ruby 2.7 support. That is because on Ruby 2.7 and below, you cannot use Rack::Lock in the concurrent but not multithread case, since it uses Mutex internally.

lib/rack/builder.rb
```diff
@@ -93,25 +117,33 @@ def self.load_file(path)
...
-    def self.new_from_string(builder_script, file = "(rackup)")
-      # We want to build a variant of TOPLEVEL_BINDING with self as a Rack::Builder instance.
-      # We cannot use instance_eval(String) as that would resolve constants differently.
-      binding, builder = TOPLEVEL_BINDING.eval('Rack::Builder.new.instance_eval { [binding, self] }')
-      eval builder_script, binding, file
+    def self.new_from_string(builder_script, file = "(rackup)", config: nil)
+      builder = self.new(config: config)
+
+      # Create a top level scope with self as the builder instance:
+      binding = TOPLEVEL_BINDING.eval('->(builder){builder.instance_eval{binding}}').call(builder)
```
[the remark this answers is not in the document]
jeremyevans: That seems fine to me.  I'd like the constant to be under `Rack::Builder`.  So I plan to use `Rack::Builder::EVAL_CONTEXT`?

lib/rack/lock.rb
jeremyevans: `config` is first, because `rackup` handles args and block in addition to app. The code used in `Builder` is:

```ruby
        if middleware.respond_to?(:rackup)
          middleware.rackup(self, *args, &block) # self is the config in this case.
```

lib/rack/lock.rb
```diff
@@ -1,32 +1,27 @@
...
+      if config.multithread?
```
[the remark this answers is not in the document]
jeremyevans: We can do this for `concurrent?`, but I think only on Ruby 3+.  I'll modify the code to handle that.

jeremyevans (commented):
@ioquatix I pushed a new commit and tried to respond to all of the points you raised.

lib/rack/builder.rb
```diff
@@ -31,6 +38,30 @@ module Rack
...
+    # Config stores settings on what the server supports, such as whether it
+    # is multithreaded.
+    class Config
+      def initialize(multithread: true, concurrent: multithread)
+        @multithread = multithread
+        @concurrent = concurrent
+      end
+
+      # Whether the application server will invoke the application from multiple threads. Generally implies {concurrent?}.
+      def multithread?; @multithread; end
```
ioquatix, line 50: This is fine, but is there a reason why we don't just use `attribution :multithread`?

lib/rack/lock.rb
ioquatix: Makes total sense, thanks for explaining.

lib/rack/builder.rb
```diff
@@ -31,6 +38,30 @@ module Rack
   # You can use +map+ to construct a Rack::URLMap in a convenient way.
 
   class Builder
+    # Config stores settings on what the server supports, such as whether it
+    # is multithreaded.
```
ioquatix, line 42: Do you think it would make sense for us to explain that we expect severs may sub-class this to provide additional details?

Also, considering this usage, does it make more sense to provide a generic options hash?

Otherwise, I imagine we might have:

```ruby
def rackup(config, app)
  if config.respond_to?(:puma_thing)
    puma_thing = config.puma_thing
    ...
```

While I'm okay with the general idea of a strong interface, I'm a little concerned about how we should use it if puma/falcon/thin/unicorn start adding server-specific interfaces.

We should also consider the case where Rack is an interface rather than a gem. To this end, it feels like `options` is a stronger contender since it's simpler. But it's also less pleasant to use.

ioquatix (changes requested):

lib/rack/builder.rb
jeremyevans: We'd have to alias it to `multithread?` (notice the `?`).  That doesn't seem worth it.  This isn't performance sensitive, it's only called during application setup.

lib/rack/builder.rb
ioquatix: Basically, I prefer your design, but I think we need to consider some specific usage scenarios.

jeremyevans: Currently, the configuration would only be used for Rack::Lock.  I'm not aware of any other use case, so making it more flexible and prone to silent failures doesn't seem like a good trade-off.  If you have ideas for how this will be used by servers, that information would definitely be helpful.  That said, I'm not strongly against the options hash approach.

ioquatix: Rack for the most part tries to separate "Rack the SPEC" and "Rack the GEM". This configuration interface touches both worlds. I personally think Rack would benefit from richer interfaces, but I also appreciate the original design which was servers could follow "Rack the SPEC" without depending on "Rack the GEM". IIRC, Passenger does not pull in "Rack the GEM" and instead just follows the spec. In that case, @FooBarWidget would need a bespoke implementation of `Rack::Builder::Config` but it's not specified anywhere except for the implementation in the GEM. Maybe this is a longer term project - we should define `config.ru` - i.e. `use`, `run`, the context for evaluation, the builder and config interfaces, etc. 

Frankly it feels a lot harder to define an interface like `Rack::Builder::Config` - the building blocks we've used for "Rack the SPEC" has been Hash, Array, String, and other simple types. In any such spec, I guess we would leave the actual class as anonymous and just define the interface, e.g.

```
Your application may respond to `rackup` in which case it will be invoked with a configuration object which contains at least the following interface:
  multithread? -> true | false
  concurrent? -> true | false
```

I guess this all hinges on how much we define `config.ru` and `Rack::Builder` as "SPEC" vs "GEM". cc @tenderlove your input would be useful too.

rafaelfranca: I think for the middleware users and server implementers using hashes for this config would be better. It is easier to check for a hash value than it is to check if the config object respond to some method the only some server configs provide.

For what I get of this implementation, the only reason why a hash as config would not work right now is because the object allows servers to implement the `rackup` method, but do we see them needing this kind of feature?

ioquatix: > I think for the middleware users and server implementers using hashes for this config would be better.

It's a fair position.

> It is easier to check for a hash value than it is to check if the config object respond to some method the only some server configs provide.

As an alternative, I proposed `config.options` to be a hash.

> For what I get of this implementation, the only reason why a hash as config would not work right now is because the object allows servers to implement the rackup method, but do we see them needing this kind of feature?

It's a fair position but I believe there is value in having a symmetric interface in this regard. My initial implementation just used `Rack::Builder` but it was considered to be too big of an interface.

ioquatix: @rafaelfranca see the original proposed implementation https://github.com/rack/rack/pull/1718/files

- `Rack::Builder#rackup` creates a nested builder with the same options.
- `Rack::Builder#options` configuration options.
- `Rack::Builder#multithread?` checks those options.

jeremyevans: Using `Rack::Builder` directly has the same issues that using `Rack::Builder::Config` does in terms of SPEC.  I agree from a SPEC perspective, rack avoids custom objects.  However, `config.ru` was never part of SPEC (which is only regarding the rack protocol), so I'm not sure it applies.

The issue with using a plain hash is that it won't have the equivalent of `Config#rackup`, so middleware cannot pass the configuration to other middleware loaded by that middleware.  However, maybe there are no such middleware that actually need that.  Switching to a plain hash seems fine, I can work on that change if that's the direction we want to go.

However, maybe we should rethink the idea of middleware autoconfiguration based on configuration parameters? Is it really needed?  Maybe we can just have `Rack::Lock` always lock, and users who know their apps are not concurrency-safe can use it.  If they use it on a non-concurrent webserver (e.g. Unicorn), it will use a lock when it doesn't need to, but uncontested mutex is not slow, so there is no real problem with doing so.  What are your thoughts on abandoning the idea of middleware configuration?  We would still remove `rack.multithread`/`rack.multiprocess`/`rack.run_once`.

rafaelfranca: > What are your thoughts on abandoning the idea of middleware configuration? We would still remove `rack.multithread`/`rack.multiprocess`/`rack.run_once`.

I'm positive for that idea. To be fair the only middleware that I saw which uses this is Rack::Lock, which Rails doesn't include anymore by default and uses it own config `config.allow_concurrency` to configure it.

ioquatix: Why don't we still expose `Builder#options` for server configuration/specific details at load time. We don't need to handle the `rackup` model in the same PR.

jeremyevans: I don't see the need for a per-builder side channel.  The complexity isn't worth it, IMO. Looks like @rafaelfranca agrees.  So unless another committer is in favor of adding it, we should probably just commit the removal of `rack.multithread`/`rack.multiprocess`/`rack.run_once`.

ioquatix: Okay, let's do that to start with.

ioquatix (approved):
Great.

Issue #1723, opened 2020-12-04
Asserts in Rack::Lint are very inefficient


eregon:
I was looking at the failure of truffleruby-head on a recent run:
https://github.com/rack/rack/runs/1482881957
```
[ruby] WARNING OutOfMemoryError
<internal:core> core/truffle/string_operations.rb:144:in `copy_from': failed to allocate memory (NoMemoryError)
	from <internal:core> core/string.rb:530:in `block in inspect'
	from <internal:core> core/string.rb:529:in `each'
	from <internal:core> core/string.rb:529:in `inspect'
	from /Users/runner/work/rack/rack/lib/rack/lint.rb:758:in `block in each'
	from /Users/runner/work/rack/rack/lib/rack/lint.rb:756:in `each'
	from /Users/runner/work/rack/rack/lib/rack/lint.rb:756:in `each'
	from /Users/runner/work/rack/rack/lib/rack/response.rb:277:in `buffered_body!'
	from /Users/runner/work/rack/rack/lib/rack/mock.rb:191:in `initialize'
	from /Users/runner/work/rack/rack/lib/rack/mock.rb:85:in `request'
	from /Users/runner/work/rack/rack/lib/rack/mock.rb:59:in `post'
	from /Users/runner/work/rack/rack/test/spec_show_exceptions.rb:58:in `test_0003_handles invalid POST data exceptions'
```

And noticed https://github.com/rack/rack/blob/d3225f7c201320ed272a2fa7b000c5850e4a5f88/lib/rack/lint.rb#L758-L760 which seems very innefficient.
Notably it will call `String#inspect` on each part of the body, even if it is correctly a String.

The general `assert` pattern like:
```ruby
        assert("Body yielded non-string value #{part.inspect}") {
          part.kind_of? String
        }
```
seems to have the same problem, the error message can be very expensive to compute, and should only be computed on error, not if the condition passes.

So I guess we should swap the arguments here, and have the message computed in the block, and the condition as an argument since the condition must be always evaluated, but the message needs not:
```ruby
        assert(part.kind_of? String) {
          "Body yielded non-string value #{part.inspect}"
        }
```
or maybe keep it simple and:
```ruby
unless part.kind_of? String
  raise LintError, "Body yielded non-string value #{part.inspect}"
end
```
This avoids many String allocations and whatever interpolations are done, etc.

I guess Rack::Lint is never meant to be used in production, but in the current state I would say it's so slow and inefficient that it will often be unusable.

---

[...]

PR #1724, opened 2020-12-08
Fix inefficient assert pattern in Rack::Lint

eregon:
Fixes https://github.com/rack/rack/issues/1723

### DIFF
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -17,6 +17,8 @@ module Rack
     class LintError < RuntimeError; end
     module Assertion
       def assert(message)
+        warn("Rack::Lint::Assertion#assert is deprecated as it is inherently inefficient. " \
+          "Use `raise Rack::Lint::LintError, 'msg' unless condition` instead", uplevel: 1)
         unless yield
           raise LintError, message
         end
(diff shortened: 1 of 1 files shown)


eregon:
@jeremyevans Could you review and merge?

jeremyevans (approved):
I'm OK with these changes.  I'll give some other committers time for feedback, but if no feedback is received, I'll merge this next week.  Please ping me if I forget.

ioquatix:
It looks fine to me.

Did you consider reversing the assert aguments?

eregon:
@ioquatix I followed Jeremy's advice from https://github.com/rack/rack/issues/1723#issuecomment-738877460

skipkayhil:
Would y'all accept a (partial) backport of this change to Rack `2-2-stable`? While working on skipkayhil/rails#5 we've run into a situation that makes a test worse because of the eager `inspect`.

Basically, in Rails there is a class put in the `env` that may raise when `to_s` is called on it: https://github.com/rails/rails/blob/0e5a99db5ac445e2bdfd19c911f62e65bbe0e45d/actionpack/lib/action_dispatch/middleware/remote_ip.rb#L168

When adding Rack::Lint to some of its tests, the error is now raised during the request when using Rack 2 instead of when the `to_s` is explicitly called later on.

ioquatix:
I’d be okay with that.

PR #1726, opened 2020-12-27
Freeze Rack::VERSION

kirs:
The [Ractor primitive](https://github.com/ruby/ruby/blob/master/doc/ractor.md) from Ruby 3.0 requires all constants to be immutable and frozen.

This PR makes `Rack::VERSION` and `Rack.version` to be immutable objects.

### DIFF
--- a/lib/rack/version.rb
+++ b/lib/rack/version.rb
@@ -13,11 +13,12 @@
 
 module Rack
   # The Rack protocol version number implemented.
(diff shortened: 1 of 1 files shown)


ioquatix:
Thanks. This makes sense. However, it seems slightly odd that `VERSION` has been co-opted for the protocol compatible version rather than the gem version. I think we should consider fixing that in Rack 3.0.

PR #1727, opened 2021-01-04
Ruby 3

beanieboi:
This adds Ruby 3 to GitHub Actions.

`webrick` is no longer bundled and need to be specified in the Gemfile.

> The following libraries are no longer bundled gems or standard libraries. Install the corresponding gems to use these features.
> sdbm
> webrick
> net-telnet
> xmlrpc

source: https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -8,7 +8,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest]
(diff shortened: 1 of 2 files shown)


beanieboi:
The CI failure didn't happen when it ran on my fork (https://github.com/beanieboi/rack/actions/runs/460996946) maybe they are intermittent? I haven't found a button to restart the test for `truffleruby-head`

PR #1731, opened 2021-01-16
Update 2.2.0 changelog with a breaking change [ci skip]

sergey-alekseev:
In my view if [you choose](https://github.com/rack/rack/pull/1672#issuecomment-643295672) to have a minor bug fixed and a major bug introduced at least a prominent mention is necessary.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,7 +52,7 @@ All notable changes to this project will be documented in this file. For info on
 
 ## [2.2.0] - 2020-02-08
 
-- BREAKING CHANGE: streaming in Rails no longer works without a workaround. See [#1416](https://github.com/rack/rack/pull/1416) and [#1619](https://github.com/rack/rack/issues/1619).
+- BREAKING CHANGE: if you are using the `Rack::ETag` middleware, streaming no longer works without a workaround, see [#1619 (comment)](https://github.com/rack/rack/issues/1619#issuecomment-848460528).
 
 ### SPEC Changes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,8 @@ All notable changes to this project will be documented in this file. For info on
 
 ## [2.2.0] - 2020-02-08
 
+- BREAKING CHANGE: streaming in Rails no longer works without a workaround. See [#1416](https://github.com/rack/rack/pull/1416) and [#1619](https://github.com/rack/rack/issues/1619).
+
 ### SPEC Changes
 
 - `rack.session` request environment entry must respond to `to_hash` and return unfrozen Hash. ([@jeremyevans](https://github.com/jeremyevans))


jeremyevans:
I think there is consensus among the rack committers that the current behavior is not a bug in Rack, but that it was a bug in Rails (https://github.com/rack/rack/issues/1619#issuecomment-648477608). Did Rails ever fix the bug?  I think I'd be fine with adding this to the 2.2.0 changelog if we could tell people which version of Rails they need to upgrade to.

sergey-alekseev:
I'm fine with blaming Rails if it's guilty. I'm not fine with people upgrading Rack to >= 2.2.0 having broken streaming and/or losing their time debugging why. That's the point of the changelog entry, what's written there can be adjusted.

jeremyevans:
In order to adjust what is written, we would want to recommend people upgrade to a Rails version that works.  So I'll ask again, did Rails fix the bug, and if so, which Rails version has streaming that works with Rack 2.2+? That way we can tell people which Rails version to upgrade to.

sergey-alekseev:
Thanks for your help. I understand the desire to recommend to upgrade to a specific Rails version, that's indeed the best solution – using latest Rails and Rack while everything works by default without any workaround necessary. However many projects are still not up-to-date with the latest Rails version. In my case I upgraded the app from Rails 4.2 (!) to 5.0 and upgraded Rack to fix the latest related CVE. After the upgrade my teammate found a bug with streaming which took time to find out a root cause and fix. With a prominent mention in the gem's changelog I'd probably catch this during the upgrade. So even without recommending a Rails version with a fix (if there is no fix yet) a mention should be helpful.

jeremyevans:
First you say "using latest Rails and Rack while everything works by default without any workaround necessary".  Then you say "So even without recommending a Rails version with a fix (if there is no fix yet)".  Is the second sentence only a hypothetical ("if there wasn't a fix yet")?

If using latest Rails and Rack everything works by default, I assume there is a Rails version with a fix, and that is the version we should recommend upgrading to. Can you find out which Rails version that is (the first Rails version that works), and modify this PR to include it?

sergey-alekseev:
I'd say the first phrase is hypothetical: "using latest Rails and Rack while everything works by default without any workaround necessary".

jeremyevans:
If Rails streaming still doesn't work with Rack 2.2.0 after almost a year, it seems like something they should address.  Looks like Rails's closed the initial bug report thinking it was a bug in Rack, even though @tenderlove confirmed it isn't.  I added a comment to the Rails ticket asking them to reopen the issue: https://github.com/rails/rails/issues/38842#issuecomment-762063366

Once Rails releases a version with a fix, I'm fine with adding updating the CHANGELOG to point users to a version with a fix. Note that if another committer wants to merge this as-is, I have no objection to that.

sergey-alekseev:
Thanks for submitting the comment.

For me https://github.com/rack/rack/issues/1619#issuecomment-648477608 doesn't sound like "tenderlove confirmed it isn't". "(in Rails?)" sounds like it's not clear to him at that point whether it's a bug in Rack or Rails. Actually I'd not spend time on this kind of discussion and would let others who are more competent on that to dive into specifics and judge. I assume you are right about this being a bug in Rails.

The point of such a changelog entry is to save time for other developers.

rafaelfranca:
I'm trying to reproduce the problem people are seeing and I'm not able to. Can someone help me to understand what exactly is failing. This passes without any problem:

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem "rails", github: "rails/rails", branch: "main"
end

require "action_controller/railtie"

class TestApp < Rails::Application
  config.root = __dir__
  config.hosts << "example.org"
  secrets.secret_key_base = "secret_key_base"

  config.logger = Logger.new($stdout)
  Rails.logger  = config.logger

  config.middleware.use Rack::ETag

  routes.draw do
    get "/" => "test#index"
  end
end

class TestController < ActionController::Base
  include ActionController::Live
  include Rails.application.routes.url_helpers

  def index
    response.headers["X-Accel-Buffering"] = "no"
    response.headers['Content-Type'] = 'text/event-stream'

    5.times do |i|
      puts "stream"

      @sse = SSE.new(response.stream, event: 'test')
      @sse.write "hello #{i}"
      sleep 1
    end
  rescue ActionController::Live::ClientDisconnected => e
    # ignore client disconnections
    nil
  ensure
    response.stream.close
  end
end

require "minitest/autorun"
require "rack/test"

class BugTest < Minitest::Test
  include Rack::Test::Methods

  def test_returns_success
    get "/"
    assert last_response.ok?
    assert_match(/hello 4/, last_response.body)
  end

  private
    def app
      Rails.application
    end
end
```

sergey-alekseev:
1. Without `response.headers['Last-Modified'] = Time.current.httpdate`  

Code
---

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem "rails", github: "rails/rails", branch: "main"
  gem 'puma'
end

require "action_controller/railtie"

class TestApp < Rails::Application
  config.root = __dir__
  config.hosts << "example.org"
  secrets.secret_key_base = "secret_key_base"

  config.logger = Logger.new($stdout)
  Rails.logger  = config.logger

  config.middleware.use Rack::ETag

  routes.draw do
    get "/" => "test#index"
  end
end

class TestController < ActionController::Base
  include ActionController::Live
  include Rails.application.routes.url_helpers

  def index
    response.headers["X-Accel-Buffering"] = "no"
    response.headers['Content-Type'] = 'text/event-stream'

    5.times do |i|
      puts "stream"

      @sse = SSE.new(response.stream, event: 'test')
      @sse.write "hello #{i}"
      sleep 1
    end
  rescue ActionController::Live::ClientDisconnected => e
    # ignore client disconnections
    nil
  ensure
    response.stream.close
  end
end

require 'rack/handler/puma'
Rack::Handler::Puma.run(TestApp, Port: 3000)
```

Video
---

without-workaround

sergey-alekseev:
2. With `response.headers['Last-Modified'] = Time.current.httpdate`

Code
---

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }

  gem "rails", github: "rails/rails", branch: "main"
  gem 'puma'
end

require "action_controller/railtie"

class TestApp < Rails::Application
  config.root = __dir__
  config.hosts << "example.org"
  secrets.secret_key_base = "secret_key_base"

  config.logger = Logger.new($stdout)
  Rails.logger  = config.logger

  config.middleware.use Rack::ETag

  routes.draw do
    get "/" => "test#index"
  end
end

class TestController < ActionController::Base
  include ActionController::Live
  include Rails.application.routes.url_helpers

  def index
    response.headers["X-Accel-Buffering"] = "no"
    response.headers['Content-Type'] = 'text/event-stream'
    response.headers['Last-Modified'] = Time.current.httpdate

    5.times do |i|
      puts "stream"

      @sse = SSE.new(response.stream, event: 'test')
      @sse.write "hello #{i}"
      sleep 1
    end
  rescue ActionController::Live::ClientDisconnected => e
    # ignore client disconnections
    nil
  ensure
    response.stream.close
  end
end

require 'rack/handler/puma'
Rack::Handler::Puma.run(TestApp, Port: 3000)
```

Video
---

with-workaround

rafaelfranca:
Ok. I see now. So the problem is that Rack::Etag is buffering the response, not that something is raising.

Would not be a problem with any rack app that tries to stream and are using `Rack::Etag`?

@jeremyevans I don't think we have a way to fix this in Rails, unless we introduce some kind of API that would tell middleware that they can't buffer the response. I guess that is what @ioquatix was trying to find in #1619. Can we expose something in the env so middleware can check if they can buffer the response or not? That would require us to release rack and rails to solve the issue but it is the only solution I can think of right now. I'm happy to make the changes if you think that direction make sense.

jeremyevans:
This would be an issue with anything trying to stream using `Rack::ETag`.  Calculating the ETag for the response header requires reading the body, which prevents streaming.  The issue with Rails is more pronounced as I think Rails includes `Rack::ETag` by default in the middleware stack.

I'm fine with adding a special header that `Rack::ETag` will recognize and not calculate the ETag, and then Rails could set that header.  You can already do that by setting either the `ETag` or `Last-Modified` headers, but a new special header would allow skipping ETag calculation without having to set one of those headers.  Is that what you meant?

rafaelfranca:
Yes change ETag, but I was thinking in a more general env value in the Rack API to tell any middleware to not read the body. Something like `env['rack.streaming'] = true`

jeremyevans:
What if the middleware needs to read the body to function, as `Rack::ETag` does?

Personally, I don't like the idea of reading `env` after the app returns a result (expecting the app to add it).  Not sure whether we do that in other cases, but it seems like a bad idea.  Is it possible to make this a response header instead of a `env` key?

I think there is a fundamental problem with skipping a middleware's behavior just because the response happens to be streamed.  I suppose renaming the key to something like `rack.forbid_body_buffering` would be OK with me.  Then at least it is specific to middleware what it shouldn't do.

rafaelfranca:
I believe most middleware that need to read the body to function would gladly be skip thenselves if they knew the response was supposed to be a stream. Do you see cases where the middleware will always chose to disregard if the application is streaming or not just to keep its behavior? In any case, we are allowing them to make a choice with this new env, so if they really want, they can.

It is very common to middleware to expect app to change the env, at least in Rails, but I bet we can find more example in other libraries like sinatra and hanami. We actually encourage in the spec that application changes some of the `rack` keys in the env like `rack.hijack`. We also have "The application is free to modify the environment." there.

I'm ok with `rack.forbid_body_buffering` as name.

jeremyevans:
`rack.hijack` as an `env` key is something the server sets and the application or middleware uses to do request hijacking, not something the application sets and middleware uses.  If response hijacking is done, `rack.hijack` is a response header set by the application.  We should follow the same approach for this, and have `rack.forbid_body_buffering` be a special response header, not an `env` key.

jeremyevans:
As an update to this, I believe the issue was fixed by 90be4d94dea9e371a4df14929f66aee6e11787b7. Rack::ETag now does not iterate over the body unless the body is already an array, so it no longer causes issues when streaming.  I don't think we should modify the CHANGELOG for previous versions, so I'm inclined to close this unless anyone expresses objections to doing so.

dentarg:
> As an update to this, I believe the issue was fixed by [90be4d9](https://github.com/rack/rack/commit/90be4d94dea9e371a4df14929f66aee6e11787b7)

Has that been released?

> I don't think we should modify the CHANGELOG for previous versions

Why? I can't see anything bad with that, in this case. It (only?) provides more information about the release. It was done in https://github.com/rack/rack/pull/1675 and people seem to have appreciated it. 

(Yes, the notice this PR brings probably very late for most users, and doesn't affect as many, I see that.)

jeremyevans:
> > As an update to this, I believe the issue was fixed by [90be4d9](https://github.com/rack/rack/commit/90be4d94dea9e371a4df14929f66aee6e11787b7)
> 
> Has that been released?

No, it hasn't been released yet.  Rack 3.0 will include it.

> > I don't think we should modify the CHANGELOG for previous versions
> 
> Why? I can't see anything bad with that, in this case. It (only?) provides more information about the release. It was done in #1675 and people seem to have appreciated it.

In #1675, an existing CHANGELOG entry was marked as breaking, a new CHANGEDLOG entry was not added.  If you want to mark the existing CHANGELOG entry for the ETag change as breaking, with an update that it breaks streaming, that should be acceptable.

dentarg:
> If you want to mark the existing CHANGELOG entry for the ETag change as breaking, with an update that it breaks streaming, that should be acceptable.

Okay 👍 See https://github.com/rack/rack/pull/1778

PR #1733, opened 2021-01-20
Default to & as separator instead of either & or ;

### LINKED ISSUE #1732: Web cache poisoning - semicolon as a separator
The `request.rb` file treats semicolon as a separator (https://github.com/rack/rack/commit/6af5f928f3a860e9bb74fc0bdd83b30be9d722a9) - whereas most proxies today only take ampersands as separators. Link to a blog post explaining this vulnerability: https://snyk.io/blog/cache-poisoning-in-popular-open-source-packages/

When the attacker can separate query parameters using a semicolon (;), they can cause a difference in the interpretation of the request between the proxy (running with default configuration) and the server. This can result in malicious requests being cached as completely safe ones, as the proxy would usually not see the semicolon as a separator, and therefore would not include it in a cache key of an unkeyed parameter - such as `utm_*` parameters, which are usually unkeyed. Let’s take the following example of a malicious request:				

```
GET /?link=http://google.com&utm_content=1;link='><t>alert(1)</script> HTTP/1.1

Host: somesite.com

Upgrade-Insecure-Requests: 1		

User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/85.0.4183.83 Safari/537.36

Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,imag e/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9 Accept-Encoding: gzip, deflate			

Accept-Language: en-US,en;q=0.9 Connection: close			
```

The backend sees 3 parameters here: `link`, `utm_content` and then `link` again. On the other hand, the proxy considers this full string: `1;link='><t>alert(1)</script>` as the value of `utm_content`, which is why the cache key would only contain `somesite.com/?link=http://google.com`. 

A possible solution could be to allow developers to specify a separator, defaulting to an ampersand.

jeremyevans:
Allowing ; as separator by default can lead to web cache poisoning.

Fixes #1732

### DIFF
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -4,7 +4,7 @@ module Rack
   class QueryParser
     (require_relative 'core_ext/regexp'; using ::Rack::RegexpExtensions) if RUBY_VERSION < '2.4'
 
-    DEFAULT_SEP = /[&;] */n
+    DEFAULT_SEP = /[&] */n
     COMMON_SEP = { ";" => /[;] */n, ";," => /[;,] */n, "&" => /[&] */n }
 
     # ParameterTypeError is the error that is raised when incoming structural
@@ -29,10 +29,9 @@ module Rack
     end
 
     # Stolen from Mongrel, with some small modifications:
-    # Parses a query string by breaking it up at the '&'
-    # and ';' characters.  You can also use this to parse
-    # cookies by changing the characters used in the second
-    # parameter (which defaults to '&;').
+    # Parses a query string by breaking it up at the '&'.  You can also use this
+    # to parse cookies by changing the characters used in the second parameter
+    # (which defaults to '&').
     def parse_query(qs, d = nil, &unescaper)
       unescaper ||= method(:unescape)
 
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -32,12 +32,12 @@ module Rack
     # Parses a query string by breaking it up at the '&'.  You can also use this
     # to parse cookies by changing the characters used in the second parameter
     # (which defaults to '&').
-    def parse_query(qs, d = nil, &unescaper)
+    def parse_query(qs, separator = nil, &unescaper)
       unescaper ||= method(:unescape)
 
       params = make_params
 
-      (qs || '').split(d ? (COMMON_SEP[d] || /[#{d}] */n) : DEFAULT_SEP).each do |p|
+      (qs || '').split(separator ? (COMMON_SEP[separator] || /[#{separator}] */n) : DEFAULT_SEP).each do |p|
         next if p.empty?
         k, v = p.split('=', 2).map!(&unescaper)
 
@@ -60,11 +60,11 @@ module Rack
     # query strings with parameters of conflicting types, in this case a
     # ParameterTypeError is raised. Users are encouraged to return a 400 in this
     # case.
-    def parse_nested_query(qs, d = nil)
+    def parse_nested_query(qs, separator = nil)
       params = make_params
 
       unless qs.nil? || qs.empty?
-        (qs || '').split(d ? (COMMON_SEP[d] || /[#{d}] */n) : DEFAULT_SEP).each do |p|
+        (qs || '').split(separator ? (COMMON_SEP[separator] || /[#{separator}] */n) : DEFAULT_SEP).each do |p|
           k, v = p.split('=', 2).map! { |s| unescape(s) }
 
           normalize_params(params, k, v, param_depth_limit)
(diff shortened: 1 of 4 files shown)


lib/rack/query_parser.rb
```diff
@@ -29,10 +29,9 @@ def initialize(params_class, key_space_limit, param_depth_limit)
...
 
     # Stolen from Mongrel, with some small modifications:
-    # Parses a query string by breaking it up at the '&'
-    # and ';' characters.  You can also use this to parse
-    # cookies by changing the characters used in the second
-    # parameter (which defaults to '&;').
+    # Parses a query string by breaking it up at the '&'.  You can also use this
+    # to parse cookies by changing the characters used in the second parameter
+    # (which defaults to '&').
     def parse_query(qs, d = nil, &unescaper)
```
rafaelfranca, line 35: Should we rename `qs` and `d` to something more descriptive? I was never think that `d` would mean separator.

ioquatix:
The next release should be a 3.0 release, so this is fine to merge.

rafaelfranca (approved):

jeremyevans:
Ubuntu CI is failing due to the inability to install packages, before it runs tests.  All tests pass on Mac, and this change is probably platform independent, so I'll merge it.

hctan:
Hi, any chances this fix can be rolled out for 2.x? I'm currently using 2.2.3. When is 3.0 expected to be released?

ioquatix:
Probably by the end of the year.

blakegearin:
Any updates?

jeremyevans:
There is no estimated release date yet.  We are nearing the end of making major feature changes, I think, but there are few more pull requests and issues we need to work through. Due to many changes in 3.0, we'll probably need to start with a prerelease version and allow for a substantial period for people to test before the final release.  So it don't think we can provide a good estimate yet for the final release of 3.0.

blakegearin:
It appears a05f8d5 was cherry-picked out of the release of `2.2.3.1` and `2.2.4`? Please correct me if I'm wrong. 

Was wondering why it would be excluded since this commit patches the [CWE-444](https://cwe.mitre.org/data/definitions/444.html) vulnerability ([Snyk report](https://security.snyk.io/vuln/SNYK-RUBY-RACK-1061917)). So applications wanting to avoid the vulnerability will have to depend on the `main` branch instead of a version which isn't ideal.

jeremyevans:
> It appears [a05f8d5](https://github.com/rack/rack/commit/a05f8d56f9ac4da14dddb8f312a3b43644f73397) was cherry-picked out of the release of `2.2.3.1` and `2.2.4`? Please correct me if I'm wrong.

The commit was only made to the main branch, it was never applied to the 2-2-stable branch.  So the phrase `cherry-picked out` certainly doesn't apply.  The commit wasn't backported to 2.2.  I think it is too backwards incompatible to be backported, and it looks like @ioquatix feels the same way.  Maybe @tenderlove can give his opinion.

> Was wondering why it would be excluded since this commit patches the [CWE-444](https://cwe.mitre.org/data/definitions/444.html) vulnerability ([Snyk report](https://security.snyk.io/vuln/SNYK-RUBY-RACK-1061917)). So applications wanting to avoid the vulnerability will have to depend on the `main` branch instead of a version which isn't ideal.

Again, this wasn't excluded, it just wasn't backported due to backwards compatibility concerns.

blakegearin:
Understood. The reason I said "cherry-picked" was because I'm used to versions being different slices of `main` instead of the multiple stable branches you all have going. So since it was merged to `main` I expected any new release to include it. My bad!

> Again, this wasn't excluded, it just wasn't backported due to backwards compatibility concerns.

I guess I consider any version with a known vulnerability to be functionally dead in the water and backwards compatibility not particularly relevant. But that's just my opinion and we can disagree. Backpacking is easier said than done.

At the places I've worked when a vulnerability was found in a dependency a new version with a patch was typically released within 24-48 hours or at most a week. If not, the dependency was removed altogether because it's not an acceptable risk and pointing at `main` is a hard sell due to lack of confidence in future vulnerability fixes. If Rack doesn't aim for/cater to a lot of commercial usage though, I understand. Just wanted to share my experiences for context.

robopolomsky:
Hi, could somebody please cherry-pick this issue into 2.2? It's a vulnerability issue known for 2 years https://security.snyk.io/vuln/SNYK-RUBY-RACK-1061917 and unfortunately [ruby-rails](https://rubyonrails.org/) still didn't include rack 3 and I don't know when it will happen.

PR #1734, opened 2021-01-27
📂 Add option for `gzip_compression_level` for `Rack::Deflater`. 

bayleedev:
This would be useful on a system with smaller memory to be set to a lower number, and a system with higher compression needs to compress up to level 9.

`:gzip_compression_level ::` An integer from 0 to 9 controlling the
level of compression; 1 is fastest and produces the least compression,
and 9 is slowest and produces the most compression. 0 is no compression.
The default is 6. This is only needed for the `gzip` type encoding.

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -24,6 +24,10 @@ jobs:
         restore-keys: |
           bundle-use-ruby-${{matrix.os}}-${{matrix.ruby}}-
 
+    - name: Updating packages (ubuntu)
+      if: matrix.os == 'ubuntu-latest'
+      run: sudo apt-get update
+
     - name: Installing packages (ubuntu)
       if: matrix.os == 'ubuntu-latest'
       run: sudo apt-get install libfcgi-dev libmemcached-dev
@@ -60,6 +64,10 @@ jobs:
         key: bundle-use-ruby-${{matrix.os}}-${{matrix.ruby}}-${{hashFiles('**/Gemfile')}}
         restore-keys: |
           bundle-use-ruby-${{matrix.os}}-${{matrix.ruby}}-
+
+    - name: Updating packages (ubuntu)
+      if: matrix.os == 'ubuntu-latest'
+      run: sudo apt-get update
 
     - name: Installing packages (ubuntu)
       if: matrix.os == 'ubuntu-latest'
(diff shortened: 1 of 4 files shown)


bayleedev:
cc @jeremyevans 572fe6203e6a75375d6d2a67993e340561dc0a43 [rack/rack/pull/1733 comment](https://github.com/rack/rack/pull/1733#issuecomment-764769802)

ioquatix:
This looks good to me.

Do you think `gzip_compression_level` is too specific, or maybe `compression_level` is sufficient?

jeremyevans:
This also looks good to me.  Currently, only gzip encoding is supported, but I could see `br` possibly being supported in the future, so `gzip_compression_level` may be fine as an option name.  @ioquatix are you OK with the current option name?

ioquatix:
I think if the compression could be different, which is possible, we should not embed the name of the compression key in the option.

bayleedev:
@ioquatix what were you thinking? `gzip_options: { compression_level: 3 }` ? Making it `compression_level` could make its use ambiguous with multiple compression algorithms.

nateberkopec:
Just want to point out that brotli has a `10` compression level, zlib only goes up to `9` (insert spinal tap reference here)

ioquatix:
I took a look at how this is implemented in other places.

In `protocol-http` (falcon) we use `Zlib::Deflate` which is non-algorithm-specific.

It would be nice to set options in a generic way.

I checked node.js and they expose gzip, deflate, brotli all through the same interface, but there is a way to set options. https://nodejs.org/api/zlib.html#zlib_brotli_constants

My suggestion would be something like:

```
deflated = Rack::Deflater.new(..., **options)
  @options = options

# In the deflater:
Zlib::Deflate.new(..., **options)
```

This is just pseudo code. This means we are not interpreting the value of the options which means we don't worry about naming or anything else.

bayleedev:
I love the idea. I'm unsure how to implement that idea. Since `ZLib::Deflate` and `Zlib::GzipWriter` both take `level` as an unnamed positional argument, not options. Thoughts?

bayleedev:
@ioquatix @nateberkopec any update on this? Happy to change the code. I'm mostly looking for the functionality to exist more than how I want the code to be structured. Let me know. [:

PR #1736, opened 2021-02-17
Fix generation of multipart filename with space.

muirdm:
The multipart generator was escaping filenames using
Rack::Utils.escape, but the parser was using Rack::Utils.unescape_path.
If the file name contained spaces such as "foo bar.txt",
escape would encode as "foo+bar.txt" and then unescape_path would
decode to "foo+bar.txt", incorrectly leaving the plus sign. Fix by
changing the generator to use escape_path to match the parser.


muirdm:
I realized that this isn't even the code that was giving me problems. The same issue exists in the rack-test gem, so I will address it there. We can continue with this PR if you think it has value.

jeremyevans:
I think it makes sense to change the behavior in Rack if it fixes a bug.  Does the new test you added fail if you don't also changing the escaping?  If so, I think it makes sense.

muirdm:
> Does the new test you added fail if you don't also changing the escaping?

Yes, it fails.

For reference, here is the rack-test PR: https://github.com/rack/rack-test/pull/275.

jeremyevans:
I'm fine with merging this, but I'll leave this open for a while to allow time for other committers to review and provide feedback.

PR #1738, opened 2021-02-27
Make tests pass on hosts that have no ipv4 connectivity

terceiro:
Hosts with no ipv4 connectivity are going to be increasingly common in
the future. In such hosts, one will see errors like the following when
running the rack tests:

Errno::EADDRNOTAVAIL: Cannot assign requested address - bind(2) for 127.0.0.1:9210

Binding to localhost instead of explicitly to
127.0.0.1 solves the problem - `localhost` is resolved to all
_available_ local addresses and those are used.  Client connections
need to do the same, and it all just works regardless of the local
networking stack.

### DIFF
--- a/test/spec_server.rb
+++ b/test/spec_server.rb
@@ -89,14 +89,14 @@ describe Rack::Server do
   end
 
   it "get options from ARGV" do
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch. This seems reasonable to me.  I'm OK merging it as long as no other committer expresses an objection.

ioquatix (approved):

PR #1739, opened 2021-03-14
CI: Workaround YAML gotcha in Actions

dentarg:
To avoid unexpectedly stop testing Ruby 3.0 when Ruby 3.1 is released.

See https://github.com/actions/runner/issues/849

At https://github.com/rack/rack/runs/2041788658?check_suite_focus=true#step:3:3 we can see that the setup-ruby action ran with just `3` as the input and not `3.0`.


ioquatix:
Thanks!

PR #1742, opened 2021-04-14
Allow ws and wss as forwarded scheme/proto headers

### LINKED ISSUE #1730: scheme wss is not allowed
Hello,

my ActionCable connections stopped working after I have upgraded Traefik. After some investigation, I think the problem is the changed `X-Forwarded-Proto` header which now ist set to `wss` instead of `https`. ActionCable verifies the origin of a request and if the request is allowed with the help of the [`Rack::Request#ssl?`](https://github.com/rails/rails/blob/master/actioncable/lib/action_cable/connection/base.rb#L204) method. 
The method [`ssl?`](https://github.com/rack/rack/blob/master/lib/rack/request.rb#L350) would return `true` if the `scheme` is detected as `https` or `wss`. 
The problem is that [`scheme` ](https://github.com/rack/rack/blob/master/lib/rack/request.rb#L210) returns `nil` although the header `X-Forwarded-Proto` is set to `wss`.

I suggest to add `wss` and maybe `ws` to the [`ALLOWED_SCHEMES`](https://github.com/rack/rack/blob/master/lib/rack/request.rb#L20) constant so that the [`forwarded_scheme`](https://github.com/rack/rack/blob/master/lib/rack/request.rb#L626) method can extract it. 

What do you think?

erwanst:
This change improves SSL detection in apps running behind some reverse-proxies.

Fixes #1730

### DIFF
--- a/test/spec_request.rb
+++ b/test/spec_request.rb
@@ -667,6 +667,14 @@ class RackRequestTest < Minitest::Spec
     request.scheme.must_equal "http"
     request.wont_be :ssl?
 
+    request = make_request(Rack::MockRequest.env_for("/", 'HTTP_X_FORWARDED_SCHEME' => 'ws'))
+    request.scheme.must_equal "ws"
+    request.wont_be :ssl?
+
+    request = make_request(Rack::MockRequest.env_for("/", 'HTTP_X_FORWARDED_PROTO' => 'ws'))
+    request.scheme.must_equal "ws"
+    request.wont_be :ssl?
+
     request = make_request(Rack::MockRequest.env_for("/", 'HTTPS' => 'on'))
     request.scheme.must_equal "https"
     request.must_be :ssl?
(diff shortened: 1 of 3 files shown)


test/spec_request.rb
```diff
@@ -695,13 +703,21 @@ def initialize(*)
...
     request.scheme.must_equal "https"
     request.must_be :ssl?
 
     request = make_request(Rack::MockRequest.env_for("/", 'HTTP_X_FORWARDED_PROTO' => 'https, http, http'))
     request.scheme.must_equal "https"
     request.must_be :ssl?
+
+    request = make_request(Rack::MockRequest.env_for("/", 'HTTP_X_FORWARDED_PROTO' => 'wss'))
+    request.scheme.must_equal "wss"
+    request.must_be :ssl?
```
erwanst, line 720: I wasn't sure here: the test is getting rather large, should I split it into several tests (for example one for each scheme ?).

ioquatix: It's up to you.

ioquatix (approved):

jeremyevans (approved):

erwanst:
Thanks !

PR #1744, opened 2021-04-17
deflater.rb - optimize GzipStream#each

MSP-Greg:
The current code in the method is below, and was added in [2008](https://github.com/rack/rack/commit/cabc20df5298), [2009](https://github.com/rack/rack/commit/c81542e28aa7) & [2011](https://github.com/rack/rack/commit/37d2b2fe4066).  When `@body` is a File object, `@body.each` is equivalent to `@body.gets`.  For a large 'text' file (css, js, html), this results in many iterations.

A lot has changed in the ten years, so added code to read the `@body` in 128kB chunks.  Old code is used if `@body` isn't a File object.

This results in a large speed increase.

Assuming that `@body.read` throws the same errors as `@body.each`, so no additional error handling is needed.

Note that `@body.is_a? ::File` is used for detection.  I'm not sure if `@body.is_a? ::IO` would be better...

```ruby
@body.each { |part|
  # Skip empty strings, as they would result in no output,
  # and flushing empty parts would raise Zlib::BufError.
  next if part.empty?
  gzip.write(part)
  gzip.flush if @sync
}
```

### DIFF
--- a/lib/rack/deflater.rb
+++ b/lib/rack/deflater.rb
@@ -77,6 +77,9 @@ module Rack
 
     # Body class used for gzip encoded responses.
     class GzipStream
+
+      BUFFER_LENGTH = 128 * 1_024
+
       # Initialize the gzip stream.  Arguments:
       # body :: Response body to compress with gzip
       # mtime :: The modification time of the body, used to set the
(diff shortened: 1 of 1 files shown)


MSP-Greg:
Using a 234kB file with ~ 4.5k lines, I tested this using wrk and Puma.  With this PR, requests per second was over 100 times faster.

jeremyevans (approved):
Sorry that nobody has responded to this until now.  This change looks good to me.

ioquatix:
LGTM.

PR #1745, opened 2021-04-19
Support callable body for explicit streaming support.

ioquatix:
Allow a rack response body to explicitly opt into streaming by implementing `#call` instead of `#each`. This interface is desirable for a number of reasons. However, this PR should be considered an experimental spike as there are also a number of downsides, and this essentially represents a breaking change to some typical but unwise use cases.

An example would look something like this:

```
run lambda{|env|
  body = lambda do |stream|
    stream.write("Hello World")
  ensure
    stream.close
  end
  
  [200, [], body]
}
```

Semantically, it is almost identical to partial hijack. Full hijack is removed. Of course, servers are welcome to support both Rack 2.x hijack, and 3.x whatever.

- [x] Falcon PoC: https://github.com/socketry/falcon/pull/140
- [x] Puma PoC: https://github.com/puma/puma/pull/2740
- [ ] Passenger PoC:
- [ ] Unicorn PoC:

## Current Design

Rack currently supports streaming using `rack.hijack`. There are two ways to hijack the underlying socket for a given request: a partial hijack and a full hijack. Not all servers support both modes of operation and neither method is fully compatible with HTTP/2.

### Partial Hijack

A partial hijack allows the web server to respond with the response status, headers, and then yields the stream back to the application code. It does this by using a special `rack.hijack` header:

``` ruby
def call(env)
	body = proc do |stream|
		stream.write("Hello World")
		stream.close
	end
	
	if env['rack.hijack?']
		return [200, {'rack.hijack' => body}, nil]
	end
end
```

- It is not clear whether the body needs to handle the formatting of the response, i.e. chunked encoding, compression, transfer encoding, etc.
- It mixes `rack.` specific headers with normal HTTP headers which is computationally expensive for the server to detect.
- It's not clear what the returned body should be.

### Full Hijack

A full hijack allows the web server to completely take control of the underlying socket. This occurs at the point the `rack.hijack` is invoked.

``` ruby
def call(env)
	if hijack = env['rack.hijack']
		socket = hijack.call
		
		Thread.new do
			socket.write("Hello World")
			socket.close
		end
		
		return nil # ???
	end
end
```

- The socket might not actually be a raw socket, if the server is using TLS.
- Extreme care is required in order to handle the protocol correctly.
- It's not clear if such a design can work with anything other than HTTP/1.
- The user code **must** return from the `call(env)` method in order for the server continue handling requests.
- It's not clear what the response should be.

### Supporting HTTP/2

[...]

PR #1748, opened 2021-04-26
Clarify streaming response body behavior in SPEC

wjordan:
This PR stems from the discussion in #1745 concerning how to distinguish between streaming and buffered bodies in the Rack interface. The proposal here adds the following two sentences to the specification:

> Strings must be processed individually as they are yielded by `each`.
However, if the Body responds to `to_ary` it can be implicitly coerced to an Array, which may then be processed all at once.

I consider this more of a 'clarification' than a breaking change to the interface, as previous discussions and implementations in the Rack project have suggested that treating `Array` bodies (or objects coercible to `Array` via `#to_ary`) as fixed-length content, in contrast to any other response bodies which may be possibly-streaming, indeterminate content, had been implicit in the interface since at least Dec 2008 (7a0476be9289b6e38dbc27946168d41894439f42).

For example, see https://github.com/rack/rack/pull/166#commitcomment-391926 from May 2011:

> While to_ary is not mentioned in the rack spec, whether a body is an array has up until now signaled the difference between bodies of known size and bodies of unknown size, at least as far as choosing between chunked and non-chunked transfer encodings is concerned.

This PR also updates `ContentLength` and `ETag` middleware to demonstrate how this spec clarification can guide a more consistent implementation of 'buffering' middleware/servers in practice. There are a few steps:

1. Check `body.respond_to?(:to_ary)`, and skips the buffering/transformation logic if the response is not coercible to `Array`
2. Call `body = body.to_ary`, which coerces the response body to ensure that it is an `Array` of `Strings`.
3. Operate on the body as needed (calculate SHA256 digest, sum bytesize of strings, etc).

Some notes:

- Note that `BodyProxy` is no longer needed in middleware using this pattern, because `Array` objects never respond to `close`. (If `#close` is required by an object that can be coerced to an `Array`, it should close as part of its `to_ary` implementation. For example, see the update to the ['close bodies that need to be closed'](https://github.com/rack/rack/compare/master...wjordan:to_ary-buffered?expand=1#diff-7f91c8639db9f03cf387d37f31c3215617aa0f595c50d2f62ebaa49357807051R56-R67) content length spec test.)
- Web servers may also adopt this pattern as an optimization to buffer content into a String before writing to the socket (or write an array of Strings with a single call to `#write`), to reduce the number of socket writes.
- This doesn't affect and is unrelated to the (partial/full) hijack APIs (#481), though the clarification allows for more well-defined behavior for one of the use-cases (streaming responses) that hijack was also originally designed for, so it may make the latter slightly more redundant.
- There is a slight but significant difference between the proposed behavior and similar historical behavior relying on `respond_to?(:ary)` in `ContentLength`. Previously, the `to_ary` coercion was never actually _performed_ on the body before eager-loading, it was assumed the content was already buffered as long as the `respond_to?` was satisfied. However, this allowed for overriding `to_ary` to return anything, even `nil` (see e.g., the workaround in rails/rails#16793), just to trigger the eager-loading behavior without actually producing the required `Array` of `String`s representing fully-buffered content. In this proposal (and implementation), the coercion is actually performed which prevents this kind of confusion.
- Updates to `Rack::Lint` are included to help enforce this new clarification.

### DIFF
--- a/SPEC.rdoc
+++ b/SPEC.rdoc
@@ -283,6 +283,10 @@ to that produced by calling +each+; this may be used by the
 server as an alternative, possibly more efficient way to
 transport the response.
 
+Strings must be processed individually as they are yielded by +each+.
+However, if the Body responds to +to_ary+ it can be implicitly coerced to
+an Array, which may then be processed all at once.
+
 The Body commonly is an Array of Strings, the application
 instance itself, or a File-like object.
 == Thanks
--- a/SPEC.rdoc
+++ b/SPEC.rdoc
@@ -273,19 +273,26 @@ and must only yield String values.
 The Body itself should not be an instance of String, as this will
 break in Ruby 1.9.
 
+Middleware must not call +each+ directly on the Body.
+Instead, middleware can return a new Body that calls +each+ on the
+original Body, yielding at least once per iteration.
+
+If the Body responds to +to_ary+, it must return an Array whose
+contents are identical to that produced by calling +each+.
+Middleware may call +to_ary+ directly on the Body and return a new Body in its place.
+In other words, middleware can only process the Body directly if it responds to +to_ary+.
+
 If the Body responds to +close+, it will be called after iteration. If
-the body is replaced by a middleware after action, the original body
-must be closed first, if it responds to close.
+the original Body is replaced by a new Body, the new Body
+must close the original Body after iteration, if it responds to +close+.
+If the Body responds to both +to_ary+ and +close+, its
+implementation of +to_ary+ must call +close+ after iteration.
 
 If the Body responds to +to_path+, it must return a String
 identifying the location of a file whose contents are identical
 to that produced by calling +each+; this may be used by the
 server as an alternative, possibly more efficient way to
 transport the response.
-
-Strings must be processed individually as they are yielded by +each+.
-However, if the Body responds to +to_ary+ it can be implicitly coerced to
-an Array, which may then be processed all at once.
 
 The Body commonly is an Array of Strings, the application
 instance itself, or a File-like object.
--- a/lib/rack/content_length.rb
+++ b/lib/rack/content_length.rb
@@ -19,17 +19,11 @@ module Rack
 
       if !STATUS_WITH_NO_ENTITY_BODY.key?(status.to_i) &&
          !headers[CONTENT_LENGTH] &&
-         !headers[TRANSFER_ENCODING]
+         !headers[TRANSFER_ENCODING] &&
+         body.respond_to?(:to_ary)
 
-        obody = body
-        body, length = [], 0
-        obody.each { |part| body << part; length += part.bytesize }
-
-        body = BodyProxy.new(body) do
-          obody.close if obody.respond_to?(:close)
-        end
-
-        headers[CONTENT_LENGTH] = length.to_s
+        body = body.to_ary
+        headers[CONTENT_LENGTH] = body.sum(&:bytesize).to_s
       end
 
       [status, headers, body]
--- a/lib/rack/etag.rb
+++ b/lib/rack/etag.rb
@@ -27,12 +27,9 @@ module Rack
       status, headers, body = @app.call(env)
       headers = Utils::HeaderHash[headers]
 
-      if etag_status?(status) && etag_body?(body) && !skip_caching?(headers)
-        original_body = body
-        digest, new_body = digest_body(body)
-        body = Rack::BodyProxy.new(new_body) do
-          original_body.close if original_body.respond_to?(:close)
-        end
+      if etag_status?(status) && body.respond_to?(:to_ary) && !skip_caching?(headers)
+        body = body.to_ary
+        digest = digest_body(body)
         headers[ETAG_STRING] = %(W/"#{digest}") if digest
       end
 
@@ -53,24 +50,18 @@ module Rack
         status == 200 || status == 201
       end
 
-      def etag_body?(body)
-        !body.respond_to?(:to_path)
-      end
-
       def skip_caching?(headers)
         headers.key?(ETAG_STRING) || headers.key?('Last-Modified')
       end
 
       def digest_body(body)
-        parts = []
         digest = nil
 
         body.each do |part|
-          parts << part
           (digest ||= Digest::SHA256.new) << part unless part.empty?
         end
 
-        [digest && digest.hexdigest.byteslice(0, 32), parts]
+        digest && digest.hexdigest.byteslice(0,32)
       end
   end
 end
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -40,6 +40,7 @@ module Rack
     end
 
     def _call(env)
+      @env = env
       ## It takes exactly one argument, the *environment*
       raise LintError, "No env given" unless env
       check_env env
@@ -67,6 +68,13 @@ module Rack
       check_content_type status, headers
       check_content_length status, headers
       @head_request = env[REQUEST_METHOD] == HEAD
+
+      @lint = (env['rack.lint'] ||= []) << self
+
+      if (env['rack.lint.body_iteration'] ||= 0) > 0
+        raise LintError, "Middleware must not call #each directly"
+      end
+
       [status, headers, self]
     end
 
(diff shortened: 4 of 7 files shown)


ioquatix (approved):
This looks good to me.

wjordan:
Note: this change will require a corresponding change in Rails's `ActionDispatch::Response::RackBody` to effectively revert rails/rails#16793 (overriding `#to_ary` to return `nil`), which is a hacky work-around for #734, which was caused by the hacky fix to #419 that has since been fixed more properly in #1453. 😅

jeremyevans (commented):
In terms of the idea, I think it is reasonable.  It's conservative in the sense that if the body is already an array, its size is known and fixed.  This does fix cases where middleware that ship with Rack will buffer when buffering is not be desireable.  The downside is the breakage of backwards compatibility in cases where the current behavior of the middleware handling non-Array responses is desired.  However, for a major version change, I guess that is acceptable.

In terms of the implementation, I think the wording should be more specific.  The phrases `must be processed individually as they are yielded` and `may be processed all at once` are too vague IMO. I'm guessing the first phrase means by the web server and the second phrase means by middleware, but that's not obvious. Maybe something like:

```
Middleware must not call +each+ directly on the Body.  Middleware can
return a new Body which has an +each+ method that calls +each+ on the
returned Body.

If the Body responds to +to_ary+, middleware can call +to_ary+ on the
Body, can call +each+ on the resulting array, and can return the resulting
array as the new Body.  In other words, middleware can only process the Body
before returning if the body responds to +to_ary+.
```

Second, `SPEC.rdoc` should not be modified directly.  Instead `lib/rack/lint.rb` should be modified to add appropriately formatted comments, and `rake spec` should be run to regenerate `SPEC.rdoc`.

Finally, the class-level comments for the ContentLength and ETag middleware should be modified to explicitly state that they only set the headers if the body responds to `to_ary`.  I suppose this isn't necessary if that's stated in `SPEC.rdoc`, but I think being more specific would help.

matthewd:
I'd consider downgrading to a "should" -- if someone wants to create a CacheAllBodiesEvenBigOnes middleware, to me that should be "consciously deviating from the recommended path" not "breaking the letter of the spec".

I also don't think we should actually call the detected `to_ary`... that seems avoidable, yet opens up weirdness where the actual response could differ based on whether your parent middleware used `to_ary.each` or plain `each` -- which is fixable by defining that those must produce identical results, as we do with `to_path`... but at that point the body-author can handle that themselves. (I also don't think `to_ary` should absolve you of the need to check for `close`.)

I noodled with a few possible phrasings, but couldn't come up with a compelling specific suggestion that conveys "[if you replace the body, and you are consuming the original] you should yield approximately as often [or at least not less often] than the body you're wrapping; if that wrapped body responds to `to_ary`, then you may ignore this and iterate as eagerly as you like". 

I think we might need a BodyProxy subclass that defines `to_ary` too -- generally, I don't think "you _may_ iterate this body freely if you have need" does or should necessarily imply "please do so at your earliest convenience"... and contrariwise, a middleware that needs after-close handling should not hide the eager-iterability of its enclosed body.

wjordan:
Thanks all for the comments and feedback so far!

@jeremyevans thanks for suggesting more specific wording (and for reminding me that the spec was generated from the lint comments!).
- I really like the line about middleware not calling `#each` directly, I think it makes the interface very clear.
- I think the line about what exactly can be done with the object returned by `to_ary` could be left out. The intent of the proposal is to make `to_ary` an action that fully 'consumes' a Rack body, so the return value from that method is just a standard Ruby Array of Strings and no longer the Rack 'Body' at that point. I don't think anything prescriptive beyond that is needed.
- I moved the spec changes into Lint comments, and added some Lint behavior (plus spec tests) as an implementation of this proposal.

@matthewd thanks for your detailed comments as well.
- I think downgrading 'must' to 'should' muddles the proposal enough to make it not worth adding to the Rack spec in the first place (since it's primarily concerned with formal requirements, not a 'recommended path'). Specifically, it would make the `Rack::Lint` implementation impossible (or at least much more difficult). I'd be interested in others weighing in on this as well however.
- I think actually calling `to_ary` is necessary. Beyond the specific case I already mentioned (to prevent confusion caused by narrow workarounds such as `#to_ary` returning `nil`), more generally, yielding individual strings via `#each` isn't always implemented in the same way as producing an array of strings, even if the end result is equal. For example, consider an Array (a common Rack body type) - `to_ary` is a no-op, while iterating via `#each` to produce a new array can be an expensive operation for large content. I do like the suggestion that the content produced must be identical to `#each` to match `#to_path`, though.
- I think it should be the `to_ary` implementation's responsibility to call `close` after iterating through its content. I think it makes sense for a custom content class to encapsulate all of the details of its own coercion (iterating plus closing), without leaking any of that out to the code consuming the returned array. It makes things much easier to reason about.
- Rather than needing _more_ `BodyProxy` subclasses with this proposal, I believe that with `close` better encapsulated, `BodyProxy` actually becomes _less_ necessary for middleware to eagerly-consume content. See the changes to `ETag` and `ContentLength` in this PR.

I did another pass on the proposed spec changes while trying to consider all the comments raised. It's quite a bit more verbose, but less ambiguous:

```
Middleware must not call +each+ directly on the Body.
Instead, middleware can return a new Body that calls +each+ on the
original Body, yielding at least once per iteration.

If the Body responds to +to_ary+, it must return an Array whose
contents are identical to that produced by calling +each+.
Middleware may call +to_ary+ directly on the Body and return a new Body in its place.
In other words, middleware can only process the Body directly if it responds to +to_ary+.

If the Body responds to +close+, it will be called after iteration. If
the original Body is replaced by a new Body, the new Body
must close the original Body after iteration, if it responds to +close+.
If the Body responds to both +to_ary+ and +close+, its
implementation of +to_ary+ must call +close+ after iteration.
```

Note also that the final paragraph replaces the existing paragraph on `+close+`. (In effect, this PR clarifies exactly how 'the body is replaced by a middleware after action'.)

ioquatix:
This looks great to me, the clarity is excellent. I really like the direction here.

ioquatix (approved):

jeremyevans:
I'm mostly OK with the revised language, except that I think we should change:

```
If the Body responds to both +to_ary+ and +close+, its
implementation of +to_ary+ must call +close+ after iteration
```

to:

```
If the Body responds to both +to_ary+ and +close+, middleware
that call +to_ary+ on the Body do not need to call +close+.
```

We shouldn't describe how Body objects must work internally, only the interface that middleware should use to interact with them.

I think we should also change:

```
Middleware may call +to_ary+ directly on the Body and return a new Body in its place.
In other words, middleware can only process the Body directly if it responds to +to_ary+.
```

to:

```
Middleware may call +to_ary+ on the Body and use the result as a new Body.
In other words, middleware can only process the Body if it responds to +to_ary+,
and must use the result of +to_ary+ instead of operating on the Body directly.
```

The reason for this change is to make it clear that you cannot operate on the Body itself just because the body responds to `to_ary`.  You may make sure you are dealing with the result of `to_ary`.  I want to forbid this:

```ruby
if body.respond_to?(:to_ary) && body.respond_to?(:[])
  body[0] = 'foo'
end
```

You have to do instead:

```ruby
if body.respond_to?(:to_ary)
  body = body.to_ary
  body[0] = 'foo'
end
```

Additionally, you already mentioned that `I think actually calling to_ary is necessary`, but I think the previous language was too ambiguous for that, and language I am recommending makes it more clear.

Finally, we should probably use a separate heading for this new section such as `Middleware interaction with the Body`, because I want to be clear that this section is not relevant to servers, who should always use `each` and `close` and not `to_ary`.

wjordan:
@jeremyevans thanks again for your comments and for taking a close look at the proposed language. I think we're getting very close!

Agree with leaving out the line about `to_ary` calling `close` internally, since that part isn't strictly a concern of the middleware interface. From the interface perspective, perhaps we can say that the original Body must be closed only if it's "replaced by middleware without calling `to_ary`".

Also agree with saying that middleware "must use the result of `to_ary`" - perhaps this could be made even more precise by saying middleware "must not call any further methods on the original Body" after calling `to_ary`?

It might be hard to _forbid_ code like `body[0] = 'foo'` as in your example, without a stricter rule like "Middleware must not call any methods on the original Body not listed here" - which might work but sounds a bit drastic (maybe worth thinking about separately).

> Finally, we should probably use a separate heading for this new section such as Middleware interaction with the Body, because I want to be clear that this section is not relevant to servers, who should always use `each` and `close` and not `to_ary`.

I actually expect servers to use `to_ary` when available as well. I mentioned one use-case which is to buffer non-streaming responses to minimize the number of `write` syscalls on the socket. Many Rack servers already do optimizations like this in practice, typically by checking for an `Array` body (sometimes a single-element Array) as a special-case. That said, separating the body-object interface from the middleware/server interface in this section could make the added parts a bit clearer, I can try out a draft of that soon.

For now, here's another draft with the changes mentioned:

```
Middleware must not call +each+ directly on the Body.
Instead, middleware can return a new Body that calls +each+ on the
original Body, yielding at least once per iteration.

If the Body responds to +to_ary+, it must return an Array whose
contents are identical to that produced by calling +each+.
Middleware may call +to_ary+ on the Body and use the result as a new Body,
and must not call any further methods on the original Body.

If the Body responds to +close+, it will be called after iteration. If
the Body is replaced by middleware without calling +to_ary+,
the original Body must be closed first, if it responds to +close+.
```

ioquatix:
> Many Rack servers already do optimizations like this in practice

Can confirm.

ioquatix:
@wjordan I love the work you are doing here. It is clear to me that you (and everyone else) is drawing on a huge level of knowledge and the result is a clearer, more precise and easier to implement SPEC. Thanks so much and I look forward to when we can merge this!

simi:
> I don't see this in the changelog. In which version is this included?

@swrobel This wasn't released yet and you can find it only in `main` branch. It wasn't backported to current "production" branch of `2-2-stable`.

swrobel:
> @swrobel This wasn't released yet and you can find it only in `main` branch. It wasn't backported to current "production" branch of `2-2-stable`.

Thanks for clarifying @simi! I do see there's a [changelog section for 3.0 unreleased items](https://github.com/rack/rack/blob/main/CHANGELOG.md#300---unreleased), so it seems it would be helpful to have there.

PR #1749, opened 2021-04-28
Output pid if the server is already running

kyoshidajp:
I sometimes find myself trying to run a duplicate server with the same pid file. When this happens, the startup fails and the path to the pid file is displayed, but it is a bit of a pain to open the file to check the pid each time.

Instead of showing the file path, how about showing the pid itself?

Before output:

```
A server is already running. Check /file/to/pid_file.
```

After output:

```
A server is already running (pid: 69431).
```

### DIFF
--- a/lib/rack/server.rb
+++ b/lib/rack/server.rb
@@ -438,7 +438,8 @@ module Rack
       def check_pid!
         case pidfile_process_status
         when :running, :not_owned
-          $stderr.puts "A server is already running. Check #{options[:pid]}."
+          pid = ::File.read(options[:pid])
+          $stderr.puts "A server is already running (pid: #{pid}, file: #{options[:pid]})."
           exit(1)
         when :dead
           ::File.delete(options[:pid])
--- a/lib/rack/server.rb
+++ b/lib/rack/server.rb
@@ -436,30 +436,23 @@ module Rack
       end
 
       def check_pid!
-        case pidfile_process_status
-        when :running, :not_owned
-          pid = ::File.read(options[:pid])
-          $stderr.puts "A server is already running (pid: #{pid}, file: #{options[:pid]})."
-          exit(1)
-        when :dead
-          ::File.delete(options[:pid])
-        end
-      end
-
-      def pidfile_process_status
-        return :exited unless ::File.exist?(options[:pid])
+        return unless ::File.exist?(options[:pid])
 
         pid = ::File.read(options[:pid]).to_i
-        return :dead if pid == 0
+        raise Errno::ESRCH if pid == 0
 
         Process.kill(0, pid)
-        :running
+        exit_with_pid(pid)
       rescue Errno::ESRCH
-        :dead
+        ::File.delete(options[:pid])
       rescue Errno::EPERM
-        :not_owned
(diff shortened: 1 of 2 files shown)


ioquatix:
Great idea, you could even show both?

lib/rack/server.rb
```diff
@@ -438,7 +438,8 @@ def write_pid
       def check_pid!
         case pidfile_process_status
         when :running, :not_owned
-          $stderr.puts "A server is already running. Check #{options[:pid]}."
+          pid = ::File.read(options[:pid]).to_i
```
ioquatix, line 441: You technically don't need to call `to_i` here.

kyoshidajp: Thanks. I will fix.

kyoshidajp:
> Great idea, you could even show both?

@ioquatix For example, like this?

```
# pattern1
A server is already running (pid: 69431, file: /file/to/pid_file).

# pattern2
A server is already running (pid=69431, file=/file/to/pid_file).
```

I don't have any particular preference, but I rather prefer pattern1.

ioquatix:
> I don't have any particular preference, but I rather prefer pattern1.

Haha. I don't mind, let's go with the first one then.

ioquatix (approved):

ioquatix:
IF the file doesn't exist, what would happen?

ioquatix:
i.e. `File.read("invalid")`

kyoshidajp:
A pid file doesn't exist, then a server will will start normally. According to the code, `pidfile_process_status` in `Rack::Server` returns `:exited`, and not making any special by caller (`check_pid!`).

https://github.com/rack/rack/blob/v2.2.2/lib/rack/server.rb#L440-L451

If the file cannot be read, it will fail to start and a message to that effect will be displayed.

```
% ll cannot_read.pid
--w-------  1 root  staff  0  4 28 12:51 cannot_read.pid
```

```ruby
Rack::Server.new(
  environment: 'development',
  pid: '/file/to/cannot_read.pid',
  Port: 3000,
  Host: 'localhost',
  config: 'config.ru',
).start
```

The execution logs is the following.

```
Traceback (most recent call last):

(traceback log)

 - /file/to/cannot_read.pid (Errno::EACCES)
```

Also if the server started same port, `Errno::EADDRINUSE` is raised like this.

```
Address already in use - bind(2) for 127.0.0.1:3000 (Errno::EADDRINUSE)
```

In other cases as well, I think the error occurs when the read fails.

jeremyevans (changes requested):
This reads the pid file twice, as `pidfile_process_status` also reads it.  I would remove the `pidfile_process_status` method and integrate it into the `check_pid!` method (`pidfile_process_status` is only called by `check_pid!`).  `pidfile_process_status` is private so this shouldn't be a backwards compatibility concern.

kyoshidajp:
@jeremyevans Thanks. I have fixed the code as you pointed out. Could you review again?

jeremyevans (approved):
Changes look good, thanks!

ioquatix:
Thanks and well done @kyoshidajp

PR #1751, opened 2021-05-25
Remove unnecessary call to `super()` in `Rack::Request::Env`

agrberg:
b2d73960e9ea6b8b15321ef190f13a290d1aedf0 changed `Rack::Request#initialize` to call `super(env)` so that the newly added `Rack::Request::Env#initialize` could be called. Previously `Rack::Request#initialize` called `super()` so `Rack::Request::Helpers#initialize` was called. This method was removed but the `super()` call was left in `Rack::Request::Env#initialize` as if it was still needed.

I don't believe this is hurting anything, it just begs the question of why there is a `super()` call that essentially goes to `BasicObject#initialize` (this is according to IRB diving, as this method isn't documented in the [Ruby docs](https://ruby-doc.org/core-3.0.1/BasicObject.html)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ All notable changes to this project will be documented in this file. For info on
 - Handle cookies with values that end in '=' ([#1645](https://github.com/rack/rack/pull/1645), [@lukaso](https://github.com/lukaso))
 - Fix multipart filename generation for filenames that contain spaces. Encode spaces as "%20" instead of "+" which will be decoded properly by the multipart parser. ([#1736](https://github.com/rack/rack/pull/1645), [@muirdm](https://github.com/muirdm))
 - `Rack::Request#scheme` returns `ws` or `wss` when one of the `X-Forwarded-Scheme` / `X-Forwarded-Proto` headers is set to `ws` or `wss`, respectively. ([#1730](https://github.com/rack/rack/issues/1730), [@erwanst](https://github.com/erwanst))
-- `Rack::Request::Env#initialize` does not need to call `super()`. This was leftover from the creation of the `Env` module. ([#1751](https://github.com/rack/rack/pull/1751)), [@agrberg](https://github.com/agrberg))
+- `Rack::Request::Env#initialize` does not need to call `super()`. This was leftover from the creation of the `Env` module. Flatten the `Env` module into `Request` as it is unused outside. ([#1751](https://github.com/rack/rack/pull/1751)), [@agrberg](https://github.com/agrberg))
 
 ## [2.2.3] - 2020-06-15
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ All notable changes to this project will be documented in this file. For info on
 - Handle cookies with values that end in '=' ([#1645](https://github.com/rack/rack/pull/1645), [@lukaso](https://github.com/lukaso))
 - Fix multipart filename generation for filenames that contain spaces. Encode spaces as "%20" instead of "+" which will be decoded properly by the multipart parser. ([#1736](https://github.com/rack/rack/pull/1645), [@muirdm](https://github.com/muirdm))
 - `Rack::Request#scheme` returns `ws` or `wss` when one of the `X-Forwarded-Scheme` / `X-Forwarded-Proto` headers is set to `ws` or `wss`, respectively. ([#1730](https://github.com/rack/rack/issues/1730), [@erwanst](https://github.com/erwanst))
+- `Rack::Request::Env#initialize` does not need to call `super()`. This was leftover from the creation of the `Env` module. ([#1751](https://github.com/rack/rack/pull/1751)), [@agrberg](https://github.com/agrberg))
 
 ## [2.2.3] - 2020-06-15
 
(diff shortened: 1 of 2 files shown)


ioquatix:
I don't have a strong opinion about this, but it looks okay given what you've said.

jeremyevans (approved):
This seems fine with me. Most Ruby subclasses do not call `super` in `initialize` if they are subclasses of `Object`.

Not really related, but I don't see the point of `Env` and `Helpers`  as separate modules, we should consider just defining the methods directly in `Request`.

agrberg:
@jeremyevans it appears you or another maintainer need to approve running the workflow before this can be merged 🙏 
image

jeremyevans:
Sorry about that. Approval to run CI for PR is still new to me. :)

matthewd:
I do feel moderately icky about having a _module_ swallow `initialize`, even though we "know" it's only included into one class, and that class's ancestors don't define one.

It's not something I feel super (heh) strongly about, though -- and that hesitation evaporates if the module is inlined.

ioquatix:
I agree all discussion here. @agrberg would you like to remove the module entirely? Do you have time to rework the PR?

agrberg:
@matthewd for my own future learning and knowledge sharing, do you have any resources on best practices for modules w/ constructors? The only thing I've found so far is [this cop from Rubocop](https://docs.rubocop.org/rubocop/1.1/cops_lint.html#lintmissingsuper) on constructors and lifecycle callbacks. Its example is short and to the point but it doesn't mention anything specific about modules.

ioquatix:
@agrberg Ruby 3 solves parts of this issue by allowing you do do this:

``` ruby
module MyModule
  def initialize(...)
    super
    # my stuff
  end
end

MyClass.prepend(MyModule)
```

agrberg:
Sorry for the delay, work gets the better of you even w/ a day off 😅.

I've removed the `Env` module as requested. I felt it wouldn't be the best choice to remove the `Helpers` as there are a number of methods and constants that are more readable/understandable being kept separate.

I've also applied some of the more modern Ruby practices to increase code readability and community commonality.

jeremyevans:
Personally, I would prefer if the last two commits were removed.  It think it is not a good idea to mix style changes and code changes in the same pull request.

In general, I'm against unnecessary style changes outside of egregious cases.  They make `git blame` significantly more difficult to use, for no real benefit.  I'm more strongly against `@env` -> `env` changes, which will make things slower for no significant benefit (do we really care to support someone overriding `Request#env`?).

Other committers may feel differently about this, I'm just giving my opinion. I've heard @tenderlove dislikes `@` signs. :)

tenderlove:
> In general, I'm against unnecessary style changes outside of egregious cases.

+1. We should roll back the last two commits.  Style can be discussed in a separate PR.

> I'm more strongly against `@env` -> `env` changes, which will make things slower for no significant benefit 

Also +1 here.  The request object is supposed to be hiding `@env`, and I'm really not comfortable adding a method people can use to access it.

> (do we really care to support someone overriding `Request#env`?).

No, absolutely not.

agrberg:
Removed the last two commits. I understand the annoyances of git blame issues. Personally I prioritize them lower than readability to bring more folks into the code base but I'm happy to leave them out in this case.

Would it be best to remove `attr_reader :env`? I didn't as it was originally part of the `Env` module but I'm pretty sure nothing is depending on it. Are you saying that using an `attr_reader` is slower than directly accessing an ivar?

tenderlove:
> Many tests fail when this is removed 🙃

Yes. I'd like to remove the reader someday 😆

> Are you saying that using an `attr_reader` is slower than directly accessing an ivar?

Yes, it's slower. There is some overhead when checking the method call inline cache.  I wouldn't convert all attr_reader calls in an app to ivar reads because it probably isn't a bottleneck.  However, I wouldn't convert all ivar reads to method calls unless there's some important reason to do it.  I don't want to support people overriding `Request#env`, and besides enabling that I'm not sure what the improvement would be. (also to be clear, I'm fine with `@` signs

agrberg:
Appreciate the knowledge drop 👍 That's not something I've benchmarked before but turning almost all ivar reads into a private `attr_reader` has been a pattern of mine for sometime that I should rethink.

jeremyevans (approved):

ioquatix:
Thanks everyone, LGTM.

tenderlove:
> has been a pattern of mine for sometime that I should rethink.

Only after you've profiled and found it to be a bottleneck though!!!  If you like that style, you should keep using it. ❤️

tenderlove:
Btw, this change will break Rails since we use the env module [here](https://github.com/rails/rails/blob/5ddff2beb1a819d6b300fbef2731cba9c5f71db2/actionpack/lib/action_dispatch/http/request.rb#L27). cc @matthewd

ioquatix:
I believe Ruby has a much better chance of optimising `@...` to direct pointer read than a method call which is indirect.

agrberg:
> Btw, this change will break Rails since we use the env module [here](https://github.com/rails/rails/blob/5ddff2beb1a819d6b300fbef2731cba9c5f71db2/actionpack/lib/action_dispatch/http/request.rb#L27).

Would a simple fix be `include Rack::Request`? I.e. was `Rack::Request::Env` included because the former wouldn't work or because it was less, more specific behavior to include?

PR #1752, opened 2021-05-28
Fix MockRequest cookie parsing for hash with array value

robinbortlik:
### Problem:
`Rack::Response.new` allows to set multiple cookies just by passing hash with key 'Set-Cookie' and value as an array containing cookies we want to set.
example:
```ruby
Rack::Response.new(
  [ "Authentication failed" ],
  401,
  {
    "Content-type" => "text/error",
    "Set-Cookie" => ["foo=bar", "baz=ban"]
  }
)
```
Unfortunately `MockRequest` class is not able to parse such definition of cookies as it expects cookie value to be always a string.

### Solution:
Wrap every cookie value to `Array()` which converts string values to array. So we can safely iterate over the values. In case the value is already an array it simply return it.

### DIFF
--- a/lib/rack/mock.rb
+++ b/lib/rack/mock.rb
@@ -233,18 +233,20 @@ module Rack
       cookies = Hash.new
       if headers.has_key? 'Set-Cookie'
         set_cookie_header = headers.fetch('Set-Cookie')
-        set_cookie_header.split("\n").each do |cookie|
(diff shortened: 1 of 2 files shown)


robinbortlik:
Hi @tenderlove , 

may I ask for a feedback on this PR? Are there any objections or we can merge? Sorry for asking directly.

jeremyevans (approved):
Looks good to me.

PR #1753, opened 2021-06-04
Properly set the expires attribute for the mock response cookie

walro:
Prior to this change, cookies with the `Expires` attribute would have the attribute stored as a `String` object  rather than a `Time` object. `CGI::Cookie` expects a `Time` object [[1]]. Having strings could lead to confusing errors later on. For example, calling `Rack::MockResponse#inspect` would lead to ``undefined method `gmtime' for "Fri, 03 Jun 2022 19:37:33 GMT":String``.

As per RFC 6265, if a cookie has both the `Max-Age` and the `Expires` attribute, `Max-Age` has precedence. [[2]]

[1]: https://ruby-doc.org/stdlib-3.0.1/libdoc/cgi/rdoc/CGI/Cookie.html
[2]: https://datatracker.ietf.org/doc/html/rfc6265#section-4.1.2.2

### DIFF
--- a/lib/rack/mock.rb
+++ b/lib/rack/mock.rb
@@ -4,6 +4,7 @@ require 'uri'
 require 'stringio'
 require_relative '../rack'
 require 'cgi/cookie'
+require 'time'
 
 module Rack
   # Rack::MockRequest helps testing your Rack application without
@@ -260,14 +261,18 @@ module Rack
         if bit.include? '='
           cookie_attribute, attribute_value = bit.split('=', 2)
           cookie_attributes.store(cookie_attribute.strip, attribute_value.strip)
-          if cookie_attribute.include? 'max-age'
-            cookie_attributes.store('expires', Time.now + attribute_value.strip.to_i)
-          end
         end
         if bit.include? 'secure'
           cookie_attributes.store('secure', true)
         end
       end
+
+      if cookie_attributes.key? 'max-age'
+        cookie_attributes.store('expires', Time.now + cookie_attributes['max-age'].to_i)
+      elsif cookie_attributes.key? 'expires'
+        cookie_attributes.store('expires', Time.httpdate(cookie_attributes['expires']))
+      end
+
       cookie_attributes
     end
 
--- a/test/spec_mock.rb
+++ b/test/spec_mock.rb
@@ -21,6 +21,8 @@ app = Rack::Lint.new(lambda { |env|
   response.set_cookie("session_test", { value: "session_test", domain: ".test.com", path: "/" })
   response.set_cookie("secure_test", { value: "secure_test", domain: ".test.com",  path: "/", secure: true })
   response.set_cookie("persistent_test", { value: "persistent_test", max_age: 15552000, path: "/" })
+  response.set_cookie("persistent_with_expires_test", { value: "persistent_with_expires_test", expires: Time.httpdate("Thu, 31 Oct 2021 07:28:00 GMT"), path: "/" })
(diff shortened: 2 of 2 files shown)


lib/rack/mock.rb
```diff
@@ -257,9 +257,13 @@ def identify_cookie_attributes(cookie_filling)
       cookie_bits.drop(1).each do |bit|
         if bit.include? '='
           cookie_attribute, attribute_value = bit.split('=', 2)
-          cookie_attributes.store(cookie_attribute.strip, attribute_value.strip)
           if cookie_attribute.include? 'max-age'
             cookie_attributes.store('expires', Time.now + attribute_value.strip.to_i)
+          elsif cookie_attribute.include? 'expires'
+            next if cookie_attributes.key? 'expires' # max-age attribute has precedence
+            cookie_attributes.store('expires', Time.httpdate(attribute_value.strip))
```
jeremyevans, line 264: I think this would be easier to understand if it used the more idiomatic form:

```ruby
cookie_attributes['expires'] ||= Time.httpdate(attribute_value.strip)
```

I realize that makes it less consistent with the rest of the method, but the rest of the method should probably switch from `store` to the more idiomatic `[]=`.

lib/rack/mock.rb
```diff
@@ -257,9 +257,13 @@ def identify_cookie_attributes(cookie_filling)
       cookie_bits.drop(1).each do |bit|
         if bit.include? '='
           cookie_attribute, attribute_value = bit.split('=', 2)
-          cookie_attributes.store(cookie_attribute.strip, attribute_value.strip)
           if cookie_attribute.include? 'max-age'
             cookie_attributes.store('expires', Time.now + attribute_value.strip.to_i)
```
jeremyevans, line 261: This appears to change behavior for `max-age`, where `it only stores `expires` instead of storing `expires` and `max-age`.  Is that intentional?

lib/rack/mock.rb
```diff
@@ -257,9 +257,13 @@ def identify_cookie_attributes(cookie_filling)
...
+          elsif cookie_attribute.include? 'expires'
+            next if cookie_attributes.key? 'expires' # max-age attribute has precedence
+            cookie_attributes.store('expires', Time.httpdate(attribute_value.strip))
+          else
+            cookie_attributes.store(cookie_attribute.strip, attribute_value.strip)
           end
```
jeremyevans, line 267: I wonder if a better algorithm for this would be to avoid checking for `max-age` and `expires` inside the `cookie_bits.drop(1).each` loop.  Instead, after the loop, if `cookie_attributes['max-age']` is set, then set `cookie_attributes['expires']`.  If it isn't set and `cookie_attributes['expires']` is set, convert it to a `Time`.

jeremyevans (changes requested):
Apologies for not responding to this sooner. I agree this is a bug that should be fixed.  Looking at this and comparing it to #1759, I think we should add `require 'time'` to `lib/rack/mock.rb`.  Also, please see inline comments for `identify_cookie_attributes`.  I lean toward modifying the algorithm to avoid checking inside the loop.

lib/rack/mock.rb
walro: Nice catch, it was not intentional. I will do as you suggested in https://github.com/rack/rack/pull/1753/files#r742316074 so it will once again be stored then.

lib/rack/mock.rb
walro: Yes, that will definitely be easier to understand.

walro: Pushed new version, I kept the "style" of the old code for the sake of consistency. E.g. using `#store` over `#[]=`.

jeremyevans (approved):
This looks good to me.

PR #1754, opened 2021-06-04
Update Rubocop and apply autofixes

agrberg:


agrberg:
@jeremyevans let me know if I should just close this one as well. I opened it as I noticed that I couldn't run `rubocop` w/o the change and there were only a handful of auto corrections to add.

lib/rack/lint.rb
```diff
@@ -749,10 +749,10 @@ def each
         unless part.kind_of? String
           raise LintError, "Body yielded non-string value #{part.inspect}"
         end
-        ##
```
ioquatix, line 752: I think this is necessary for the RDoc generation.

ioquatix: (But I could be wrong, maybe can you check it?)

jeremyevans:
@ioquatix I checked the rake spec task and it uses `/^\s*## ?(.*)/`, so lines ending in `##` should still be copied. I didn't test to confirm, though.

Personally, I don't think we should be pedantic about whitespace.  Again, other committers may feel differently.  

I do think the change to the TargetRubyVersion looks wrong.  Rack targets Ruby 2.3+ currently.  If Rubocop doesn't support that, I support dropping Rubocop.  To be fair, even if Rubocop does support that, I support dropping Rubocop. :)

ioquatix (approved):

agrberg:
I view Rubocop (and linting) similar to static typing. A bit more work but beneficial as it removes an entire class of disagreements over style/errors. However, just as one can write types to be more permissive, linters are configurable to work for your team.

This isn't my show and I'm comfortable in the chorus 🎤 . Let me know if you'd like me to close this or if we should run the workflows and merge.

ioquatix:
Thanks but I feel this is not a big enough advantage for us to adopt it.

PR #1759, opened 2021-06-30
Parse 'expires' to Time in MockResponse cookies

### LINKED ISSUE #1758: Rack::MockRequest should be parsing 'expires' to a Time instance
I encountered the error

	NoMethodError: undefined method `gmtime' for "Thu, 29 Jun 2023 10:44:23 GMT":String

from the following expectation (which was failing):

	_(last_response).must_be :redirection?

Evidently this was due to the `CGI::Cookie#inspect` method being called.

[CGI::Cookie#inspect](https://github.com/ruby/cgi/blob/c42f91561548f29416c7b6d89020983b519903b1/lib/cgi/cookie.rb#L143) calls `#to_s`, which passes the value of 'expires' to [CGI.rfc1123_date](https://github.com/ruby/cgi/blob/c42f91561548f29416c7b6d89020983b519903b1/lib/cgi/util.rb#L193), which expects a Time object (or at least something that responds to `#gmtime`).  While `Rack::MockRequest#identify_cookie_attributes` sets 'expires' to a Time instance if 'max-age' is set, it doesn't parse 'expires' itself into a Time instance.

If I monkeypatch the code like so, it works as expected:

```ruby
module FixCookies
  private

    super.tap do |cookie_attributes|
      if cookie_attributes['expires'].is_a? String
        cookie_attributes['expires'] = Time.parse cookie_attributes['expires']
      end
    end
  end
end

Rack::MockResponse.prepend FixCookies
```

speckins:
CGI::Cookie expects the 'expires' attribute, if set, to be an instance
of Time.  While this was being correctly set for 'max-age', 'expires'
was set as-is, a String.  This caused errors when inspecting the
response, as CGI::Cookie#inspect calls #gmtime on the 'expires' value.

This commit fixes MockResponse to parse 'expires' as an RFC1123
formatted date/time when parsing cookies from the header.

Fixes #1758.

### DIFF
--- a/lib/rack/mock.rb
+++ b/lib/rack/mock.rb
@@ -4,6 +4,7 @@ require 'uri'
 require 'stringio'
 require_relative '../rack'
 require 'cgi/cookie'
+require 'time'  # for Time.httpdate
 
 module Rack
   # Rack::MockRequest helps testing your Rack application without
@@ -265,6 +266,10 @@ module Rack
         if bit.include? 'secure'
           cookie_attributes.store('secure', true)
         end
+      end
+      # the 'expires' attribute, if set, must be an instance of Time
+      if (expires = cookie_attributes['expires']).is_a? String
+        cookie_attributes.store('expires', Time.httpdate(expires))
       end
       cookie_attributes
     end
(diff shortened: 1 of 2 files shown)


walro:
Found the same problem and did something similar in #1753 :)

speckins:
Oops.  I searched issues before doing this but didn't think to search PRs.

jeremyevans:
I'm going to close this, as it looks like we'll go with the approach in #1753.

PR #1763, opened 2021-07-14
Remove an example that passes options on the first `#\` line [ci skip]

y-yagi:
This feature was deprecated since #1574.


jeremyevans:
Thanks for the patch!

ioquatix:
Awesome, thanks for catching this!

PR #1765, opened 2021-07-19
Don't ary.inspect in the lint assertions (backport)

ziggythehamster:
This is a backport to 2.2 of #1626 (fixing #1625)

### DIFF
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -48,10 +48,10 @@ module Rack
 
       ## and returns an Array of exactly three values:
       ary = @app.call(env)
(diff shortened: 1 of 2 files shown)


jeremyevans (changes requested):
In terms of code, this is fine.  However, it changes both files to be executable (100644 → 100755), which is definitely wrong.

Just to set expectations, I don't think we're planning on releasing 2.2.4 unless there is a security issue.  However, I see no problem backporting this so that if we do release 2.2.4, it is included.

ziggythehamster:
In case someone is curious, this made it to Rack 2.2.5 and newer (but the history doesn't seem to reflect that and I didn't add it to the changelog)

PR #1780, opened 2021-11-24
Fix MockResponse encoding

gudel98:
During my work connected with rack update, I've noticed that Rack::MockResponse#body returns a string with default ASCII-8BIT encoding. That behavior differs from rack version < 2 where implementation of MockResponse#body was just `super.join`
I decided to implement encoding accept for `body` method for testing encoding of real response.

### DIFF
--- a/lib/rack/mock.rb
+++ b/lib/rack/mock.rb
@@ -200,6 +200,10 @@ module Rack
     end
 
     def body
+      # By default String.new creates an object in ASCII-8BIT encoding
+      # We should specify the encoding according to Rack::Response#body encoding
+      encoding = super.join.encoding.to_s
+
       # FIXME: apparently users of MockResponse expect the return value of
       # MockResponse#body to be a string.  However, the real response object
       # returns the body as a list.
@@ -210,7 +214,7 @@ module Rack
       #     ...
       #     res.body.should == "foo!"
       #   end
-      buffer = String.new
+      buffer = String.new.force_encoding(encoding)
 
       super.each do |chunk|
         buffer << chunk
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch.  Unfortunately, I don't think this is worth doing.  Ruby will automatically change the encoding of 7-bit ASCII strings into other compatible encodings:

```ruby
x = "\xaf".force_encoding('Windows-1251')
b = 'b'.b
x.encoding
# => #<Encoding:Windows-1251>
b.encoding
# => #<Encoding:ASCII-8BIT>
b << x
b.encoding
# => #<Encoding:Windows-1251>
```

Can you come up with a failing test that involves non 7-bit ASCII strings and doesn't require checking the encoding directly?

gudel98:
@jeremyevans hm, weird. If u change the code od MockResponse with force_encoding and run the test-case I wrote, u'll see that it fails, cause ruby cannot change the encoding

jeremyevans:
What I'm saying is your test shouldn't matter in practice.  Can you come up with a failing test that involves non 7-bit ASCII strings and doesn't require checking the encoding directly?

lib/rack/mock.rb
```diff
@@ -200,6 +200,10 @@ def match(other)
     end
 
     def body
+      # By default String.new creates an object in ASCII-8BIT encoding
+      # We should specify the encoding according to Rack::Response#body encoding
+      encoding = super.join.encoding.to_s
```
ioquatix, line 205: This is too inefficient.

lib/rack/mock.rb
```diff
@@ -210,7 +214,7 @@ def body
       #     ...
       #     res.body.should == "foo!"
       #   end
-      buffer = String.new
+      buffer = String.new.force_encoding(encoding)
```
ioquatix, line 217: It's better to do `String.new(encoding: ...)`. But not sure about backwards compatibility.

ioquatix (changes requested):
I don't think we can accept this as is, I agree with @jeremyevans.

jeremyevans:
@ioquatix If you agree with me, then we should close this.  My point is that there is nothing to fix.  Encoding of 7-bit strings as binary doesn't matter, because they will auto-convert to other encodings if necessary.

PR #1781, opened 2021-11-25
Update trusted_proxy? to match on 127.0.0.0/8

snbloch:
RFC-3330 defines the entire 127.0.0.0/8 network as internal and as such I believe that the ip_filter regex should honor this instead of only 127.0.0.1 when determining remote IP address: https://datatracker.ietf.org/doc/html/rfc3330

Rails has already made this modification to the ActionDispatch RemoteIp TRUSTED_PROXIES default configuration as of v6.2: https://edgeapi.rubyonrails.org/classes/ActionDispatch/RemoteIp.html

This behavior was observed when introducing a rails application in to an Istio service mesh, where an envoy proxy is injected in front of the application by way of iptables and forwards inbounds requests from a "magic" IP address of 127.0.0.6 in order to break loopback routing loops: https://github.com/istio/istio/issues/29603

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -16,7 +16,7 @@ module Rack
       attr_accessor :ip_filter
     end
 
-    self.ip_filter = lambda { |ip| /\A127\.0\.0\.1\Z|\A(10|172\.(1[6-9]|2[0-9]|30|31)|192\.168)\.|\A::1\Z|\Afd[0-9a-f]{2}:.+|\Alocalhost\Z|\Aunix\Z|\Aunix:/i.match?(ip) }
+    self.ip_filter = lambda { |ip| /\A(127|10|172\.(1[6-9]|2[0-9]|30|31)|192\.168)\.|\A::1\Z|\Afd[0-9a-f]{2}:.+|\Alocalhost\Z|\Aunix\Z|\Aunix:/i.match?(ip) }
     ALLOWED_SCHEMES = %w(https http wss ws).freeze
     SCHEME_WHITELIST = ALLOWED_SCHEMES
     if Object.respond_to?(:deprecate_constant)
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -16,19 +16,24 @@ module Rack
       attr_accessor :ip_filter
     end
 
-    TRUSTED_PROXIES = /\A127\.(([1-9]?\d|[12]\d\d)\.){2}([1-9]?\d|[12]\d\d)\Z|\A::1\Z|\A[fF][cCdD][0-9a-fA-F]{2}(?:[:][0-9a-fA-F]{0,4}){0,7}\Z|\A10\.(([1-9]?\d|[12]\d\d)\.){2}([1-9]?\d|[12]\d\d)\Z|\A172\.(1[6-9]|2\d|3[01])(\.([1-9]?\d|[12]\d\d)){2}\Z|\A192\.168\.([1-9]?\d|[12]\d\d)\.([1-9]?\d|[12]\d\d)\Z|\Alocalhost\Z|\Aunix\Z|\Aunix:/i
+    TRUSTED_PROXIES = /\A127\.(([1-9]?\d|[12]\d\d)\.){2}([1-9]?\d|[12]\d\d)\Z
+                      |\A::1\Z
+                      |\A[fF][cCdD][0-9a-fA-F]{2}(?:[:][0-9a-fA-F]{0,4}){0,7}\Z
+                      |\A10\.(([1-9]?\d|[12]\d\d)\.){2}([1-9]?\d|[12]\d\d)\Z
+                      |\A172\.(1[6-9]|2\d|3[01])(\.([1-9]?\d|[12]\d\d)){2}\Z
+                      |\A192\.168\.([1-9]?\d|[12]\d\d)\.([1-9]?\d|[12]\d\d)\Z
+                      |\Alocalhost\Z|\Aunix\Z|\Aunix:/ix
     
     # For readability and comprehension, here is the preceding regex represented in parts:
-    # TRUSTED_PROXIES = 
-    #  [
-    #    /\A127\.(([1-9]?\d|[12]\d\d)\.){2}([1-9]?\d|[12]\d\d)\Z/, # localhost IPv4 range, per RFC-3330
-    #    /\A::1\Z/, # localhost IPv6
-    #    /\A[fF][cCdD][0-9a-fA-F]{2}(?:[:][0-9a-fA-F]{0,4}){0,7}\Z/, # private IPv6 range fc00::/7
-    #    /\A10\.(([1-9]?\d|[12]\d\d)\.){2}([1-9]?\d|[12]\d\d)\Z/, # private IPv4 range 10.x.x.x
-    #    /\A172\.(1[6-9]|2\d|3[01])(\.([1-9]?\d|[12]\d\d)){2}\Z/, # private IPv4 range 172.16.0.0 .. 172.31.255.255
-    #    /\A192\.168\.([1-9]?\d|[12]\d\d)\.([1-9]?\d|[12]\d\d)\Z/, # private IPv4 range 192.168.x.x
-    #    /\Alocalhost\Z|\Aunix\Z|\Aunix:/i
-    #  ]
+    #
+    #    \A127\.(([1-9]?\d|[12]\d\d)\.){2}([1-9]?\d|[12]\d\d)\Z     # localhost IPv4 range 127.x.x.x, per RFC-3330
+    #    \A::1\Z                                                    # localhost IPv6 ::1
+    #    \A[fF][cCdD][0-9a-fA-F]{2}(?:[:][0-9a-fA-F]{0,4}){0,7}\Z   # private IPv6 range fc00::/7
+    #    \A10\.(([1-9]?\d|[12]\d\d)\.){2}([1-9]?\d|[12]\d\d)\Z      # private IPv4 range 10.x.x.x
+    #    \A172\.(1[6-9]|2\d|3[01])(\.([1-9]?\d|[12]\d\d)){2}\Z      # private IPv4 range 172.16.0.0 .. 172.31.255.255
+    #    \A192\.168\.([1-9]?\d|[12]\d\d)\.([1-9]?\d|[12]\d\d)\Z     # private IPv4 range 192.168.x.x
+    #    \Alocalhost\Z|\Aunix\Z|\Aunix:                             # localhost hostname, and unix domain sockets
+    #
     
     self.ip_filter = lambda { |ip| TRUSTED_PROXIES.match?(ip) }
 
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -16,24 +16,21 @@ module Rack
       attr_accessor :ip_filter
     end
 
-    TRUSTED_PROXIES = 
-      [
-        "127.0.0.0/8", # localhost IPv4 range, per RFC-3330
-        "::1", # localhost IPv6
-        "fc00::/7", # private IPv6 range fc00::/7
-        "10.0.0.0/8", # private IPv4 range 10.x.x.x
-        "172.16.0.0/12", # private IPv4 range 172.16.0.0 .. 172.31.255.255
-        "192.168.0.0/16", # private IPv4 range 192.168.x.x
-      ].map { |proxy| IPAddr.new(proxy) }
+    TRUSTED_PROXIES = /\A127\.(([1-9]?\d|[12]\d\d)\.){2}([1-9]?\d|[12]\d\d)\Z|\A::1\Z|\A[fF][cCdD][0-9a-fA-F]{2}(?:[:][0-9a-fA-F]{0,4}){0,7}\Z|\A10\.(([1-9]?\d|[12]\d\d)\.){2}([1-9]?\d|[12]\d\d)\Z|\A172\.(1[6-9]|2\d|3[01])(\.([1-9]?\d|[12]\d\d)){2}\Z|\A192\.168\.([1-9]?\d|[12]\d\d)\.([1-9]?\d|[12]\d\d)\Z|\Alocalhost\Z|\Aunix\Z|\Aunix:/i
     
-    self.ip_filter = lambda do |ip| 
-      return true if TRUSTED_PROXIES.any? { |tp| tp.include? ip } 
-      false
-    rescue IPAddr::InvalidAddressError
-      # Not actually an IP address
-      return true if /\Alocalhost\Z|\Aunix\Z|\Aunix:/i.match?(ip)
-      false
-    end
+    # For readability and comprehension, here is the preceding regex represented in parts:
+    # TRUSTED_PROXIES = 
+    #  [
+    #    /\A127\.(([1-9]?\d|[12]\d\d)\.){2}([1-9]?\d|[12]\d\d)\Z/, # localhost IPv4 range, per RFC-3330
+    #    /\A::1\Z/, # localhost IPv6
+    #    /\A[fF][cCdD][0-9a-fA-F]{2}(?:[:][0-9a-fA-F]{0,4}){0,7}\Z/, # private IPv6 range fc00::/7
+    #    /\A10\.(([1-9]?\d|[12]\d\d)\.){2}([1-9]?\d|[12]\d\d)\Z/, # private IPv4 range 10.x.x.x
+    #    /\A172\.(1[6-9]|2\d|3[01])(\.([1-9]?\d|[12]\d\d)){2}\Z/, # private IPv4 range 172.16.0.0 .. 172.31.255.255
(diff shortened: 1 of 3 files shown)


ioquatix:
This looks okay but are there any security issues we should be concerned about?

snbloch:
I'm struggling to think of any.  With that said, I am happy for this to be further scrutinized by any member of the maintaining team or the community at large with an alternative perspective of how this could possibly introduce any security concerns.

ioquatix:
I also believe it's okay, but we should handle such changes with care since it could introduce security issue.

Looking at the regex, do we need to concern about use of `\Z`?

```
\z Matches the end of the string.

\Z Matches the end of the string unless the string ends with a "\n", in which case it matches just before the "\n".
```

snbloch:
I was trying to match the existing implementation of class-A networks defined in the regex, but to be honest a regex is not the correct tool for this job when CIDR ranges can be evaluated as IPAddr objects directly.

I've reworked this, defining a list of CIDR ranges, and then returning true if the proxy IP address that is being tested falls inside of any of them.

I am preserving the regex in a rescue block to evaluate for the special cases of `localhost`, `unix`, or a unix socket such as `unix:/tmp/sock` as defined before and included in the test suite.

ioquatix:
What is the overhead of this implementation? Faster or slower?

snbloch:
Thanks for looking.  Good point, the new implementation is slower particularly in the case where an IP address is _not_ a known proxy, which is likely going to be the vast majority of invocations in the real world.  I will take a look and see if I can optimize this some.

snbloch:
Back to using a single line regex for the performance, but this is now much more representative of proper IP ranges rather than just checking a first or second octet, which should address any security concerns.

100,000 iterations:

ruby old_impl.rb  0.11s user 0.07s system 94% cpu 0.186 total

ruby new_impl.rb  0.12s user 0.07s system 94% cpu 0.193 total

ioquatix:
This looks okay to me, but I believe we would benefit from a multi-line regex `/x` modifier IIRC, just to keep things understandable. WDYT?

snbloch:
Great idea, converted the regex to multi-line and cleaned up the comment block a bit.

ioquatix:
Great work, In believe you can embed the comments directly in the regexp for better readability:

https://ruby-doc.org/core-3.0.2/Regexp.html#class-Regexp-label-Free-Spacing+Mode+and+Comments

snbloch:
haha great, I was trying to get the inline comments to work earlier and it turns out that the `fc00/7` in the third line comment had the `/` character being interpreted as the regex terminator even though it was after the `#`

I just went ahead and removed that character for the sake of finishing this up easily.  Escaping it inside of a comment felt strange.

ioquatix:
Oh, I can see how that would be a problem.

I'm fine with what you've got here, but one other option which you might then prefer if comments were a pain is something like this:

``` ruby
Regexp.union(
  /\A127\.(([1-9]?\d|[12]\d\d)\.){2}([1-9]?\d|[12]\d\d)\Z/i,     # localhost IPv4 range 127.x.x.x, per RFC-3330
  /\A::1\Z/i,                                                    # localhost IPv6 ::1
  /\A[fF][cCdD][0-9a-fA-F]{2}(?:[:][0-9a-fA-F]{0,4}){0,7}\Z/i,   # private IPv6 range fc00 .. fdff
  /\A10\.(([1-9]?\d|[12]\d\d)\.){2}([1-9]?\d|[12]\d\d)\Z/i,      # private IPv4 range 10.x.x.x
  /\A172\.(1[6-9]|2\d|3[01])(\.([1-9]?\d|[12]\d\d)){2}\Z/i,      # private IPv4 range 172.16.0.0 .. 172.31.255.255
  /\A192\.168\.([1-9]?\d|[12]\d\d)\.([1-9]?\d|[12]\d\d)\Z/i,     # private IPv4 range 192.168.x.x
  /\Alocalhost\Z|\Aunix\Z|\Aunix:/i,                             # localhost hostname, and unix domain sockets
)
```

OI just eyeballed this but I think it should be equivalent. It might be nicer since the comment is not part of the regexp, and you can comment them out individually a bit more easily/explicitly?

Up to you.

snbloch:
Thanks so much for all of your advice on this.  I _really_ like the way the Regexp.union ended up, it even allows the IDE to syntax highlight the individual regexps, which got lost with the multi-line formatting.

I think we're ready with this one, thanks again!

ioquatix (approved):

lib/rack/request.rb
```diff
@@ -16,7 +16,18 @@ class << self
       attr_accessor :ip_filter
     end
 
-    self.ip_filter = lambda { |ip| /\A127\.0\.0\.1\Z|\A(10|172\.(1[6-9]|2[0-9]|30|31)|192\.168)\.|\A::1\Z|\Afd[0-9a-f]{2}:.+|\Alocalhost\Z|\Aunix\Z|\Aunix:/i.match?(ip) }
+    TRUSTED_PROXIES = Regexp.union(
+      /\A127\.(([1-9]?\d|[12]\d\d)\.){2}([1-9]?\d|[12]\d\d)\Z/i,     # localhost IPv4 range 127.x.x.x, per RFC-3330
+      /\A::1\Z/i,                                                    # localhost IPv6 ::1
+      /\A[fF][cCdD][0-9a-fA-F]{2}(?:[:][0-9a-fA-F]{0,4}){0,7}\Z/i,   # private IPv6 range fc00 .. fdff
```
jeremyevans, line 22: You don't need `[fF]` and similar in a case insensitive regexp.  Stick to just the lowercase or uppercase character.

lib/rack/request.rb
```diff
@@ -16,7 +16,18 @@ class << self
       attr_accessor :ip_filter
     end
 
-    self.ip_filter = lambda { |ip| /\A127\.0\.0\.1\Z|\A(10|172\.(1[6-9]|2[0-9]|30|31)|192\.168)\.|\A::1\Z|\Afd[0-9a-f]{2}:.+|\Alocalhost\Z|\Aunix\Z|\Aunix:/i.match?(ip) }
+    TRUSTED_PROXIES = Regexp.union(
+      /\A127\.(([1-9]?\d|[12]\d\d)\.){2}([1-9]?\d|[12]\d\d)\Z/i,     # localhost IPv4 range 127.x.x.x, per RFC-3330
```
jeremyevans, line 20: Can we explain the reasoning for using `([1-9]?\d|[12]\d\d)` as a valid IPv4 segment?  Is it just to prevent preceding `0`?  That you can handle more simply with `[1-9]\d\d`.  If it is only allow valid IPv4 segments, then it isn't strict enough, since `256` will match, and you'll need a more complicated regexp, such as `([1-9]?\d|1\d\d)|2[0-4]\d|25[0-5]`.

Whatever regexp is used for IPv4 segments, it should be assigned to a local variable, and interpolated into each regexp used in the union to reduce duplication.

Additionally, IPv4 regexps don't need the case insensitive regexp flag, since they don't have any alphabetic characters, only numeric characters.

jeremyevans (changes requested):
I'm OK with this behavior change, though I do have a few requested changes to the implementation (see inline comments).

I think CHANGELOG needs to be updated to mention this as a breaking change.  I don't expect it will affect many people, but it is a change of behavior with potential security implications for anyone it does affect.

lib/rack/request.rb
snbloch: Thank you @jeremyevans this is all very valid feedback.

I think I have addressed all of your concerns, and I have added more negative test cases to confirm that the IPv4 octet regexp is properly validating.  Please give it another look at your convenience.

lib/rack/request.rb
```diff
@@ -16,7 +16,20 @@ class << self
       attr_accessor :ip_filter
     end
 
-    self.ip_filter = lambda { |ip| /\A127\.0\.0\.1\Z|\A(10|172\.(1[6-9]|2[0-9]|30|31)|192\.168)\.|\A::1\Z|\Afd[0-9a-f]{2}:.+|\Alocalhost\Z|\Aunix\Z|\Aunix:/i.match?(ip) }
+    VALID_IPV4_OCTET = '(25[0-5]|2[0-4][0-9]|[01]?[0-9]?[0-9])'.freeze
```
jeremyevans, line 19: No reason to make this a constant.  As I mentioned earlier, a local variable is better.  We want to avoid introducing new public comments for implementation details.

lib/rack/request.rb
```diff
@@ -16,7 +16,20 @@ class << self
...
+    
+    TRUSTED_PROXIES = Regexp.union(
+      /\A127(\.(#{VALID_IPV4_OCTET})){3}$\Z/,                         # localhost IPv4 range 127.x.x.x, per RFC-3330
+      /\A::1$\Z/,                                                      # localhost IPv6 ::1
+      /\A[f][cd][0-9a-f]{2}(?:[:][0-9a-f]{0,4}){0,7}$\Z/i,             # private IPv6 range fc00 .. fdff
```
jeremyevans, line 24: `[f]` -> `f`, `[:]` -> `:`

lib/rack/request.rb
```diff
@@ -16,7 +16,20 @@ class << self
       attr_accessor :ip_filter
     end
 
-    self.ip_filter = lambda { |ip| /\A127\.0\.0\.1\Z|\A(10|172\.(1[6-9]|2[0-9]|30|31)|192\.168)\.|\A::1\Z|\Afd[0-9a-f]{2}:.+|\Alocalhost\Z|\Aunix\Z|\Aunix:/i.match?(ip) }
+    VALID_IPV4_OCTET = '(25[0-5]|2[0-4][0-9]|[01]?[0-9]?[0-9])'.freeze
+    
+    TRUSTED_PROXIES = Regexp.union(
+      /\A127(\.(#{VALID_IPV4_OCTET})){3}$\Z/,                         # localhost IPv4 range 127.x.x.x, per RFC-3330
```
jeremyevans, line 22: Is there a reason for `$\Z` instead of just `\Z`? If not, these can be changed to just `\Z`.

jeremyevans (changes requested):
This looks better.  I still have a few minor changes requested to the implementation, please see inline comments.

lib/rack/request.rb
snbloch: Without the $ it is allowing the regexp to match when there is a trailing character, for example `127.0.0.300` matches without the $ because valid IP address `127.0.0.30` is inside of it.

lib/rack/request.rb
snbloch: Changed to local variable `valid_ipv4_octet`.

lib/rack/request.rb
snbloch: Fixed.

lib/rack/request.rb
jeremyevans: This sounded weird to me, so I tested it:

```ruby
VALID_IPV4_OCTET = '(25[0-5]|2[0-4][0-9]|[01]?[0-9]?[0-9])'
re = /\A127(\.(#{VALID_IPV4_OCTET})){3}$\Z/
re2 = /\A127(\.(#{VALID_IPV4_OCTET})){3}\Z/
re.match('127.0.0.300') # => nil
re2.match('127.0.0.300') # => nil 
```

The Ruby documentation for `\Z` states: `\Z - Matches end of string. If string ends with a newline, it matches just before newline`.  `0` is not a newline, so `\Z` should not match before the final `0`.

Note that in general, you should use `\z` (end of string), and not `\Z`.  `\Z` should only be used if you want to allow a trailing newline, but have the trailing newline not be matched.  So if you want to match against end of string, you should switch everything to `\z`.  My initial assumption from reading the code is you wanted to handle trailing newlines specially, since both `$` and `\Z` are related to newlines.  But from this discussion, it seems that was not the reason, and you should probably switch to `\z`.

snbloch: Converted all instances of `\Z` to `\z` and all tests still pass.

jeremyevans: Just `\z`, not `$\z`.  `$` before `\z` has no effect.

snbloch: Ah, I see, thank you.  I was using an external regexp tester to develop the patterns, and there must be a difference here between ruby and some other flavor like PCRE.

Removed the `$` and all tests still pass using only `\z`

lib/rack/request.rb
```diff
@@ -16,7 +16,20 @@ class << self
       attr_accessor :ip_filter
     end
 
-    self.ip_filter = lambda { |ip| /\A127\.0\.0\.1\Z|\A(10|172\.(1[6-9]|2[0-9]|30|31)|192\.168)\.|\A::1\Z|\Afd[0-9a-f]{2}:.+|\Alocalhost\Z|\Aunix\Z|\Aunix:/i.match?(ip) }
+    valid_ipv4_octet = '(25[0-5]|2[0-4][0-9]|[01]?[0-9]?[0-9])'.freeze
+    
+    TRUSTED_PROXIES = Regexp.union(
```
jeremyevans, line 21: `TRUSTED_PROXIES` is also better as a local variable and not a constant.

lib/rack/request.rb
```diff
@@ -16,7 +16,20 @@ class << self
       attr_accessor :ip_filter
     end
 
-    self.ip_filter = lambda { |ip| /\A127\.0\.0\.1\Z|\A(10|172\.(1[6-9]|2[0-9]|30|31)|192\.168)\.|\A::1\Z|\Afd[0-9a-f]{2}:.+|\Alocalhost\Z|\Aunix\Z|\Aunix:/i.match?(ip) }
+    valid_ipv4_octet = '(25[0-5]|2[0-4][0-9]|[01]?[0-9]?[0-9])'.freeze
```
jeremyevans, line 19: Looking at everywhere that `valid_ipv4_octet` is used, it is preceded by a `.`.  So it seems better to do:

```ruby
# no need to freeze this local variable
# Also makes sense to use a regexp instead of a string
valid_ipv4_octet = /\.(25[0-5]|2[0-4][0-9]|[01]?[0-9]?[0-9])/
```

That will allow you to simplify usage to:

```ruby
/\A127#{valid_ipv4_octet}{3}\z/, 
```

With similar changes to the other IPv4 regexps.

lib/rack/request.rb
```diff
@@ -16,7 +16,20 @@ class << self
...
+    
+    TRUSTED_PROXIES = Regexp.union(
+      /\A127(\.(#{valid_ipv4_octet})){3}\z/,                         # localhost IPv4 range 127.x.x.x, per RFC-3330
+      /\A::1\z/,                                                      # localhost IPv6 ::1
+      /\Af[cd][0-9a-f]{2}(?::[0-9a-f]{0,4}){0,7}\z/i,             # private IPv6 range fc00 .. fdff
+      /\A10(\.(#{valid_ipv4_octet})){3}\z/,                          # private IPv4 range 10.x.x.x
+      /\A172\.(1[6-9]|2[0-9]|3[01])(\.(#{valid_ipv4_octet})){2}\z/,     # private IPv4 range 172.16.0.0 .. 172.31.255.255
+      /\A192\.168(\.(#{valid_ipv4_octet})){2}\z/,                     # private IPv4 range 192.168.x.x
+      /\Alocalhost\z|\Aunix\z|\Aunix:/i,                              # localhost hostname, and unix domain sockets
```
jeremyevans, line 28: `/\Alocalhost\z|\Aunix\z|\Aunix:/i,` => `/\Alocalhost\z|\Aunix(\z|:)/i,`

jeremyevans (changes requested):
I noticed a few more minor issues while reviewing, please see the inline comments.

lib/rack/request.rb
snbloch: Oh, nice catch, thank you.  That makes everything much more readable.

lib/rack/request.rb
```diff
@@ -16,7 +16,20 @@ class << self
...
+    
+    trusted_proxies = Regexp.union(
+      /\A127#{valid_ipv4_octet}{3}\z/,                         # localhost IPv4 range 127.x.x.x, per RFC-3330
+      /\A::1\z/,                                                      # localhost IPv6 ::1
+      /\Af[cd][0-9a-f]{2}(?::[0-9a-f]{0,4}){0,7}\z/i,             # private IPv6 range fc00 .. fdff
+      /\A10#{valid_ipv4_octet}{3}\z/,                          # private IPv4 range 10.x.x.x
+      /\A172\.(1[6-9]|2[0-9]|3[01])#{valid_ipv4_octet}{2}\z/,     # private IPv4 range 172.16.0.0 .. 172.31.255.255
+      /\A192\.168#{valid_ipv4_octet}{2}\z/,                     # private IPv4 range 192.168.x.x
+      /\Alocalhost\z|\Aunix(\z|:)/i,                              # localhost hostname, and unix domain sockets
+    ).freeze
```
jeremyevans, line 29: No need to freeze regexps, especially since this is a local variable and not a constant, but doing so causes no problems.

jeremyevans (approved):
I think this is fine to merge.  Lining up the comments would be nice, but it's not a blocker.

lib/rack/request.rb
snbloch: Removed unnecessary freeze

snbloch:
Comments got mis-aligned with the earlier interpolation simplification, they're fixed up now.

PR #1790, opened 2022-01-24
Drop support for Ruby 2.3

jeremyevans:
We don't test Ruby 2.3 in CI, and tests are already broken because we use Array#sum.

Dropping Ruby 2.3 also us to drop the Regexp core_ext we were using, as well.

While here, fix the specs to run correctly the version of psych that ships with Ruby.  We should always make sure rack will work correctly with the default gems shipped with each supported Ruby version.

Start testing Ruby 3.1 on CI.


rafaelfranca (approved):

test/spec_mock.rb
```diff
@@ -3,6 +3,12 @@
 require_relative 'helper'
 require 'yaml'
 
+unless YAML.respond_to?(:unsafe_load)
```
ioquatix, line 6: Can you add a note to explain why this is added and when it can be removed?

ioquatix (changes requested):
LGTM. Just think we should document the monkey patch a bit.

Gemfile
```diff
@@ -11,8 +11,6 @@ group :maintenance, optional: true do
   gem "rubocop-packaging", require: false
 end
 
-gem "psych", "~> 4.0"
```
ioquatix, line 14: Do you mind explaining why we should remove this? Is it breaking something? Current CI seems all green to me.

jeremyevans: IMO, it is unreasonable to expect people to upgrade stdlib gems.  Tests should pass on any supported Ruby installation with the stdlib gems of that Ruby version (specifying non stdlib gems, such as webrick in Ruby 3+, is fine).  To make sure that tests still pass with the stdlib gems, it should not be specified in the Gemfile used for CI.  The tests were broken in Ruby 2.4-2.7 with stdlib.

test/spec_mock.rb
jeremyevans: Done.

Gemfile
ioquatix: Fair enough. I don't fully agree with this, but it's fine. The problem here is the stdlib interface was basically broken in a way that requires a monkey patch for compatibility if we need to support old versions.

ioquatix (approved):
Thanks for this!

PR #1791, opened 2022-01-25
Remove `Rack::MockResponse#body` and introduce `Rack::MockResponse#join` as a replacement.

ioquatix:
The subclass interface is not compatible with the base class (Liskov substitution principle). There is a comment which basically cries for help. This retains backwards compatibility with a warning, and provides an alternative convenient interface for joining the body (which is not deprecated).

I propose for Rack 3.1 or 4.0 we remove the keyword argument.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,10 @@ All notable changes to this project will be documented in this file. For info on
 - `Rack::Utils.secure_compare` uses OpenSSL's faster implementation if available. ([#1711](https://github.com/rack/rack/pull/1711), [@bdewater](https://github.com/bdewater))
 - BREAKING CHANGE: Updated `trusted_proxy?` to match full 127.0.0.0/8 network. ([#1781](https://github.com/rack/rack/pull/1781), [@snbloch](https://github.com/snbloch))
 
+### Removed
+
+- `Rack::MockResponse#body` violates Liskov substitution principle. Use `Rack::MockResponse#join` for the current behaviour. ([#1791][https://github.com/rack/rack/pull/1791], [@ioquatix][https://github.com/ioquatix])
+
 ### Fixed
 
 - Make Rack::MockResponse handle non-hash headers. ([#1629](https://github.com/rack/rack/issues/1629), [@jeremyevans](https://github.com/jeremyevans))
--- a/lib/rack/mock.rb
+++ b/lib/rack/mock.rb
@@ -199,24 +199,23 @@ module Rack
       body.match other
     end
 
-    def body
-      # FIXME: apparently users of MockResponse expect the return value of
-      # MockResponse#body to be a string.  However, the real response object
-      # returns the body as a list.
-      #
-      # See spec_showstatus.rb:
-      #
-      #   should "not replace existing messages" do
-      #     ...
-      #     res.body.should == "foo!"
-      #   end
-      buffer = String.new
+    def body(join: true)
+      if join
+        buffer = String.new
 
-      super.each do |chunk|
-        buffer << chunk
+        super().each do |chunk|
+          buffer << chunk
+        end
+
+        return buffer
+      else
+        super
       end
+    end
 
-      return buffer
+    # @returns [String] The response body parts joined into a single string.
+    def join
+      body(join: true)
     end
 
     def empty?
--- a/lib/rack/mock.rb
+++ b/lib/rack/mock.rb
@@ -192,19 +192,29 @@ module Rack
       buffered_body!
     end
 
-    # @returns [String] The response body parts joined into a single string.
-    def join
-      unless @joined
-        @joined = String.new
+    def body(join = true)
+      body = super()
+
+      # This is considered deprecated behaviour, prefer to use #join.
+      if join && body.respond_to?(:each)
+        # warn 'Prefer MockResponse#join for response body as string!' if deprecated
+
+        buffer = String.new
 
         body.each do |chunk|
-          @joined << chunk
+          buffer << chunk
         end
 
-        @joined.freeze
+        return buffer
+      else
+        # Everything else:
+        return body
       end
+    end
 
-      return @joined
+    # @returns [String] The response body parts joined into a single string.
+    def join
+      @joined ||= self.body(true)
     end
 
     def =~(other)
--- a/lib/rack/mock.rb
+++ b/lib/rack/mock.rb
@@ -182,8 +182,9 @@ module Rack
     attr_accessor :errors
 
     def initialize(status, headers, body, errors = StringIO.new(""))
+      @joined = false
       @original_headers = headers
-      @errors           = errors.string if errors.respond_to?(:string)
+      @errors = errors.string if errors.respond_to?(:string)
 
       super(body, status, headers)
 
@@ -191,16 +192,10 @@ module Rack
       buffered_body!
     end
 
-    def =~(other)
-      body =~ other
-    end
-
-    def match(other)
-      body.match other
-    end
-
-    def body(join: true)
+    def body(join = (deprecated = true))
       if join
+        warn 'Prefer MockResponse#join for response body as string!' if deprecated
+
         buffer = String.new
 
         super().each do |chunk|
@@ -215,7 +210,15 @@ module Rack
 
     # @returns [String] The response body parts joined into a single string.
     def join
-      body(join: true)
+      @joined ||= self.body(true)
+    end
+
+    def =~(other)
+      join =~ other
+    end
+
+    def match(other)
+      join.match other
     end
 
     def empty?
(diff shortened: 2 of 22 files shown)


ioquatix:
After thinking about this, I believe we are better to make a clean break. Users can monkey patch Rack 2.x:

```
unless Rack::MockResponse.instance_methods.include?(:join)
  class Rack::MockResponse
    alias join body
  end
end
```

and/or we could backport this interface to 2.x

jeremyevans (commented):
I don't think it's worth breaking backwards compatibility for this. As the removed comment indicates, it appears users expect the existing behavior. However, as I'm not a direct user, I won't object.

ioquatix:
I agree with your assessment. The failing external tests show the extent of the issue, but fortunately this only impacts tests and I cannot imagine any reasonable situation where this impacts non-test systems.

Therefore, I feel like this is the right thing to do, especially because we now define `body` to have multiple shapes `#each` or `#call`. At some point we are going to have to face the pain of this change since it was designed incorrectly in the first place. Even with my proposed argument, we break Liskov substitution principle, so I believe we should just go all in on this and fix it once and for all, and provide a monkey patch for backwards compatibility is good enough.

jeremyevans:
> I agree with your assessment. The failing external tests show the extent of the issue, but fortunately this only impacts tests and I cannot imagine any reasonable situation where this impacts non-test systems.

I think you are correct in regards to this affecting only tests.  What you are missing is that, to my knowledge, MockResponse is only used for testing, so you are breaking the most important use case.  When testing, you generally want something like `response.body.must_equal 'foo'`.  In almost all cases when testing, you don't care if the original body was `['foo']`, `['f', 'o', 'o']`, or `['', 'foo', '']`.

> Therefore, I feel like this is the right thing to do, especially because we now define `body` to have multiple shapes `#each` or `#call`. At some point we are going to have to face the pain of this change since it was designed incorrectly in the first place. Even with my proposed argument, we break Liskov substitution principle, so I believe we should just go all in on this and fix it once and for all, and provide a monkey patch for backwards compatibility is good enough.

I think `MockResponse#body`, if used with a callable body, should attempt mimic the behavior it currently has, returning a single string.  Alternatively, not as nicely, you could return callable bodies as-is, keeping the current behavior for other bodies.  I think it's a poor idea to break the expected behavior of `MockResponse#body` simply because callable bodies were introduced.

Let's study history in this case. `MockResponse` was added in 2007, specifically for easier testing (0fac8599660f45d50c4ffeac780804879fcf549d), with body returning a string, since that makes by far the most sense when testing. It was not a subclass of `Response` until 2011 (812ac75b327c9f3d6ff5074def9721bf6f19e1de), where it kept the original behavior of body returning a string, because that is what all users expect.

Can you please explain why you think users would prefer original bodies and not string bodies when testing? What changed in testing between then and now?  If you are going to break backwards compatibility, it should be for the benefit of the users, not because of a dogmatic adherence to the Liskov substitution principle (which only happened due to a refactoring, not by design).

ioquatix:
I don't think users should care whether they have a `MockResponse` or a real `Response` in their tests but right now this is impossible because of the interface. I agree there is a convenience factor but that argument breaks down with more complex responses.

The use case with `rack.hijack` and streaming bodies will get more tricky going forward. You can't turn SSE/WebSockets into a string. So at best we could return it if it doesn't `respond_to?(:each)`.

> Can you please explain why you think users would prefer original bodies and not string bodies when testing?

Actually I agree with you - most users DO just want to make some assertion against the result... but this begs the question why `MockResponse` exists at all. Why can't we just have `Rack::MockRequest` return instances of `Rack::Response`? The answer to this seems to be all the extra "assertion helpers" but honestly it would make way more sense for these to be within respective test libraries like `rack-test` etc.

I personally see this as an inconsistency in the interface and I'd like to think we can fix it, achieving simpler code and semantics.

Ignoring the design choices that lead to `MockResponse`, we can retain backwards compatibility by skipping 2bd6bc2 and we have a forward path, i.e.

1. Introduce alternative interface and retain backwards compatibility (`#join`)
2. Tests move to alternative interface.
3. (optionally) remove deprecated interface (`body(join: true)`).

Longer term, I'd prefer if we could drop `MockResponse` entirely. I don't see why it should exist.

jeremyevans:
> I don't think users should care whether they have a `MockResponse` or a real `Response` in their tests but right now this is impossible because of the interface. I agree there is a convenience factor but that argument breaks down with more complex responses.

Almost all responses are not "more complex responses".  Currently, MockResponse doesn't handle hijacked responses at all, it just doesn't break because they respond to `each`.

Please do not underestimate the convenience factor of returning string bodies. I see string bodies for MockResponse as a huge convenience compared to enumerable bodies.

> The use case with `rack.hijack` and streaming bodies will get more tricky going forward. You can't turn SSE/WebSockets into a string. So at best we could return it if it doesn't `respond_to?(:each)`.

That seems reasonable to me.

> > Can you please explain why you think users would prefer original bodies and not string bodies when testing?
> 
> Actually I agree with you - most users DO just want to make some assertion against the result... but this begs the question why `MockResponse` exists at all. Why can't we just have `Rack::MockRequest` return instances of `Rack::Response`?

Specifically because having body return an enumerable instead of a string is almost never what you want when testing.

> The answer to this seems to be all the extra "assertion helpers" but honestly it would make way more sense for these to be within respective test libraries like `rack-test` etc.

Well, having the code in rack allows the various test libraries to share the mock implementation.  This is basically the same reason for all of rack's code.  Pretty much all test libraries are going to want to work with string bodies, and it seems unnecessary to have them all recreate the wheel, especially a wheel that already exists in rack and that they are currently relying on.

> I personally see this as an inconsistency in the interface and I'd like to think we can fix it, achieving simpler code and semantics.

I think we should be pragmatic instead of dogmatic.  There is a very good reason for the current inconsistency, and I think making consistency a goal in this case is foolish.

> Ignoring the design choices that lead to `MockResponse`, we can retain backwards compatibility by skipping [2bd6bc2](https://github.com/rack/rack/commit/2bd6bc209b41a375325664db7634affb3ab3d4f5) and we have a forward path, i.e.
> 
> 1. Introduce alternative interface and retain backwards compatibility (`#join`)
> 2. Tests move to alternative interface.
> 3. (optionally) remove deprecated interface (`body(join: true)`).

I think we should keep things as-is.  However, I've said my piece and don't have more to add. Hopefully @rafaelfranca or @tenderlove can chime in.

ioquatix:
I think this PR is pragmatic, with the discussed changes all the external tests as passing which means we have retained compatibility while providing a forward path.

tenderlove:
I've run in to the problem that this PR fixes: namely I *thought* I was getting the enumerable thing from `body` but it was a `MockResponse` instance and did something different (I wanted to assert the body had deferred chunks in it).

TBH I find it really odd that `MockRequest` [is constructed with an app](https://github.com/rack/rack/blob/d15dd728440710cfc35ed155d66a98dc2c07ae42/lib/rack/mock.rb#L52-L54) and calls the app.  I would expect `MockRequest` to know nothing about about app instance and instead just be a helper object for constructing the right `env` to pass to `call`.  IOW something like this:

```ruby
mr = MockRequest.new
mr.get("/whatever")
status, headers, body = app.call mr.to_env
response = MockResponse.new(status, headers, body)
```

Then leave it up to the caller to construct any kind of wrapper around the response they want.  IOW neither `MockRequest` nor `MockResponse` really make any sense to me.  We should have a `RequestBuilder` (a nice way of constructing the env) and maybe some kind of convenience wrapper for the response triple.  IMO it's a mistake for either of these objects to know about the app instance.

Can we consider leaving these classes as-is (essentially abandoning / deprecating) and create something better?

tenderlove:
(btw, I think we could consider changing the existing classes, but I really don't like the `join` method name)

ioquatix:
Rather than building this here, why don't we create `rack-mock` as a separate gem. The namespace should be free for us to use and then we have a clear migration path.

jeremyevans:
I'd prefer keeping things as-is for Rack::MockResponse (and MockRequest).  Adding a rack-mock gem for a new approach also seems fine to me.  Maybe we can use Rack::Mock::{Request,Response} in such a gem so there is no namespace collision.

ioquatix:
I agree with @tenderlove I think the interface here is cumbersome, how about we aim to deprecate it in Rack 3 and remove it in Rack 4? Provided we have an alternative `rack-mock` gem.

tenderlove:
> I agree with @tenderlove I think the interface here is cumbersome, how about we aim to deprecate it in Rack 3 and remove it in Rack 4? Provided we have an alternative `rack-mock` gem.

Ya, I think that's fine.  We can even pull the existing code in to a gem that folks can install if they depend on it but can't / won't upgarde.

ioquatix:
Awesome we have a path forward, thanks everyone, I'll close this.

PR #1793, opened 2022-01-25
Allow header value to be an Array of String instances.

ioquatix:
Addresses https://github.com/rack/rack/issues/1598

### DIFF
--- a/lib/rack/handler/webrick.rb
+++ b/lib/rack/handler/webrick.rb
@@ -95,15 +95,15 @@ module Rack
         begin
           res.status = status.to_i
           io_lambda = nil
-          headers.each { |k, vs|
-            if k == RACK_HIJACK
-              io_lambda = vs
-            elsif k == "set-cookie"
-              res.cookies.concat vs.split("\n")
+          headers.each { |key, value|
+            if key == RACK_HIJACK
+              io_lambda = value
+            elsif key == "set-cookie"
+              res.cookies.concat(Array(value))
             else
               # Since WEBrick won't accept repeated headers,
               # merge the values per RFC 1945 section 4.2.
-              res[k] = vs.split("\n").join(", ")
+              res[key] = Array(value).join(", ")
             end
           }
 
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -655,7 +655,7 @@ module Rack
           raise LintError, "headers object should not be frozen, but is"
         end
 
-        headers.each { |key, value|
+        headers.each do |key, value|
           ## The header keys must be Strings.
           unless key.kind_of? String
             raise LintError, "header key must be a string, was #{key.class}"
@@ -663,7 +663,7 @@ module Rack
 
           ## Special headers starting "rack." are for communicating with the
           ## server, and must not be sent back to the client.
-          next if key =~ /^rack\..+$/
+          next if key.start_with?("rack.")
 
           ## The header must not contain a +Status+ key.
           raise LintError, "header must not contain status" if key == "status"
@@ -673,19 +673,23 @@ module Rack
           ## Header keys must not contain uppercase ASCII characters (A-Z).
           raise LintError, "uppercase character in header name: #{key}" if key =~ /[A-Z]/
 
-          ## The values of the header must be Strings,
-          unless value.kind_of? String
-            raise LintError, "a header value must be a String, but the value of '#{key}' is a #{value.class}"
+          ## Header values must be either a String instance,
+          if value.kind_of?(String)
+            check_header_value(key, value)
+          elsif value.kind_of?(Array)
+            ## or an Array of String instances,
+            value.each{|value| check_header_value(key, value)}
+          else
+            raise LintError, "a header value must be a String or Array of Strings, but the value of '#{key}' is a #{value.class}"
           end
-          ## consisting of lines (for multiple header values, e.g. multiple
-          ## <tt>Set-Cookie</tt> values) separated by "\\n".
-          value.split("\n").each { |item|
-            ## The lines must not contain characters below 037.
-            if item =~ /[\000-\037]/
-              raise LintError, "invalid header value #{key}: #{item.inspect}"
-            end
-          }
-        }
+        end
+      end
+
+      def check_header_value(key, value)
+        ## such that each String instance must not contain characters below 037.
+        if value =~ /[\000-\037]/
+          raise LintError, "invalid header value #{key}: #{value.inspect}"
+        end
       end
 
       ##
--- a/lib/rack/response.rb
+++ b/lib/rack/response.rb
@@ -143,19 +143,19 @@ module Rack
 
     def has_header?(key)
       raise ArgumentError unless key.is_a?(String)
-      headers.key? key
+      @headers.key?(key)
     end
     def get_header(key)
       raise ArgumentError unless key.is_a?(String)
-      headers[key]
+      @headers[key]
     end
-    def set_header(key, v)
+    def set_header(key, value)
       raise ArgumentError unless key.is_a?(String)
-      headers[key] = v
+      @headers[key] = value
     end
     def delete_header(key)
       raise ArgumentError unless key.is_a?(String)
-      headers.delete key
+      @headers.delete key
     end
 
     alias :[] :get_header
@@ -186,7 +186,7 @@ module Rack
       def redirect?;            [301, 302, 303, 307, 308].include? status; end
 
       def include?(header)
-        has_header? header
+        has_header?(header)
       end
 
       # Add a header that may have multiple values.
@@ -198,15 +198,23 @@ module Rack
       #   assert_equal 'Accept-Encoding,Cookie', response.get_header('Vary')
       #
       # http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2
-      def add_header(key, v)
+      def add_header(key, value)
         raise ArgumentError unless key.is_a?(String)
 
-        if v.nil?
-          get_header key
-        elsif has_header? key
-          set_header key, "#{get_header key},#{v}"
+        if value.nil?
+          return get_header(key)
+        end
+
+        value = value.to_s
+
+        if header = get_header(key)
+          if header.is_a?(Array)
+            header << value
+          else
+            set_header(key, [header, value])
+          end
         else
-          set_header key, v
+          set_header(key, value)
         end
       end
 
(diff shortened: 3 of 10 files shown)


ioquatix:
@jeremyevans @tenderlove Looking at the [RFC](https://datatracker.ietf.org/doc/html/rfc6265), there should be no need to delete an existing cookie by scanning the list of `set-cookie` headers before adding one with `value=` and `expires` in the past. The browser will simply add and then remove a cookie.

I believe that there would be very few cases where this is actually useful behaviour. i.e. someone would have to add and then delete a cookie with the same details to hit this case.

Bearing in mind, this means that every time someone deletes the cookie, it's O(N) scan, and even worse on the current implementation which has to unpack, parse and then recombine all the cookies.

I'd personally like to simplify this in this PR.

If someone deletes a cookie, we simply append the appropriate deletion (`expires` in past, `value=`) without trying to unpack or parse existing cookies.

The only disadvantage I can see with this is if someone sets, say, a 1MiB cookie and then deletes it, in my proposed implementation it would send still 1MiB of cookie data.

I think the simplicity (and probably performance) win is big enough to justify the difference. That being said, if there are buggy behaviours that don't process `set-cookie` in order, it could be an issue... but then it's a bug in the browser.

ioquatix:
By the way, I've also removed a bunch of `Utils` methods that don't really make sense and were very specific to the newline encoded format. We can try to add compatibility shims, not sure if it's necessary?

lib/rack/utils.rb
```diff
@@ -248,71 +248,59 @@ def add_cookie_to_header(header, key, value)
...
+                   if path
+                     /\A#{key}=.*(?:domain=#{domain}(?:;|$).*path=#{path}(?:;|$)|path=#{path}(?:;|$).*domain=#{domain}(?:;|$))/
+                   else
+                     /\A#{key}=.*domain=#{domain}(?:;|$)/
+                   end
+                 elsif path
+                   /\A#{key}=.*path=#{path}(?:;|$)/
                  else
-                   /\A#{key}=.*domain=#{domain}(?:;|$)/
+                   /\A#{key}=/
```
ioquatix, line 293: This could all go if we don't care about deleting cookies from the `set-header` header. Unless there is some RFC that specifically calls out this behaviour, I don't think we should do it like this.

ioquatix:
Doing some digging, found that it looks like we did previously support header value being an Array: https://github.com/rack/rack/blob/f23e8330085e6967b9e57a26b293bd08fa4b0d53/lib/rack/response.rb#L84-L96

ioquatix:
@leahneukirchen I see your first implementation here: https://github.com/rack/rack/commit/7ed819ad716661a67ada21d04e7815e493ccdb9d#diff-cfcc6d2645b6f77c5ec05acd448f280ce3004d3dba85b8aef3b3c81faf6d1373R53-R55

It looks like you decided to delete cookies with the given name using a regexp. While I understand that it's a fairly "complete" implementation, my impression of reading the RFC is that it's unnecessary, as adding a later header with `value=` and `expires=1970` is sufficient as the user agent should process it in order. Do you remember why you implemented it this way?

lib/rack/lint.rb
```diff
@@ -673,19 +673,23 @@ def check_headers(headers)
...
+            ## or an Array of String instances,
+            value.each{|value| check_header_value(key, value)}
+          else
+            raise LintError, "a header value must be a String or Array of Strings, but the value of '#{key}' is a #{value.class}"
+          end
+        end
+      end
+
+      def check_header_value(key, value)
+        ## such that each String instance must not contain characters below 037.
```
jeremyevans, line 689: If servers want to support both Rack 1/2 and Rack 3 applications, they'll need to still check for newlines in strings.  As newlines in strings are not valid in Rack 3, the server can assume if they are present, they are dealing with Rack 1/2.  This isn't a problem, I just want to confirm that's how we expect servers to handle this.

tenderlove:
This is definitely better than splitting on `\n`, but I have a few concerns.  First, we should make the webrick handler backwards compatible.  Second, it's very very bad if an outgoing header contains a newline as it can lead to header splitting attacks.  If an attacker is able to control header values, they can use \n to set cookies or whatever.  The fact that the Rack SPEC requires webservers to split on `\n` surprisingly helps us avoid that particular issue as it means output header values can never have a newline.

I realize the spec says values below `037` are not allowed but it looks like we're not verifying that in the webrick server.

I'm not sure what we should do exactly, but the spec needs to be quite strongly worded with regard to the data allowed in headers, and we might want to add exceptions and checking in `Rack::Response`.

lib/rack/handler/webrick.rb
```diff
@@ -95,15 +95,15 @@ def service(req, res)
...
-          headers.each { |k, vs|
-            if k == RACK_HIJACK
-              io_lambda = vs
-            elsif k == "set-cookie"
-              res.cookies.concat vs.split("\n")
+          headers.each { |key, value|
+            if key == RACK_HIJACK
+              io_lambda = value
+            elsif key == "set-cookie"
+              res.cookies.concat(Array(value))
```
jeremyevans, line 102: Because this handler is included with Rack, it's fine to support only Rack 3 protocol and not Rack 1/2 protocol. As @tenderlove mentioned, passing `\n` directly through on the connection is bad idea for security.  We should print an error message to stderr, and probably skip that header, or cut off the value at the first `\n`.

lib/rack/response.rb
```diff
@@ -143,19 +143,19 @@ def empty?
 
     def has_header?(key)
       raise ArgumentError unless key.is_a?(String)
-      headers.key? key
+      @headers.key?(key)
```
jeremyevans, line 146: In the future, probably best to leave unrelated changes in a separate commit.

lib/rack/utils.rb
```diff
@@ -222,7 +222,7 @@ def parse_cookies_header(header)
       end
     end
 
-    def add_cookie_to_header(header, key, value)
```
jeremyevans, line 225: The public methods that you would like to remove should be properly deprecated in Rack 3.0 before removal in Rack 3.1.  If you don't think that is feasible, please explain why for each case.

jeremyevans (changes requested):
I'm OK with the protocol change.  I think we need to properly deprecate the Utils methods, unless doing so is not feasible.

lib/rack/lint.rb
ioquatix: I think that’s fair, what are the implications?

lib/rack/utils.rb
ioquatix: I’ll do my best to add compatible shims.

lib/rack/handler/webrick.rb
ioquatix: I don’t have a strong opinion about this but what do you think of the performance implications?

ioquatix:
@tenderlove I actually had a similar discussion and we introduced checks here as a result (which is used by Falcon): 

https://github.com/socketry/protocol-http1/blob/main/lib/protocol/http1/connection.rb#L160

This problem doesn’t apply to HTTP/2 w.r.t. potential security issues so I don’t bother with a similar check there.

lib/rack/lint.rb
jeremyevans: I guess that the implications for servers would be that in case `\n` was accidentally included in the header value, instead of skipping or truncating the header, we would end up with multiple headers (per the Rack 2 protocol).

lib/rack/handler/webrick.rb
jeremyevans: I doubt anyone running webrick cares too much about performance.  I don't think the performance differences will be major.  One `String#index` call per header value, and either skipping or truncating the header value if it doesn't return nil.

ioquatix:
@tenderlove I checked this and Webrick already fails if you give it a header containing an invalid value:

e.g.

```
# config.ru

run lambda{|env| [200, {"test" => "a\nb: c"}, {}]}

```

Result:

```
> curl -v http://localhost:9292
*   Trying 127.0.0.1:9292...
* Connected to localhost (127.0.0.1) port 9292 (#0)
> GET / HTTP/1.1
> Host: localhost:9292
> User-Agent: curl/7.80.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 500 Internal Server Error
< Content-Type: text/html; charset=ISO-8859-1
* no chunk, no close, no size. Assume close to signal end
< 
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.0//EN">
<HTML>
  <HEAD><TITLE>Internal Server Error</TITLE></HEAD>
  <BODY>
    <H1>Internal Server Error</H1>
    WEBrick::HTTPResponse::InvalidHeader
    <HR>
    <ADDRESS>
     WEBrick/1.7.0 (Ruby/3.0.3/2021-11-24) at
     localhost:9292
    </ADDRESS>
  </BODY>
</HTML>
* Closing connection 0
```

So, I don't think we need to do anything here, as it's already correctly failing. WDYT?

lib/rack/handler/webrick.rb
ioquatix: As outlined below, Webrick rejects invalid header keys/values already, so I don't think we need to add redundant validations here.

ioquatix:
> This is definitely better than splitting on \n, but I have a few concerns. First, we should make the webrick handler backwards compatible

As @jeremyevans mentioned, if someone includes Rack 2 they will get the Rack 2 implementation, and if they include Rack 3 they will get the Rack 3 implementation. I don't think backwards compatibility is needed here?

> I realize the spec says values below 037 are not allowed but it looks like we're not verifying that in the webrick server.

Webrick appears to validate this internally now.

> I'm not sure what we should do exactly, but the spec needs to be quite strongly worded with regard to the data allowed in headers, and we might want to add exceptions and checking in `Rack::Response`.

My feeling is it's better to check this in the server since it reduces the performance cost (doesn't double up on checks).

lib/rack/utils.rb
ioquatix: Okay, I've added compatibility shims.

lib/rack/utils.rb
```diff
@@ -223,6 +223,21 @@ def parse_cookies_header(header)
     end
 
     def add_cookie_to_header(header, key, value)
+      warn "add_cookie_to_header is deprecated and will be removed in Rack 3.1"
```
jeremyevans, line 226: Need to add `uplevel: 1` to this warn call and all other warn calls added.

jeremyevans (changes requested):
Looks mostly good, just need to modify the `warn` calls.

lib/rack/utils.rb
ioquatix: Oh, thanks, damn, I didn't learn from this last time you mentioned it.

jeremyevans (approved):

PR #1794, opened 2022-01-25
Don't create garbage StringIO.

ioquatix:
Does this break compatibility? Currently if no argument is supplied it defaults to `""`, but now it defaults to `nil`. However, all usage within Rack supplies an argument AFAIK.

### DIFF
--- a/lib/rack/mock.rb
+++ b/lib/rack/mock.rb
@@ -188,7 +188,7 @@ module Rack
       if errors
         @errors = errors.string if errors.respond_to?(:string)
       else
-        @errors = nil
+        @errors = ""
       end
 
       super(body, status, headers)
--- a/lib/rack/mock.rb
+++ b/lib/rack/mock.rb
@@ -182,9 +182,14 @@ module Rack
     # Errors
     attr_accessor :errors
 
-    def initialize(status, headers, body, errors = StringIO.new(""))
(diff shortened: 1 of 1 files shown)


lib/rack/mock.rb
```diff
@@ -182,9 +182,14 @@ class << self
     # Errors
     attr_accessor :errors
 
-    def initialize(status, headers, body, errors = StringIO.new(""))
+    def initialize(status, headers, body, errors = nil)
```
jeremyevans, line 185: I think we should default to `''` instead of `nil`.  This still avoids allocations, since it is a frozen string.  I think it results in the same behavior as well, since StringIO#string returns the input string (which will be the literal frozen string passed to it).

jeremyevans (changes requested):

lib/rack/mock.rb
```diff
@@ -182,9 +182,14 @@ class << self
...
       @original_headers = headers
-      @errors           = errors.string if errors.respond_to?(:string)
+
+      if errors
+        @errors = errors.string if errors.respond_to?(:string)
+      else
+        @errors = nil
```
ioquatix, line 191:
```suggestion
        @errors = ""
```

jeremyevans (approved):

PR #1795, opened 2022-01-25
Cache Rack::Request#POST result if input content type is not parseable (Fixes #749)

jeremyevans:
In all other cases, the result was cached, so not caching in this
case is inconsistent, and can result in unexpected behavior if
POST is called multiple times on the same request.


ioquatix (approved):
LGTM.

PR #1797, opened 2022-01-27
Avoid use of regexps for parsing parameter keys

### LINKED ISSUE #1704: <RegexpError: failed to allocate memory: /\A[\[\]]*([^\[\]]+)\]*/>  error during upload of file
steps to reproduce
 1) dd of=data_178mb if=/dev/zero bs=1M count=178
 2) curl -v --data-binary @data_178mb https://localhost:3000/bulk_create_verification
 3) 

```
2020-08-28 13:11:26 +0000: Rack app error handling request { POST /bulk_create_verification }
#<RegexpError: failed to allocate memory: /\A[\[\]]*([^\[\]]+)\]*/>
/opt/panel/vendor/bundle/ruby/2.5.0/gems/rack-2.1.4/lib/rack/query_parser.rb:87:in `normalize_params'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/rack-2.1.4/lib/rack/query_parser.rb:73:in `block in parse_nested_query'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/rack-2.1.4/lib/rack/query_parser.rb:70:in `each'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/rack-2.1.4/lib/rack/query_parser.rb:70:in `parse_nested_query'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/rack-2.1.4/lib/rack/request.rb:490:in `parse_query'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/rack-2.1.4/lib/rack/request.rb:364:in `POST'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/rack-2.1.4/lib/rack/method_override.rb:45:in `method_override_param'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/rack-2.1.4/lib/rack/method_override.rb:29:in `method_override'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/rack-2.1.4/lib/rack/method_override.rb:17:in `call'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/rack-2.1.4/lib/rack/runtime.rb:24:in `call'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/activesupport-5.2.4.3/lib/active_support/cache/strategy/local_cache_middleware.rb:29:in `call'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/actionpack-5.2.4.3/lib/action_dispatch/middleware/executor.rb:14:in `call'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/rack-2.1.4/lib/rack/sendfile.rb:113:in `call'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/utf8-cleaner-0.2.5/lib/utf8-cleaner/middleware.rb:21:in `call'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/sentry-raven-2.9.0/lib/raven/integrations/rack.rb:51:in `call'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/railties-5.2.4.3/lib/rails/engine.rb:524:in `call'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/puma-4.3.5/lib/puma/configuration.rb:228:in `call'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/puma-4.3.5/lib/puma/server.rb:713:in `handle_request'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/puma-4.3.5/lib/puma/server.rb:472:in `process_client'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/puma-4.3.5/lib/puma/server.rb:328:in `block in run'
/opt/panel/vendor/bundle/ruby/2.5.0/gems/puma-4.3.5/lib/puma/thread_pool.rb:134:in `block in spawn_thread'
```

jeremyevans:
This avoids a RegexpError when trying to parse long key input.
It also avoids an invalid InvalidParameterError when trying
to parse non-UTF8 keys, which was only raised previously
because regexps were used without marking them as ASCII-8BIT.

Flip the depth parameter to QueryParser#normalize_params to
be the current parsing depth, instead of a downward counter
from the maximum depth.

Fix a bunch of questionable behavior in parameter parsing
when using [ and ] outside cases that are expected. Treat
[ and ] as normal characters if the occur outside expected
usage.

This leaves one questionable parameter parsing behavior that
also existed previously, which is that: a[b]c is parsed the
same as a[b][c].

Fixes #1704.

### DIFF
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -72,7 +72,7 @@ module Rack
         (qs || '').split(separator ? (COMMON_SEP[separator] || /[#{separator}] */n) : DEFAULT_SEP).each do |p|
           k, v = p.split('=', 2).map! { |s| unescape(s) }
 
-          normalize_params(params, k, v, param_depth_limit)
+          normalize_params(params, k, v, 0)
         end
       end
 
@@ -85,11 +85,29 @@ module Rack
     # the structural types represented by two different parameter names are in
     # conflict, a ParameterTypeError is raised.
     def normalize_params(params, name, v, depth)
-      raise RangeError if depth <= 0
+      raise RangeError if depth >= param_depth_limit
 
-      name =~ %r(\A[\[\]]*([^\[\]]+)\]*)
-      k = $1 || ''
-      after = $' || ''
+      if !name
+        k = after = ''
+      elsif depth == 0
+        if start = name.index('[', 1)
+          k = name[0, start]
+          after = name[start, name.length]
+        else
+          k = name
+          after = ''
+        end
+      elsif name.start_with?('[]')
+        k = '[]'
+        after = name[2, name.length]
+      elsif name.start_with?('[') && (start = name.index(']', 1))
+        k = name[1, start-1]
+        after = name[start+1, name.length]
+      else
+        k = name
+        after = ''
+      end
+
       v ||= String.new
 
       if k.empty?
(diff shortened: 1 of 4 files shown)


ioquatix:
This looks good to me but what are the performance implications?

jeremyevans:
Improves performance from brief testing.  Using this example from the specs:

```
Rack::Utils.parse_nested_query( "x[][id]=1&x[][y][a]=5&x[][y][b]=7&x[][z][id]=3&x[][z][w]=0&x[][id]=2&x[][y][a]=6&x[][y][b]=8&x[][z][id]=4&x[][z][w]=0")
```

Benchmark.ips shows an increase of 3.024k i/s to 3.989k i/s, an improvement of about 32%.  This matches my expectation, since in previous cases where I have replaced regexp usage with string methods in Roda, I've generally also seen performance improvements.

rafaelfranca (approved):
The implementation is harder to understand though, but it solves a lot of issues, so I think is worth it.

lib/rack/query_parser.rb
```diff
@@ -85,11 +85,29 @@ def parse_nested_query(qs, separator = nil)
     # the structural types represented by two different parameter names are in
     # conflict, a ParameterTypeError is raised.
     def normalize_params(params, name, v, depth)
-      raise RangeError if depth <= 0
+      raise RangeError if depth >= param_depth_limit
+
+      if !name
```
ioquatix, line 90: Can youn please add documentation explaining the behaviour of the implementation, even if just high level, e.g.

```
# Parse and consume the X token:
```

ioquatix (changes requested):

lib/rack/query_parser.rb
ioquatix: Basically, if anyone else needs to work on this code, let's make sure they are with relative ease, able to understand the high level structure and how it's working.

jeremyevans: That makes sense, I pushed a commit that adds some comments.

ioquatix (approved):

ioquatix:
I've approved this but I wonder given how critical this code path is, whether we should fuzz both the old and new implementation and compare the outputs to confirm identical behaviours.

jeremyevans:
> I've approved this but I wonder given how critical this code path is, whether we should fuzz both the old and new implementation and compare the outputs to confirm identical behaviours.

I thought the commit message indicated that the new behavior was not identical to the old behavior and not intended to be.  The previous behavior was nonsensical in a lot of corner cases.  The new behavior passes all of the old tests (except one test expecting an encoding exception).  The old behavior won't pass the new tests.

PR #1800, opened 2022-02-03
Add Rack::Headers to support lower-case header keys

jeremyevans:
This implements a Hash subclass that lower-cases keys passed to it.
The expected usage is for Rack 2 libraries/frameworks that want to be
compatible with Rack 3, assuming that we require response header
keys to be lower-case in Rack 3 (as discussed in #1592).

I've tested this implementation in Roda (a web framework), and only
minimal changes were needed to get Roda's tests passing with it,
even though Roda uses mixed-case headers throughout its implementation
and tests.  It was simple to get the tests passing both with the
implementation and without, showing that it's possible to get a
framework that can support both Rack 2 and Rack 3.

Additionally, I tested Rodauth (an authentication framework built on
top of Roda), with the modified version of Roda, and again, only
minimal changes were needed to keep the tests passing.

I also tested a handful of small applications based on Roda using the
modified version of Roda, and none of those applications needed any
changes to keep their tests passing.  I think that shows that
requiring lower-case response header keys is probably acceptable,
since it should result in minimal if any churn to applications, and
only minor churn to libraries.

The implementation of Rack::Headers is based on ruby-cicphash
(a case-insensitive, case-preserving hash implementation),
but streamlined as case doesn't need to be preserved, and it
is a Hash subclass (as required by Rack 3 SPEC), instead of
an Object subclass that uses a hash internally.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file. For info on
 
 ### Added
 
+- `Rack::Headers` added to support lower-case header keys. ([@jeremyevans](https://github.com/jeremyevans))
 - `Rack::RewindableInput` supports size. ([@ahorek](https://github.com/ahorek))
 - Rack::Session::Pool now accepts `:allow_fallback` option to disable fallback to public id. ([#1431](https://github.com/rack/rack/issues/1431), [@jeremyevans](https://github.com/jeremyevans))
 
--- a/lib/rack.rb
+++ b/lib/rack.rb
@@ -90,6 +90,7 @@ module Rack
   autoload :ForwardRequest, "rack/recursive"
   autoload :Handler, "rack/handler"
   autoload :Head, "rack/head"
+  autoload :Headers, "rack/headers"
   autoload :Lint, "rack/lint"
   autoload :Lock, "rack/lock"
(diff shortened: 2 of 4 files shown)


ioquatix:
I like this a lot.

Are we operating under the assumption that Rack 3 native applications can just use a `Hash` instance, and Rack 2 / mixed case applications can use `Rack::Headers`?

I personally think this is fine, but thinking about it, I have a slight feeling as to whether we should make this use case slightly more ugly to encourage correct usage. However, I think the performance impact alone should hopefully be enough to encourage people to switch.

Some of the use cases I want to try and help address with this spec/implementation change are issues like https://github.com/rack/rack-test/issues/229 which really don't have a good solution in the current Rack 2 spec. Because the only solution is to normalise the case of the headers before doing any kind of lookup. I prefer the approach proposed here because I believe it correctly addresses the real issue in an efficient way. HTTP/2 server implementation will be (slightly) more efficient as a result.

ioquatix (approved):

jeremyevans:
Correct, this is to allow applications/middleware designed for Rack 2 to more easily transition to the Rack 3 API.  I think the performance impact, while as small I as could make it, is still probably high enough that people would want to switch.

PR #1801, opened 2022-02-04
Enable directly requiring Rack components

jeremyevans:
Add test:separate task to test run each test file separately.
This is to test that directly requiring a rack component will work.

With test:separate, the test helper will not require rack, and each
test file should call separate_testing with a block that uses
require_relative to only require that component being tested.

Add the appropriate relative requires to get all tests passing
when run separately.

Note that this isn't foolproof. In many cases, the tests use
Rack components that are not under test (such as lint and mock),
and both of those other other rack components, so it's possible
the tests are loading hidden dependencies.  I attempted to check
each component to try to find the hidden dependencies, but it is
certainly possible I missed something.

Implements #1621.


stevenharman:
Wow, this is great. Thank you, Jeremy, for tackling this.

I will (hopefully soon) get back to fixing how we qualify the constants. Life and `$JOB` have had me a bit scattered lately.

ioquatix (approved):
I basically like this change, however some things I think we can improve on, but this is a really great starting point. Thanks for your effort.

tenderlove:
Wow.  Thank you!

jeremyevans:
> I basically like this change, however some things I think we can improve on, but this is a really great starting point. Thanks for your effort.

Agreed.  One advantage of separate testing is it showed quite a few verbose warnings indicating things we should probably fix.

Issue #1777, opened 2021-10-22
Can rack.after_reply be added to the spec?


BlakeWilliams:
Both [Puma](https://github.com/puma/puma/blob/56a25f709604240fb32310f15126d47d5a6541ca/lib/puma/request.rb#L72-L176) and [Unicorn](https://github.com/defunkt/unicorn/blob/158e9aad11ee2ed7dc01182da150e803f7cdbfef/lib/unicorn/http_server.rb#L632-L657) have support for `rack.after_reply` which has been extremely useful but it's not part of the official Rack spec. 

Given its usefulness and existing support, would it make sense to make this an official part of the spec to help encourage other rack based webservers to implement this functionality? If not, is there a path forward to making this an official part of the spec?

ioquatix:
@tenderlove I know you have some ideas around this, do you mind reviewing this request?

tenderlove:
I agree we should add this and I like `rack.response_finished`.  I'm not 100% sure what to do in the case of errors though.  I think we should send this message even in the case an error occurs, but pass a value (maybe an exception object?).  The thing is no webserver can really guarantee this thing will get executed (maybe the process segv's, or gets `kill`d or something).  I don't think that's a big deal, we just need to document it.  Also each webserver should probably document its guarantees wrt executing this.

ioquatix:
@BlakeWilliams do you want to have a go at cutting the spec for this? Update `lib/rack/lint.rb`

nateberkopec:
Just linking for posterity - an excellent use of `after_reply`: https://github.com/testdouble/maybe_later

BlakeWilliams:
> @BlakeWilliams do you want to have a go at cutting the spec for this? Update `lib/rack/lint.rb`

Definitely, I'll make some time to work on that.

BlakeWilliams:
Just opened up a draft PR that has a few open questions here: https://github.com/rack/rack/pull/1802

ioquatix:
Merged in https://github.com/rack/rack/pull/1952

PR #1802, opened 2022-02-04
 Add rack.response_finished to Rack::Lint

BlakeWilliams:
This updates Rack::Lint to validate that `rack.response_finished` is an
array of callables when present in the `env`. e.g. procs, lambdas, or
objects that respond to `call`.

This validates that:

* `rack.response_finished` is an array
* The contents of the array all respond to `call`

Part of https://github.com/rack/rack/issues/1777

### DIFF
--- a/SPEC.rdoc
+++ b/SPEC.rdoc
@@ -137,6 +137,7 @@ There are the following restrictions:
   set. <tt>PATH_INFO</tt> should be <tt>/</tt> if
   <tt>SCRIPT_NAME</tt> is empty.
   <tt>SCRIPT_NAME</tt> never should be <tt>/</tt>, but instead be empty.
+<tt>rack.response_finished</tt>:: An array of callables run after the HTTP response has been sent.
 
 === The Input Stream
 
--- a/SPEC.rdoc
+++ b/SPEC.rdoc
@@ -139,7 +139,9 @@ There are the following restrictions:
   <tt>SCRIPT_NAME</tt> never should be <tt>/</tt>, but instead be empty.
 <tt>rack.response_finished</tt>:: An array of callables run after the HTTP response has been sent.
 The array of callables should be called directly after the HTTP response has
-been closed. The callables should be called sequentially and synchronously.
+been completely sent. The callables should be called sequentially and synchronously
+in the same execution context as the response. If an exception is raised, it will
+be ignored and will not impact the execution of other callables.
 
 
 === The Input Stream
--- a/SPEC.rdoc
+++ b/SPEC.rdoc
@@ -137,7 +137,15 @@ There are the following restrictions:
   set. <tt>PATH_INFO</tt> should be <tt>/</tt> if
   <tt>SCRIPT_NAME</tt> is empty.
   <tt>SCRIPT_NAME</tt> never should be <tt>/</tt>, but instead be empty.
-<tt>rack.response.finished</tt>:: An array of callables run after the HTTP response has been sent.
+<tt>rack.response.finished</tt>:: An array of objects responding to #call with one argument, the env hash for the request, that will be called after the HTTP response has been sent to the client.
+The callables are called directly after the HTTP response has been sent to the
+client. The callables should be called sequentially and synchronously in the
+same execution context as the the response. If an exception is raised, it will
+be ignored and will not impact the execution of the other callables. The
+callbacks will be called event if the request is cancelled (e.g. a user closing
+the browser tab before the request completes). Servers supporting this
+functionality will prepopulate env with an empty array.
+
 
 === The Input Stream
 
--- a/SPEC.rdoc
+++ b/SPEC.rdoc
@@ -146,7 +146,6 @@ callbacks will be called event if the request is cancelled (e.g. a user closing
 the browser tab before the request completes). Servers supporting this
 functionality will prepopulate env with an empty array.
 
-
 === The Input Stream
 
 The input stream is an IO-like object which contains the raw HTTP
--- a/SPEC.rdoc
+++ b/SPEC.rdoc
@@ -138,6 +138,9 @@ There are the following restrictions:
   <tt>SCRIPT_NAME</tt> is empty.
   <tt>SCRIPT_NAME</tt> never should be <tt>/</tt>, but instead be empty.
 <tt>rack.response_finished</tt>:: An array of callables run after the HTTP response has been sent.
+The array of callables should be called directly after the HTTP response has
+been closed. The callables should be called sequentially and synchronously.
+
 
 === The Input Stream
 
--- a/SPEC.rdoc
+++ b/SPEC.rdoc
@@ -137,12 +137,7 @@ There are the following restrictions:
   set. <tt>PATH_INFO</tt> should be <tt>/</tt> if
   <tt>SCRIPT_NAME</tt> is empty.
   <tt>SCRIPT_NAME</tt> never should be <tt>/</tt>, but instead be empty.
-<tt>rack.response_finished</tt>:: An array of objects responding to #call with one argument, the env hash for the request, that will be called after the HTTP response has been sent to the client.
-The callables are called directly after the HTTP response has been sent to the
-client. The callables should be called sequentially and synchronously in the
-same execution context as the the response. If an exception is raised, it will
-be ignored and will not impact the execution of the other callables.
-
+<tt>rack.response.finished</tt>:: An array of callables run after the HTTP response has been sent.
 
 === The Input Stream
 
--- a/SPEC.rdoc
+++ b/SPEC.rdoc
@@ -137,11 +137,11 @@ There are the following restrictions:
   set. <tt>PATH_INFO</tt> should be <tt>/</tt> if
   <tt>SCRIPT_NAME</tt> is empty.
   <tt>SCRIPT_NAME</tt> never should be <tt>/</tt>, but instead be empty.
-<tt>rack.response_finished</tt>:: An array of callables run after the HTTP response has been sent.
-The array of callables should be called directly after the HTTP response has
-been completely sent. The callables should be called sequentially and synchronously
-in the same execution context as the response. If an exception is raised, it will
-be ignored and will not impact the execution of other callables.
+<tt>rack.response_finished</tt>:: An array of objects responding to #call with one argument, the env hash for the request, that will be called after the HTTP response has been sent to the client.
+The callables are called directly after the HTTP response has been sent to the
+client. The callables should be called sequentially and synchronously in the
+same execution context as the the response. If an exception is raised, it will
+be ignored and will not impact the execution of the other callables.
 
 
 === The Input Stream
--- a/lib/rack/constants.rb
+++ b/lib/rack/constants.rb
@@ -52,6 +52,7 @@ module Rack
   RACK_RECURSIVE_INCLUDE              = 'rack.recursive.include'
   RACK_MULTIPART_BUFFER_SIZE          = 'rack.multipart.buffer_size'
   RACK_MULTIPART_TEMPFILE_FACTORY     = 'rack.multipart.tempfile_factory'
+  RACK_RESPONSE_FINISHED              = 'rack.response_finished'
   RACK_REQUEST_FORM_INPUT             = 'rack.request.form_input'
   RACK_REQUEST_FORM_HASH              = 'rack.request.form_hash'
   RACK_REQUEST_FORM_VARS              = 'rack.request.form_vars'
--- a/lib/rack/constants.rb
+++ b/lib/rack/constants.rb
@@ -52,7 +52,7 @@ module Rack
   RACK_RECURSIVE_INCLUDE              = 'rack.recursive.include'
   RACK_MULTIPART_BUFFER_SIZE          = 'rack.multipart.buffer_size'
   RACK_MULTIPART_TEMPFILE_FACTORY     = 'rack.multipart.tempfile_factory'
-  RACK_RESPONSE_FINISHED              = 'rack.response_finished'
+  RACK_RESPONSE_FINISHED              = 'rack.response.finished'
   RACK_REQUEST_FORM_INPUT             = 'rack.request.form_input'
   RACK_REQUEST_FORM_HASH              = 'rack.request.form_hash'
   RACK_REQUEST_FORM_VARS              = 'rack.request.form_vars'
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -359,6 +359,15 @@ module Rack
         unless env[SCRIPT_NAME] != "/"
           raise LintError, "SCRIPT_NAME cannot be '/', make it '' and PATH_INFO '/'"
         end
+
+        ## <tt>rack.response_finished</tt>:: An array of callables run after the HTTP response has been sent.
+        if callables = env[RACK_RESPONSE_FINISHED]
+          raise LintError, "rack.response_finished must be an array of callable objects" unless callables.is_a?(Array)
+
+          callables.each do |callable|
+            raise LintError, "rack.response_finished values must respond to call" unless callable.respond_to?(:call)
+          end
+        end
       end
 
       ##
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -360,12 +360,12 @@ module Rack
           raise LintError, "SCRIPT_NAME cannot be '/', make it '' and PATH_INFO '/'"
         end
 
-        ## <tt>rack.response_finished</tt>:: An array of callables run after the HTTP response has been sent.
+        ## <tt>rack.response.finished</tt>:: An array of callables run after the HTTP response has been sent.
         if callables = env[RACK_RESPONSE_FINISHED]
-          raise LintError, "rack.response_finished must be an array of callable objects" unless callables.is_a?(Array)
+          raise LintError, "rack.response.finished must be an array of callable objects" unless callables.is_a?(Array)
 
           callables.each do |callable|
-            raise LintError, "rack.response_finished values must respond to call" unless callable.respond_to?(:call)
+            raise LintError, "rack.response.finished values must respond to call" unless callable.respond_to?(:call)
 
             arity = if callable.respond_to?(:arity)
               callable.arity
@@ -373,7 +373,7 @@ module Rack
               callable.method(:call).arity
             end
 
-            raise LintError, "rack.response_finished values must accept an env argument" unless arity == 1
+            raise LintError, "rack.response.finished values must accept an env argument" unless arity == 1
           end
         end
       end
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -366,6 +366,14 @@ module Rack
 
           callables.each do |callable|
             raise LintError, "rack.response_finished values must respond to call" unless callable.respond_to?(:call)
+
+            arity = if callable.respond_to?(:arity)
+              callable.arity
+            else
+              callable.method(:call).arity
+            end
+
+            raise LintError, "rack.response_finished values must accept an env argument" unless arity == 1
           end
         end
       end
--- a/test/spec_lint.rb
+++ b/test/spec_lint.rb
@@ -720,7 +720,7 @@ describe Rack::Lint do
 
   it "pass valid rack.response_finished" do
     Rack::Lint.new(lambda { |env|
-                     [200, { "rack.response_finished" => [-> {}, lambda {}], "Content-length" => "3" }, ["foo"]]
+                     [200, { "rack.response_finished" => [-> {}, lambda {}], "content-length" => "3" }, ["foo"]]
                    }).call(env({})).first.must_equal 200
   end
 
--- a/test/spec_lint.rb
+++ b/test/spec_lint.rb
@@ -209,6 +209,16 @@ describe Rack::Lint do
       Rack::Lint.new(nil).call(env("SCRIPT_NAME" => "/"))
     }.must_raise(Rack::Lint::LintError).
       message.must_match(/cannot be .* make it ''/)
+
+    lambda {
+      Rack::Lint.new(nil).call(env("rack.response_finished" => "not a callable"))
+    }.must_raise(Rack::Lint::LintError).
+    message.must_match(/rack.response_finished must be an array of callable objects/)
+
+    lambda {
+      Rack::Lint.new(nil).call(env("rack.response_finished" => [->{}, "not a callable"]))
+    }.must_raise(Rack::Lint::LintError).
+    message.must_match(/rack.response_finished values must respond to call/)
   end
 
   it "notice input errors" do
@@ -706,6 +716,12 @@ describe Rack::Lint do
                        env['rack.hijack?'] = true
                        [201, { "content-type" => "text/plain", "content-length" => "0", 'rack.hijack' => lambda {|io| io }, 'rack.hijack_io' => StringIO.new }, []]
                      }).call(env({}))[1]['rack.hijack'].call(StringIO.new).read.must_equal ''
+  end
+
+  it "pass valid rack.response_finished" do
+    Rack::Lint.new(lambda { |env|
+                     [200, { "rack.response_finished" => [-> {}, lambda {}], "Content-length" => "3" }, ["foo"]]
+                   }).call(env({})).first.must_equal 200
   end
 
 end
--- a/test/spec_lint.rb
+++ b/test/spec_lint.rb
@@ -211,19 +211,19 @@ describe Rack::Lint do
       message.must_match(/cannot be .* make it ''/)
 
     lambda {
-      Rack::Lint.new(nil).call(env("rack.response_finished" => "not a callable"))
+      Rack::Lint.new(nil).call(env("rack.response.finished" => "not a callable"))
     }.must_raise(Rack::Lint::LintError).
-    message.must_match(/rack.response_finished must be an array of callable objects/)
+    message.must_match(/rack.response.finished must be an array of callable objects/)
 
     lambda {
-      Rack::Lint.new(nil).call(env("rack.response_finished" => [-> (env) {}, "not a callable"]))
+      Rack::Lint.new(nil).call(env("rack.response.finished" => [-> (env) {}, "not a callable"]))
     }.must_raise(Rack::Lint::LintError).
-    message.must_match(/rack.response_finished values must respond to call/)
+    message.must_match(/rack.response.finished values must respond to call/)
 
     lambda {
-      Rack::Lint.new(nil).call(env("rack.response_finished" => [-> () {}]))
+      Rack::Lint.new(nil).call(env("rack.response.finished" => [-> () {}]))
     }.must_raise(Rack::Lint::LintError).
-    message.must_match(/rack.response_finished values must accept an env argument/)
+    message.must_match(/rack.response.finished values must accept an env argument/)
 
     callable_object = Class.new do
       def call
@@ -231,9 +231,9 @@ describe Rack::Lint do
     end.new
 
     lambda {
-      Rack::Lint.new(nil).call(env("rack.response_finished" => [callable_object]))
+      Rack::Lint.new(nil).call(env("rack.response.finished" => [callable_object]))
     }.must_raise(Rack::Lint::LintError).
-    message.must_match(/rack.response_finished values must accept an env argument/)
+    message.must_match(/rack.response.finished values must accept an env argument/)
   end
 
   it "notice input errors" do
@@ -733,7 +733,7 @@ describe Rack::Lint do
                      }).call(env({}))[1]['rack.hijack'].call(StringIO.new).read.must_equal ''
   end
 
-  it "pass valid rack.response_finished" do
+  it "pass valid rack.response.finished" do
     callable_object = Class.new do
       def call(env)
       end
@@ -741,7 +741,7 @@ describe Rack::Lint do
 
     Rack::Lint.new(lambda { |env|
                      [200, {}, ["foo"]]
-                   }).call(env({ "rack.response_finished" => [-> (env) {}, lambda { |env| }, callable_object], "content-length" => "3" })).first.must_equal 200
+                   }).call(env({ "rack.response.finished" => [-> (env) {}, lambda { |env| }, callable_object], "content-length" => "3" })).first.must_equal 200
   end
 
 end
--- a/test/spec_lint.rb
+++ b/test/spec_lint.rb
@@ -216,9 +216,24 @@ describe Rack::Lint do
     message.must_match(/rack.response_finished must be an array of callable objects/)
 
     lambda {
-      Rack::Lint.new(nil).call(env("rack.response_finished" => [->{}, "not a callable"]))
+      Rack::Lint.new(nil).call(env("rack.response_finished" => [-> (env) {}, "not a callable"]))
     }.must_raise(Rack::Lint::LintError).
     message.must_match(/rack.response_finished values must respond to call/)
+
+    lambda {
+      Rack::Lint.new(nil).call(env("rack.response_finished" => [-> () {}]))
+    }.must_raise(Rack::Lint::LintError).
+    message.must_match(/rack.response_finished values must accept an env argument/)
+
+    callable_object = Class.new do
+      def call
+      end
+    end.new
+
+    lambda {
+      Rack::Lint.new(nil).call(env("rack.response_finished" => [callable_object]))
+    }.must_raise(Rack::Lint::LintError).
+    message.must_match(/rack.response_finished values must accept an env argument/)
   end
 
   it "notice input errors" do
@@ -719,9 +734,14 @@ describe Rack::Lint do
   end
 
   it "pass valid rack.response_finished" do
+    callable_object = Class.new do
+      def call(env)
+      end
+    end.new
+
     Rack::Lint.new(lambda { |env|
-                     [200, { "rack.response_finished" => [-> {}, lambda {}], "content-length" => "3" }, ["foo"]]
-                   }).call(env({})).first.must_equal 200
+                     [200, {}, ["foo"]]
+                   }).call(env({ "rack.response_finished" => [-> (env) {}, lambda { |env| }, callable_object], "content-length" => "3" })).first.must_equal 200
   end
 
 end


lib/rack/lint.rb
```diff
@@ -356,6 +356,15 @@ def check_environment(env)
         unless env[SCRIPT_NAME] != "/"
           raise LintError, "SCRIPT_NAME cannot be '/', make it '' and PATH_INFO '/'"
         end
+
+        ## <tt>rack.response_finished</tt>:: An array of callables run after the HTTP response has been sent.
+        if callables = env[RACK_RESPONSE_FINISHED]
+          raise LintError, "rack.response_finished must be an array of callable objects" unless callables.is_a?(Array)
+
+          callables.each do |callable|
+            raise LintError, "rack.response_finished values must respond to call" unless callable.respond_to?(:call)
```
BlakeWilliams, line 365: I was originally checking `.arity` here too to ensure it's <= 1 but that seemed like it may needlessly rigid since now objects that respond to `call` can be passed as well.

It does bring up a question about errors though. If an exception is raised do we want servers to pass it to `call`, or do we leave it up to the server to check for arity and pass the exception if it accepts an argument?

lib/rack/lint.rb
```diff
@@ -356,6 +356,15 @@ def check_environment(env)
         unless env[SCRIPT_NAME] != "/"
           raise LintError, "SCRIPT_NAME cannot be '/', make it '' and PATH_INFO '/'"
         end
+
+        ## <tt>rack.response_finished</tt>:: An array of callables run after the HTTP response has been sent.
```
BlakeWilliams, line 360: Should there be some mention of upstream servers implementing this functionality similar to `rack.hijack`? How much do we want to leave up to the implementation?

lib/rack/lint.rb
ioquatix: Regarding arity, I think one way to handle this is to wrap the array with a linter which validates the calls and responses.

Regarding errors, I think we need to make a decision about this:

1. Either we care about errors which occur when writing the response, and provide a mechanism for dealing with them via this interface.
2. We ignore errors or provide some other interface for receiving errors.

The difference between (1) and (2) is that (2) is less well specified and thus server implementations have more flexibility as to how they implement the specification.

lib/rack/lint.rb
ioquatix: My feeling is, if we care about this feature, it should be a mandatory requirement of Rack 3. However, maybe @tenderlove can give his opinion?

lib/rack/lint.rb
BlakeWilliams: > Regarding arity, I think one way to handle this is to wrap the array with a linter which validates the calls and responses.

I think I see what you mean, but let me know if this isn't correct. The idea is to make sure each value is called and doesn't raise? There's no specific return value, at least in the current implementation of `rack.after_reply` which I'm basing this on.

> Regarding errors, I think we need to make a decision about this:
> 
> 1. Either we care about errors which occur when writing the response, and provide a mechanism for dealing with them via this interface.
> 2. We ignore errors or provide some other interface for receiving errors.
> 
> The difference between (1) and (2) is that (2) is less well specified and thus server implementations have more flexibility as to how they implement the specification.

Yeah, I'm on the fence here. The implementations I've seen/implemented myself don't need to check for errors since the code would have been run anyways, but I can imagine there's at least a few use-cases where it would be helpful to have the error.

If we omit it from the spec, I think there's a few different ways the error could be retrieved. One thought would be to catch the exception in a middleware and store it in the env since the callable may have access to the env.

jeremyevans:
> `Rack::Lint` seems to validate that middleware implements the spec correctly but `rack.response_finished` is completely dependent on the server implementation. Is there a standard way to validate that a server implements specific behavior?

No.  `Rack::Lint` doesn't validate that a server handles responses correctly, it only validates inputs (`env`) and outputs (`[status, headers, body]`).  This implementation looks fine to me.  I think we should keep the `rack.after_reply` name, as Unicorn and Puma already implement it.  If we are adding something to SPEC that is already supported and de facto standardized by multiple implementations, we should use the existing name, even if we think our name is better.

ioquatix:
Is `rack.after_reply` compatible with the proposed spec?

jeremyevans:
> Is `rack.after_reply` compatible with the proposed spec?

Other than the key, it seems to be compatible: 

* Unicorn: `env["rack.after_reply"].each(&:call)`
* Puma: `after_reply.each { |o| o.call }`

There is no advantage in adding a different name for behavior already de facto standardized in existing webservers.

jeremyevans:
Unless we can identify significant problems, I believe we should adopt existing de facto standards as already implemented.  @tenderlove agrees (https://github.com/rack/rack/pull/1701#issuecomment-666525691):

> If popular web servers all do it, then it's really the defacto spec. Rack's job is to define the interface, and if all web servers agree this is a fine interface then I think we're good to go?

Here the question is just about what env key to use, which is purely a bikeshed issue.  The bikeshed is already painted, there is no point in painting it a different color, even if the different color looks nicer.

ioquatix:
@jeremyevans that's not what @tenderlove said: https://github.com/rack/rack/issues/1777#issuecomment-1023768075

lib/rack/lint.rb
ioquatix: > I think I see what you mean, but let me know if this isn't correct. The idea is to make sure each value is called and doesn't raise? There's no specific return value, at least in the current implementation of rack.after_reply which I'm basing this on.

Yes, wrap the values with an internal validation layer.

I don't think I mind what you do here, but lets define something clearly in the spec w.r.t. how well behaved code should execute.

Let's also make it clear when we generally expect execution to occur (in the spec).

ioquatix:
I would like to add, we CAN validate that these are called by the server, by wrapping them - the same way we validate the body was either called or enumerated IIRC.

jeremyevans:
> @jeremyevans that's not what @tenderlove said: [#1777 (comment)](https://github.com/rack/rack/issues/1777#issuecomment-1023768075)

What I quoted is what @tenderlove said in issue #1701.  It is a general statement about how we should treat a de facto spec.  It looks like what @tenderlove states in #1777 is in favor of this particular case, even though this case goes against his advice in #1701.  I will leave it to @tenderlove to explain why the general statement does not apply to this particular case.

> I would like to add, we CAN validate that these are called by the server, by wrapping them - the same way we validate the body was either called or enumerated IIRC.

I suppose we could do that, but I don't see the advantage.  Lint is not about validating how servers handle responses, merely that the response provided to the server is in the correct format.

Even assuming we wanted Lint to do this, when could you actually validate them?  You could use a timeout that fails if it isn't called.  How long would such a timeout be?   Would it start when the body was closed?  Not sure how such a timeout can reliably raise an appropriate exception, and I think it's a fools errand to try such an approach

ioquatix:
@jeremyevans I think these are all really good questions. Does the design being hard to test indicate a more fundamental issue?

All the issues you mention seem like reasonable things... like, when does this execute? After 1 year? When the server shuts down? Should we allow it to be deferred? Should it run on a background thread? Does this indicate design issues we need to address?

BlakeWilliams:
> All the issues you mention seem like reasonable things... like, when does this execute? After 1 year? When the server shuts down? Should we allow it to be deferred? Should it run on a background thread? Does this indicate design issues we need to address?

I can attempt to answer some of these questions in the spec. I'll try adding some details about how/when these callables should be called under `rack.response_finished` in `SPEC.rdoc`, but happy to add it elsewhere if there's a better place.

I'm also happy to revert the name back to `rack.after_reply` if that's what we want to do. There's a good amount of value keeping the old name since existing implementations will continue to work and servers won't have to go through a full deprecation process due to the name change. Happy to defer to y'all though.

jeremyevans:
@tenderlove I think your guidance is needed here to decide whether we should officially bless the de facto standard `rack.after_reply` (implemented by Puma and Unicorn), or standardize a different (perhaps more accurate) `rack.response_finished`.  The semantics will be the same in both cases, so this just depends on whether the benefit of a potentially more accurate name outweighs the cost of forcing changes in Puma, Unicorn, and apps that currently use `rack.after_reply` with Puma or Unicorn.

SPEC.rdoc
```diff
@@ -138,6 +138,9 @@ There are the following restrictions:
   <tt>SCRIPT_NAME</tt> is empty.
   <tt>SCRIPT_NAME</tt> never should be <tt>/</tt>, but instead be empty.
 <tt>rack.response_finished</tt>:: An array of callables run after the HTTP response has been sent.
+The array of callables should be called directly after the HTTP response has
+been closed. The callables should be called sequentially and synchronously.
```
ioquatix, line 142: This is great, but I'd suggest some minor terminology adjustments:

```suggestion
The array of callables should be called directly after the HTTP response has
been sent. The callables should be called sequentially and synchronously in
the same execution context as the response. If an exception is raised, it will
be ignored and will not impact the execution of other callables.
```

I'm not strong on the exception handling clause, the alternative is to propagate it up and fail the response (i.e. log).

ioquatix: A quick survey:

## Puma

Doesn't look like there is any error handling:

https://github.com/puma/puma/blob/master/lib/puma/request.rb#L185

Based on the caller, it looks like the entire connection will be closed:

https://github.com/puma/puma/blob/db751ba82f8980bc3095fc01cbd4458480b804fc/lib/puma/server.rb#L445

## Unicorn

Doesn't look like any error handling:

https://github.com/defunkt/unicorn/blob/master/lib/unicorn/http_server.rb#L657

Based on the caller it looks like it would take out the entire connection:

https://github.com/defunkt/unicorn/blob/a08e4a6d613d9f3b4c2afb57f2691f0710a9d920/lib/unicorn/http_server.rb#L732-L751

ioquatix: I think we need to consider the safety trade off that is being made by the current implementation of `after_reply`. I can see three approaches:

1. The current approach - do nothing and the entire connection might be trashed.
2. `begin...rescue` around ALL callables.
3. `begin...rescue` around EACH callable.

1 is most efficient / least safe <-> 3 is least efficient / most safe.

ioquatix: Testing this out:

```
run ->(env) do
  env['rack.after_reply'] << lambda do
    raise "Fail"
  end

  [200, [], []]
end
```

```
> puma 
Puma starting in single mode...
* Puma version: 5.6.2 (ruby 3.1.1-p18) ("Birdie's Version")
*  Min threads: 0
*  Max threads: 5
*  Environment: development
*          PID: 19925
* Listening on http://0.0.0.0:9292
Use Ctrl-C to stop
2022-04-09 11:57:26 +1200 Read: #<RuntimeError: Fail>
```

Puma logs the error, but:

```
> curl -v http://localhost:9292 http://localhost:9292
*   Trying 127.0.0.1:9292...
* Connected to localhost (127.0.0.1) port 9292 (#0)
> GET / HTTP/1.1
> Host: localhost:9292
> User-Agent: curl/7.80.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Transfer-Encoding: chunked
< 
* Leftovers after chunking:  38u bytes
* Connection #0 to host localhost left intact
* Found bundle for host localhost: 0x6000036d9410 [serially]
* Connection 0 seems to be dead!
* Closing connection 0
* Hostname localhost was found in DNS cache
*   Trying 127.0.0.1:9292...
* Connected to localhost (127.0.0.1) port 9292 (#1)
> GET / HTTP/1.1
> Host: localhost:9292
> User-Agent: curl/7.80.0
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 200 OK
< Transfer-Encoding: chunked
< 
* Leftovers after chunking:  38u bytes
* Connection #1 to host localhost left intact
```

Note that:

```
* Connection #0 to host localhost left intact
* Found bundle for host localhost: 0x6000036d9410 [serially]
* Connection 0 seems to be dead!
```

which indicates the connection has failed internally.

I suspect something similar for Unicorn.

BlakeWilliams: > I think we need to consider the safety trade off that is being made by the current implementation of `after_reply`. I can see three approaches:
> 
> 1. The current approach - do nothing and the entire connection might be trashed.
> 2. `begin...rescue` around ALL callables.
> 3. `begin...rescue` around EACH callable.
> 
> 1 is most efficient / least safe <-> 3 is least efficient / most safe.

If we go with 2 or 3 that might be a good reason to change the name of the feature since the existing implementations won't be compatible with the spec.

I view this as a low-level API, so I think the existing behavior is likely fine. Internally we're using a small wrapper on-top of `rack.after_reply` to handle exceptions, reporting, metrics, etc. There's also [maybe_later](https://github.com/testdouble/maybe_later) which is a gem that implements similar logic.

I'm not familiar with how Puma handles dead connections, will it restart the connection or replace it so there's no impact on the running application?

jeremyevans: Agree that the existing behavior is fine as this is a low-level method. This still needs a decision on naming from @tenderlove (or another committer).

I don't think option 2 makes sense.  What would a rescue around all callables do?  Maybe just allow skipping the webserver error logging?

A rescue around each callable (option 3) has some utility, in that you can add your callback and be fairly sure it will run even if another callback fails.  

Note that if want to standardize different behavior (that the server must rescue all/each callable), then I don't have an objection to changing the `env` key.  In that case, it would be best to pick a `env` key that wasn't already used, so that we don't make existing webservers violate the spec.

ioquatix: > A rescue around each callable (option 3) has some utility, in that you can add your callback and be fairly sure it will run even if another callback fails.

I personally think this is the safest option.

Regarding option 2, the reality is servers don't do anything at all and it looks to be breaking (e.g. Puma) internal request handling. So at least stating that; "If a callable fails, it won't break the entire connection" is something we should be clear on in the spec IMHO if we opt to go in that direction.

ioquatix: > I'm not familiar with how Puma handles dead connections, will it restart the connection or replace it so there's no impact on the running application?

It looks like a server bug as the error is totally unexpected and basically seems to cause strange handling of the connection internally.

BlakeWilliams: > Regarding option 2, the reality is servers don't do anything at all and it looks to be breaking (e.g. Puma) internal request handling. So at least stating that; "If a callable fails, it won't break the entire connection" is something we should be clear on in the spec IMHO if we opt to go in that direction.

I agree. I think if we're going in that direction providing a guarantee that your `callable` will run is a logical next step to that.

I could imagine there being some cases where you'd want to stop the callable chain early, specifically since it's taking up worker time but I think building abstractions like the `maybe_later` gem will largely cover that. The question I have is where should guarantees about each callable running live?

* Should we define the spec to declare that `callable`s will all be run, even if previous `callable`s fail and rely on abstractions like `maybe_later` to enable short-circuiting running `callable`s?
* Should we define the spec to declare that `callable`s that raise won't break the server, and rely on abstractions like `maybe_later` to enable guarantees that each of your callables will run?

Alternatively, we could go with #3 (the first option above) and have a special error class that when raised will prevent subsequent callables from running.

jeremyevans: > Alternatively, we could go with #3 (the first option above) and have a special error class that when raised will prevent subsequent callables from running.

All of the rack SPEC is designed so you do not need to include rack the library.  So an exception class defined by rack the library would not work.  One way around this would be to say you can throw `:skip_response_finished`  inside a `rack.response_finished` entry to skip the remaining entries.  This is easy for servers to implement via `catch` around the code that calls each of the entries.

BlakeWilliams: > One way around this would be to say you can throw `:skip_response_finished` inside a `rack.response_finished` entry to skip the remaining entries. This is easy for servers to implement via `catch` around the code that calls each of the entries.

Ah I see, thank you for the context. 

That general approach makes sense to me. Maybe instead of throwing, we could rely on returning `:skip_response_finished` from the `callable`?

ioquatix: I don't personally believe that there is any good use case where one callable should be able to cancel subsequent callables as this seems order dependent and pretty fragile.

I'd be okay with exceptions terminating all remaining callables as a performance/safety middle ground.

tenderlove:
I like the name `rack.response_finished`.  It makes sense that the callbacks are an array.  I'm not really in favor of being able to skip callbacks, and if a callback raises an exception I'd still like the rest of the callbacks to be called.

The usecase I think of is closing database connections after a request is finished.  If some callback said everything should be skipped, it might mean leaked connections, and that wouldn't be a good thing.

Open ended question: do we want to do anything with the return value of the callbacks, and do we want to pass anything to them?  Or should we just punt that for later?

@BlakeWilliams I think this is a great feature and I'd like to see it in Rack 3.0 (If you have the time of course!!)

matthewd:
> Open ended question: do we want to do anything with the return value of the callbacks, and do we want to pass anything to them?

I doubt there's much to be done with the return value, but passing `env` might allow some users to avoid needing a per-request allocation, instead of having to rely on closures? 🤔

jeremyevans:
> I doubt there's much to be done with the return value, but passing `env` might allow some users to avoid needing a per-request allocation, instead of having to rely on closures? thinking

I think this is a good idea, but I recommend it be called with both `env` and the response array.  We could even add that to Lint, checking that if it is called, those are what it is called with.  Since that is different from how `rack.after_reply` works in Puma and Unicorn, I fully support using `rack.response_finished` as the key in that case.

tenderlove:
> I doubt there's much to be done with the return value

Ya, probably not.  I guess I was thinking it might be of use if callbacks knew the status of previous ones.  But it's probably best not to share that info.

> I think this is a good idea, but I recommend it be called with both `env` and the response array. We could even add that to Lint, checking that if it is called, those are what it is called with. Since that is different from how `rack.after_reply` works in Puma and Unicorn, I fully support using `rack.response_finished` as the key in that case.

Passing the env hash makes sense.  I like the idea of passing the response array because the callbacks would probably want to know the response code and possibly the headers.  The only thing I'm worried about is passing a body which may have already been iterated over and is no good (but maybe that doesn't matter).  I'd guess the callbacks wouldn't be concerned with what's inside the body, so it makes me wonder why pass it in the first place (besides that it's easy to implement).

BlakeWilliams:
> @BlakeWilliams I think this is a great feature and I'd like to see it in Rack 3.0 (If you have the time of course!!)

Rad, I'll make some time for it this week and try to get some changes pushed up shortly. I'm 👍 on passing the `env` to the handlers, especially now that we're explicitly diverging from the current implementations in puma/unicorn.

I had similar thoughts about passing `body` so I'll start by just passing `env` to the handlers, but happy to update the PR to pass the response array too if we decide that's the way we want to go.

ioquatix:
One part of this that I want us to consider is how virtual requests work. I've used these in the past and I won't be the only one.

Essentially, it looks like multiplexing within a Rack application - the ability to inject one or more new request into the rack middleware. These requests may append to the `rack.response_finished` callback array.

For such a reason, I believe this should be a non-optional feature of Rack 3 as it will make it easier for us to have a guarantee for users that such a thing exists which makes usage easier (e.g. in virtual requests).

This also raises the question about arguments - whether `env` should be the original request `env` or the final request `env`. We should be clear about when these callbacks should be invoked and how they work in relation to other in-flight sub-requests. My current expectation is they should be called after the body is read, but we might want to say "Even if a request is cancelled, (i.e. the client terminates the request before the full response is written), the callback will still be invoked.

The reason why this matters is because multiple requests on the same thread, with ActiveRecord, adding multiple `response_finished` entries might cause chaos. So we should consider carefully how this fits together.

BlakeWilliams:
I just pushed an update to the spec and linter to support `ENV` being passed as an argument, as well as explicitly supporting objects that respond to `#call`.


> Essentially, it looks like multiplexing within a Rack application - the ability to inject one or more new request into the rack middleware. These requests may append to the `rack.response_finished` callback array.

What does this look like in a web server? Here's the non-multiplexed version might (very roughly) look like once implemented:

```ruby
# unicorn code, but simplified a lot
def process_client(client)
  env = @request.read(client)
  env["rack.response_finished"] = []

  status, headers, body = @app.call(env)
  # handle closing body/client, etc
rescue => e
  handle_error(client, e)
ensure
  env["rack.response_finished"].map { |c| begin; c.call(env); rescue; end }
end
```

Looking at the above code and thinking about it in terms of multiplexing, I have a few questions:

* Should virtual/multiplexed requests know they're multiplexed or be treated different? If so:
  * Do they need a way to check if they're part of a virtual request or a real request?
  * Do we need to pass the `env["rack.response_finished"]` value from one request to the next? (this has concurrency concerns)
  * Is it reasonable to expect applications to require code changes to support multiplexing? (e.g. not duplicating `rack.response_finished` array entries like closing the database connection twice)

The concern about virtual requests/multiplexing is really valid, but I wonder if `rack.response_finished` is the correct abstraction for that specific case, or if there's another abstraction required, like `rack.multiplexed_response_finished`. I imagine that `rack.response_finished` being defined as "run once per `env`/`@app.call`" (with better terminology ofc) would still be valuable for both cases.

> This also raises the question about arguments - whether `env` should be the original request `env` or the final request `env`.

I think some of the questions about would help inform us here, assuming this is the right abstraction for the multiplexing scenario. I could also see the spec defining it somewhat like "The callable will be passed the `env` hash it was originally defined on".

> My current expectation is they should be called after the body is read, but we might want to say "Even if a request is cancelled, (i.e. the client terminates the request before the full response is written), the callback will still be invoked.

I think that's good to call out in the spec, I'll add that. 👍

ioquatix:
Where we ran into this problem before is apps which create virtual requests, and then call `rack.hijack` and expect `env['rack.hijack_io']` to be defined. But there is no way for the server to know what env should be updated, so in practice updating env was a terrible idea since `env` can be cloned/duped/completely different from what the server initially created.

e.g.

```
class MyWebServer
  def handle_request(client, request)
    env = make_rack_env(client, request)
    response = @app.call(env)
    
    env['rack.response_finished'].each do |callback|
      callback.call(env) # and maybe response
    end
  end
end

class MyApp
  def call(env)
    localizations.each do |locale|
      response = @app.call(make_fake_env(env, locale))
      if response[0] != 404
        return response
      end
    end
  end
end
```

This is a very simple example, but shows some hypothetical virtual requests. It seems like we need to either state:
(1) Your callbacks might not be called or
(2) A user is responsible for calling callbacks in some cases.

Not sure I really have any deep insight into the correct way forward here.

matthewd:
I think this setup is safe, unlike the `hijack_io` one, because either:
* the middleware has passed along the server-supplied `response_finished` array (and the server will handle any callbacks there added), or 
* the middleware has deliberately supplied its own `response_finished` (in which case it's taking on the obligation to comply with the optional spec: as a "server" [which it is acting as] it's optional that you define the key, but if you've defined the key, it's not optional that you process its contents), or 
* the middleware is supplying an `env` without any `response_finished` (in which case the app believes/knows it's running inside a server that doesn't support the feature, and will do its thing in-line via a BodyProxy or whatever*).

The unique trouble with hijack was that the server was trying to retroactively inject a new key into the `env` itself, which breaks as a mechanism the moment any middleware passes along an `env` that is not the exact Hash instance it originally received.

The one notable consequence of the above reasoning is that the application cannot be certain that a `response_finished` will be run before the next request occurs on the current thread/fiber/whatever: if a multiplexing middleware passes `response_finished` through to the outer server, then all subrequests' callbacks will occur together at the end of the main request. I think I would claim that is the _better_ result, though. If something must occur at an exact time (such as AR connection cleanup) you should run it yourself; to me `response_finished` is a better fit for something that should happen (to the best of the server's ability, processes do sometimes die, etc) at some point in the near future _when the client is not waiting_. In the multiplex case, that last criteria is only satisfied after all the subrequests have been handled, so it follows that that's when all the callbacks will occur. (If something AR-cleanup-shaped was slow enough to want to defer, an application could choose to use in-line execution to detach the connection from the current thread, but then leave it in a separate pool to be fully slow-cleaned later by the callback.)

\* as a follow-up to this spec change, Rack::BodyProxy should probably grow an API that means "run this block after the response, using `response_finished` if available but otherwise via BodyProxy"

ioquatix:
@matthewd those are excellent points.

lib/rack/constants.rb
```diff
@@ -52,6 +52,7 @@ module Rack
   RACK_RECURSIVE_INCLUDE              = 'rack.recursive.include'
   RACK_MULTIPART_BUFFER_SIZE          = 'rack.multipart.buffer_size'
   RACK_MULTIPART_TEMPFILE_FACTORY     = 'rack.multipart.tempfile_factory'
+  RACK_RESPONSE_FINISHED              = 'rack.response_finished'
```
ioquatix, line 55: @BlakeWilliams @tenderlove looking at other keys, should this be `rack.response.finished` or `rack.response_finished`? We don't use `rack.request_form_input` but `rack.request.form_input`.

Is there any precend for appending `_callbacks`? I don't mind either way, just calling it out.

ioquatix:
I was thinking about this a bit more and I do think we should make handling this field mandatory Rack 3 feature because it ultimately makes it much easier to deal with.

However, unlike `rack.hijack` or other metadata, I can't see why there is value having this as part of the request. Shouldn't it be part of the response headers?

What's the justification for having this as part of `env` and not part of `response[1]` (response headers).

If it's part of `env`, it means the array must always be allocated even if never used.

matthewd:
Hmmm. Putting anything that's not an actual header value into the response headers feels like a pretty big / far-reaching change, both conceptually and in impact to anyone who might be reading that hash.

ioquatix:
I was under the impression we already put rack specific metadata in the response headers. Let me check.

ioquatix:
I'm going to manually merge this.

jeremyevans:
I also don't understand why this was closed and #1932 was opened.  As far as I can see, the name `rack.closed` was never discussed prior to this being closed.  I think `rack.after_reply`, `rack.response_finished`, `rack.response.finished`, and `rack.response.finished_callbacks` are all better names than `rack.closed`.  Are there any other differences between this pull request and #1932 other than the name?

ioquatix:
I started playing around with real implementations and realised we are missing a critically useful part which is the ability to feed any client errors back to the callback. I actually don’t mind what we call it but I after actually trying it I want something better. Semantically this isn’t about a response or reply or even when it’s finished. It can also be when there is an error and for internal or virtual requests it has nothing to do with an actual response or reply.

After investigating the naming scheme, the `rack.response` prefix is inconsistent with `rack.request` usage. So I want to propose `rack.closed` which is semantically the same as `body.close` except it is always called even if there is a client error which is given as an optional argument. This is fundamentally advantageous over the existing approach, since before there was no way to tell if a client disconnected or the response failed for some other reason. In the case of an error I think `after_reply` or `response.finished` is misleading.

ioquatix:
Also if it was up to me I’d seriously consider adding an argument to `body.close` so we could pass through an error. This can be critically
Important for some usages like we sockets and streaming.

ioquatix:
By the way I don't want to seem like I'm trying to make a unilateral decision, but I can't add commits to this PR. I was planning on merging this PR largely as is, but found that in practice it's not sufficient. Anyway, feel free to re-open this PR if we want to discuss it or work on this design further, that's totally fine, I just didn't feel comfortable asking @BlakeWilliams to continue working on something when I don't really know what I want to propose yet - I need to spend some time working through the implementation in Falcon too.

jeremyevans:
Reopening as #1932 was closed.

ioquatix:
There has been a lot of discussion on this.

I'm a strong believer that this needs to be compatible with the existing semantics where possible.

The main use cases we have been informed of is:

1. GitHub using it for flushing metrics after the response was sent.
2. Rails releasing resources after a response was generated/sent.

Both use cases would be satisfied by an optional `rack.after_reply` which could be relatively efficiently implemented by:

```ruby
module Rack
  def self.Ensure(env, app, &block)
    if after_reply = env['rack.after_reply']
      after_reply << block

      return app.call(env)
    else
      begin
        response = app.call(env)
        response[2] = BodyProxy.new(response[2]) {block.call}
      rescue => error
        block.call(error)
        raise error
      end
    end
  end
end

response = Rack::Ensure(env, @app) do |error|
  # Best effort to be called, error may be set if the response failed to be sent.
end
```

The biggest limitation of this is the inability to provide error information to the callback when the `BodyProxy` semantic is used.

I would hypothesize that there is little performance difference between the use of `BodyProxy` and `rack.after_response`. It would be nice to have some evidence.

ioquatix:
If we are willing to bend the rules a little bit:

```ruby
response[2] = BodyProxy.new(response[2]) {block.call($!)}
```

could be acceptable / best effort.

ioquatix:
I couldn't help but flesh this out. Not convinced by any of the method names or `rack.after_reply` itself, but I think this is a reasonable semantic. Zero-cost if not in use.

```ruby

module Rack
  # Usage:
  #   # In config.ru:
  #   use Rack::AfterReply # unless feature?(:after_reply)
  #
  #   # In application:
  #   response = Rack::AfterReply.call(env, @app) do |error|
  #     # Best effort to be called, error may be set if the response failed to be sent.
  #   end
  class AfterReply
    def initialize(app)
      @app = app
    end

    RACK_AFTER_REPLY = 'rack.after_reply'

    # Expected to be called by the server.
    def self.apply(env, error)
      return unless after_reply = env.delete(RACK_AFTER_REPLY)

      callbacks.each do |callback|
        begin
          callback.call(error)
        rescue => callback_error
          env['rack.errors'].puts(callback_error.full_message)
        end
      end
    end

    # Provides a defacto implementation for servers that don't support it.
    def call(env)
      env[RACK_AFTER_REPLY] ||= []
      
      @app.call(env)
    rescue => error
    ensure
      self.class.apply(env, error)
    end

    # Expected to be used by the middleware/application:
    def self.call(env, app, &block)
      if after_reply = env[RACK_AFTER_REPLY]
        after_reply << block
        return app.call(env)
      else
        begin
          response = app.call(env)
          response[2] = BodyProxy.new(response[2]) {block.call($!)}
          return response
        rescue => error
          block.call(error)
          raise error
        end
      end
    end
  end
end
```

BlakeWilliams:
> The main use cases we have been informed of is:
> 
> 1. GitHub using it for flushing metrics after the response was sent.
> 2. Rails releasing resources after a response was generated/sent.

GitHub is also using it in a few other places, like persisting some cache values at the end of a request which ended up making some of our pages a good bit more performant iirc. Puma has also been utilizing this since 2015 for logs using the `commonlogger` here: https://github.com/puma/puma/commit/537bc21593182cd9c4c0079a3936d05b1f91fe14#diff-aaa1392bef19f789281e5b8c8650529f5caf26a139df7a32631da6592dbe25d3R38

> The biggest limitation of this is the inability to provide error information to the callback when the `BodyProxy` semantic is used. I'm not sure if I care much about this, but I think it's a big reason why servers might choose to implement `after_reply`.

I'm not quite convinced that we need error information in the callbacks, at least based on the use-cases I've run into. Typically by the time you've created the callback I think it's reasonable to expect that the callback will run, regardless of error status. e.g. If we want to flush metrics at the end of a request, or close some kind of connection to a service, it doesn't matter if the client closed the connection.

> I would hypothesize that there is little performance difference between the use of `BodyProxy` and `rack.after_response`. It would be nice to have some evidence. If there is no performance difference, does it reduce the value of this specification?

I think it would reduce some of the value, but there's also value in having an explicit API for this behavior that's at least a bit more ergonomic than using `BodyProxy`. It doesn't feel great that you have to replace a result in the response array, and it limits where you can create callbacks (for better or worse).

Also, when I originally wanted to implement the stats flushing behavior I used `Rack::Events` which I believe uses some form of `BodyProxy` under the hood.  Unfortunately I found that it would leave the request open until the provided `on_finish` callbacks completed. That means that in some scenarios, it's possible that `BodyProxy` might not provide the functionality you'd expect to get from it. The docs seemed to indicate that it would run after the response, despite it not doing so in at least our configuration:

> The webserver has closed the response, and all data has been written to the response socket. The request and response object should both be read-only at this point. The body MAY NOT be available on the response object as it may have been flushed to the socket.

BlakeWilliams:
> I'm a strong believer that this needs to be compatible with the existing semantics where possible, and I'm no longer convinced this should be a required feature. I changed my opinion after implementing it in Falcon and reviewing existing use cases.

For what it's worth, both Puma and Unicorn support this functionality today, which results in a pretty large number of applications having `rack.after_reply` available, but as an unofficial extension (disclaimer, I implemented the version in Unicorn).

The primary reasoning I had for opening this issue and this PR is that it would be ideal if we could make the implementation of those two webservers an official spec, since it is providing value to existing applications. By making it required/official, it means applications depending on that behavior are now able to more easily swap between server implementations if they desired (barring other issues, like thread safety) and have a slightly more consistent feature-set available. I'd find it unfortunate if we decided not to move forward with some variation on this functionality, but it would still exist in Puma/Unicorn and be usable as-is going forward.

BlakeWilliams:
oh, there was also this gem, [`maybe_later`](https://github.com/testdouble/maybe_later) that I remembered that also utilizes this functionality. 🙂

ioquatix:
> The primary reasoning I had for opening this issue and this PR is that it would be ideal if we could make the implementation of those two webservers an official spec, since it is providing value to existing applications. By making it required/official,

That's a totally reasonable position. However, at best, I feel this can be optional feature with a standard interface, and I don't think there is any point to introduce it in isolation.

I would say that the value for me would be, defining a standard (optional) interface in Rack would allow us to update `Rack::Events` to take advantage of this interface. The resulting performance or robustness improvement should be testable and then servers that opt into that provide that advantage.

The counter point is, if we can't show a significant advantage, I'd prefer we don't introduce such a feature, because every optional feature or alternative way of doing things adds an extra dimension of complexity to applications and server implementations.

Things like `Rack::Event` callbacks not firing at the right time (e.g. keeping the request open) seem like a bug of the implementation, not the interface. I'd suggest those should be fixed independently of this proposal (and would welcome such fixes).

Regarding handling errors, it can be useful to know if the request was sent to the client completely or not. One simple use case would be logging the status of the response (sent successfully or not), another would be detecting unusual client behaviour (e.g. `rack-attack`), if you are streaming something from upstream (e.g. a HTTP/2 client), knowing how to close the stream (with an error or not) can be important. If it's easy to implement, I'd recommend we do it.

jeremyevans:
From the long discussion we had with @tenderlove, it is obvious both he and I are in favor of including `rack.after_reply` or similar as an optional feature in the Rack 3 SPEC.  Since Puma and Unicorn already support that name, I recommend we keep it.  @tenderlove mentioned he liked the name `rack.response_finished`, so I'd like him to reply here and let us know if he is OK with `rack.after_reply` (the de facto standard), or whether he would like to switch to `rack.response_finished`.

If we keep `rack.after_reply`, we should keep the same API standardized by Puma and Unicorn, not calling the block with any arguments.  If we switch to `rack.response_finished`, then we can consider passing arguments to the block, such as `env`, the response array (or just the status and headers), and a possible exception. 

The question at this point is not whether this feature will be in Rack 3, but what key will be used for it, and if switching to `rack.response_finished`, what arguments will be passed to the callbacks.

ioquatix:
If we introduce an optional interface, but it's impossible to achieve the same behaviour with non-optional interfaces (i.e. `BodyProxy`), we haven't introduced an optional feature, we've bifurcated the specification and server implementations as applications/frameworks will have no choice but to either 1/ require the optional feature to function (no equivalent non-optional feature) or not use it at all. That means servers which don't support the optional feature won't be able to support some frameworks. I'm completely against any approach that leads to that outcome.

jeremyevans:
My understanding is both @tenderlove and me are OK with it being optional or required.  So if you would prefer it be required and not optional, we can make it required.

tenderlove:
> An example of this would be GitHub's own application which apparently depends on this optional interface. Therefore, any server which does not implement it cannot host GitHub's application, because there is no abstraction which allows them to use "either rack.after_reply if it exists or BodyProxy".

GitHub's own application also isn't thread safe or fiber safe, meaning any webserver that uses either of those models can't host GitHub's application out of the box.  In the Rack project, we should add specifications that are generally useful for _users_, and _can_ be implemented by webservers.  Whether a webserver chooses to implement an optional spec or not is up to them.  But if users find that particular feature (be it threads, fibers, processes, or an `after_reply` callback) to be useful, it makes sense for them to choose a webserver that implements that feature.

> That means servers which don't support the optional feature won't be able to support some frameworks [correctly]. I'm completely against any approach that leads to that outcome for what is positioned as "optional".

Making this feature required makes sense to me from a webserver compatibility perspective.

> If we keep `rack.after_reply`, we should keep the same API standardized by Puma and Unicorn, not calling the block with any arguments. If we switch to `rack.response_finished`, then we can consider passing arguments to the block, such as `env`, the response array (or just the status and headers), and a possible exception.

I think we should keep the new name and pass `env`, the response status and headers, and a possible exception.  I like the idea of keeping the old name / interface so that nobody has to change their code, but that API was unofficial, adding the additional information makes sense, and if we're updating the spec we should just add this nice stuff.

@BlakeWilliams sorry to keep dragging on with this, but could you update the spec such that the callbacks take 4 params, `env`, `status`, `headers`, and `error`?  The webserver should pass 4 parameters, but `error` will usually just be `nil`.

ioquatix:
I would personally like to see `Rack::Events` adopt whatever hook we provide, and I don't think that's unreasonable, since we should probably make it easy for users to consume optional features with a fallback to the minimum level of compatibility.

BlakeWilliams:
> @BlakeWilliams sorry to keep dragging on with this, but could you update the spec such that the callbacks take 4 params, `env`, `status`, `headers`, and `error`? The webserver should pass 4 parameters, but `error` will usually just be `nil`.

No worries! Happy to update this to match that signature. I do wonder if would make sense to avoid passing `headers` directly and guide folks towards putting any relevant values into `env` though.

> it would be nice to avoid having to retain `env` and `headers` as these can be a significant source of memory usage/garbage.

Is there a benchmark that could measure how much extra memory is in-use to quantify the impact? I'd be curious to know what the baseline usage is vs the usage when headers and env are retained for each request.

ioquatix:
@BlakeWilliams yes I benchmarked this extensively and it's a non-trivial amount of data per request. I don't have the benchmarks in front of me but I'll endeavour to share this with you later this week.

To give you some rough numbers (my memory is not great because it was like 2-3 years ago I did this), it was something like 20-30GiB of memory per client and server handling 1 million connections (so about 40-50GiB in total), and IIRC it was on the order of 50-100 objects per connection... so it doesn't take much to, say double the amount of memory used. An `env` with 20-30 string instances, a response `headers` with another 5-10 objects... it all adds up.

jeremyevans:
> > @BlakeWilliams sorry to keep dragging on with this, but could you update the spec such that the callbacks take 4 params, `env`, `status`, `headers`, and `error`? The webserver should pass 4 parameters, but `error` will usually just be `nil`.
> 
> No worries! Happy to update this to match that signature. I do wonder if would make sense to avoid passing `headers` directly and guide folks towards putting any relevant values into `env` though.

One usage for such a callback is for writing an access log, which will typically include the `Content-Length` header.  I don't think we should push people to duplicate response header values in `env`.

> > it would be nice to avoid having to retain `env` and `headers` as these can be a significant source of memory usage/garbage.
> 
> Is there a benchmark that could measure how much extra memory is in-use to quantify the impact? I'd be curious to know what the baseline usage is vs the usage when headers and env are retained for each request.

If we do want to benchmark this, we should pick a benchmark that aims to reflect typical production workloads, such as Railsbench.

Note that if the callback doesn't receive `env` and `headers`, in a lot of cases, a per-request proc will be allocated, which would likely result in the implicit retention of `env`, `headers` (due to them being reachable in some way from the scope of the proc), and likely additional objects, and add more overhead due to the proc allocation.  So the idea that passing these variables to the callback will necessarily increase object retention and memory usage seems dubious to me.  There are probably cases were it would result in additional retention, but I doubt it would for most production uses.  There are definitely cases where passing `env` and `headers` as arguments would decrease both object retention and memory usage, by avoiding the creation of per-request procs.

ioquatix:
If this feature is not required, thus by implication not required on every request, if I see that the request likely to be long lived, I can avoid providing the callback array and avoid the memory costs, it's acceptable. I did the same thing for rewindability in Rack 2.x - I sniffed content type and only did it on requests where it was going to be used to avoid the overhead on every request.

I think you'll have a hard time showing a performance impact on something like Rails, because it's lack of scalability inherent in the design, e.g. puma only handling 8-16 simultaneous requests. Falcon is designed to handle 10,000 or more where the impact is felt more heavily.

So, introducing this feature as optional sets the bar much lower and is probably more acceptable since server can opt in depending on the situation it expects.

ioquatix:
Okay, this was merged in https://github.com/rack/rack/pull/1952.

Issue #1148, opened 2017-03-07
Proposal for dealing with body streams differently in 2017


thesmart:
I'm curious about Rack's requirement for rewindable input.  I'm bothered that Rack will buffer a read-once stream to disk just so `rewind` is available.  Modern cloud-based NAS throughput imposes limitation on how well this feature scales.  I was wondering... 

**Is it a good time for future versions of Rack 2.x to drop disk buffering as a default behavior, and move it to an optional Rack middleware specifically for handling "large" bodies / file uploads?**

I'd like to propose an alternative strategy for handling "large" bodies.

(1) Provide a configurable post body buffer limit `max_body_buffer_length`. If `content_length.present? && content_length <= max_body_buffer_length` then read it ASAP (ideally non-blocking) into `rack.input` as `StringIO` or other.  Else, if `content_length.nil? || content_length > max_body_buffer_length`, then do not read the stream.  This would mean that no chunked encodings are handled by default, and no buffering to disk for streams larger than the configured limit.

(2) When `request.params` or `request.body` is accessed, if body was not parsed due to length or chunked encoding, report a warning message that can be disabled.

(3) Provide automatic "large body" parsing as a middleware module plugin that can be optionally installed.

tenderlove:
> Is it a good time for future versions of Rack 2.x to drop disk buffering as a default behavior, and move it to an optional Rack middleware specifically for handling "large" bodies / file uploads?

Yes.  I plan on revising the spec for Rack 3 (master should be 3).  It will remove the requirement for a rewindable body.  If you have time to do that work, then please start sending PRs (starting with updates to the SPEC file).  Thanks!

ioquatix:
I recently ran into this limitation when implementing streaming requests/responses with `async-http` and the rack-compatible server `falcon`.

With HTTP/2, one can stream a request/response indefinitely, and the requirement for supporting `#rewind` is a bit of a pain. In addition, `read(length)` also requires buffering within individual HTTP2 DATA frames which makes things slightly more complicated.

I'd suggest that the `rack.input` API is good, but 1/ drop `#rewind` and limit `#read` to having no arguments, i.e. it just returns the next available chunk of data from the stream. This would suit HTTP/2 perfectly.

ioquatix:
I would be happy to work on the SPEC file but do you want to have a discussion here first? What, if any, progress/decisions have already been made?

ioquatix:
[...]

PR #1803, opened 2022-02-05
Add Rack::RewindableInput::Middleware

jeremyevans:
This will automatically wrap rack.input with Rack::RewindableInput,
for compatibility with middleware and applications that expect
rewindable input.

Related to #1148, but this does not contain any SPEC changes. It's
possible for servers targetting Rack 2 compatibility to use this
middleware to implement the compatibility.


ioquatix (approved):

ioquatix:
I personally think we should move this to `rack-contrib` but let's merge this.

PR #1808, opened 2022-02-15
Add default app which returns 404.

ioquatix:
I believe this is a very convenient default which many applications end up implementing. I don't know if there is a good reason why we shouldn't do this.

cc @leahneukirchen since this was your original design.

### DIFF
--- a/lib/rack/builder.rb
+++ b/lib/rack/builder.rb
@@ -107,17 +107,19 @@ module Rack
       return builder.to_app
     end
 
+    DEFAULT_APP = lambda{|env| [404, {}, []]}
+
     # Initialize a new Rack::Builder instance.  +default_app+ specifies the
     # default application if +run+ is not called later.  If a block
     # is given, it is evaluated in the context of the instance.
-    def initialize(default_app = nil, &block)
+    def initialize(default_app = DEFAULT_APP, &block)
       @use, @map, @run, @warmup, @freeze_app = [], nil, default_app, nil, false
       instance_eval(&block) if block_given?
     end
 
     # Create a new Rack::Builder instance and return the Rack application
     # generated from it.
-    def self.app(default_app = nil, &block)
+    def self.app(default_app = DEFAULT_APP, &block)
       self.new(default_app, &block).to_app
     end
 
--- a/test/spec_builder.rb
+++ b/test/spec_builder.rb
@@ -26,6 +26,13 @@ class NothingMiddleware
 end
 
 describe Rack::Builder do
+  it "has a default app" do
+    app = Rack::Builder.app
+    
+    response = Rack::MockRequest.new(app).get("/")
+    response.status.must_equal 404
+  end
+
   def builder(&block)
     Rack::Lint.new Rack::Builder.new(&block)
   end
@@ -225,12 +232,6 @@ describe Rack::Builder do
     Rack::MockRequest.new(app).get("/a").must_be :server_error?
     Rack::MockRequest.new(app).get("/b").must_be :server_error?
     Rack::MockRequest.new(app).get("/c").status.must_equal 200
-  end
-
-  it 'complains about a missing run' do
-    proc do
-      Rack::Lint.new Rack::Builder.app { use Rack::ShowExceptions }
-    end.must_raise(RuntimeError)
   end
 
   describe "parse_file" do


jeremyevans:
If we do this, I think we should have `content-type` and `content-length` headers.  Length will obviously be 0, and type can be text/html or text/plain.

ioquatix:
What is the motivation for adding those headers?

jeremyevans:
Responses should in general have content-length unless you are streaming them (chunked encoding in HTTP 1.1).  I'm not sure how all browsers will handle 0-byte responses without a content-type (maybe they will download them as 0-byte files?), so it seems safer to specify a text based content type so the browser will just show an empty page.  I'm not positive the headers are needed, but it seems safer to include them.

ioquatix:
Falcon ignores user supplied content length. Streaming is done based on the shape of the body. It’s too unreliable to depend on content length and in addition it doesn’t make much sense for HTTP/2 so in most cases the header has to be stripped. So  not sure there is any point to include it by default.

Regarding content type, since there is no content, and it’s a 404, most browsers will display their own error, or a rack middleware will do the same job. I’m not sure content type is really useful in this context but I want to check what other servers do.

leahneukirchen:
Are empty 404 messages rendered in a useful manner these days? Else I think it's not a sensible default.

ioquatix:
I can check this.

ioquatix:
Chrome (error page):

image

Safari (blank page):

image

matthewd:
Perhaps a new instance method would give more of a middle ground?

That would retain the current behaviour where `run` [or an app argument] is required, and accidentally forgetting to include one is an error -- which seems somewhat useful to me, lest a caller get so distracted thinking about which middlewares to include that they forget to include their app, and then get confused by seeing nothing but terse 404s for all URLs.

... but would also provide a trivially-spelled "I've handled everything I care about; drop the rest", if that is a common need?

ioquatix:
The point is, an empty `config.ru` should just return 404s instead of being an error. A lot of web frameworks are basically just middleware with no final "application". So you end up with a placeholder `run ...` at the end of your middleware chain. This just basically makes `run` optional, so if people only use middleware, it works as expected. Most stacks that work this way have a error handler very early on in the stack which maps error codes into actual error pages so in practice this is a good design and will reduce the complexity of defining apps and using `Rack::Builder`. Defaulting to 404s is basically what every other web server does AFAIK.

matthewd:
> The point is, an empty `config.ru` should just return 404s instead of being an error

Disagree. An empty `config.ru` is a mistake, and thus should error.

> A lot of web frameworks are basically just middleware with no final "application"

I see Sinatra can work in a middleware mode, where it can then `forward` some requests to the nested app, but at a quick glance that's not true of Rails, Hanami, Roda, Grape, or Rack::Lobster. 

(I think I would argue that, in an application-backed web application like Rack [as distinct from a filesystem-backed one, where no matching file = 404], a request making it all the way through and past your mentioned friendlier error handler, is 500-worthy.)

I'm not particularly invested in this point, but yeah.. I don't think my experience with app definitions supports a widespread need for a zero-code default handler.

jeremyevans:
> > The point is, an empty `config.ru` should just return 404s instead of being an error
> 
> Disagree. An empty `config.ru` is a mistake, and thus should error.

I agree with @matthewd on this.  If the purpose of this change is to allow an empty `config.ru` to work, we should reject it.  The idea of an instance method to allow easily overriding the behavior seems like a good middle ground to me (keeping the current behavior as default).

> > A lot of web frameworks are basically just middleware with no final "application"
> 
> I see Sinatra can work in a middleware mode, where it can then `forward` some requests to the nested app, but at a quick glance that's not true of Rails, Hanami, Roda, Grape, or Rack::Lobster.

I'm not aware of any web framework that runs as middleware with no final application by default.  Maybe there is one, but I highly doubt there is a lot.

To clarify, Roda supports running as middleware (using the middleware plugin it ships with).  This is what allows the Rodauth authentication framework (built with Roda) to run on all Ruby web frameworks.

leahneukirchen:
I don't have any stakes in this really, but I also think forgetting `run` should be an error.

jeremyevans:
> Am I missing something here? Is this the default model we want to encourage? What's the point of middleware if all major frameworks are building applications that can't be composed?

What you are missing is that the point of middleware is that any Rack application can use middleware, not that all Rack applications should be middleware.  If you look at middleware that ship with Rack, in most cases they aren't handling requests by themselves, but wrapping the process of handling requests.

Middleware that handle a subset of requests, and call the next middleware with other requests, are fairly rare.  I don't think that's a bad thing.  I'm not saying it's a bad idea to implement middleware that does that, but I also don't think there is any reason Rack should encourage that design.

Note that if you want composable applications, you can use Rack::Cascade.  I think few people actually have a need for it.

If `run` is missing in a `config.ru`, I think almost all users would prefer us raising an error, since they probably forgot to use the application they want to use.  If we don't, almost all users will be submitting requests and getting 404s for all submitted requests, and it will take them a lot longer to debug the cause of the problem.

ioquatix:
Thanks for explaining that point of view @jeremyevans.

> Middleware that handle a subset of requests, and call the next middleware with other requests, are fairly rare.

There are middleware shipped with Rack which work this way.

- https://github.com/rack/rack/blob/main/lib/rack/conditional_get.rb
- https://github.com/rack/rack/blob/main/lib/rack/directory.rb
- https://github.com/rack/rack/blob/main/lib/rack/static.rb

`Rack::Static` is a fairly important middleware design and I think it's design has informed other similar implementations.

That being said, if everyone is against this change, I'll close the PR.

PR #1810, opened 2022-02-18
Correct handling of streaming body in `Rack::Response`.

ioquatix:
The current implementation does not correctly handle bodies that don't respond to `#each`.

### DIFF
--- a/lib/rack/response.rb
+++ b/lib/rack/response.rb
@@ -36,7 +36,7 @@ module Rack
     # Initialize the response object with the specified body, status
     # and headers.
     #
-    # @param body [nil, #each, #call, #to_s] the response body.
+    # @param body [nil, #each, #call, #to_str] the response body.
     # @param status [Integer] the integer status as defined by the
     # HTTP protocol RFCs.
     # @param headers [#each] a list of key-value header pairs which
--- a/lib/rack/response.rb
+++ b/lib/rack/response.rb
@@ -36,7 +36,7 @@ module Rack
     # Initialize the response object with the specified body, status
     # and headers.
     #
-    # @param body [nil, #each, #to_str] the response body.
+    # @param body [nil, #each, #call, #to_s] the response body.
     # @param status [Integer] the integer status as defined by the
     # HTTP protocol RFCs.
     # @param headers [#each] a list of key-value header pairs which
@@ -62,7 +62,7 @@ module Rack
         @length = body.to_str.bytesize
       else
         @body = body
-        @buffered = false
+        @buffered = nil # undetermined as of yet.
         @length = 0
       end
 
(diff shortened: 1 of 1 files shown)


jeremyevans (changes requested):
I think this is way too much backwards compatibility breakage for little gain.  My guess would be most users using Rack::Response are calling `new` with a string body.  One thing to understand is that Rack::Response is used for constructing a rack response array, it's not designed to be called with an already valid rack response array.

Can you change pull request so it's backwards compatible, and only adds the ability to handle of a streaming body (as indicated by the subject)?  It would be best to submit separate pull requests for the other parts (removal of `header`, not allowing string bodies, etc.), so they can be considered in isolation.

ioquatix:
> I think this is way too much backwards compatibility breakage for little gain. My guess would be most users using Rack::Response are calling new with a string body.

I understand. We can move that change to a separate PR, but it was already documented as being deprecated. If we aren't going to remove that feature in Rack 3 after it was deprecated in Rack 2, maybe we should un-deprecate it?

> One thing to understand is that Rack::Response is used for constructing a rack response array, it's not designed to be called with an already valid rack response array.

This is exactly what Rack::MockResponse does? I agree, the use cases are confusing and as a result the implementation has to handle both cases.

jeremyevans:
> > I think this is way too much backwards compatibility breakage for little gain. My guess would be most users using Rack::Response are calling new with a string body.
> 
> I understand. We can move that change to a separate PR, but it was already documented as being deprecated. If we aren't going to remove that feature in Rack 3 after it was deprecated in Rack 2, maybe we should un-deprecate it?

IMO, something is only deprecated if it raises a deprecation warning when used.  If the documentation only marks it as deprecated, it isn't really deprecated.  I don't think it's fair to users to mark something as deprecated in the documentation, but not issue a warning when used, and then remove it completely.  Deprecation warnings should always come first, outside of exceptional circumstances. I'm fine raising a deprecation warning in Rack 3.0 and removing in Rack 3.1 (assuming you are referring to the `header` alias).

ioquatix:
Based on your explanation, I think `Rack::Response` shouldn't break with streaming responses but it shouldn't directly support the construction of streaming responses either.

I personally think using a proxy object would make more sense for this use case, i.e. `Rack::ResponseWithEnumerableBody` or `Rack::ResponseWithStreamingBody` might be one possible design, with the user writing `Rack::Response.buffered` for the Array backed enumerable body and `Rack::Response.streaming(proc)` for streaming.

lib/rack/response.rb
```diff
@@ -36,7 +36,7 @@ def self.[](status, headers, body)
     # Initialize the response object with the specified body, status
     # and headers.
     #
-    # @param body [nil, #each, #to_str] the response body.
+    # @param body [nil, #each, #call, #to_s] the response body.
```
jeremyevans, line 39: This should probably be `to_str` and no be changed to `to_s`, since that's the method the code checks for.

jeremyevans (changes requested):
This looks good, and I'm OK with merging after one small documentation fix.

lib/rack/response.rb
ioquatix: Ohhh my bad, thanks for catching that.

ioquatix:
```suggestion
    # @param body [nil, #each, #call, #to_str] the response body.
```

jeremyevans (approved):

PR #1811, opened 2022-02-18
Remove ancient (and confusing/buggy) `Rack::File` compatibility wrapper.

ioquatix:


jeremyevans:
`Rack::File` was not properly deprecated in the current version.  I think we should add a non-verbose deprecation warning in Rack 3.0, and remove in Rack 3.1.

jeremyevans (approved):

PR #1812, opened 2022-02-19
Change SPEC to not allow uppercase header keys

jeremyevans:
Also, update CHANGELOG with other spec changes made since 2.2.

Implements #1592

External tests are broken because this is a huge breaking change.  If we are sure that we want this, we should disable the external CI jobs until the related projects have been fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,12 @@ All notable changes to this project will be documented in this file. For info on
 ### SPEC Changes
 
 - rack.input is no longer required to be rewindable.
+- rack.multithread/rack.multiprocess/rack.run_once are no longer required environment keys.
+- Response status must now be an integer greater than or equal to 100.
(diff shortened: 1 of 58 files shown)


ioquatix (approved):
There will be a lot of small details but I approve the general direction and think this is a great first step.

ioquatix:
Fixing external tests seems straight forward so I don't think this should be a blocker. However, I'd suggest we merge this and then clean up the remaining external issues. Once we confirm the direction, it won't be hard to sort out the downstream dependencies.

ioquatix:
Okay @jeremyevans all external tests are passing. Falcon was already 100% compatible :)

ioquatix:
@tenderlove do you have any reservations regarding this?

tenderlove:
> @tenderlove do you have any reservations regarding this?

No, I think this is great.  Thanks for doing this work @jeremyevans!!

PR #1817, opened 2022-02-23
Add deprecation warning to `Rack::Response#headers`.

ioquatix:
See discussion in https://github.com/rack/rack/pull/1810


jeremyevans (changes requested):
Should add `uplevel: 1`.  OK with merging after that change.

jeremyevans (approved):

PR #1818, opened 2022-02-23
Add deprecation warning to `Rack::Chunked`.

ioquatix:
`Rack::Chunked` is HTTP/1.0 version specific and a poor fit for Rack middleware. It should be an implementation detail of the server.


jeremyevans (changes requested):
Likewise, please add `uplevel: 1`.  OK with merging after that change.

jeremyevans (approved):

PR #1819, opened 2022-02-23
Use Rack::Headers for headers in Rack::Files

jeremyevans:
This allows for case insensitive access, allowing for greater
compatibility with code using Rack::Files.

My opinion is the backwards compatibility benefits outweigh the small
performance cost.

### DIFF
--- a/test/spec_files.rb
+++ b/test/spec_files.rb
@@ -73,6 +73,15 @@ describe Rack::Files do
     res["last-modified"].must_equal File.mtime(path).httpdate
   end
 
+  it "supports case-insensitive header access" do
+    res = Rack::MockRequest.new(files(DOCROOT)).get("/cgi/test")
+
+    path = File.join(DOCROOT, "/cgi/test")
+
+    res.must_be :ok?
+    res["Last-Modified"].must_equal File.mtime(path).httpdate
+  end
+
   it "return 304 if file isn't modified since last serve" do
     path = File.join(DOCROOT, "/cgi/test")
     res = Rack::MockRequest.new(files(DOCROOT)).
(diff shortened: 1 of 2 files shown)


test/spec_files.rb
```diff
@@ -73,6 +73,15 @@ def files(*args)
     res["last-modified"].must_equal File.mtime(path).httpdate
   end
 
+  it "supports case-insensitive header access" do
+    res = Rack::MockRequest.new(files(DOCROOT)).get("/cgi/test")
+
+    path = File.join(DOCROOT, "/cgi/test")
+
+    res.must_be :ok?
+    res["Last-Modified"].must_equal File.mtime(path).httpdate
```
ioquatix, line 82: Consider adding a comment here explaining that you are explicitly testing that the non-normalised upper case form of then header is being used.

ioquatix:
> This allows for case insensitive access

Do you mind explaining exactly what the problem is you are addressing with this PR, it's not immediately obvious to me.

However, it seems reasonable as I believe the cost of `Rack::Headers` would be dwarfed by any kind of file access.

jeremyevans:
My personal desire for this are it keeps Roda's tests passing.  In general, most testing of Rack::Files by current code would probably benefit from this.  I think the main production benefit would be middleware that operates on response headers that hasn't yet been upgraded to Rack::Files.

ioquatix:
Can you give a brief example of a failing test?

I don't fully understand why `Rack::Files` response headers being an instance of `Rack::Headers` is more compatible - is this because you are sub-classing `Rack::Files`?

jeremyevans:
I'm going to close this.  It's simple enough to work around this issue when testing, and this change is not necessary in production use if middleware are Rack 3 compliant.

PR #1821, opened 2022-02-24
Deprecate Rack::Auth::Digest

jeremyevans:
To only issue a single deprecation warning, combine all digest auth
related files into a single file.  Have all previous files use
require_relative to load the new file, so previous requires all
still work.

The reason for this combining of files is if you have 4 separate
files, you cannot be sure you warn for all possible usage unless
you have deprecation warnings in all four files, and that would
result in 4 deprecation warnings, which could be annoying.

The multipart parser was requiring auth/digest/params to
use a single method.  Inline that method into the multipart parser
to avoid the problem.


ioquatix (approved):
This looks fine to me but we might want to get input from other maintainers too.

jeremyevans:
No objections in 14 days.  I'm going to merge this now.  If another committer really wants to keep Auth::Digest, we can revert it.

Issue #1823, opened 2022-03-01
Eager loading mode?


casperisfine:
For context there's a discussion in Ruby's bug tracker about replacing the global constant cache by a more finely grained cache: https://bugs.ruby-lang.org/issues/18589

This led me to investigate global constant cache busts in our app, and one source turned out to be `rack`.

For instance the first time we access `Request#media_type` it triggers the autoloading of `Rack::MediaType` which bust the cache, but also means this code wasn't loaded pre-fork, so slightly degraded Copy on Write performance.

I get that the library make use of `autoload` so that you only pay for what you actually need, but if I could chose I think I'd rather load a bit of unused code.

@jeremyevans would you be open to a `Rack.eager_load!` method?

casperisfine:
Another example is `Rack::Auth::Digest`:

```
    from gems/rack-2.2.3/lib/rack/auth/digest/params.rb:6:in `<module:Digest>'
    from gems/rack-2.2.3/lib/rack/auth/digest/params.rb:5:in `<module:Auth>'
    from gems/rack-2.2.3/lib/rack/auth/digest/params.rb:4:in `<module:Rack>'
    from gems/rack-2.2.3/lib/rack/auth/digest/params.rb:3:in `<main>'
    from gems/bootsnap-1.10.3/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:30:in `require'
    from gems/bootsnap-1.10.3/lib/bootsnap/load_path_cache/core_ext/kernel_require.rb:30:in `require'
    from gems/zeitwerk-2.5.4/lib/zeitwerk/kernel.rb:35:in `require'
    from gems/rack-2.2.3/lib/rack/multipart/parser.rb:244:in `handle_mime_head'
    from gems/rack-2.2.3/lib/rack/multipart/parser.rb:211:in `block in run_parser'
    from gems/rack-2.2.3/lib/rack/multipart/parser.rb:204:in `loop'
    from gems/rack-2.2.3/lib/rack/multipart/parser.rb:204:in `run_parser'
    from gems/rack-2.2.3/lib/rack/multipart/parser.rb:188:in `on_read'
    from gems/rack-2.2.3/lib/rack/multipart/parser.rb:73:in `parse'
    from gems/rack-2.2.3/lib/rack/multipart.rb:54:in `extract_multipart'
    from gems/rack-2.2.3/lib/rack/request.rb:594:in `parse_multipart'
[...]

PR #1824, opened 2022-03-01
Eager load Rack::MediaType from Rack::Request

casperisfine:
Closes: https://github.com/rack/rack/issues/1823

This avoid triggering some lazy autoloading when `Request#media_type` is accessed. If you use `Rack::Request` it's lamost certain you'll use `Rack::MediaType` so it makes sense to load them together.

cc @jeremyevans

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require_relative 'media_type'
+
 module Rack
   # Rack::Request provides a convenient interface to a Rack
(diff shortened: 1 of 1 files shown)


jeremyevans:
Looks like you need to update.  This require is already there: https://github.com/rack/rack/blob/main/lib/rack/request.rb#L5

casperisfine:
Ah, this change just wasn't released 🤦

jeremyevans:
There have been a lot of changes since Rack 2.2.  The next Rack release (modulo security releases) will be Rack 3.0.

PR #1825, opened 2022-03-02
Remove deprecated `Rack::Utils#byte_ranges`.

ioquatix:


jeremyevans (changes requested):
The warning is only issued in verbose mode.  We should replace `if $VERBOSE` with `, uplevel: 1` in 3.0, and remove in 3.1.

jeremyevans (approved):
Looks good, but can you update `test/spec_utils.rb` to switch `byte_ranges` to `get_byte_ranges`?

jeremyevans:
No need to update the specs, I took care of it while working on #1826.

PR #1828, opened 2022-03-04
Undeprecate set_cookie_header!/delete_cookie_header!

jeremyevans:
There is not a replacement for these methods that updates a
header hash in place.  We should not be deprecating these
methods unless we are providing a suitable replacement for them.

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -268,8 +268,6 @@ module Rack
     end
 
     def set_cookie_header!(headers, key, value)
-      warn("set_cookie_header! is deprecated and will be removed in Rack 3.1", uplevel: 1)
-
       if header = headers[SET_COOKIE]
         if header.is_a?(Array)
           header << set_cookie_header(key, value)
@@ -298,8 +296,6 @@ module Rack
     end
 
     def delete_cookie_header!(headers, key, value = {})
-      warn("delete_cookie_header! is deprecated and will be removed in Rack 3.1", uplevel: 1)
-
       headers[SET_COOKIE] = delete_set_cookie_header!(headers[SET_COOKIE], key, value)
 
       return nil


ioquatix:
@tenderlove asked to deprecate/remove any methods which effectively depend on `env` as an argument, rather than the specific header, same as `byte_ranges`.

jeremyevans:
These methods have nothing to do with `env`, they modify a response header hash.

ioquatix:
> These methods have nothing to do with env, they modify a response header hash.

Same set of issues.

jeremyevans:
The alternative is more repetitive, as you at least have to provide the header key twice.  This isn't about testing, it's about ease of use by applications and frameworks.

ioquatix:
That's why I said, I'm on the fence, but that being said, I don't think we should have the similar wrappers around header manipulation when a general one (add this header, remove this header) is sufficient. In any case, I see both sides and you probably need input from @tenderlove about what direction he wants to take.

jeremyevans:
From a brief review, `Utils.parse_cookies` is not deprecated, and that accepts `env` as an argument.  So if the deprecation consideration is to remove methods that accept `env`, then `parse_cookies` should be deprecated as well.  However, I prefer the approach in this pull request.

ioquatix:
I think we need @tenderlove to give input here. I actually don't have a strong opinion but the way headers are currently managed is a bit cumbersome. Especially how cookies are deleted is something we should revisited since it's inefficient.

tenderlove:
Honestly it's fine if we keep it.  I'd really like to avoid mutations in general, but since these are mutating a parameter the user passed in (and they have bang methods) its probably fine.

These methods are kind of funny though because `set_cookie_header!` seem to be additive.  I'd expect a setter method to only set, not add.  `add_cookie` seems more accurate.

Anyway, GitHub code search shows [lots of usage](https://cs.github.com/?q=set_cookie_header%21), so lets undeprecate this until we can find a better replacement.

Issue #1829, opened 2022-03-14
Different request.host results


ylecuyer:
Hello,

Is it expected that this commit https://github.com/rack/rack/commit/290523f67cc43c5847b2be2d12964d1232061fe1 changed the behavior of requet.host ? The new behavior is not documented in the commit message nor in the project changelog.

Here is an example repo where you can reproduce the issue: https://github.com/ylecuyer/bugged-rack-x-forwarded-host

Basically now it always returns host2 whereas before it was hostN (last host)

Before

```
ylecuyer@lenovo:~$ curl -X GET -H "X-Forwarded-Host: host1" -H "X-Forwarded-Host: host2" -H "X-Forwarded-Host: host3" -H "X-Forwarded-Host: host4" http://localhost:9292
host4
ylecuyer@lenovo:~$ curl -X GET -H "X-Forwarded-Host: host1" -H "X-Forwarded-Host: host2" -H "X-Forwarded-Host: host3" http://localhost:9292
host3
ylecuyer@lenovo:~$ curl -X GET -H "X-Forwarded-Host: host1" -H "X-Forwarded-Host: host2" http://localhost:9292
host2
ylecuyer@lenovo:~$ curl -X GET -H "X-Forwarded-Host: host1" http://localhost:9292
host1
```

Now:

```
ylecuyer@lenovo:~$ curl -X GET -H "X-Forwarded-Host: host1" -H "X-Forwarded-Host: host2" -H "X-Forwarded-Host: host3" -H "X-Forwarded-Host: host4" http://localhost:9292
host1
ylecuyer@lenovo:~$ curl -X GET -H "X-Forwarded-Host: host1" -H "X-Forwarded-Host: host2" -H "X-Forwarded-Host: host3" http://localhost:9292
host1
ylecuyer@lenovo:~$ curl -X GET -H "X-Forwarded-Host: host1" -H "X-Forwarded-Host: host2" http://localhost:9292
host1
ylecuyer@lenovo:~$ curl -X GET -H "X-Forwarded-Host: host1" http://localhost:9292
host1
```

ylecuyer:
Also I did the following patch:

```diff
diff --git a/lib/rack/request.rb b/lib/rack/request.rb
index e633540b..9cc83923 100644
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -359,7 +359,7 @@ module Rack
 
       def forwarded_authority
         if value = get_header(HTTP_X_FORWARDED_HOST)
[...]

PR #1832, opened 2022-03-16
Restore: changed behavior for request.host when multiple X-Forwarded-Host are used

ylecuyer:
As explained in https://github.com/rack/rack/issues/1829 commit https://github.com/rack/rack/commit/290523f67cc43c5847b2be2d12964d1232061fe1 changed the behavior of request.host when multiple X-Forwarded-Host are used. This PR restores the previous behavior.

I don't know which behavior is the correct one, but this is a breaking change. If we really want to change from `last` to `first` we should issue the proper deprecation warning and include the change in the v3 release. I didn't update the changelog yet because I think it would be best to have this fix in 2.2.4 though I'd prefer to have your insight.

Here is the before after:

Before:

```
ylecuyer@lenovo:~$ curl -X GET -H "X-Forwarded-Host: host1" -H "X-Forwarded-Host: host2" -H "X-Forwarded-Host: host3" -H "X-Forwarded-Host: host4" http://localhost:9292
host1
ylecuyer@lenovo:~$ curl -X GET -H "X-Forwarded-Host: host1" -H "X-Forwarded-Host: host2" -H "X-Forwarded-Host: host3" http://localhost:9292
host1
ylecuyer@lenovo:~$ curl -X GET -H "X-Forwarded-Host: host1" -H "X-Forwarded-Host: host2" http://localhost:9292
host1
ylecuyer@lenovo:~$ curl -X GET -H "X-Forwarded-Host: host1" http://localhost:9292
host1
```

After

```
ylecuyer@lenovo:~$ curl -X GET -H "X-Forwarded-Host: host1" -H "X-Forwarded-Host: host2" -H "X-Forwarded-Host: host3" -H "X-Forwarded-Host: host4" http://localhost:9292
host4
ylecuyer@lenovo:~$ curl -X GET -H "X-Forwarded-Host: host1" -H "X-Forwarded-Host: host2" -H "X-Forwarded-Host: host3" http://localhost:9292
host3
ylecuyer@lenovo:~$ curl -X GET -H "X-Forwarded-Host: host1" -H "X-Forwarded-Host: host2" http://localhost:9292
host2
ylecuyer@lenovo:~$ curl -X GET -H "X-Forwarded-Host: host1" http://localhost:9292
host1
```

Also it should be noted that rails is also using last https://github.com/rails/rails/blob/main/actionpack/lib/action_dispatch/http/url.rb#L216 so for consistency if one is changed the other should be changed too.

cc @ioquatix as you are the author of the initial commit and maybe cc @tenderlove for the rails note above.

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -359,7 +359,7 @@ module Rack
 
       def forwarded_authority
         if value = get_header(HTTP_X_FORWARDED_HOST)
-          wrap_ipv6(split_header(value).first)
+          wrap_ipv6(split_header(value).last)
         end
       end
 
--- a/test/spec_request.rb
+++ b/test/spec_request.rb
@@ -168,6 +168,11 @@ class RackRequestTest < Minitest::Spec
     req.hostname.must_equal "example.org"
 
     req = make_request \
+      Rack::MockRequest.env_for("/", "HTTP_HOST" => "localhost:81", "HTTP_X_FORWARDED_HOST" => "somehost:9292, anotherhost:9292, example.org:9292")
+    req.host.must_equal "example.org"
+    req.hostname.must_equal "example.org"
+
+    req = make_request \
       Rack::MockRequest.env_for("/", "HTTP_HOST" => "localhost:81", "HTTP_X_FORWARDED_HOST" => "[2001:db8:cafe::17]:47011")
     req.host.must_equal "[2001:db8:cafe::17]"
     req.hostname.must_equal "2001:db8:cafe::17"


ioquatix:
I don't have a strong opinion either way, but is there an RFC which explains the best value to use? I assume the earlier forwarded for details are closer to the client while later ones are more likely to be proxies.

misdoro:
Ah, we've also had this issue.
We ended up adding a middleware like 
```
class CustomRemoteIp < ActionDispatch::RemoteIp
  X_REAL_IP_HEADER = 'HTTP_TRUE_CLIENT_IP'

  # Since the IP address may not be needed, we store the object here
  # without calculating the IP to keep from slowing down the majority of
  # requests. For those requests that do need to know the IP, the
  # GetCustomIp#calculate_ip method will calculate the memoized client IP address.
  def call(env)
    req = ActionDispatch::Request.new env
    req.remote_ip = GetCustomIp.new(req, check_ip, proxies)
    @app.call(req.env)
  end

  class GetCustomIp < ActionDispatch::RemoteIp::GetIp
    def calculate_ip
      @req.get_header(X_REAL_IP_HEADER).presence || super
    end
  end
end
```
as a workaround.

ioquatix:
Would it make sense to add an interface so that users can ask for all forwarded hosts and then can select either the first or last as is useful to their own use case?

matthewd:
My quick thought would be to push the "clean up something that's nominally a list of IP/host entries" behaviour into a public Rack::Util method, and then say that if someone has a more complex need they can read the header themselves [and we've supplied that cleanup help]. My theory there is that `first` is a pretty nuanced value, given the question of how many proxies you trust to exist and accurately copy a once-true `Host` header across, vs complete fabrication from an end client. IMO if you don't want last, you perhaps actually want the Nth-last entry, which might _generally_ be first given well-behaved clients. (And I'd consequently prefer to nudge people toward using last as authoritative, by having to work a bit more to get something else, lest they pick arbitrarily and get surprised later.)

... again, all of that applies less strongly to forwarded host in particular, but I guess I'm reluctant to add a list-of-hosts feature because of the precedent it would create for a matching list-of-client-IPs, and _that_ feels like leaving a foot-gun on the table, where I'd rather just note that there is one kept in the safe.

ioquatix:
I think the value in standardising an interface here, apart from the ones discussed already, is that internally HTTP has got a defacto standard of `x-forwarded-whatever` and an official standard `forwarded`. It would be nice for us to have some way to handle this transparently and security for the application since it's a bit of a mess and I think we prefer avoiding bespoke code that ultimately could end up being insecure, i.e. let's guide applications to do the right thing.

ioquatix:
Based on discussion here, it sounds like it would be a security issue to change this behaviour. Thus we will close it. Thanks for your ideas and contributions.

ioquatix:
Sorry, I may have mis-understood the advice here.

> We strictly must use the last (and thus nearest-to-us proxy) details for X-Forwarded-For, because any nominally-further value cannot be trusted: any originating client can claim to be proxying on behalf of anyone they like (including things like private IPs, or localhost).

This means we should apply this PR.

jeremyevans:
I fixed this as part of #1834, can you please review that instead?: https://github.com/rack/rack/pull/1834/files#diff-7ce97931f18a63a4d028696a6f4ba81991644dbc2d70eaa664285264e9a5cd64L360

jeremyevans:
Implemented by #1834

Issue #1809, opened 2022-02-15
Web cache poisoning -- `X-Forwarded-Scheme` vs `X-Forwarded-Proto`


directionless:
Hi! I believe there default rack behavior of [`forwarded_scheme`](https://github.com/rack/rack/blob/42aff22f708123839ba706cbe659d108b47c40c7/lib/rack/request.rb#L641-L644) creates a cache poisoning vulnerability in common configurations. This was initially disclosed almost a year ago in the blog post [Cache Poisoning At Scale](https://youst.in/posts/cache-poisoning-at-scale/), and I am starting to see it used by pentesters. As I don't an issue or fix, I wanted to document it here. 

In short, many CDNs use `X-Forwarded-Proto` to convey the original request protocol. But, rack prefers `X-Forwarded-Scheme`, this creates a simple opportunity for a cache poisoning attack.

This is a bit hard to show, but if you assume the `rack` URL has a CDN in front of it:

```shell
URL=https://rack.example.com/assets/image.png?cachebust=${RANDOM}

curl -i  $URL
# Observe that the `x-cache` response is MISS. 

curl -i  $URL
# Observe that the `x-cache` response is now HIT

##
## Attack
## 
URL=https://rack.example.com/assets/image.png?cachebust=${RANDOM}

curl -H 'x-forwarded-scheme: http' -i $URL
# Observe the 301 response, and `x-cache: MISS`

curl  -i $URL
# observe a cached 301
```

This is something I had to patch on our CDN, and it sounds like something many sites have had to patch as well. (The previously linked blog post lists several notable sites. Here is the [hackerone discussion](https://hackerone.com/reports/1181946))

ioquatix:
Do you have any suggestions about how we can fix it?

jeremyevans:
The obvious approach to "fixing it" would be respecting `x-forwarded-proto` in preference to `x-forwarded-scheme` if both are present.  Of course, doing that will break existing safe usage where the CDN/proxy uses `x-forwarded-scheme` (since then an attacker could specify `x-forwarded-proto`).  There is no completely safe default.  We would need to make this configurable, and probably choose whichever is more popular as the default (and I don't know which is more popular).

matthewd:
I think the long term answer is https://github.com/rack/rack/pull/1423, but that only helps when relevant upstreams have also adopted it.

directionless:
I was thinking about what an ideal solution is, unfortunately I suspect that any time you check multiple headers, the site 
[...]

PR #1834, opened 2022-03-17
Support RFC 7239 Forwarded header and configurable priorities for checking forwarding headers

### LINKED ISSUE #1829: Different request.host results

jeremyevans:
This continues @fatkodima's work from #1423, rebasing it against the master branch and making changes so it applies.

To allow Rack users to choose the appropriate forwarding headers to inspect for their environment, this sets up configurable priorities using `Rack::Request.forwarded_priority`.  This defaults to `[:forwarded, :x_forwarded]`, to check the `Forwarded` header before the `X-Forwarded-*` headers.

Additionally, to handle the `X-Forwarded-Proto` vs. `X-Forwarded-Scheme` security issue (#1809), this sets up configurable priorities for that using `Rack::Request.x_forwarded_proto_priority`.  This defaults to `[:proto, :scheme]`, to check `X-Forwarded-Proto` before `X-Forwarded-Scheme`.

In both cases, you can omit one of the priorities to never check the related header, or use an empty array to not check any forwarding headers, which is useful if your application does not use a reverse proxy, and you don't want to allow clients to forge forwarded information.

While here, fix #1829 by making `forwarded_authority` use the last forwarded authority and not the first, since previous forwarded authorities can be forged by clients.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ All notable changes to this project will be documented in this file. For info on
 - `Rack::Headers` added to support lower-case header keys. ([@jeremyevans](https://github.com/jeremyevans))
 - `Rack::RewindableInput` supports size. ([@ahorek](https://github.com/ahorek))
 - `Rack::RewindableInput::Middleware` added for making `rack.input` rewindable. ([@jeremyevans](https://github.com/jeremyevans))
-- Rack::Session::Pool now accepts `:allow_fallback` option to disable fallback to public id. ([#1431](https://github.com/rack/rack/issues/1431), [@jeremyevans](https://github.com/jeremyevans))
+- The RFC 7239 Forwarded header is now supported and considered by default when looking for information on forwarding, falling back to the the X-Forwarded-* headers. `Rack::Request.forwarded_priority` accessor has been added for configuring the priority of which header to check.  ([#1423](https://github.com/rack/rack/issues/1423), [@jeremyevans](https://github.com/jeremyevans))
 
 ### Changed
 
@@ -47,6 +47,8 @@ All notable changes to this project will be documented in this file. For info on
 - Explicitly deprecate `Rack::File` which was an alias for `Rack::Files`. ([#1811](https://github.com/rack/rack/pull/1720), [@ioquatix](https://github.com/ioquatix)).
 - Moved `Rack::Session` into [separate gem](https://github.com/rack/rack-session). ([#1805](https://github.com/rack/rack/pull/1805), [@ioquatix](https://github.com/ioquatix))
 - rackup -D option to daemonizes no longer changes the working directory to the root. ([#1813](https://github.com/rack/rack/pull/1813), [@jeremyevans](https://github.com/jeremyevans))
+- The X-Forwarded-Proto header is now considered before the X-Forwarded-Scheme header for determining the forwarded protocol. `Rack::Request.x_forwarded_proto_priority` accessor has been added for configuring the priority of which header to check.  ([#1809](https://github.com/rack/rack/issues/1809), [@jeremyevans](https://github.com/jeremyevans))
+- `Rack::Request.forwarded_authority` (and methods that call it, such as `host`) now returns the last authority in the forwarded header, instead of the first, as earlier forwarded authorities can be forged by clients. This restores the Rack 2.1 behavior. ([#1829](https://github.com/rack/rack/issues/1809), [@jeremyevans](https://github.com/jeremyevans))
 
 ### Fixed
 
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -152,6 +152,8 @@ module Rack
       # The contents of the host/:authority header sent to the proxy.
       HTTP_X_FORWARDED_HOST = 'HTTP_X_FORWARDED_HOST'
 
+      HTTP_FORWARDED          = 'HTTP_FORWARDED'
+
       # The value of the scheme sent to the proxy.
       HTTP_X_FORWARDED_SCHEME = 'HTTP_X_FORWARDED_SCHEME'
 
@@ -331,7 +333,7 @@ module Rack
         end
 
         if forwarded_port = self.forwarded_port
-          return forwarded_port.first
+          return forwarded_port.last
         end
 
         if scheme = self.scheme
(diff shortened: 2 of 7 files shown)


lib/rack/utils.rb
```diff
@@ -140,6 +140,21 @@ def q_values(q_value_header)
       end
     end
 
+    FORWARDED_PAIR_REGEX = /\A\s*(by|for|host|proto)\s*=\s*"?([^"]+)"?\s*\Z/i
```
ioquatix, line 143: Should we make this private?

jeremyevans: That's a good idea, I'll make that change.

jeremyevans: This constant is only used in one place, so I inlined it instead.

ioquatix:
This generally looks really great to me, I just want to understand a bit more what the expected arguments and return values should be.

I also wonder how much of this should be public and whether we should make some methods private.

I am travelling from tomorrow so my review would be delayed.

lib/rack/request.rb
```diff
@@ -638,9 +707,33 @@ def reject_trusted_ip_addresses(ip_addresses)
...
+      FORWARDED_SCHEME_HEADERS = {
+        proto: HTTP_X_FORWARDED_PROTO,
+        scheme: HTTP_X_FORWARDED_SCHEME
+      }.freeze
+      private_constant :FORWARDED_SCHEME_HEADERS
       def forwarded_scheme
-        allowed_scheme(get_header(HTTP_X_FORWARDED_SCHEME)) ||
-        allowed_scheme(extract_proto_header(get_header(HTTP_X_FORWARDED_PROTO)))
+        forwarded_priority.each do |type|
+          case type
```
ioquatix, line 717: I feel using a case statement here is a little bit cumbersome. Like, we know in advance what code should be executed. 

Thinking longer term, does the symbol list in `forwarded_priority` make it hard to evolve this implementation? If we wanted to get rid of the case statement and make the internal implementation more streamlined.

Does it matter? My main thought is performance. I guess probably not. If it was a big issue, maybe later we can construct a lambda based on the expected behaviour and just cache it?

I guess the alternative would be, instead of an array (which is flexible, and could even contain the same entry more than once even thought it doesn't make sense).

Maybe an alternative would be `self.send(:"forwareded_scheme_#{type}")`???

jeremyevans: I'm not attached to the particular approach used.  If we want to change to use `send` instead of `case`, that seems fine with me.  The important part is that the solution implemented allows the user to choose which header to respect in which order, and allows configuring to explicitly ignore either header or both headers.

ioquatix: Do you have the energy to make that change and do you prefer it? I'm happy to defer to your judgement here.

jeremyevans: I'm fine keeping the `case`.  We can optimize using `send` later in a backwards compatible manner if it turns out to be a bottleneck.

ioquatix (approved):

ioquatix:
Thanks for your hard work.

CHANGELOG.md
```diff
@@ -47,6 +47,8 @@ All notable changes to this project will be documented in this file. For info on
...
 - Moved `Rack::Session` into [separate gem](https://github.com/rack/rack-session). ([#1805](https://github.com/rack/rack/pull/1805), [@ioquatix](https://github.com/ioquatix))
 - rackup -D option to daemonizes no longer changes the working directory to the root. ([#1813](https://github.com/rack/rack/pull/1813), [@jeremyevans](https://github.com/jeremyevans))
+- The X-Forwarded-Proto header is now considered before the X-Forwarded-Scheme header for determining the forwarded protocol. `Rack::Request.x_forwarded_proto_priority` accessor has been added for configuring the priority of which header to check.  ([#1809](https://github.com/rack/rack/issues/1809), [@jeremyevans](https://github.com/jeremyevans))
+- `Rack::Request.forwarded_authority` (and methods that call it, such as `host`) now returns the last authority in the forwarded header, instead of the first, as earlier forwarded authorities can be forged by clients. This restores the Rack 2.1 behavior. ([#1829](https://github.com/rack/rack/issues/1809), [@jeremyevans](https://github.com/jeremyevans))
```
ylecuyer, line 51: > This restores the Rack 2.1 behavior.

Would it be possible to clearly mark this change as a breaking change in the changelog of the release where this different behavior was included?

Would it also be possible to backport the fix to the v2 branch?

ioquatix: Do you mean backport this entire PR?

PR #1835, opened 2022-03-21
Tighten authority matching

jeremyevans:
Tighten up IPv6 parsing rules using regexp extracted from resolv in stdlib,
simplified to avoid creating additional groups.

Tighten up hostname matching to graphical characters, except square brackets
(so it doesn't overlap with IPv6 parsing).

Avoid unnecessary IPv4 matching, since anything that matches as an IPv4
address would match as a hostname.

Remove unnecessary named group creation.

Don't allow trailing newlines in host names.

Fixes #1607

Co-authored-by: Pieter van de Bruggen <pvande@gmail.com>

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -605,28 +605,48 @@ module Rack
         value ? value.strip.split(/[,\s]+/) : []
       end
 
+      # ipv6 extracted from resolv stdlib, simplified
+      # to remove numbered match group creation.
+      ipv6 = Regexp.union(
+        /(?:[0-9A-Fa-f]{1,4}:){7}
+         [0-9A-Fa-f]{1,4}/x,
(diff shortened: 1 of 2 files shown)


lib/rack/request.rb
```diff
@@ -605,28 +605,48 @@ def split_header(value)
         value ? value.strip.split(/[,\s]+/) : []
       end
 
+      # ipv6 extracted from resolv stdlib, simplified
+      # to remove numbered match group creation.
+      ipv6 = Regexp.union(
```
ioquatix, line 610: I like this approach. However, should this be a private constants or are we okay with it being a local variable? Either is okay, just wondering out loud.

ioquatix (approved):
Great, thanks for cleaning this up and the great addition to the tests!

lib/rack/request.rb
jeremyevans: I prefer local variable when we aren't using the value inside a method.

PR #1836, opened 2022-03-21
Add :escape_key option to set_cookie_header

### LINKED ISSUE #1796: Consider enabling a subset of cookie names to be unescaped.
Reviewing https://github.com/rack/rack/commit/1f5763de6a9fe515ff84992b343d63c88104654c I wonder if we can find a better middle-ground.

One option is allow for unescaping of cookie names but only a specific subset of characters that can't be otherwise represented.

Another option is to explicitly disallow unescaping names that start with `__` in any form.

This would largely mitigate the compatibility issue that the security fix introduced i.e. <https://github.com/rack/rack/issues/1674>.

cc @tenderlove

jeremyevans:
This can be set to false to avoid escaping the key.

This is a very explicit approach to allowing some cookie names to
not be escaped.

Fixes #1796

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ All notable changes to this project will be documented in this file. For info on
 ### Added
 
 - `Rack::Headers` added to support lower-case header keys. ([@jeremyevans](https://github.com/jeremyevans))
+- `Rack::Utils#set_cookie_header` now supports `escape_key: false` to avoid key escaping.  ([@jeremyevans](https://github.com/jeremyevans))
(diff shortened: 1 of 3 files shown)


ioquatix:
This seems pretty reasonable to me, are options handled by `Rack::Response#set_cookie` too?

ioquatix:
Also does cookie deletion need the same option?

jeremyevans:
Looks like all code paths run through `set_cookie_header`, so this should take care of all cases.  Options in this case are providing the value as a hash instead of a string, so all methods support the `escape_key: false` option.

ioquatix (approved):

PR #1837, opened 2022-04-03
Use custom exception on params too deep error.

simi:
Hello, I have been working on fixing some malicious requests causing server 500s responses at rails application.

There is list of default exceptions transformed into proper responses in Rails.

https://github.com/rails/rails/blob/926b803297c4bc529ee46296a7cd5cc02bb21100/actionpack/lib/action_dispatch/middleware/exception_wrapper.rb#L6-L24

As you can see, the other errors (`ParameterTypeError` and `InvalidParameterError`) are already part of this list and they result into `400 Bad request` response by default. But breaking the rack `param_depth_limit` setting results into general Ruby `RangeError` exception currently and it is hard to recognise what's the cause (since any other code can raise this general Ruby exception) and that's (probably) the reason why it is not handled by default in Rails.

I suggest to introduce Rack custom exception for this problem as well (inherited from originally used `RangeError`) to be able to  catch this easily. Next step (if approved) would be to add this exception to default Rails list as well. In the end I think it needs to be added in here - https://github.com/rails/rails/blob/926b803297c4bc529ee46296a7cd5cc02bb21100/actionpack/lib/action_dispatch/http/request.rb#L382-L384.

It is hard to fix currently in the app, the only solution I was able to find out is to create simple middleware to read the params and handle the error in there.

Since the exception is inherited from the original one (and tests were passing with no change as well) I would to propose to backport this to 2.x series as well to be able to fix this as soon as possible.

### DIFF
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -14,6 +14,10 @@ module Rack
     # sequence.
     class InvalidParameterError < ArgumentError; end
 
+    # ParamsTooDeepError is the error that is raised when params are recursively
+    # nested over the specified limit.
+    class ParamsTooDeepError < RangeError; end
+
     def self.make_default(_key_space_limit=(not_deprecated = true; nil), param_depth_limit)
       unless not_deprecated
         warn("`first argument `key_space limit` is deprecated and no longer has an effect. Please call with only one argument, which will be required in a future version of Rack", uplevel: 1)
@@ -91,7 +95,7 @@ module Rack
     end
 
     private def _normalize_params(params, name, v, depth)
-      raise RangeError if depth >= param_depth_limit
+      raise ParamsTooDeepError if depth >= param_depth_limit
 
       if !name
         # nil name, treat same as empty string (required by tests)
--- a/test/spec_request.rb
+++ b/test/spec_request.rb
@@ -400,7 +400,7 @@ class RackRequestTest < Minitest::Spec
   it "limit the allowed parameter depth when parsing parameters" do
     env = Rack::MockRequest.env_for("/?a#{'[a]' * 40}=b")
     req = make_request(env)
-    lambda { req.GET }.must_raise RangeError
+    lambda { req.GET }.must_raise Rack::QueryParser::ParamsTooDeepError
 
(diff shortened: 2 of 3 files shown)


simi:
CI error seems unrelated, would it be possible to try restart?

Btw. this is the middleware I was able to use to gracefully fail in my Rails app on this error.

```ruby
module MyApp::Middleware
  class DeepParamsHandler
    def initialize(app)
      @app = app
    end

    def call(env)
      begin
        Rack::Request.new(env).params
        status, headers, response = @app.call(env)
        [status, headers, response]
      rescue RangeError => e
        [ 302, {'Location' =>"/500.html"}, [] ]
      end
    end
  end
end
```

jeremyevans (approved):

simi:
Thanks @tenderlove. What do you think about backporting this to Rack 2.2.x? I can open PR if welcomed.

rafaelfranca:
@simi if the goal is to allow Rails apps to handle that, I think it is fine to backport. Please open the PR.

ioquatix:
Since this is a user visible change, I would like to see an entry to the change log for this (as well as the backport). Thanks.

simi:
@ioquatix changelog is definitely planned from my side. I have checked few other PRs and changelogs were not introduced, I wasn't sure when it is good time to add it. Sometimes maintainers do changelog update before the release on their own. :/

ioquatix:
Yeah, and I'm also one to forgot to do it in the past. We should definitely aim for fully self-contained PRs where possible. For the backport it will need a clear explanation of any changes if there are compatibility trade offs, so it's a good time to introduce it. One challenge is that the changelog will have slightly different organisation in the backport but ideally we'd do that as part of the backport PR.

jchestershopify:
I think consideration should be given to opening a See Vee Ee (sounded out because there are bots that scrape Github looking for the proper abbreviation). This issue appears to affect availability.

simi:
Actually rack itself do nothing wrong. It just raises an error on problem as expected. But this error is not handled in Rails and thanks to the nature of the general nature being raised, it is hard to update Rails code to handle this now. Your proposal maybe will make sense later at Rails side. I'll ask once I'll open PR in there as well.

PR #1838, opened 2022-04-04
Use custom exception on params too deep error.

simi:
- backport of df23f57407426a85245e35d8c0c7c90d53dcb74a

/cc @rafaelfranca

### DIFF
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -16,6 +16,10 @@ module Rack
     # sequence.
     class InvalidParameterError < ArgumentError; end
 
+    # ParamsTooDeepError is the error that is raised when params are recursively
+    # nested over the specified limit.
+    class ParamsTooDeepError < RangeError; end
+
     def self.make_default(key_space_limit, param_depth_limit)
       new Params, key_space_limit, param_depth_limit
     end
@@ -81,7 +85,7 @@ module Rack
     # the structural types represented by two different parameter names are in
     # conflict, a ParameterTypeError is raised.
     def normalize_params(params, name, v, depth)
-      raise RangeError if depth <= 0
+      raise ParamsTooDeepError if depth <= 0
 
       name =~ %r(\A[\[\]]*([^\[\]]+)\]*)
       k = $1 || ''
@@ -168,7 +172,7 @@ module Rack
 
       def []=(key, value)
         @size += key.size if key && !@params.key?(key)
-        raise RangeError, 'exceeded available parameter key space' if @size > @limit
+        raise ParamsTooDeepError, 'exceeded available parameter key space' if @size > @limit
         @params[key] = value
       end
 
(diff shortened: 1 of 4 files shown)


simi:
@rafaelfranca thanks. Are there any plans to release this with patch version bump? I can help to prepare the release. Once released, I can open the last PR to handle this exception properly in Rails.

ioquatix:
This to me looks like a breaking change for existing Rack 2.x - @jeremyevans what do you think?

simi:
@ioquatix As I have pointed at https://github.com/rack/rack/pull/1837, specs are passing with no change (even I have changed them to use the new exception). New exception is inherited from the old one. Can you elaborate a little more what do you think is breaking change in here?

simi:
I can add additional spec to ensure `ParamsTooDeepError` inherits from `RangeError` if welcomed with small comment explaining on this. That should cover unintentional change of the parent class in the future. WDYT?

ioquatix:
I think it's a great idea to prevent someone accidentally refactoring it and not realising it was part of the original specs :)

ioquatix:
It might be pedantic, but even just a spec like, ensuring that the class is a subclass of `RangeError` could be sufficient.

ioquatix:
I thought about this a bit more, the only compatibility issue that I can think of would be systems like Honeybadger which do error reporting, it would mean the class would change and historical data might not align up.. but not sure if I personally care about that much. That being said, it is a user visible change.

simi:
@ioquatix currently default Rails app is vulnerable to 500 error (crashing the app on uncaught exception) which could be considered attack vector. Having this custom exception, it would be possible to craft patch to Rails to prevent this and gracefully fail. That's the reason for this backport. Currently it is not possible to release from main branch and update Rails dependency to Rack 3. There is no other option than this (at least I wasn't able to find out any).

A little confused error reporting is not big deal in my eyes, maybe even it is clever enough to group those errors together. No error is going to be hidden, it can just re-appear under different name.

ioquatix:
Thanks for that information. I didn't know there was a security component to this. I think the take away here is that a little bit more communication would have been helpful here, the original description doesn't give enough context. For reference, here is the original PR: https://github.com/rack/rack/pull/1837. Might I suggest in the future you start by creating an issue and then link all the relevant PRs.

The bar for security fixes is certainly lower in terms of compatibility but it still needs to be considered carefully. I don't personally have a stake in this, but my expectation is that @tenderlove probably needs to review this change before we make any release. I would also like to hear input from @jeremyevans since he has a keen sense for compatibility issues.

Like I said earlier, I personally support this PR, but that doesn't mean the bar doesn't need to be met, in fact my approval means very little in this particular space as I think other maintainers are more sensitive to the issues here :)

jeremyevans:
Replacing an exception class with a subclass of the exception class is generally considered a backwards compatible change, as all rescue clauses will still match.  Anything it breaks pretty much deserves to be broken.

I don't consider this a security issue, though I'm sure it is annoying to receive bug reports for 500 errors instead of just reporting 4xx errors to the client.  That said, I'm fine with the backport to Rack 2.2.

PR #1839, opened 2022-04-04
Make specs compatible with later Ruby and setup CI.

simi:
- targets `2-2-stable` to make it easier to cut new releases
- ci is heavily inspired by `main` branch one, but includes only ubuntu Ruby (since other rubies do fail for CGI dependency fail or (JRuby) fails the same way as at master)
- I'll try to fix `main` JRuby CI, but I can't reproduce the problem locally.
-  Time::RFC2822_MONTH_NAME was removed at https://github.com/ruby/time/commit/6b8cc4799e, better to not rely on that (I have inlined those two constants)

/cc @rafaelfranca


ioquatix:
Thanks for this. In the future can you please give current maintainers time to review, which would include myself, @jeremyevans and @tenderlove :)

simi:
:thinking: I think I have missed Ruby 2.3 on CI, which is still supported for 2.2 series. I'll update the CI to add 2.3 as well (I'll need to somehow bandaid the psych dependency since Psych 4.x is not compatible with 2.3).

rafaelfranca:
@ioquatix since when I was demoted to not be a current maintainer?

rafaelfranca:
Being more constructive, if you see anything wrong with the code, please let me know, I'll fix, now setting those nonsense rules in a project that I'm maintainer more time than you, doesn't solve any problem. If you don't have any problem with the code that was merged, let's keep focusing in the real problems, not in trying to gatekeep fixes to real problems.

ioquatix:
@rafaelfranca I apologise for what I said, it was not intended to cause stress.

I'm happy that you and @simi made this contribution, and I don't personally have any problem with the code that was merged, but given how critical some of this stuff is, it's good to have several sets of eyes on it. I actually really value @jeremyevans and @tenderlove's input input because they often bring important perspectives to changes like this (including compatibility, security, etc) and I try to involve them in every PR I create and try to get consensus of both of them where possible.

This PR was proposed and merged within 10 minutes. I felt like this didn't give a chance to anyone else to review the code. There were a couple of other rapid fire PRs yesterday too.

Let's figure out a process for ensuring PRs have the appropriate number of approvals before getting merged: https://github.com/rack/rack/discussions/1842 I welcome your feedback on what this should look like.

simi:
Thanks @ioquatix I'll backport this.

Issue #1840, opened 2022-04-04
Fix cookie deletion.


ioquatix:
Cookie deletion uses a regexp to delete cookies using a linear scan. I don't believe this is useful. So, let's remove it.

ioquatix:
I tested this on chromium and it behaved as expected:

```ruby
# config.ru

run ->(env) do
	headers = {}
	
	# These methods add and remove a cookie.
	# In the browser response was:
	# HTTP/1.1 200 OK
	# set-cookie: ["my_cookie=world", "my_cookie=; max-age=0; expires=Thu, 01 Jan 1970 00:00:00 GMT"]
	# The result was on the next request, `request.cookies` was empty (as expected).
	Rack::Utils.set_cookie_header!(headers, "my_cookie", {value: "world"})
	Rack::Utils.delete_cookie_header!(headers, "my_cookie")
	
	request = Rack::Request.new(env)
	
	[200, headers, ["Cookies: #{request.cookies.inspect}"]]
end
```

PR #1844, opened 2022-04-06
Don't bother pattern matching existing set-cookie for deletion.

ioquatix:
Fixes https://github.com/rack/rack/issues/1840

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ All notable changes to this project will be documented in this file. For info on
 
 - Remove `rack.multithread`/`rack.multiprocess`/`rack.run_once`. These variables generally come too late to be useful. ([#1720](https://github.com/rack/rack/pull/1720), [@ioquatix](https://github.com/ioquatix), [@jeremyevans](https://github.com/jeremyevans)))
 - Remove deprecated Rack::Request::SCHEME_WHITELIST. ([@jeremyevans](https://github.com/jeremyevans))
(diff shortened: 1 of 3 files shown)


ioquatix:
I'm honestly not sure what these tests were trying to achieve. Our deletion logic doesn't matter at all if browsers don't implement it correctly. Because I imagine 99% of people trying to delete cookies would be doing so after the fact it was set in a prior request, so the tests really don't reflect reality.

leahneukirchen:
Please don't commit .DS_Store files..

One use case could be to have middle-ware that removes cookies from some other layer, but I'm not sure how useful that is.

ioquatix:
Oh lol, I don't even know how that happened, I was committing from Linux, but I think I copied the entire repo from my old development laptop haha. Thanks for pointing it out!

tenderlove (approved):

tenderlove:
If you're sending sensitive information, and then relying on this to delete it, I think your app has bigger problems.  Specifically adding sensitive info, then trying to remove it later seems like a "your app problem" not a "rack should deal with this" problem.  Additionally, the regex match used for detecting if it's the right cookie could be problematic.  Since the app developer has access to the array and can mutate it themselves, why is it our responsibility to try and parse their strings?

(My main gripe with this method are the regular expressions.  If we kept the cookies in a more structured format like objects, I think removing existing ones would be fine)

jeremyevans:
I agree the method implementation is ugly. However, the only reason for `delete_set_cookie_header!` is to do this removal.  If we don't want to do this removal, it would be better to deprecate the method, and just tell people to switch to `delete_set_cookie_header`.  Most internal callers of `delete_set_cookie_header!` are already deprecated, and `delete_cookie_header!` could switch to calling `delete_set_cookie_header`.

ioquatix:
I'm happy to revert this PR and just deprecate the entire method if we all agree to that, I'll do it today.

PR #1846, opened 2022-04-06
Don't overwrite other cookie attributes when building deletion cookie.

ioquatix:
I noticed this when going through the cookie utils methods. We were previously supplying nil for path and domain but I think it's fair game that they could be specified for the cookie deletion to allow for deleting a very specific cookie.

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -298,11 +298,7 @@ module Rack
     # Adds a cookie that will *remove* a cookie from the client.  Hence the
     # strange method name.
     def delete_set_cookie_header(key, value = {})
-      set_cookie_header(key, {
-        value: '', path: nil, domain: nil,
-        max_age: '0',
-        expires: Time.at(0)
-      }.merge(value))
+      set_cookie_header(key, value.merge(max_age: '0', expires: Time.at(0), value: ''))
     end
 
     def make_delete_cookie_header(header, key, value)
--- a/test/spec_utils.rb
+++ b/test/spec_utils.rb
@@ -648,9 +648,33 @@ describe Rack::Utils, "cookies" do
     header = []
 
     Rack::Utils.delete_set_cookie_header!(header, 'name2')
-    header.must_equal ["name2=; max-age=0; expires=Thu, 01 Jan 1970 00:00:00 GMT"]
+    header.must_equal [
+      "name2=; max-age=0; expires=Thu, 01 Jan 1970 00:00:00 GMT"
+    ]
+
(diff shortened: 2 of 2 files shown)


jeremyevans:
I think this change is wrong.  The value hash passed to `delete_set_cookie_header` should take precedence.  With this change, you cannot overwrite the `max_age: '0', expires: Time.at(0), value: ''` settings.  We want `default_hash.merge(argument)` (previous implementation), not `argument.merge(default_hash)` (this PR).

ioquatix:
```
> value = {value: "hello world", path: "/a/b"}
=> {:value=>"hello world", :path=>"/a/b"}
> value.merge(max_age: '0', expires: Time.at(0), value: '')
=> {:value=>"", :path=>"/a/b", :max_age=>"0", :expires=>1970-01-01 12:00:00 +1200}
```

I feel like this is the correct behaviour. What do you think?

jeremyevans:
I think we should always prefer the values in the given hash, and never have the default values override them.

ioquatix:
> I think we should always prefer the values in the given hash, and never have the default values override them.

That will mean the cookie may not get deleted, is that logical?

ioquatix:
As an example, someone might go:

```
cookie = request.get_cookie("my_cookie")
request.delete_cookie("my_cookie", cookie)
```

Does it make sense that it won't delete the cookie?

ioquatix:
Maybe you can give an example of an actual use case where your behaviour is the desired one?

rafaelfranca:
I also fail to see when we want to allow people to set `value`, `max_age` and `expire` in this method. We could say the whole point of calling this method is so the values of those arguments are set by our code.

jeremyevans:
I can see usage such as:

```ruby
# Delete cookie after 60 seconds
delete_set_cookie_header('key', value: 'foo', max_age: 60, expires: Time.now + 60)
```

I guess you could argue this is misuse of the API.  If we are OK breaking this use case, I'm fine with this change.

ioquatix:
@jeremyevans I don't have a strong opinion, your example seems reasonable but I think they'd just be better off using `set_cookie_header` in that case. I think @rafaelfranca is on point, specifically:

> We could say the whole point of calling this method is so the values of those arguments are set by our code.

Agree.

jeremyevans (approved):
Since @rafaelfranca appears to be OK with breaking compatibility here, I guess I'm OK with this.  The change to not remove an existing cookie with the same name (dc72e607d9fbfd39c8362eafd5d3be951f3b2fc0) is a far greater break to backwards compatibility than this, and we already made that.

PR #1847, opened 2022-04-06
Remove Rack 3.1 Deprecations

ioquatix:
This is a tentative PR to remove deprecated functionality.


rafaelfranca:
Are we planning to remove deprecations in minor releases? I'm personally fine with that, and it is what Rails does, but that is not what rack did in the past and what semver recommends. Just want to align which version schema we want to use.

ioquatix:
This is the approach that was proposed by @jeremyevans. I guess the other option is just to remove it in 3.0? What do you prefer?

jeremyevans:
The other approach is removing in 4.0 instead of 3.1.  I'm fine removing in 4.0 instead of 3.1 if that is was @rafaelfranca prefers. I'm not OK removing in 3.0 without a deprecation period.

ioquatix:
What about pushing deprecation notices into a final 2.x release?

ioquatix:
I want to make sure we are doing the right thing here.

As far as I see it, there are a set of trade offs.

1. Remove in 3.0 = users have clear knowledge that 3.0 has breaking changes and is consistent with all 3.x releases = semver & communication & predictability.
2. Don't remove in 3.0, remove in 3.1 = don't follow semver, +1 experience for users who start on 3.0 = clear picture of deprecations, bad experience for someone who migrates from 2.x to 3.1+ = missing features & no deprecations = confusion?
3. Don't remove in 3.0, don't remove in 3.1 = we have to maintain a ton of legacy interfaces until a 4.0 release, which encourages us to release 4.0 sooner rather than later = more burden for server and app developers = slow development / more effort maintaining legacy stuff that really should be removed.

Am I missing another option?

jeremyevans:
> What about pushing deprecation notices into a final 2.x release?

I'm strongly against introducing deprecation warnings in a tiny release.

I'm fine with removal of deprecated features in either 3.1 or 4.0.

ioquatix:
I'm not really fine with deprecating in 4.0 as it introduces a huge maintenance burden but honestly that seems like the most semver way to do it.

dentarg:
> * Don't remove in 3.0, remove in 3.1 = don't follow semver, +1 experience for users who start on 3.0 = clear picture of deprecations, bad experience for someone who migrates from 2.x to 3.1+ = missing features & no deprecations = confusion?

I think this should be avoided, most people familiar with Rack doesn't expect this (as has been spelled out above)

> * Don't remove in 3.0, don't remove in 3.1 = we have to maintain a ton of legacy interfaces until a 4.0 release, which encourages us to release 4.0 sooner rather than later = more burden for server and app developers = slow development / more effort maintaining legacy stuff that really should be removed.

I agree with the desire of lessen the maintenance burden, so I would encourage a 4.0 release rather sooner than later, if that helps that goal. Yes, sure, that's churn for server and app developers, but I think most people would pick that over unexpected breaking changes.

jeremyevans:
@ioquatix and I discussed this yesterday.  I think we agreed to remove the code in 3.1, unless @tenderlove would prefer to keep support until 4.0. Rack doesn't advertise semantic versioning. I don't think we should have a quick release of 4.0 just to remove deprecations.  4.0 should be released when we have made major changes (as we did in 3.0).

ioquatix:
Replacing with https://github.com/rack/rack/pull/1966

sheputis:
> Rack doesn't advertise semantic versioning.

@jeremyevans, hello 👋 Just an honest question about this: why don't you use semantic versioning? [rubygems guidelines](https://guides.rubygems.org/patterns/#semantic-versioning:~:text=The%20RubyGems%20team%20urges%20gem%20developers%20to%20follow%20the%20Semantic%20Versioning%20standard%20for%20their%20gem%E2%80%99s%20versions.%20The%20RubyGems%20library%20itself%20does%20not%20enforce%20a%20strict%20versioning%20policy%2C%20but%20using%20an%20%E2%80%9Cirrational%E2%80%9D%20policy%20will%20only%20be%20a%20disservice%20to%20those%20in%20the%20community%20who%20use%20your%20gems.) CI pipelines all over the world depend on the fact that everyone sticks to not releasing breaking changes in minor releases, yet there are occasional gems that do not follow this pattern, which leads CI pipelines suddenly breaking out of nowhere, making pipelines less stable and giving unnecessary overhead.

So my question is, out of curiosity, why do you do this? :)

jeremyevans:
To be clear, I'm not the one who decides rack's versioning policy, and can only speak for myself and not other Rack maintainers. 

Rack has historically broken backwards compatibility even in tiny versions for security fixes, and removing deprecated code in minor releases is fairly common in Ruby gems (as mentioned earlier in this thread, Rails does it).

Plenty of gems advertise semantic versioning, and still break applications and libraries depending on them in minor and sometimes tiny updates.  "All bug fixes are incompatibilities".

For minimal gems (i.e. JavaScript type libraries exposing 1 function), I think semantic versioning makes perfect sense.  For gems with a large API (e.g. Rack), where most users of the gem are only using a small subset of the API, it's much more a gray area, and the tradeoffs need to be carefully considered.  Because of rubygems recommendations to use pessimistic version (upper bounds on dependency versions), bumping the major results in a huge maintenance burden, since all dependent gems must independently update their dependency versions, even if no code changes are required.  I would guess the total maintenance burden for all dependencies for a major bump is larger than the total maintenance burden when removing deprecated methods in a minor release.

sheputis:
thank you for elaborating on this @jeremyevans, makes more sense to me now 👍

Issue #1848, opened 2022-04-06
Cookie attribute names should be matched case insensitively.


ioquatix:
After reviewing more of the cookie code, I noticed we are case sensitive on some parts, namely `secure` and `httponly` in some cases.

https://tools.ietf.org/search/rfc6265 p18-20 discuss that these fields should match case insensitively but we use case sensitive string keys in `mock.rb` and possibly other places.

I suggest we lower case these to normalise the fields, since it seems like this is probably what we expect 99% of the time.

ioquatix:
https://tools.ietf.org/search/rfc6265#section-5.2.6

PR #1849, opened 2022-04-06
Use lower case normalisation for cookie attributes.

ioquatix:
Fixes https://github.com/rack/rack/issues/1848


CHANGELOG.md
```diff
@@ -51,6 +51,7 @@ All notable changes to this project will be documented in this file. For info on
 - rackup -D option to daemonizes no longer changes the working directory to the root. ([#1813](https://github.com/rack/rack/pull/1813), [@jeremyevans](https://github.com/jeremyevans))
 - The X-Forwarded-Proto header is now considered before the X-Forwarded-Scheme header for determining the forwarded protocol. `Rack::Request.x_forwarded_proto_priority` accessor has been added for configuring the priority of which header to check.  ([#1809](https://github.com/rack/rack/issues/1809), [@jeremyevans](https://github.com/jeremyevans))
 - `Rack::Request.forwarded_authority` (and methods that call it, such as `host`) now returns the last authority in the forwarded header, instead of the first, as earlier forwarded authorities can be forged by clients. This restores the Rack 2.1 behavior. ([#1829](https://github.com/rack/rack/issues/1809), [@jeremyevans](https://github.com/jeremyevans))
+- Prefer lower case handling of cookie attributes, including `secure` and `httponly`. ([#1849](https://github.com/rack/rack/pull/1849), [@ioquatix](https://github.com/ioquatix))
```
jeremyevans, line 54: Prefer sounds like if both a lower case and non-lower value are present, one should be preferred, which isn't the case.  Maybe `Use lower case cookie attributes when creating cookies, and fold cookie attributes to lower case when reading cookies.`

jeremyevans (changes requested):
One small change to CHANGELOG, otherwise looks fine.

CHANGELOG.md
ioquatix: Thanks, that looks like a great clarification.

jeremyevans (approved):

PR #1850, opened 2022-04-06
Add documentation to key `Rack::Utils` cookie methods.

ioquatix:
On reviewing these methods, I added documentation.


jeremyevans (changes requested):
We don't use yard `@parameter` or `@returns` anywhere else in rack currently, and I don't think we should start.  It looks weird when you aren't using yard, since it shows up as a regular sentence in the documentation.

I think it would be an improvement to show code examples instead.  This has the advantage of not only showing the types, but gives a better indication of what the methods actual do.

lib/rack/utils.rb
```diff
@@ -218,16 +218,13 @@ def select_best_encoding(available_encodings, accept_encoding)
       (encoding_candidates & available_encodings)[0]
     end
 
-    def parse_cookies(env)
-      parse_cookies_header env[HTTP_COOKIE]
-    end
-
+    # Parse cookies according to RFC6265.
+    # The syntax for cookie headers only supports semicolons.
+    # @parameter header [String] The serialised cookie string from the `cookie` header.
```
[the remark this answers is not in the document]
ioquatix: Yeah let's add that.

rafaelfranca:
👍🏽  for the docs, but we use regular rdoc format.

ioquatix:
What is the canonical guideline for rdoc format?

I have a strong inclination to propose we move to markdown for our documentation.

lib/rack/utils.rb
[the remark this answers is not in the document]
ioquatix: Done.

jeremyevans:
> What is the canonical guideline for rdoc format?

https://docs.ruby-lang.org/en/3.1/RDoc/Markup.html#class-RDoc::Markup-label-RDoc+Markup+Reference

> I have a strong inclination to propose we move to markdown for our documentation.

I'm strongly against switching to markdown.  All documentation is currently in RDoc and should stay that way.

jeremyevans (changes requested):
Not sure why a review was rerequested. All of the yard-style `@param`/`@return` should be converted to example Ruby code.

ioquatix:
I don't know how RDoc works, can you explain it to me or point me at a guide.

ioquatix:
> All documentation is currently in RDoc and should stay that way.

How do we mark up argument types and return types?

How do we attach descriptions to arguments and return values?

This seems like a step backwards.

jeremyevans:
> Sorry, I'm going to need a bit more help regarding what this should look like, since I've never written RDoc before.

Maybe look at Rails method documentation, since Rails uses rdoc for documentation.

> How do we mark up argument types and return types?

In general, you don't need to.  With clear example code showing how the methods are used, the types are obvious in the example.

> How do we attach descriptions to arguments and return values?

To the extent that you need to, a written description is fine.

lib/rack/utils.rb
```diff
@@ -250,6 +251,33 @@ def add_cookie_to_header(header, key, value)
...
+    #
+    # Returns a map of cookie key to cookie value.
+    def parse_cookies(env)
+      parse_cookies_header env[HTTP_COOKIE]
+    end
+
+    # :call-seq:
+    #   set_cookie_header(key, value) -> encoded String
+    #   set_cookie_header("session", "id") -> "session=id"
+    #   set_cookie_header("session", {value: "id", max_age: 3600}) -> "session=id; max-age=3600"
```
jeremyevans, line 267: call-seq, if you do include it, should never include example code (such as string values).  This type of code is great, but please add it as a code block.

lib/rack/utils.rb
```diff
@@ -218,16 +218,17 @@ def select_best_encoding(available_encodings, accept_encoding)
       (encoding_candidates & available_encodings)[0]
     end
 
-    def parse_cookies(env)
-      parse_cookies_header env[HTTP_COOKIE]
-    end
-
+    # :call-seq:
+    #   parse_cookies_header(header) -> hash
```
jeremyevans, line 222: I personally don't think call-seq is necessary for these methods.  For example, for this method, the only thing different than the automatically generated call-seq is the return value, which is redundant considering you are documenting the return value in the text in a more descriptive manner (`map of cookie key to cookie value`).

What would be much more valuable would be adding examples for each method.  Such as for this method:

```ruby

parse_cookies_header('cookie_name=value; max-age=0')
# => {"cookie_name"=>"value", "max-age"=>"0"}
```

This should go after the text description.

jeremyevans (changes requested):
I think there is a lot textual descriptions here, but I would remove the call-seqs and replace them with example code after the textual descriptions.

lib/rack/utils.rb
ioquatix: Fixed.

lib/rack/utils.rb
ioquatix: I've added the example.

I'm on the fence about removing the return type.

I think it's useful.

I can also see that adding `:call-seq:` can be overhead since any change to the method results in a change to the documentation.

If you think that this cost outweighs the benefit, or there are other reasons I'm missing, I'm happy to go through and delete all the `:call-seq:` but I think I slightly prefer it.

lib/rack/utils.rb
```diff
@@ -250,6 +253,42 @@ def add_cookie_to_header(header, key, value)
...
+    # cookie attribute keys: +domain+, +max_age+, +expires+ (must be instance
+    # of +Time+), +secure+, +http_only+, +same_site+ and +value+. For more
+    # details about the interpretation of these fields, consult 
+    # [RFC6265 Section 5.2](https://datatracker.ietf.org/doc/html/rfc6265#section-5.2).
+    #
+    # An extra cookie attribute +escape_key+ can be provided to control whether
+    # or not the cookie key is URL encoded. If explicitly set to +false+, the
+    # cookie key name will not be url encoded (escaped). The default is +true+.
+    #
+    #   Rack::Utils.set_cookie_header("myname", "myvalue")
```
jeremyevans, line 286: We should probably be consistent about either always using Rack::Utils or not using it in example code.  Since none of the other examples use it, I would remove it here.

jeremyevans (changes requested):
Only one small change to the examples requested.  Can be merged without further review after that change has been made.

lib/rack/utils.rb
ioquatix: Ah yes, good catch.

ioquatix:
```suggestion
    #   set_cookie_header("myname", "myvalue")
```

lib/rack/utils.rb
```diff
@@ -250,6 +253,42 @@ def add_cookie_to_header(header, key, value)
...
+    #   # => "myname=myvalue"
+    #
+    #   Rack::Utils.set_cookie_header("myname", {value: "myvalue", max_age: 10})
```
ioquatix, line 289:
```suggestion
    #   set_cookie_header("myname", {value: "myvalue", max_age: 10})
```

jeremyevans (approved):
As I already mentioned, after the changes requested, this could be merged without further review.  I really dislike having to be forced to rereview something after I already said it could be merged without further review after changes.

PR #1851, opened 2022-04-06
Create codeql-analysis.yml

ioquatix:
Has anyone tried codeql for vulnerability scanning?

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -1,14 +1,3 @@
-# For most projects, this workflow file will not need changing; you simply need
-# to commit it to your repository.
-#
-# You may wish to alter this file to override the set of languages analyzed,
-# or to provide custom queries or build logic.
-#
-# ******** NOTE ********
-# We have attempted to detect the languages in your repository. Please check
-# the `language` matrix defined below to confirm you have the correct set of
-# supported CodeQL languages.
-#
 name: "CodeQL"
 
 on:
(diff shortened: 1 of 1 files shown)


jeremyevans:
I'm not familiar with codeql.  Unless other committers are familiar with it and found it to be helpful (finds real issues, no/few false positives), I'm negative on merging this.

ioquatix:
I actually don't know what it does but it seems interesting. @tenderlove do you know anything about it?

tenderlove:
@ioquatix maybe you should try it on a personal project first?  I'm not super familiar with the tool and I don't want to research it (I've seen it on some repos, but I can't say that it's done anything for me)

jeremyevans:
Funnily enough, I just had a ruby commit fail the CodeQL action due to problems with codeql: https://github.com/ruby/ruby/runs/5861396950

ioquatix:
What does that analysis actually tell you?

jeremyevans:
In that case, the analysis didn't run.  I don't actually know where to look for the results when it does run.

ioquatix:
Maybe this?

https://github.com/ruby/ruby/security/code-scanning

ioquatix:
This actually looks pretty useful?

https://github.com/ruby/ruby/security/code-scanning?query=branch%3Amaster+is%3Aclosed shows all the actionable items.

jeremyevans:
I checked the first 5 closed issues, and all were won't fix/false positive.  Maybe codeql does find real issues, but based on the sample selected, I don't think that looking at the remaining 46 issues is a good use of my time.

ioquatix:
@tenderlove has been investigating regex issues and that's exactly what those look like: https://github.com/ruby/ruby/security/code-scanning/54

PR #1853, opened 2022-04-07
Add missing changelog entries and minor formatting improvements.

ioquatix:
@jeremyevans I added myself to the lower case headers, because this is something I felt like I deeply cared about and pushed for and attempted several times to implement but you managed to do it in the end.

I changed both of us to link references to make it easier to read the markdown. I suggest we do this for other common authors going forward (and maybe PR/Issue links).

I've reformatted some of the code constants with backticks.

There was one missing spec change: response headers' values can be an array.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,6 @@ All notable changes to this project will be documented in this file. For info on
 - Relax validations around `Rack::Request#host` and `Rack::Request#hostname`. ([#1606](https://github.com/rack/rack/issues/1606), [@pvande](https://github.com/pvande))
 - Removed antiquated handlers: FCGI, LSWS, SCGI, Thin. ([#1658](https://github.com/rack/rack/pull/1658), [@ioquatix])
 - Removed options from `Rack::Builder.parse_file` and `Rack::Builder.load_file`. ([#1663](https://github.com/rack/rack/pull/1663), [@ioquatix])
-- HMAC argument for `Rack::Session::Cookie` doesn't accept a class constant anymore, but only a string recognized by OpenSSL (e.g. `"SHA256"`) or compatible instance (e.g. `OpenSSL::Digest.new("SHA256")`) ([#1676](https://github.com/rack/rack/pull/1676), [@bdewater](https://github.com/bdewater))
 - `Rack::HTTP_VERSION` has been removed and the `HTTP_VERSION` env setting is no longer set in the CGI and Webrick handlers. ([#970](https://github.com/rack/rack/issues/970), [@jeremyevans])
 - `Rack::Request#[]` and `#[]=` now warn even in non-verbose mode. ([#1277](https://github.com/rack/rack/issues/1277), [@jeremyevans])
 - Decrease default allowed parameter recursion level from 100 to 32. ([#1640](https://github.com/rack/rack/issues/1640), [@jeremyevans])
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -6,60 +6,62 @@ All notable changes to this project will be documented in this file. For info on
...
-- Remove `rack.multithread`/`rack.multiprocess`/`rack.run_once`. These variables generally come too late to be useful. ([#1720](https://github.com/rack/rack/pull/1720), [@ioquatix](https://github.com/ioquatix), [@jeremyevans](https://github.com/jeremyevans)))
-- Remove deprecated Rack::Request::SCHEME_WHITELIST. ([@jeremyevans](https://github.com/jeremyevans))
+- Remove `rack.multithread`/`rack.multiprocess`/`rack.run_once`. These variables generally come too late to be useful. ([#1720](https://github.com/rack/rack/pull/1720), [@ioquatix], [@jeremyevans]))
+- Remove deprecated `Rack::Request::SCHEME_WHITELIST`. ([@jeremyevans])
 
 ### Added
 
-- `Rack::Headers` added to support lower-case header keys. ([@jeremyevans](https://github.com/jeremyevans))
-- `Rack::Utils#set_cookie_header` now supports `escape_key: false` to avoid key escaping.  ([@jeremyevans](https://github.com/jeremyevans))
+- `Rack::Headers` added to support lower-case header keys. ([@jeremyevans], [@ioquatix])
```
jeremyevans, line 29: I believe only people who wrote the related code for the CHANGELOG entry should be listed.  While you did propose the idea of enforcing lower case response header keys, as far as I am aware, you did not write any code related to `Rack::Headers`, so I think you should remove yourself from this line.

CHANGELOG.md
```diff
@@ -6,60 +6,62 @@ All notable changes to this project will be documented in this file. For info on
...
 
 ### Fixed
 
-- Make Rack::MockResponse handle non-hash headers. ([#1629](https://github.com/rack/rack/issues/1629), [@jeremyevans](https://github.com/jeremyevans))
-- TempfileReaper now deletes temp files if application raises an exception. ([#1679](https://github.com/rack/rack/issues/1679), [@jeremyevans](https://github.com/jeremyevans))
-- Fix using Rack::Session::Cookie with coder: Rack::Session::Cookie::Base64::{JSON,Zip}. ([#1666](https://github.com/rack/rack/issues/1666), [@jeremyevans](https://github.com/jeremyevans))
+- Make Rack::MockResponse handle non-hash headers. ([#1629](https://github.com/rack/rack/issues/1629), [@jeremyevans])
+- TempfileReaper now deletes temp files if application raises an exception. ([#1679](https://github.com/rack/rack/issues/1679), [@jeremyevans])
+- Fix using Rack::Session::Cookie with coder: Rack::Session::Cookie::Base64::{JSON,Zip}. ([#1666](https://github.com/rack/rack/issues/1666), [@jeremyevans])
```
jeremyevans, line 61: Considering the session code was extracted, we should drop post rack 2.2 CHANGELOG entries related to it, just keeping the single existing CHANGELOG entry that session code has been extracted to `rack-session` gem.

jeremyevans (changes requested):
Thanks for working on this.  Looks mostly good, see two inline comments.

CHANGELOG.md
ioquatix: Should we move those entries somewhere else?

CHANGELOG.md
ioquatix: Okay, fair enough. I did spend a huge amount of time on this proposal though and it was really something I cared about.

ioquatix:
```suggestion
- `Rack::Headers` added to support lower-case header keys. ([@jeremyevans])
```

jeremyevans (approved):

PR #1857, opened 2022-04-08
Check the pre-condition before asserting it changed.

ioquatix:
Minor improvement to the spec - assert the negative before making the change and asserting the positive.


ioquatix:
Sorry, this was a total non-starter.

PR #1858, opened 2022-04-08
Avoid potential Array glob/unglob operations in `Rack::Builder`.

ioquatix:
I'll admit this is a bit pedantic but this syntax can be simplified and it's also better for performance (not that it's a critical path).

A -> current approach.
B -> proposed approach.

```
                   A   578.374k i/100ms
                   B   686.813k i/100ms
                   A      5.830M (± 0.8%) i/s -     29.497M in   5.060201s
                   B      6.919M (± 0.8%) i/s -     35.027M in   5.062755s

Comparison:
                   B:  6919058.9 i/s
                   A:  5829588.9 i/s - 1.19x  (± 0.00) slower
```

### DIFF
--- a/lib/rack/builder.rb
+++ b/lib/rack/builder.rb
@@ -111,7 +111,12 @@ module Rack
     # default application if +run+ is not called later.  If a block
     # is given, it is evaluated in the context of the instance.
     def initialize(default_app = nil, &block)
-      @use, @map, @run, @warmup, @freeze_app = [], nil, default_app, nil, false
+      @use = []
+      @map = nil
+      @run = default_app
+      @warmup = nil
+      @freeze_app = false
+
       instance_eval(&block) if block_given?
     end
 


ioquatix:
For reference here is my benchmark:

```ruby
require 'benchmark/ips'

class A
  # default application if +run+ is not called later.  If a block
  # is given, it is evaluated in the context of the instance.
  def initialize(default_app = nil, &block)
    @use, @map, @run, @warmup, @freeze_app = [], nil, default_app, nil, false
  end
end

class B
  # default application if +run+ is not called later.  If a block
  # is given, it is evaluated in the context of the instance.
  def initialize(default_app = nil, &block)
    @use = []
    @map = nil
    @run = default_app
    @warmup = nil
    @freeze_app = false
  end
end

require 'benchmark/ips'

Benchmark.ips do |x|
  x.report("A") {|times| while times > 0; A.new; times -= 1; end;}
  x.report("B") {|times| while times > 0; B.new; times -= 1; end;}
  x.compare!
end
```

matthewd:
> here is my benchmark

Ah, I was confused for a moment there. You need a nil: that one-line `initialize` has to create the array as its return value, which the original code does not.

Performance is still unimportant either way.. I just felt the need to defend parallel assignment's honour 😅

ioquatix:
@matthewd thanks for that comment, I appreciate being corrected and educated, and in this case it looks like you are correct:

```
                   A   684.172k i/100ms
                   B   684.250k i/100ms
                   A      6.802M (± 0.9%) i/s -     34.209M in   5.029420s
                   B      6.916M (± 0.7%) i/s -     34.897M in   5.046151s

Comparison:
                   B:  6915877.7 i/s
                   A:  6802254.5 i/s - 1.02x  (± 0.00) slower
```

Still seems like B is slightly faster but in this case it's hardly a matter to be concerned with but very interesting none the less if it was in a hot code path.

ioquatix:
@matthewd if you agree with this PR do you mind approving it? Thanks!

jeremyevans:
I'm against this.  It does not improve performance, just adds unnecessary churn.

ioquatix:
I prefer the syntax of one assignment per line, the current syntax choice is error-prone, hard to expand, and atypical.

simi:
I'm not sure what's the `churn` @jeremyevans refers to. I do consider the best practice to specify variables one per line in constructor. I have checked few random files and it seems uncommon to use this "mass assignment" in the codebase as well. Even the performance is probably not the problem here (who's going to initialize bulider milions times per second), code consistency argument seems reasonable to me.

simi (approved):

jeremyevans:
> I'm not sure what's the `churn` @jeremyevans refers to.

Churn in this context just means lines changed.  If the lines are changed just for formatting, it makes using things like `git blame` harder to use.  In general, I think lines should only be changed if there is a good reason for doing so.  I realize that "good reason" has a subjective definition.

While I'm against this, I'm not strongly against it.  If @ioquatix can find another maintainer to approve this, that's fine.

simi:
Thanks for the explanation @jeremyevans. It seems similar approach as Rails' "cosmetic changes" policy. IMHO in Rails it makes more sense since the higher "code change traffic", anyway this PR idea can be at least kept around somehow and applied with other changes to `Rack::Builder` around related lines later at some point. :handshake:

jeremyevans (approved):
Still don't think this is worth doing, but since @ioquatix feel stronglys about it, I'll remove my objection and approve this.

amatsuda:
AFAIK `def initialize` allocates an Array only when the multiple assignment becomes the final return value of the method. #1888 is the case, but I'm afraid, this one is not.

```ruby
class A
  def initialize
    @a, @b = 1, 2
  end
end

class B
  def initialize
    @a, @b = 1, 2
    nil
  end
end

GC.disable

p before: ObjectSpace.count_objects[:T_ARRAY]

A.new
p after_a: ObjectSpace.count_objects[:T_ARRAY]

B.new
p after_b: ObjectSpace.count_objects[:T_ARRAY]
```
=>
```
{:before=>1798}
{:after_a=>1799}
{:after_b=>1799}
```

Issue #1859, opened 2022-04-08
Noticing `Net::HTTP` `EOFError` in CI.


ioquatix:
I've been noticing this recently and don't really know why it's coming up.

<img width="817" alt="Screen Shot 2022-04-08 at 6 13 10 PM" src="https://user-images.githubusercontent.com/30030/162375499-8b8c60ae-929c-4ff4-a317-aa860ccad305.png">

It looks like race condition... maybe?

ioquatix:
I wonder if it's related to this:

https://bugs.ruby-lang.org/issues/14972

ioquatix:
Blowing up on this line:

https://github.com/jruby/jruby/blob/9.3.4.0/lib/ruby/stdlib/net/protocol.rb#L225

ioquatix:
@headius @eregon do you have any advice about what might be going wrong here? I've seen it affect both JRuby and TruffleRuby... I'm a bit suspicious that the busy waiting here is not very robust. Check #1860 for a proposed solution... also concerned about `Process.kill :INT, $$` which seems really unproductive. I mean, it appears to work, but it also seems like it might introduce other issues.

jeremyevans:
I think this is related to code moved into the `rackup` gem, so I'm closing this.  I'm guessing the issue still exists in the `rackup` gem, so maybe an issue should be opened in that repository.

ioquatix:
I'm fine with that.

PR #1860, opened 2022-04-08
More robust/thread safe test server startup sequence.

ioquatix:
Upon reviewing the test code, I'm not convinced the startup "join" is really that great. This uses a `Thread::Queue` to wait until the server starts up successfully (or crashes).

Potential fix for https://github.com/rack/rack/issues/1859 ?

### DIFF
--- a/test/spec_server.rb
+++ b/test/spec_server.rb
@@ -334,6 +334,26 @@ describe Rack::Server do
     c.start.must_equal [c, :started]
   end
 
+  def start_server(server)
+    begin
+      ready = Thread::Queue.new
+      thread = Thread.new do
+        begin
+          server.start { |s| ready.push(true) }
+        ensure
+          ready.close
+        end
+      end
+
+      unless ready.pop
+        thread.join
+      end
+
+      return thread
+    ensure
+      ready.close
+    end
+  end
 
   it "run a server" do
     pidfile = Tempfile.open('pidfile') { |f| break f }
@@ -349,8 +369,9 @@ describe Rack::Server do
       daemonize: false,
       server: 'webrick'
     )
-    t = Thread.new { server.start { |s| Thread.current[:server] = s } }
-    t.join(0.01) until t[:server] && t[:server].status != :Stop
+
+    t = start_server(server)
+
     body = if URI.respond_to?(:open)
              URI.open("http://localhost:#{server.options[:Port]}/") { |f| f.read }
            else
@@ -366,6 +387,7 @@ describe Rack::Server do
   it "run a secure server" do
     pidfile = Tempfile.open('pidfile') { |f| break f }
     FileUtils.rm pidfile.path
+
     server = Rack::Server.new(
       app: app,
       environment: 'none',
@@ -379,8 +401,8 @@ describe Rack::Server do
       SSLEnable: true,
       SSLCertName: [['CN', 'nobody'], ['DC', 'example']]
     )
-    t = Thread.new { server.start { |s| Thread.current[:server] = s } }
-    t.join(0.01) until t[:server] && t[:server].status != :Stop
+
+    t = start_server(server)
 
     uri = URI.parse("https://localhost:#{server.options[:Port]}/")
 


ioquatix:
Okay, this doesn't seem to solve the problem.

simi:
Is this problem only on JRuby? Maybe we can ask JRuby folks to help us in here. I was looking at this problem as well, but I wasn't able to reproduce it locally.

ioquatix:
I've asked JRuby and TruffleRuby folks for any advice.

The code looks less than ideal. It's running a server in the same process and killing it by sending itself SIGINT.

My guess is something isn't playing nice. Either thread bug or race condition, but it's impossible to tell right now. I'll mess around on this PR a bit more.

ioquatix:
I also wonder if:

```
      Port: TCPServer.open('localhost', 0){|s| s.addr[1] },
```

might be racy.

test/spec_server.rb
```diff
@@ -334,6 +334,28 @@ def start(*) [self.class, :started] end
...
+      thread = Thread.new do
+        begin
+          server.start { |s| ready.push(true) }
+        ensure
+          ready.close
+        end
+      end
+
+      unless ready.pop
+        thread.join
```
eregon, line 349: This seems very hard to follow, what is it trying to do?
Can't it just always join?

ioquatix: It only joins if it fails to start the server.

test/spec_server.rb
```diff
@@ -334,6 +334,28 @@ def start(*) [self.class, :started] end
...
+      end
+
+      sleep 10
```
eregon, line 352: This seems hacky, sleep forever or not sleep seems more reliable

ioquatix: It's a total hack to see if it's a race condition or something else.

test/spec_server.rb
```diff
@@ -349,8 +371,9 @@ def start(*) [self.class, :started] end
       daemonize: false,
       server: 'webrick'
     )
-    t = Thread.new { server.start { |s| Thread.current[:server] = s } }
-    t.join(0.01) until t[:server] && t[:server].status != :Stop
+
+    t = start_server(server)
+
     body = if URI.respond_to?(:open)
```
eregon, line 377: Why not have all that logic in the `server.start` block?
Is the server stopped when the block ends?

ioquatix: I have no idea about the deeper implementation, I just tried to move the existing shared logic into a method.

eregon:
> I also wonder if:
> 
> ```
>       Port: TCPServer.open('localhost', 0){|s| s.addr[1] },
> ```
> 
> might be racy.

Yes, it is. Can we rather pass `0`/`nil` for the Port and that then uses whatever is available?

ioquatix:
> Yes, it is. Can we rather pass 0/nil for the Port and that then uses whatever is available?

I think the problem with that is we don't know the port to connect to... well I guess we might? But there is no interface for passing it back AFAIK.

eregon:
The `Process.kill(:INT, $$)` seems fragile indeed, that basically raises an Interrupt on the main thread (unless a custom handler is set with #trap).
Maybe the test just wants to t.kill or t.raise?

ioquatix:
> Maybe the test just wants to t.kill or t.raise?

Yes I think so.

eregon:
> I think the problem with that is we don't know the port to connect to... well I guess we might? But there is no interface for passing it back AFAIK.

Should the block given to start indicate the socket is opened, and then one can get the port from socket.addr[1]?

ioquatix:
> Shouldn't the block given to start indicate the socket is opened, and then one can get the port from socket.addr[1]?

Maybe?

eregon:
The `Process.kill(:INT, $$)` is likely to trigger the custom SIGINT handler, if Rack::Server somehow sets one up?

ioquatix:
Does `trap(:int)` work with `Thread#raise(Interrupt)`?

https://github.com/rack/rack/blob/main/lib/rack/server.rb#L331-L337

eregon:
No, it's only if the signal is sent.
So this could be a bit more direct with calling `server.shutdown` directly.

ioquatix:
https://github.com/rack/rack/blob/56fe8ae3d187513aca3ba4bd905bdca2dba42d01/lib/rack/server.rb#L339

I don't think we can get any sockets out of this generally speaking.

simi:
If the problem is only within this test, wouldn't be easier to just write small bash script to start the server in one process, wait unless it is up and test request using some CLI tool? We can run it as a part of CI.

eregon:
Does `server.start { ... }` block until it processes all requests/forever (after it started and called the block)?
Seems very confusing :p

test/spec_server.rb
eregon: Another way would be to set `abort_on_exception` for that thread as the first thing it does. Then no need for the queue.

ioquatix:
Not just confusing, potentially broken... anyway, here is the webrick implementation which is what is used by this test:

https://github.com/rack/rack/blob/56fe8ae3d187513aca3ba4bd905bdca2dba42d01/lib/rack/handler/webrick.rb#L29-L45

eregon:
Does anything use Rack::Server? Seems pretty unusable.

Anyway, I'm out of ideas, but from the failing backtrace on JRuby, it sounds like it could be a problem of net/http or of that spec_server being incorrect.
FWIW I wrote a small HTTP server for testing in https://github.com/ruby/spec/blob/master/library/net/http/http/fixtures/http_server.rb, there were quite a few subtleties.

eregon:
Ah and WEBRick's shutdown is fundamentally racy and broken, so that may be part of it too, and that's why I wrote that HTTP server, WEBrick isn't reliable enough for a clean shutdown.

ioquatix:
I see so we might end up with the previous webrick server corrupting the process?

ioquatix:
Is it possible to fix WEBRick or is the problem too deeply ingrained into the design/public interface?

ioquatix:
Now I see TruffleRuby fail: https://github.com/rack/rack/runs/5884224992?check_suite_focus=true#step:8:178

image

simi:
@ioquatix I have seen this error when I was setting 2-2-stable CI as well on TruffleRuby. Seems random as well.

eregon:
That error just looks like a typical race between the exists check and the File.delete/unlink.

Re WEBrick I'd think it would be fixable, but the shutdown pipe design would need to be rethought carefully to avoid concurrent write_nonblock+close IIRC.

simi:
What about to skip this spec at JRuby for now?

ioquatix:
We could do... but @headius said he'd take a look on Monday.

Newman1515 (commented):
Got it

ioquatix:
We are going to punt on this, probably by extracting `rackup` to a separate gem.

PR #1861, opened 2022-04-08
Rework `Rack::Response#initialize` documentation to be RDoc format.

ioquatix:
@jeremyevans If you are able to review this and provide feedback that would be great.

### DIFF
--- a/lib/rack/response.rb
+++ b/lib/rack/response.rb
@@ -31,28 +31,26 @@ module Rack
     attr_accessor :length, :status, :body
     attr_reader :headers
 
-    # @deprecated Use {#headers} instead.
+    # Deprecated, use headers instead.
     def header
       warn 'Rack::Response#header is deprecated and will be removed in Rack 3.1', uplevel: 1
 
       headers
     end
 
-    # :call-seq:
-    #   Response.new(body = nil, status = 200, headers = {}) => response
+    # Initialize the response object with the specified +body+, +status+
+    # and +headers+.
     #
-    # Initialize the response object with the specified body, status
-    # and headers.
-    #
-    # If the +body+ is nil, construct an empty response object with internal
+    # If the +body+ is +nil+, construct an empty response object with internal
     # buffering.
     # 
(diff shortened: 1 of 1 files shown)


lib/rack/response.rb
```diff
@@ -38,16 +38,28 @@ def header
...
+    #
+    # Otherwise it is expected +body+ conforms to the normal requiremnets of a
+    # Rack repsonse body, typically implementing one of +#each+ (enumerable
+    # body) or +#call+ (streaming body).
+    #
+    # The +status+ defaults to +200+ which is the "OK" HTTP status code. You
+    # can provide any other valid status code.
     #
-    # Providing a body which responds to #to_str is legacy behaviour.
+    # The +headers+ must be a +Hash+ of key-value header pairs which conform to
```
ioquatix, line 60: @jeremyevans do I need to write `+Hash+` here?

lib/rack/response.rb
```diff
@@ -38,16 +38,28 @@ def header
       headers
     end
 
+    # :call-seq:
+    #   Response.new(body = nil, status = 200, headers = {}) => response
```
jeremyevans, line 42: I personally don't see the value of the call-seq here.

lib/rack/response.rb
jeremyevans: I don't think you need to.  You can use `Hash` or `hash`.  You only need to use `+Hash+` if you want to force it displayed as code.  Either way is fine with me.

jeremyevans (changes requested):
I'm OK with merging after the call-seq is removed.

lib/rack/response.rb
ioquatix: I tried out different ways and checked how it was formatted, I'm happy with it being formatted as code for now.

lib/rack/response.rb
ioquatix: Okay, removed.

jeremyevans (approved):

PR #1864, opened 2022-04-09
Ensure Rack::QueryParser::ParamsTooDeepError is inherited from RangeError.

simi:
- this is to cover backward compatibility

### DIFF
--- a/test/spec_utils.rb
+++ b/test/spec_utils.rb
@@ -133,6 +133,12 @@ describe Rack::Utils do
     Rack::Utils.parse_nested_query("foo#{"[a]" * (len - 1)}=bar")
   end
 
+  # ParamsTooDeepError was introduced in the middle of 2.2 releases
+  # and this test is here to ensure backwards compatibility
+  it "ParamsTooDeepError is inherited from originally used RangeError" do
+    (Rack::QueryParser::ParamsTooDeepError < RangeError).must_equal(true)
+  end
+
   it "parse nested query strings correctly" do
     Rack::Utils.parse_nested_query("foo").
       must_equal "foo" => nil


test/spec_utils.rb
```diff
@@ -133,6 +133,12 @@ def assert_nested_query(exp, act)
     Rack::Utils.parse_nested_query("foo#{"[a]" * (len - 1)}=bar")
   end
 
+  # ParamsTooDeepError was introduced in the middle of 2.2 releases
+  # and this test is here to ensure backwards compatibility
+  it "ParamsTooDeepError is inherited from originally used RangeError" do
+    Rack::QueryParser::ParamsTooDeepError.ancestors.must_include(RangeError)
```
ioquatix, line 139:
```suggestion
    assert Rack::QueryParser::ParamsTooDeepError < RangeError
```

https://ruby-doc.org/core-3.1.1/Module.html#method-i-3C

?

simi: I'm trying to follow the rest of the file where "spec" assertions are used.

ioquatix: I don't mind too much about this, just wondered if it was an easier way to express the relationship.

simi: What about `(Rack::QueryParser::ParamsTooDeepError < RangeError).must_equal(true)`?

ioquatix: I'm fine with that if that's the best, totally up to you.

simi: updated

ioquatix (approved):
Approved, but let's get feedback from @jeremyevans he might have a better suggestion we can learn from :)

jeremyevans (approved):
This seems unnecessary to me.  We would never change the superclass of an exception class since it would break backwards compatibility.  However, I don't object.

PR #1868, opened 2022-04-13
Fix "reject insanely long boundaries" test hang

jeremyevans:
At least on OpenBSD, this test occasionally hangs, because the
`wr.write` does not return/raise after the `rd.close` in the other
thread.  Switch to `write_nonblock` with `exception: false`, using
Thread.pass if the write would block.

With this change, the test takes less than two seconds and does
not hang.

### DIFF
--- a/test/spec_multipart.rb
+++ b/test/spec_multipart.rb
@@ -131,7 +131,11 @@ describe Rack::Multipart do
 
         # make the initial boundary a few gigs long
         longer = "0123456789" * 1024 * 1024
-        (1024 * 1024).times { wr.write(longer) }
+        (1024 * 1024).times do
+          while wr.write_nonblock(longer, exception: false) == :wait_writable
+            Thread.pass
+          end
+        end
 
         wr.write("\r\n")
         wr.write('content-disposition: form-data; name="a"; filename="a.txt"')


jeremyevans:
All CI failures are unrelated: `incompatible library version - /Users/runner/work/rack/rack/vendor/bundle/ruby/3.0.0/gems/psych-4.0.3/lib/psych.bundle (LoadError)`

ioquatix:
Is this a bug in Ruby?

jeremyevans:
> Is this a bug in Ruby?

I don't know enough to state whether it is a bug in Ruby or OpenBSD, or whether this isn't a bug and is just a behavior difference between operating systems.

ioquatix:
It looks like a bug with Ruby.

Calling write should release the GVL and cause thread context switch, what you've implemented is a busy loop calling `Thread.pass`. I understand why we can fix the test this way but it feels wrong way to me personally.

I wonder WHY this is an issue in the first place, is releasing the GVL on blocking operation insufficient?

Do you mind trying this:

```
require 'io/nonblock'

wr.nonblock = true
```

This should be very similar internal implementation as what you wrote but without busy waiting.

jeremyevans:
> It looks like a bug with Ruby.
> 
> Calling write should release the GVL and cause thread context switch, what you've implemented is a busy loop calling `Thread.pass`. I understand why we can fix the test this way but it feels wrong way to me personally.

Even if this is a Ruby bug, and we fixed it, a fix would not be backported for all previous Ruby versions supported by Rack, so I think we should still accept this PR to modify the spec.

> Do you mind trying this:
> 
> ```
> require 'io/nonblock'
> 
> wr.nonblock = true
> ```
> 
> This should be very similar internal implementation as what you wrote but without busy waiting.

Unfortunately, this doesn't work.  When I remove the `Thread.pass` implementation and use this instead. I still get the error.  When run in debug mode, I can see the Rack::Multipart::EmptyContentError being generated before the hang (same both with and without the `wr.nonblock = true`):

```
$ ruby -d test/spec_multipart.rb -n '/reject insanely long boundaries/'
Exception `LoadError' at /usr/local/lib/ruby/3.1/rubygems.rb:1325 - cannot load such file -- rubygems/defaults/operating_system
Exception `LoadError' at /usr/local/lib/ruby/3.1/rubygems.rb:1341 - cannot load such file -- rubygems/defaults/ruby
Exception `LoadError' at <internal:/usr/local/lib/ruby/3.1/rubygems/core_ext/kernel_require.rb>:85 - cannot load such file -- minitest/global_expectations/autorun
Run options: -n "/reject insanely long boundaries/" --seed 31513

# Running:

Exception `Rack::Multipart::EmptyContentError' at /home/billg/rack/lib/rack/multipart/parser.rb:366 - Rack::Multipart::EmptyContentError
```

ioquatix:
What about if you replace `Thread.pass` with `wr.wait_writable`?

ioquatix (approved):
If you have time can you please create a bug on <https://bugs.ruby-lang.org> because this doesn't look like correct behaviour to me. `wait_writable` definitely shouldn't be hanging.

jeremyevans:
> If you have time can you please create a bug on https://bugs.ruby-lang.org because this doesn't look like correct behaviour to me. `wait_writable` definitely shouldn't be hanging.

Here you go: https://bugs.ruby-lang.org/issues/18725

PR #1869, opened 2022-04-13
Remove the autoload of Rack::Handler::WEBrick

jeremyevans:
There isn't a reason the constant needs to be autoloaded.  If puma or
falcon is not installed, the try_require will still load the webrick
handler file, so register the the handler in the handler file, similar
to how puma and falcon handle it.  This makes webrick support less
special.


ioquatix (approved):
LGTM.

PR #1870, opened 2022-04-13
Remove autoloads of Multipart::UploadedFile and Multipart::Generator

jeremyevans:
With this and the webrick handler autoload removal, rack is autoload
free if you do not require rack itself, and only require the parts
of rack you are using.

Take care to avoid the circular requires when switching from
autoload to require_relative.

### DIFF
--- a/lib/rack/multipart.rb
+++ b/lib/rack/multipart.rb
@@ -8,35 +8,7 @@ module Rack
   #
   # Usually, Rack::Request#POST takes care of calling this.
   module Multipart
-    autoload :UploadedFile, 'rack/multipart/uploaded_file'
-    autoload :Generator, 'rack/multipart/generator'
-
-    EOL = "\r\n"
     MULTIPART_BOUNDARY = "AaB03x"
-    MULTIPART = %r|\Amultipart/.*boundary=\"?([^\";,]+)\"?|ni
-    TOKEN = /[^\s()<>,;:\\"\/\[\]?=]+/
-    CONDISP = /content-disposition:\s*#{TOKEN}\s*/i
-    VALUE = /"(?:\\"|[^"])*"|#{TOKEN}/
-    BROKEN_QUOTED = /^#{CONDISP}.*;\s*filename="(.*?)"(?:\s*$|\s*;\s*#{TOKEN}=)/i
-    BROKEN_UNQUOTED = /^#{CONDISP}.*;\s*filename=(#{TOKEN})/i
-    MULTIPART_CONTENT_TYPE = /content-type: (.*)#{EOL}/ni
-    MULTIPART_CONTENT_DISPOSITION = /content-disposition:.*;\s*name=(#{VALUE})/ni
-    MULTIPART_CONTENT_ID = /Content-ID:\s*([^#{EOL}]*)/ni
-    # Updated definitions from RFC 2231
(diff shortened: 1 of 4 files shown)


ioquatix:
I am fine with this, but I'd prefer moving the contents of `rack/multipart.rb` into `rack/multipart/whatever.rb` (not literally) and then using require_relative without `defined?` - does it make sense?

ioquatix:
Maybe just moving the constants into `rack/multipart/parser.rb` might be an okay name?

jeremyevans:
> Maybe just moving the constants into `rack/multipart/parser.rb` might be an okay name?

I'll check the constants.  If all are used by only Parser, that makes the most sense to me.

lib/rack/multipart.rb
```diff
@@ -60,9 +32,10 @@ def build_multipart(params, first = true)
...
       end
     end
-
   end
 end
 
 require_relative 'request' unless defined?(Rack::Request)
 require_relative 'multipart/parser'
+require_relative 'multipart/uploaded_file' unless defined?(Rack::Multipart::UploadedFile)
+require_relative 'multipart/generator' unless defined?(Rack::Multipart::Generator)
```
ioquatix, line 41: Are we able to move these to the top of the file and make them unconditional?

lib/rack/multipart/generator.rb
```diff
@@ -1,6 +1,6 @@
 # frozen_string_literal: true
 
-require_relative '../multipart'
+require_relative '../multipart' unless defined?(Rack::Multipart)
```
ioquatix, line 3: Can this now become `require_relative 'parser'`?

lib/rack/multipart/uploaded_file.rb
```diff
@@ -2,7 +2,7 @@
 
 require 'tempfile'
 require 'fileutils'
-require_relative '../multipart'
+require_relative '../multipart' unless defined?(Rack::Multipart)
```
ioquatix, line 5: Can this now become `require_relative 'parser'`?

jeremyevans:
@ioquatix I think all of your suggestions can cause circular require issues.   The issue is these four files can be required in any order: `rack/multipart`, `rack/multipart/parser`, `rack/multipart/generator`, `rack/multipart/uploaded_file`.  I found that even the previous implementation (before this PR, and before the most recent commit in this PR) could result in circular require warnings, so I pushed another commit to fix that.

I'm not against moving things around when it makes sense to do so, but I don't think moving non-parser parts into `lib/rack/multipart/parser.rb` makes sense.

If you want, open an alternative PR for this, but make sure the following commands do not emit warnings:

```
ruby -w -I lib -r rack/multipart-e ''
ruby -w -I lib -r rack/multipart/parser -e ''
ruby -w -I lib -r rack/multipart/generator -e ''
ruby -w -I lib -r rack/multipart/uploaded_file -e ''
```

ioquatix (approved):

ioquatix:
I'm happy to merge this work and I'll try to make a follow up PR.

ioquatix:
It looks like our cached gems might be incorrect.

jeremyevans:
> It looks like our cached gems might be incorrect.

I think #1871 may fix that. It hasn't been merged yet.

ioquatix:
I've just merged the CI `bundler-cache` option on main.

PR #1871, opened 2022-04-13
Add cache version to CI.

simi:


ioquatix:
This looks good to me, but just thinking would using something like https://github.com/socketry/async/blob/main/.github/workflows/development.yml#L40 easier/less config?

simi:
@ioquatix yes, that next step I'm going to test. It can handle cache versioning as well on its own if needed.

simi:
https://github.com/ruby/setup-ruby/blob/ebaea52cb20fea395b0904125276395e37183dac/action.yml#L34-L39

.github/workflows/development.yml
```diff
@@ -1,6 +1,8 @@
 name: Development
 
 on: [push, pull_request]
+env:
+  ruby-cache-version: 'v1'
```
ioquatix, line 5: Looking good, do you want to introduce `with: cache-version: 'v1'`? Is that the plan, is it required?

I do similar test setups and don't seem to have a problem, since we use different gemfiles.

simi: Ohh, that is just leftover, I'll remove it for now.

ioquatix (approved):
LGTM. Well done.

PR #1872, opened 2022-04-14
Remove `Rack::Multipart` dependency on `Rack::Request`.

ioquatix:


test/spec_request.rb
```diff
@@ -1867,8 +1872,6 @@ def initialize(req)
       end
 
       def delegate?; true; end
-
-      def env; @req.env.dup; end
```
ioquatix, line 1871: @tenderlove since you wrote this code do you mind explaining why you are calling dup here if you recall the reason?

ioquatix:
@jeremyevans my understanding is even if you approve a PR, if there is open unresolved conversations, they will block merging. So my understanding is, you can add feedback, approve the PR, and it can be merged provided those are resolved.

test/spec_request.rb
tenderlove: IIRC the idea was that if we return the hash directly, then people can muck with it without going through getters / setters.  Looks like this is in a spec though so, I'm not sure why we need it. 😅

tenderlove (approved):

test/spec_request.rb
ioquatix: Yeah, IMHO it was hiding bugs.

PR #1873, opened 2022-04-15
Add new query parser exception to Utils.

simi:
I was looking into porting new exception to Rails error handling and I have spotted Rails uses exceptions defined at `Rack::Utils`. Recently added `ParamsTooDeepError` is not defined in there. I'm adding it to keep it consistent.

https://github.com/rails/rails/blob/c9e5057de4be7e0f4365ff9c2d411cffc39c3746/actionpack/lib/action_dispatch/http/request.rb#L383

_:information_source: If approved, I'll backport this to `2-2-stable` as well._

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -18,6 +18,7 @@ module Rack
   module Utils
     ParameterTypeError = QueryParser::ParameterTypeError
     InvalidParameterError = QueryParser::InvalidParameterError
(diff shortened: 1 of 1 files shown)


ioquatix:
This looks okay to me from a consistency POV, but why are we doing this for all of them? I mean, I guess I approve, but at first glance it seems a bit odd.

ioquatix (approved):

jeremyevans (approved):

PR #1874, opened 2022-04-15
Limit file extension size used for Tempfile.

simi:
Using file with enormous file extension size can raise an exception like:

```
Errno::ENAMETOOLONG: Filename too long @ rb_sysopen - /tmp/RackMultipart20220414-6-lm7dab.xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx
```

It should be safe to cut the file extension size to 20 (we can pick also bigger number if needed). That way tempfile name size is predictable.

I have checked few online lists of file extensions and it seems limit of 20 is ok. The longest common file extension seems to be `crdownload` (10 characters).

_:information_source: I would like to backport this to `2-2-stable` if approved._

### DIFF
--- a/test/spec_request.rb
+++ b/test/spec_request.rb
@@ -1132,6 +1132,31 @@ EOF
     end
   end
 
+  it 'cuts long file extension' do
+    begin
+      long_extension = "a" * 1_000
+      filename= "testfile.#{long_extension}"
+      data = "--AaB03x\r\nContent-Type: text/plain\r\nContent-Disposition: attachment; name=#{SecureRandom.hex(10)}; filename=#{filename}\r\n\r\ncontents\r\n"
+      data += "--AaB03x--\r"
+
+      file = nil
+      options = {
+        "CONTENT_TYPE" => "multipart/form-data; boundary=AaB03x",
+        "CONTENT_LENGTH" => data.length.to_s,
+        Rack::RACK_MULTIPART_TEMPFILE_FACTORY => lambda { |filename, content_type|
+          file = Rack::Multipart::Parser::TEMPFILE_FACTORY.call(filename, content_type)
+        },
+        :input => StringIO.new(data)
+      }
+
+      request = make_request Rack::MockRequest.env_for("/", options)
+      request.POST
+
+      processed_extension = ::File.extname(file.path)
+      assert_equal Rack::Multipart::Parser::EXTENSION_SIZE_LIMIT + 1, processed_extension.size
+    end
+  end
+
   it "parse big multipart form data" do
     input = <<EOF
 --AaB03x\r
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -15,9 +15,18 @@ module Rack
 
     class Parser
       BUFSIZE = 1_048_576
+      EXTENSION_SIZE_LIMIT = 20
       TEXT_PLAIN = "text/plain"
       TEMPFILE_FACTORY = lambda { |filename, content_type|
-        Tempfile.new(["RackMultipart", ::File.extname(filename.gsub("\0", '%00'))])
+        extension = ::File.extname(filename.gsub("\0", '%00'))
+
+        # filenames with long extensions can result into invalid tempfile name due to filename size
+        # to keep it safe only last 20 characters from filename are used to find extension
+        if extension.size > EXTENSION_SIZE_LIMIT
+          extension = "." + extension.slice(extension.size - EXTENSION_SIZE_LIMIT, EXTENSION_SIZE_LIMIT)
+        end
+
+        Tempfile.new(["RackMultipart", extension])
       }
 
       BOUNDARY_REGEX = /\A([^\n]*(?:\n|\Z))/


test/spec_request.rb
```diff
@@ -1132,6 +1132,30 @@ def initialize(*)
...
+        Rack::RACK_MULTIPART_TEMPFILE_FACTORY => lambda { |filename, content_type|
+          file = Rack::Multipart::Parser::TEMPFILE_FACTORY.call(filename, content_type)
+        },
+        :input => StringIO.new(data)
+      }
+
+      request = make_request Rack::MockRequest.env_for("/", options)
+      request.POST
+
+      refute_includes(file.path, "ThisIsVeryLongExtension")
```
jeremyevans, line 1154: This seems fragile (what if we change 20 to 32 or 16)?  Since this is being written to avoid the ENAMETOOLONG, in general we should write the spec that fails with ENAMETOOLONG, and then check that the change to the multipart parser fixes it.

simi: Thanks for the review. I have find out also one problem I have introduced (missing `.` in stripped extension). Code changed. Test now fails with `Errno::ENAMETOOLONG` without change in parser.

test/spec_request.rb
```diff
@@ -1132,6 +1132,31 @@ def initialize(*)
...
+          file = Rack::Multipart::Parser::TEMPFILE_FACTORY.call(filename, content_type)
+        },
+        :input => StringIO.new(data)
+      }
+
+      request = make_request Rack::MockRequest.env_for("/", options)
+      request.POST
+
+      processed_extension = ::File.extname(file.path)
+      assert_equal processed_extension.size, Rack::Multipart::Parser::EXTENSION_SIZE_LIMIT + 1
```
jeremyevans, line 1156: I think this is reversed, as expected comes before actual (https://github.com/seattlerb/minitest/blob/14dbfed4c2dd6b7b00913ca00e085faf5afbca91/lib/minitest/assertions.rb#L216).  I assume the intent is that `processed_extension.size` is equal to `Rack::Multipart::Parser::EXTENSION_SIZE_LIMIT + 1` and not the other way around.

jeremyevans (changes requested):
One minor change requested, then this looks good to me.

test/spec_request.rb
simi: right, fixed

simi:
@jeremyevans are you ok with backport?

jeremyevans:
> @jeremyevans are you ok with backport?

I think this is OK to backport.

ioquatix:
On what OS is this a problem?

simi:
> On what OS is this a problem?

official Ruby docker Alpine image -> Alpine Linux 3.15

simi:
:information_source: CI error seems unrelated (known JRuby CI problem).

ioquatix:
After discussing it, here is my suggestion/advice.

I don't think any user data, file extension, or otherwise, should make it to the file system directory list by default. So I think we agree here, and the matter is now just compatibility.

Including the file extension has potentially become an ossified standard. I think this can be a potential security issue but initial investigation looks okay - i.e. it looks like `File.extname` protects us.

I would be much more comfortable with us doing something like what Jeremy has done in the past. Add a configuration option which we can eventually deprecate/default to a sane/secure default.

```
Rack::Multipart.file_name_extension_limit = 20 # file name extension is limited to this many characters - better than current
Rack::Multipart.file_name_extension_limit = nil # no file name extension - best/secure by default.
Rack::Multipart.file_name = "RackMultipart" # the name of the file?
```

I would suggest we have a mechanism for warning/informing people about these defaults, because it seems like right now we want to introduce a default (e.g. 20), but utimately we'd prefer to have no file extension at all.

This isn't an area I've spent much time in the code so it would be up to yourself to propose the best mechanism for configuration. The above is just an idea.

simi:
I don't see any big benefit in making this configurable. As an alternative approach, we can just skip the extension if threshold is reached or even we can just rescue `Errno::ENAMETOOLONG` and try without extension.

ioquatix:
> I don't see any big benefit in making this configurable.

The benefit is preserving backwards compatibility while providing a way forward with the new and improved code path.

jeremyevans:
I'm not sure if trying to avoid the use of an extension for the temporary file is a good long term goal.  Seems like that would break cases where the user is taking the extension of the temporary file and using `Rack::Mime` to get the mime type?  I suppose you can tell them to get the extension some other way, but that seems like pointlessly breaking their code.  I checked and the no extension supported by `Rack::Mime` is longer than 10 characters.

ioquatix:
Can't someone use the file extension of the provided file name, rather than the tempfile path name?

jeremyevans:
> Can't someone use the file extension of the provided file name, rather than the tempfile path name?

They could.  That's what I was discussing when I said "I suppose you can tell them to get the extension some other way, but that seems like pointlessly breaking their code."

ioquatix:
> "I suppose you can tell them to get the extension some other way, but that seems like pointlessly breaking their code."

That's assuming that anyone with a 20 character file extension expects to do anything meaningful with it. Remember we are talking about cases which are already exceeding anything longer than what is in the mime table.

Also, Are people actually using `tempfile.path` rather than the filename given in the params itself? I can't imagine anyone is doing that. Do you know any examples?

simi:
I got also probably final idea. What about to provide something like "rack_invalid_extension" or some similar constant for extension failing the sanity check? That way it would be clear if inspecting the problem manually, extension is not missing (so it is clear extension was provided) and there is again no need to modify the tempfile name based on original extension.

simi:
I was thinking about this again and I'm definitelly sold on my latest comment. I'll update PR. 🙏 If welcomed, I can make extension size configurable.

ioquatix:
I think it's totally reasonable.

simi:
:information_source: I got some alternative ideas, I'm going to explore more on this and open new PR later.

ioquatix:
@simi did we ever fix this?

simi:
Nope @ioquatix, still problem. I was thinking about introducing lazy multipart parsing in Rails instead to be able to react in "user land" to similar problems. There is a lot of potential problems in here. For example `.` isn't extension separator every time, it can be just part of the filename (I have seen filenames sanitized by replacing spaces with dots :fearful:).

If we would like to fix this still in Rack, I would go for random tempfile names and keeping original name around to be able to do checks (mime type?) on that instead of tempfile name.

ioquatix:
This issue cropped up again in our production environment.

I think you are right, the file name on disk should not be part of the "content type" or any other user supplied data = security risk. Let's have a think about how to deal with this. I like your proposal.

PR #1876, opened 2022-04-23
Only digest body when `body.to_ary` is present

st0012:
Since https://github.com/rack/rack/pull/1748, the result of `body.to_ary` will be sent directly into `digest_body` for generating etag digest. But `ActionDispatch::Response::RackBody#to_ary` always returns `nil`, which will cause `NoMethodError` with this approach. Therefore, I think it makes sense to check if the result of `body.to_ary` is present before trying to generate digest from it.

(It wasn't an issue before #1748 because Rack used to pass the `body` (responds to `#each`) to `digest_body` instead).

### DIFF
--- a/lib/rack/etag.rb
+++ b/lib/rack/etag.rb
@@ -29,9 +29,10 @@ module Rack
       status, headers, body = @app.call(env)
 
       if etag_status?(status) && body.respond_to?(:to_ary) && !skip_caching?(headers)
-        body = body.to_ary
-        digest = digest_body(body)
-        headers[ETAG_STRING] = %(W/"#{digest}") if digest
+        if body = body.to_ary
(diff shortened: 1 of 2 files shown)


jeremyevans:
The Rack 3 SPEC is that if body responds to `to_ary`, then `to_ary` needs to return an array:

```
If the Body responds to +to_ary+, it must return an Array whose
contents are identical to that produced by calling +each+.
```

We don't want everything that interacts with the body have to defensively check to see whether `to_ary` returns an array.  So `ActionDispatch::Response::RackBody` needs to be fixed to work with Rack 3.

st0012:
@jeremyevans Sorry I didn't see that. I'll leave Rails team to make it compatible then, thx!

PR #1880, opened 2022-04-27
Fix Rack::Request headers usage.

ioquatix:
As discussed previously there is a mismatch between `Rack::Response#get/set_headers` and `Rack::Request#get/set_headers`.

Specifically, `Rack::Response` headers are actual HTTP headers, while `Rack::Request` headers are env key pairs.

This PR is one option for changing `Rack::Request#get/set_headers` to be consistent with the semantic of HTTP headers. I think this is the best option. The proposed methods are backwards compatible but emit warnings when used with things which are likely to be env keys.

The biggest limitation is that not all HTTP header keys are allowed, specifically headers with periods (`.`) are considered env keys, even though periods are considered valid `token` characters and thus valid header keys. This allows `rack.internal.keys` to pass correctly as non-headers retaining compatibility (following the spec which specifies one period must be included in application specific keys).

I believe this PR is 100% backwards compatible and I've split it into two parts: (commit 1) is just the change itself and emits lots of warnings, and (commit 2) is reworking all Rack code to use `env` directly where it makes sense.

I believe this PR will allow Rails to remove their "patch" https://github.com/rails/rails/blob/f95c0b7e96eb36bc3efc0c5beffbb9e84ea664e4/actionpack/lib/action_dispatch/http/headers.rb#L121-L129 althought in theory it remains fully compatible.

### DIFF
--- a/lib/rack/common_logger.rb
+++ b/lib/rack/common_logger.rb
@@ -54,18 +54,18 @@ module Rack
 
       msg = FORMAT % [
         request.ip || "-",
-        request.get_header("REMOTE_USER") || "-",
+        request.env['REMOTE_USER'] || "-",
         Time.now.strftime("%d/%b/%Y:%H:%M:%S %z"),
         request.request_method,
         request.script_name,
         request.path_info,
         request.query_string.empty? ? "" : "?#{request.query_string}",
-        request.get_header(SERVER_PROTOCOL),
+        request.env[SERVER_PROTOCOL],
         status.to_s[0..3],
         length,
         Utils.clock_time - began_at ]
 
-      logger = @logger || request.get_header(RACK_ERRORS)
+      logger = @logger || request.env[RACK_ERRORS]
       # Standard library logger doesn't support write but it supports << which actually
       # calls to write on the log device without formatting
       if logger.respond_to?(:write)
--- a/lib/rack/files.rb
+++ b/lib/rack/files.rb
@@ -78,7 +78,7 @@ module Rack
         return [200, { 'allow' => ALLOW_HEADER, CONTENT_LENGTH => '0' }, []]
       end
       last_modified = ::File.mtime(path).httpdate
-      return [304, {}, []] if request.get_header('HTTP_IF_MODIFIED_SINCE') == last_modified
+      return [304, {}, []] if request.get_header('if-modified-since') == last_modified
 
       headers = { "last-modified" => last_modified }
       mime_type = mime_type path, @default_mime
@@ -90,7 +90,7 @@ module Rack
       status = 200
       size = filesize path
 
-      ranges = Rack::Utils.get_byte_ranges(request.get_header('HTTP_RANGE'), size)
+      ranges = Rack::Utils.get_byte_ranges(request.get_header('range'), size)
       if ranges.nil?
         # No ranges:
         ranges = [0..size - 1]
--- a/lib/rack/method_override.rb
+++ b/lib/rack/method_override.rb
@@ -48,9 +48,9 @@ module Rack
     def method_override_param(req)
       req.POST[METHOD_OVERRIDE_PARAM_KEY]
     rescue Utils::InvalidParameterError, Utils::ParameterTypeError
-      req.get_header(RACK_ERRORS).puts "Invalid or incomplete POST params"
+      req.env[RACK_ERRORS].puts "Invalid or incomplete POST params"
     rescue EOFError
-      req.get_header(RACK_ERRORS).puts "Bad request content body"
+      req.env[RACK_ERRORS].puts "Bad request content body"
     end
   end
 end
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -90,24 +90,65 @@ module Rack
         super()
       end
 
-      # Predicate method to test to see if `name` has been set as request
-      # specific data
+      # Predicate method to test to see if a header of the given +name+ exists.
       def has_header?(name)
         @env.key? env_key(name)
       end
 
-      # Get a request specific value for `name`.
+      # Get the value from the request environment with the specified +name+, returning +nil+ if it doesn't exist.
+      #
+      # When invoked with a lower case (canonical) header name, it will be
+      # access the HTTP header of that name from the request environment.
+      #
+      #   request.get_header('accept') # => '*/*'
+      #
+      # When invoked with a name which contains periods +.+ (e.g. 'rack.input'),
+      # it willm directly read from the request environment. This is considered
+      # legacy behaviour and you should use env directly.
+      #
+      #   # Legacy usage:
+      #   request.get_header('rack.input')
+      #
+      #   # Use env directly:
+      #   request.env['rack.input']
+      #
+      # Note that +rack.input+ is also valid HTTP header name and therefore at
+      # this time it's impossible to use headers that contain periods.
+      #
+      # When invoked with a name which matches a CGI variable (as defined in 
+      # RFC3875), it will be directly read from the request environment. This
+      # is considered legacy behaviour and you should access env directly.
+      #
+      #   # Legacy usage:
+      #   request.get_header('PATH_INFO')
+      #
+      #   # Use env directly:
+      #   request.env['PATH_INFO']
+      #
       def get_header(name)
         @env[env_key(name)]
       end
 
-      # If a block is given, it yields to the block if the value hasn't been set
-      # on the request.
+      # Fetch an HTTP header using using +Hash#fetch+. Yields the internal CGI
+      # variable key which you should use if you want to set the default
+      # value. See get_header details on how name lookup works.
       def fetch_header(name, &block)
         @env.fetch(env_key(name), &block)
       end
 
-      # Loops through each key / value pair in the request specific data.
+      # Loops through each key / value pair in the request HTTP headers. Yields
+      # header names in their canonical form.
+      #
+      #   request.each do |key, value|
+      #     [key, value] # => ["accept", "*/*"]
+      #   end
+      #
+      # Note that CGI style headers cannot represent all possible HTTP headers
+      # and thus the reconstruction of HTTP headers is intriniscally lossy.
+      # The actual behaviour will depend on your server configuration but
+      # generally, +_+ characters will be mapped to +-+ characters. Incoming
+      # headers with underscores (+_+) will thus be indistinguisable from those
+      # with dashes (+-+).
       def each_header(&block)
         @env.each do |key, value|
           if name = header_name(key)
@@ -116,38 +157,39 @@ module Rack
         end
       end
 
-      # Set a request specific value for `name` to `v`
+      # Set a request HTTP header value for +name+ to +value+ overwriting any
+      # existing value.
       def set_header(name, value)
         @env[env_key(name)] = value
       end
 
-      # Add a header that may have multiple values.
+      # Add a request HTTP header that may have multiple values.
       #
-      # Example:
-      #   request.add_header 'Accept', 'image/png'
-      #   request.add_header 'Accept', '*/*'
+      #   request.add_header 'accept', 'image/png'
+      #   request.add_header 'accept', '*/*'
       #
-      #   assert_equal 'image/png,*/*', request.get_header('Accept')
+      #   request.get_header('accept') 
+      #   # => 'image/png,*/*'
       #
-      # http://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2
+      # See get_header for details on name mapping.
       def add_header(name, value)
         key = env_key(name)
 
         if value.nil?
           @env[key]
         elsif current = env[key]
-          case current
-          when Array
-            current << value
+          if current.empty?
+            @env[key] = value
           else
-            @env[key] = [current, value]
+            @env[key] = "#{current},#{value}"
           end
         else
           @env[key] = value
         end
       end
 
-      # Delete a request specific value for `name`.
+      # Delete a request HTTP header with the specified +name+. See get_header
+      # for details on name mapping.
       def delete_header(name)
         @env.delete(env_key(name))
       end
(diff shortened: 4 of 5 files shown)


ioquatix:
There are some small details which I think we need to flesh out too.

Specifically, I noticed that the linter does not do any validation on incoming request `HTTP_` header keys/values.

I suspect it might be a good idea to add this.

Specifically, I've added (and should probably revert, but left it in for this specific discussion):

```
- set_header key, "#{get_header key},#{v}"
+ @env[key] = [current, value]
```

Generally speaking there is only one response header whihc does not allow comma separated values: `set-cookie`. On the request side, all headers (AFAIK) support comma separated form. That being said, splitting headers can be tricky and I wonder if we want to have a more general model, like `get_header_values` or something.

Regardless of what we do here, we should update `Rack::Lint` to validate that `HTTP_` header keys conform to the expected format.

ioquatix:
@jeremyevans what part of this PR is not compatible with existing code? Basically, all the specs pass.

jeremyevans:
The plan that this PR sets forth (removing the warnings and breaking code in Rack 3.1) is not backwards compatible. I think if the goal of the PR is to break backwards compatibility later, the PR should be considered as not backwards compatible, even if it technically retains immediate compatibility via code that warns that it will be removed in a later version.

Considering you like the interface in #1881, and that approach is backwards compatible, I think we should use that approach.  We can warn in the documentation against using the existing `*_header` methods because they don't operate the way you would expect based on the name.  Personally, I don't think the benefits of deprecating and removing outweigh the cost of backwards compatibility loss, but if we have to, we could deprecate now and remove in Rack 4.

In terms of implementation in this PR, `content-type` and `content-length` headers are not handled correctly.  Those are special cased in the Rack SPEC to not be prefixed via `HTTP_` in `env`, but `header_name` and `env_name` don't handle that.  For example, `get_header('content-type')` and `get_header('content-length')` should work without warning (since they were submitted as HTTP headers), but looking at the implementation, I'm guessing they currently warn.  Likewise, `each_header` will not yield `content-type` and `content-length` headers, even if they were submitted to the webserver.  I also don't see a reason for `HTTP_HEADER_PATTERN` to exclude `A-Z`, considering you are going to upcase anyway.

ioquatix:
I essentially see this as a bug fix PR where as your proposal is a new feature. Rails is already working around this behaviour. The implementation here should be safely composable with existing Rails releases and also allow them with minimal changes to remove their patch over `get/set_headers`.

>  The plan that this PR sets forth (removing the warnings and breaking code in Rack 3.1) is not backwards compatible.

With that interpretation, what about all the other deprecated functionality that is slated to be removed in 3.1? Quite a bit of of that is not backwards compatible.

> we could deprecate now and remove in Rack 4.

I'm fine with that.

> In terms of implementation in this PR, content-type and content-length headers are not handled correctly

Define "correctly"...? From https://www.rfc-editor.org/rfc/rfc3875.html#section-4.1.18

```
   The server is not required to create meta-variables for all the
   header fields that it receives.  In particular, it SHOULD remove any
   header fields carrying authentication information, such as
   'Authorization'; or that are available to the script in other
   variables, such as 'Content-Length' and 'Content-Type'.  The server
   MAY remove header fields that relate solely to client-side
   communication issues, such as 'Connection'.
```

Based on this interpretation from the CGI RFC, these are no longer headers: "remove any header fields... Content-Length... Content-Type...". Not sure what the ideal mechanism is here, but after the point they are extracted, they should no longer be present in the "headers" according to the RFC.

> Considering you like the interface in https://github.com/rack/rack/pull/1881, and that approach is backwards compatible

It doesn't solve the problem being addressed here. I like the general idea though. I think we should discuss it for Rack 3.1 Perfectly happy to punt this PR to a later Rack release too. It seems like something we should try to fix though.

jeremyevans:
> I essentially see this as a bug fix PR where as your proposal is a new feature. Rails is already working around this behaviour. The implementation here should be safely composable with existing Rails releases and also allow them with minimal changes to remove their patch over `get/set_headers`.

I do not see this as a bug fix API.  The existing `*_header` methods are poorly named, but they were never intended to do the type of mangling you are doing.  What you are doing is implementing a backwards incompatible new feature using existing methods.

> > The plan that this PR sets forth (removing the warnings and breaking code in Rack 3.1) is not backwards compatible.
> 
> With that interpretation, what about all the other deprecated functionality that is slated to be removed in 3.1? Quite a bit of of that is not backwards compatible.

We've agreed to make many backwards incompatible changes.  We knew when doing so that removal was not backwards compatible.  It's not like we accepted deprecation with a plan to remove and considered it a backwards compatible solution.
 
> > we could deprecate now and remove in Rack 4.
> 
> I'm fine with that.

To be clear, you are fine with merging #1881, closing this, and deprecating the existing `*_header` methods?  If not, can you please state what approach you are fine with?

> > In terms of implementation in this PR, content-type and content-length headers are not handled correctly
> 
> Define "correctly"...? From https://www.rfc-editor.org/rfc/rfc3875.html#section-4.1.18
> 
> ```
>    The server is not required to create meta-variables for all the
>    header fields that it receives.  In particular, it SHOULD remove any
>    header fields carrying authentication information, such as
>    'Authorization'; or that are available to the script in other
>    variables, such as 'Content-Length' and 'Content-Type'.  The server
>    MAY remove header fields that relate solely to client-side
>    communication issues, such as 'Connection'.
> ```
> 
> Based on this interpretation from the CGI RFC, these are no longer headers: "remove any header fields... Content-Length... Content-Type...". Not sure what the ideal mechanism is here, but after the point they are extracted, they should no longer be present in the "headers" according to the RFC.

My reading of this is that `HTTP_CONTENT_LENGTH` and `HTTP_CONTENT_TYPE` meta variables should not be created, because the values are already set in `CONTENT_LENGTH` and `CONTENT_TYPE`.  This matches what SPEC states. Also, SPEC only requires RFC3875 conformance for `HTTP_` variables, it doesn't say anything about RFC3875 conformance for non-`HTTP_` variables.

I think almost all rack users would assume that `content-length` and `content-type`, if submitted as request headers, with values present in `env` in the `CONTENT_LENGTH` and `CONTENT_TYPE` keys, would be included in `headers`. Do you honestly think that users expect that `content-length` and `content-type` would not show up as headers?

> > Considering you like the interface in #1881, and that approach is backwards compatible
> 
> It doesn't solve the problem being addressed here. I like the general idea though. I think we should discuss it for Rack 3.1 Perfectly happy to punt this PR to a later Rack release too. It seems like something we should try to fix though.

The problem being addressed here is that users don't have a good way to do `header` to `env` and `env` to `header` mangling.  You propose modifying `*_header` methods to do such mangling, in a backwards incompatible manner.  I propose a new `headers` method to do such mangling, in a backwards compatible manner.  It's exactly the same problem, IMO.

ioquatix:
> I do not see this as a bug fix API. The existing *_header methods are poorly named, but they were never intended to do the type of mangling you are doing.

Okay, either (1) change the name to match the function or (2) fix the function to match the name. To me, it's both bug. Since (1) introduce major compatibility issues, a careful implementation of (2) seems like the only way forward.

I think it's clear from the comment made 7 (!) years ago that there is confusion around what these methods are supposed to do, with the original documentation giving examples using raw HTTP header names which seems to imply the name is correct and the implementation is wrong.

Since generally, CGI variables, headers and metadata keys can be distinct (with some very minor/obscure overlap) we can definitely fix these methods to be both compatible and do the right thing.

> I think almost all rack users would assume that content-length and content-type, if submitted as request headers, with values present in env in the CONTENT_LENGTH and CONTENT_TYPE keys, would be included in headers. Do you honestly think that users expect that content-length and content-type would not show up as headers?

`content-length` is not a header in HTTP/2 and in theory it might not even be a header in CGI / HTTP/1 (chunked encoding could be buffered for example). So my opinion is `get_header('content-length')` should not be used as it's not reliable (and can introduce security issues too, imagine someone provides `content_length` header for example).

I find `content-type` more peculiar. I can't imagine any security issues coming out of `content-type` spoofing that aren't already present from the original provider of the request. I therefore don't know why the CGI spec chose to expose `CONTENT_TYPE` instead of `HTTP_CONTENT_TYPE`. The CGI RFC outlines several headers like this (authorization being another one) and I wonder to what extent we should try to map them back? Probably we need a general reverse mapping of CGI variable to header for some subset of CGI variables that are documented to be extracted from headers.

> The problem being addressed here is that users don't have a good way to do header to env and env to header mangling. You propose modifying *_header methods to do such mangling, in a backwards incompatible manner. I propose a new headers method to do such mangling, in a backwards compatible manner. It's exactly the same problem, IMO.

Your solution does not fix the original source of the problem, it just introduce a new interface without fixing the original issue. In order to deal with that, you'd then need to introduce blanket deprecations on `Rack::Request#*_header` which is worse than what I've proposed here which is specific, targeted deprecations based on specific questionable usage patterns.

I think we all agree that `Rack::Request#*_header` and `Rack::Response#*_header` methods should deal with headers. This PR tries to make a step towards that goal. If we want to retain compatibility indefinitely, we can just remove the warning and leave it as proposed here, which is completely compatible with all existing specs, external tests, etc.

jeremyevans:
> > I do not see this as a bug fix API. The existing *_header methods are poorly named, but they were never intended to do the type of mangling you are doing.
> 
> Okay, either (1) change the name to match the function or (2) fix the function to match the name. To me, it's both bug. Since (1) introduce major compatibility issues, a careful implementation of (2) seems like the only way forward.

I prefer (3), realize that backwards compatibility is important, don't break it, and instead introduce a better API for solving the same problem.

> I think it's clear from the comment made 7 (!) years ago that there is confusion around what these methods are supposed to do, with the original documentation giving examples using raw HTTP header names which seems to imply the name is correct and the implementation is wrong.

It's clear from when these methods were first introduced in ef5546ce355f180c68bd0ce0d326b5e2c2d610ca that they were not intended to deal with HTTP specific headers, but with `arbitrary data associated witht 
 the request object`.  Look at the commit message and tests in the commit.

> Since generally, CGI variables, headers and metadata keys can be distinct (with some very minor/obscure overlap) we can definitely fix these methods to be both compatible and do the right thing.

You can 100% not fix these methods to be backwards compatible and "do the right thing".  Either the methods mangle, or they don't.  If they mangle, they are not backwards compatible.

> > I think almost all rack users would assume that content-length and content-type, if submitted as request headers, with values present in env in the CONTENT_LENGTH and CONTENT_TYPE keys, would be included in headers. Do you honestly think that users expect that content-length and content-type would not show up as headers?
> 
> `content-length` is not a header in HTTP/2 and in theory it might not even be a header in CGI / HTTP/1 (chunked encoding could be buffered for example). So my opinion is `get_header('content-length')` should not be used as it's not reliable (and can introduce security issues too, imagine someone provides `content_length` header for example).

99% of users are using HTTP/1 and would expect that `get_header('content-length')` returns the length submitted in the `content-length` header.

The `content-length` vs `content_length` is not specific to this header, but to all headers.  Most front-end webservers like Nginx would block/ignore `content_length`, and I wouldn't have a problem updating SPEC to say that webservers should ignore headers with underscores.

> I find `content-type` more peculiar. I can't imagine any security issues coming out of `content-type` spoofing that aren't already present from the original provider of the request. I therefore don't know why the CGI spec chose to expose `CONTENT_TYPE` instead of `HTTP_CONTENT_TYPE`. The CGI RFC outlines several headers like this (authorization being another one) and I wonder to what extent we should try to map them back? Probably we need a general reverse mapping of CGI variable to header for some subset of CGI variables that are documented to be extracted from headers.

`content-length` and `content-type` are the only headers treated specially by not being prefixed with `HTTP_`.  Here's some examples from popular web servers:

* https://github.com/defunkt/unicorn/blob/2c347116305338710331d238fefa23f00e98cf54/ext/unicorn_http/common_field_optimization.h#L91
* https://github.com/puma/puma/blob/34c3c59809d39ed9c03cf8845b6461ec18642b88/ext/puma_http11/puma_http11.c#L99
* https://github.com/puma/puma/blob/15b079439222e489915380021b9dca7b9fa5ff65/ext/puma_http11/org/jruby/puma/Http11.java#L40

> > The problem being addressed here is that users don't have a good way to do header to env and env to header mangling. You propose modifying *_header methods to do such mangling, in a backwards incompatible manner. I propose a new headers method to do such mangling, in a backwards compatible manner. It's exactly the same problem, IMO.
> 
> Your solution does not fix the original source of the problem, it just introduce a new interface without fixing the original issue. In order to deal with that, you'd then need to introduce blanket deprecations on `Rack::Request#*_header` which is worse than what I've proposed here which is specific, targeted deprecations based on specific questionable usage patterns.

I disagree.  My solution fixes the problem of allowing users to easily deal with HTTP headers.  You want to modify `*_header` to do that, in a backwards incompatible manner, instead of adding a new backwards compatible method to do it.

You are fixated on the fact that `*_header` isn't a good name.  The fact that it isn't a good name is not a bug, and this is not a bug fix. 

We don't need to introduce blanket deprecations at all.  We can keep the existing methods, and just indicate via documentation that it is a bad idea to use them in new code.  Old code stays working, new code uses `headers`, nobody's application breaks just to satisfy a foolish desire for consistency.

> I think we all agree that `Rack::Request#*_header` and `Rack::Response#*_header` methods should deal with headers.

No, we don't all agree about that.  We probably all agree that absent backwards compatibility concerns, it would make more sense for `Request#*_header` to deal with HTTP headers.  However, backwards compatibility is important.  We should aim to be like Ruby itself, and only break backwards compatibility when the benefits outweigh the costs.

> This PR tries to make a step towards that goal. If we want to retain compatibility indefinitely, we can just remove the warning and leave it as proposed here, which is completely compatible with all existing specs, external tests, etc.

This approach is backwards incompatible if you change the behavior that currently warns to always treating as an HTTP header.  Until you do that, the semantics of the `*_header` methods are questionable.  For example, if I submit a request with a `path-info` header, `get_header('path-info')` does not do what I want.  So this approach is either backwards incompatible or has even worse semantics than just a bad method name.

ioquatix:
Thanks for the great investigation. I don't think consistency is a foolish desire. Being inconsistent leads to real world bugs, hard to use code, and security issues.

By choosing (3) we don't fix any existing code paths and end up with confusing/inconsistent interfaces - so I think (3) needs to be combined with warnings & a deprecation strategy.

> This approach is backwards incompatible if you change the behavior that currently warns to always treating as an HTTP header. 

Yes.

> Until you do that, the semantics of the *_header methods are questionable. 

The semantics are already questionable and will remain questionable unless changed. This is an incremental step towards being consistent with other methods of the same name in `Rack::Response` and being less questionable. The middle step proposed here is backwards compatible. The end goal is not.

Not sure if there are any alternatives to fixing this except doing nothing at all to the implementation and adding documentation saying it's questionable.

Do you mind explaining what your expectations are for `get_header('path-info')` because I'm not sure how that wouldn't work, but in any case, let me add a spec for it.

ioquatix:
Ok, never mind about `get_header('path-info')`, I've fixed that case and added a spec, thanks for pointing it out.

jeremyevans:
My preference is to leave code as-is and discourage use in documentation.  However, I'm open to deprecation and removal.  I'm against changing behavior in a backwards-incompatible manner.

ioquatix:
Okay, I agree. I am specifically in favour of deprecation and removal, but maybe that is far off. I agree with introducing some kind of `Rack::Request.headers` and maybe we can do the same for `Rack::Response.headers` for consistency's sake.

PR #1881, opened 2022-04-27
Add Rack::Request#headers for getting access to HTTP headers

jeremyevans:
In addition to being backwards compatible, I think this offers
a nicer API:

```ruby
  request.headers['host']           # env['HTTP_HOST']
  request.headers['content-type']   # env['CONTENT_TYPE']
  request.headers['content-length'] # env['CONTENT_LENGTH']

  request.headers['allow'] = 'GET'     # env['HTTP_ALLOW'] = 'GET'
  request.headers.add('allow', 'POST') # env['HTTP_ALLOW'] #=> %w'GET POST'

  request.headers.has_key?('allow')
  request.headers.fetch('allow')
  request.headers.delete('allow')

  request.headers.each{|key, value| }
  request.headers.to_h
```

You can compare this to a backwards incompatible proposal to modify
the related *_header methods:

```ruby
  # This proposal
  request.headers['host']

  # Backwards incompatible alternative
  request.get_header('host')
```

This only allocates potentially one Rack::Request::Headers object
per request, on the first call to Rack::Request#headers.  I don't
think this will be a performance issue.  Any code desiring maximum
performance should access env directly.

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -83,6 +83,88 @@ module Rack
       v
     end
 
+    def headers
+      @headers ||= Headers.new(@env)
+    end
+
+    class Headers
+      def initialize(env)
+        @env = env
+      end
+
+      def [](k)
+        @env[header_to_env_key(k)]
+      end
+
+      def []=(k, v)
+        @env[header_to_env_key(k)] = v
+      end
+
+      def add(k, v)
+        k = header_to_env_key(k)
+        case existing = @env[k]
+        when nil
+          @env[k] = v
+        when String
+          @env[k] = [existing, v]
+        when Array
+          existing << v
+        end
+      end
+
+      def delete(k)
+        @env.delete(header_to_env_key(k))
+      end
+
+      def each
+        return to_enum(:each) unless block_given?
+
+        @env.each do |k, v|
+          next unless k = env_to_header_key(k)
+          yield k, v
+        end
+      end
+
+      def fetch(k, &block)
+        @env.fetch(header_to_env_key(k), &block)
+      end
+
+      def has_key?(k)
+        @env.has_key?(header_to_env_key(k))
+      end
+
+      def to_h
+        h = {}
+        each{|k, v| h[k] = v}
+        h
+      end
+
+      private
+
+      def env_to_header_key(k)
+        case k
+        when /\AHTTP_/
+          k = k[5..-1]
+          k.downcase!
+          k.tr!('_', '-')
+          k
+        when "CONTENT_LENGTH", "CONTENT_TYPE"
+          k = k.downcase
+          k.tr!('_', '-')
+          k
+        end
+      end
+
+      def header_to_env_key(k)
+        k = k.upcase
+        k.tr!('-', '_')
+        unless k == "CONTENT_LENGTH" || k == "CONTENT_TYPE"
+          k = "HTTP_#{k}"
+        end
+        k
+      end
+    end
+
     module Env
       # The environment of the request.
       attr_reader :env
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ All notable changes to this project will be documented in this file. For info on
 - Add `deflaters` option to `Rack::Deflater` to enable custom compression algorithms like zstd. ([#2168](https://github.com/rack/rack/issues/2168), [@alexanderadam](https://github.com/alexanderadam))
 - Add `Rack::Request#prefetch?` for identifying requests with `Sec-Purpose: prefetch` header set. ([#2405](https://github.com/rack/rack/pull/2405), [@glaszig](https://github.com/glaszig))
 - Add `rack.request.trusted_proxy` environment key to indicate whether the request is coming from a trusted proxy.
+- Add `Rack::Request#headers` for simpler access to request headers by header name. ([#1881](https://github.com/rack/rack/pull/1881), [@jeremyevans](https://github.com/jeremyevans))
 
 ### Changed
 
--- a/test/spec_request.rb
+++ b/test/spec_request.rb
@@ -1282,6 +1282,147 @@ EOF
     req2.params.must_equal({})
   end
 
+  it "allows access to specific HTTP header via headers[]" do
+    e = Rack::MockRequest.env_for("",
+      "HTTP_FOO_BAR" => "baz",
+      "CONTENT_LENGTH" => "1",
+      "CONTENT_TYPE" => "text/html")
+    req = make_request(e)
+
+    req.headers['foo-bar'].must_equal 'baz'
+    req.headers['FOO-BAR'].must_equal 'baz'
+    req.headers['content-type'].must_equal 'text/html'
+    req.headers['Content-Length'].must_equal '1'
+    req.headers['bar-foo'].must_be_nil
+  end
+
+  it "allows access to specific HTTP header via headers.fetch" do
+    e = Rack::MockRequest.env_for("",
+      "HTTP_FOO_BAR" => "baz",
+      "CONTENT_LENGTH" => "1",
+      "CONTENT_TYPE" => "text/html")
+    req = make_request(e)
+
+    req.headers.fetch('foo-bar').must_equal 'baz'
+    req.headers.fetch('FOO-BAR').must_equal 'baz'
+    req.headers.fetch('content-type').must_equal 'text/html'
+    req.headers.fetch('Content-Length').must_equal '1'
+
+    proc{req.headers.fetch('bar-foo')}.must_raise KeyError
+    req.headers.fetch('bar-foo'){'1'}.must_equal '1'
+  end
+
+  it "allows checking for specific HTTP header via headers.has_key?" do
+    e = Rack::MockRequest.env_for("",
+      "HTTP_FOO_BAR" => "baz",
+      "CONTENT_LENGTH" => "1",
+      "CONTENT_TYPE" => "text/html")
+    req = make_request(e)
+
+    req.headers.has_key?('foo-bar').must_equal true
+    req.headers.has_key?('FOO-BAR').must_equal true
+    req.headers.has_key?('content-type').must_equal true
+    req.headers.has_key?('Content-Length').must_equal true
+    req.headers.has_key?('bar-foo').must_equal false
+  end
+
+  it "allows deleting specific HTTP header via headers.delete" do
+    e = Rack::MockRequest.env_for("",
+      "HTTP_FOO_BAR" => "baz",
+      "CONTENT_LENGTH" => "1",
+      "CONTENT_TYPE" => "text/html")
+    req = make_request(e)
+
+    req.headers.delete('foo-bar').must_equal 'baz'
+    req.headers.delete('content-type').must_equal 'text/html'
+    req.headers.delete('Content-Length').must_equal '1'
+    req.headers.delete('bar-foo').must_be_nil
+
+    e.has_key?("HTTP_FOO_BAR").must_equal false
+    e.has_key?("CONTENT_LENGTH").must_equal false
+    e.has_key?("CONTENT_TYPE").must_equal false
+
+    req.headers.has_key?('foo-bar').must_equal false
+    req.headers.has_key?('content-type').must_equal false
+    req.headers.has_key?('Content-Length').must_equal false
+  end
+
+  it "set HTTP header in env via headers[]=" do
+    e = Rack::MockRequest.env_for("",
+      "CONTENT_LENGTH" => "12",
+      "CONTENT_TYPE" => "text/plain")
+    req = make_request(e)
+
+    req.headers['foo-bar'] = 'baz'
+    req.headers['Content-Type'] = 'text/html'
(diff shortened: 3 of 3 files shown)


ioquatix:
I personally like this kind of interface, but I'm not sure if this fixes the existing semantic issues with `get/set_header`.

jeremyevans:
> I personally like this kind of interface, but I'm not sure if this fixes the existing semantic issues with `get/set_header`.

It doesn't attempt to fix the semantic issues.  We should document `*_header` methods as only for backwards compatibility, and recommend using `headers` instead.  If this is accepted, I'll handle the documentation updates, and remove internal use of the `*_header` methods (Similar to your second commit in #1880).

This allows current code that uses the `*_header` methods to continue to work. New code can use the `headers` method for an even nicer interface than previously provided by `*_headers`.  There should be no semantic issues with the `headers` method.

As long as you like this interface, I think it is a much better path forward than #1880.

ioquatix:
How do we deprecate the existing methods without introducing the same backwards incompatible change? Basically this doesn't seem to solve the original problem.

I'm also interested in a solution where some hypothetical `Rack::Request::Headers` can internally multiplex around different `env` shapes. I think `Rack::Request::Headers` is insufficient for that purpose, but it's very close to being a good solution. Basically, it's too narrow. The problem that should be solved here is, "given an env, give me a structured interface".

By adopting this PR, it will be more tricky later on to hide this implementation detail. That's why I prefer direct methods on `Rack::Request` rather than a separate object/instance/interface.

For that reason, I'd like more time to think about the implications of this design and don't want to rush into it for Rack 3.0.

As I've outlined several times, my main concern is the semantic inconsistency between `Request#*_headers` and `Response#*headers` and this PR does not solve that problem.

jeremyevans:
> How do we deprecate the existing methods without introducing the same backwards incompatible change? Basically this doesn't seem to solve the original problem.

As I've stated, I don't think we should deprecate.  The benefits are very small, and the costs are significant.  It's better to just document the methods as existing only for backwards compatibility, and introduce this as a replacement.

> I'm also interested in a solution where some hypothetical `Rack::Request::Headers` can internally multiplex around different `env` shapes. I think `Rack::Request::Headers` is insufficient for that purpose, but it's very close to being a good solution. Basically, it's too narrow. The problem that should be solved here is, "given an env, give me a structured interface".

Assuming `Rack::Request` can deal with different `env` shapes, this is trivial because `#headers` could return an instance of a different class with the same API, for that specific `env` shape.  So that seems simple to address.

> By adopting this PR, it will be more tricky later on to hide this implementation detail. That's why I prefer direct methods on `Rack::Request` rather than a separate object/instance/interface.

I disagree.  The polymorphic approach described (instance of a separate class) would solve it very simply.  Even if you didn't want to use that approach, any way that Rack::Request handled separate shapes inside the `*_headers` method could be handled by `Rack::Request::Headers`.

> For that reason, I'd like more time to think about the implications of this design and don't want to rush into it for Rack 3.0.

That's fine, as long as you aren't pushing for any `*_header` method changes in Rack 3.0.

ioquatix:
I'm fine punting both PRs for later in the Rack 3.x release cycle, but I'd like to hear input from others before making a final decision. I think introducing a new interface has a slightly higher burden of review than fixing an existing one.

> I disagree. The polymorphic approach described (instance of a separate class) would solve it very simply. Even if you didn't want to use that approach, any way that Rack::Request handled separate shapes inside the *_headers method could be handled by Rack::Request::Headers.

I agree with you generally that the approach you've described is the right one, but I think we want to be a bit more nuanced with the actual interface and naming, since this is about wrapping an environment (of different shapes), not specifically about headers. I would prefer to consider introducing something like `Rack::Environment` which has methods for `path`, `method`, `scheme`, `get_header`, `set_header`, `each_header` and a few others. The scope of this interface is a big bigger than just "headers". I like the general idea, but I think we should be a little bit more ambitious and as such I think we should punt it to a later release rather than holding up a 3.0 release.

jeremyevans:
> I'm fine punting both PRs for later in the Rack 3.x release cycle, but I'd like to hear input from others before making a final decision. I think introducing a new interface has a slightly higher burden of review than fixing an existing one.

IMO, you are not fixing an existing interface.  What you are proposing here is not a bug fix.  You are implementing a new feature by changing the behavior of existing methods in a backwards incompatible manner.  IMO, backwards incompatible changes should require a much higher burden of review than backwards compatible new features.

That being said, I agree with you that we should probably focus on Rack 3.0 and revisit these PRs after 3.0 is released.

ahx:
I like the interface, because it talks like a Hash. I am currently working on a piece of code that picks certain headers from a request and I think this interface would help me a lot making unit tests and implementation easier to understand, because I can just write something like `my_code(headers)` where `headers` is `request.headers` or just a normal Hash in unit test.

ioquatix:
Yeah, I too would like to see us fix this in 3.1 - I'm okay with this proposal, but I'd like to bounce it around a little more before making a final decision.

jeremyevans:
I'm still in favor of merging this.

ioquatix:
After consideration, I'm okay with this PR, and I think it's a reasonable step forward.

However, I'd like to discuss some of the semantics.

I'm okay with this implementation mutating the underlying environment. However, I wonder if we should disallow that. The reason is as follows:

`falcon` and potentially other servers already have an internal list of headers, e.g. a key-value mapping or some rich object that provides the proposed interface, e.g. `request.headers['foo'] => 'bar'`. Right now, those servers must unwrap all of those headers into Rack's `env`, e.g. `HTTP_FOO`. It would be nice if there was an efficient way to provide this, e.g.

```
# request in this context is from the server itself, not rack.
def call(request)
  env = {
    # ...
    'rack.headers' => request.headers
  }

  @app.call(env)
end
```

We should specify the `rack.headers` interface as part of the SPEC, rather than an implementation detail of `Rack::Request`.

Later on, it might be possible for an application to say "I don't want `HTTP_` headers, `rack.headers` is sufficient". This will be a non-trivial performance improvement.

If `request.headers['foo'] = 'bar'` ossifies to manipulating `env['HTTP_FOO']`, it will be more tricky to implement such an improvement later on.

Do you have any thoughts on how we can achieve this?

jeremyevans:
> I'm okay with this implementation mutating the underlying environment. However, I wonder if we should disallow that. The reason is as follows:
> 
> `falcon` and potentially other servers already have an internal list of headers, e.g. a key-value mapping or some rich object that provides the proposed interface, e.g. `request.headers['foo'] => 'bar'`. Right now, those servers must unwrap all of those headers into Rack's `env`, e.g. `HTTP_FOO`. It would be nice if there was an efficient way to provide this, e.g.
> 
> ```
> # request in this context is from the server itself, not rack.
> def call(request)
>   env = {
>     # ...
>     'rack.headers' => request.headers
>   }
> 
>   @app.call(env)
> end
> ```

I'm OK memoizing the Headers instance in `env`, and using it if already present.  I would use `rack.request.headers` as the key, similar to the other `rack.request.*` keys.

> We should specify the `rack.headers` interface as part of the SPEC, rather than an implementation detail of `Rack::Request`.

Considering the interface, I'm reluctant to add `Rack::Headers` to SPEC.  If we are considering SPEC changes, I would make `rack.headers` optional, and if present, it must be a plain unfrozen hash, and must have lower case string keys (treat request headers similar to response headers).  `Rack::Request#headers` could use `rack.headers` if present using a different class, modifying the `rack.headers` hash and not `env` itself.

> Later on, it might be possible for an application to say "I don't want `HTTP_` headers, `rack.headers` is sufficient". This will be a non-trivial performance improvement.

I'm not sure how much this would help, at least in HTTP/1.  You would need to do the header key string mangling anyway for case insensitive access (or build some case insensitive index).  I guess in HTTP 2/3 this may be significantly faster as the keys must be lower case already and shouldn't need mangling.  I think the `rack.headers` idea may solve that.

> If `request.headers['foo'] = 'bar'` ossifies to manipulating `env['HTTP_FOO']`, it will be more tricky to implement such an improvement later on.
> 
> Do you have any thoughts on how we can achieve this?

Would the `rack.headers` SPEC idea as an unfrozen hash with lower case string keys solve that?

ioquatix:
I am okay with having a separate key for `rack.headers` and `rack.request.headers`.

Regarding memoization, I'm slightly against the internals of `Rack::Request` performing memoization as we've experienced significant pain in the past. I'd prefer we work towards a mechanism for memoizing the entire `Rack::Request` instance later, and instead keep the simplicity of your approach here.

Regarding `rack.headers`, I'd prefer if we focused on the interface rather the object type. The reason is, I'd prefer if I can substitute in my own, already existing, rich object for headers.

I think it would be fine to have `#to_h` or `#to_hash` to provide access to the hash of the style you are considering.

If you don't mind, maybe I'll cut a PR for `rack.headers` in the spec, and we can decide if it makes sense or not, or you may have an alternative suggestion.

jeremyevans:
> I am okay with having a separate key for `rack.headers` and `rack.request.headers`.
> 
> Regarding memoization, I'm slightly against the internals of `Rack::Request` performing memoization as we've experienced significant pain in the past. I'd prefer we work towards a mechanism for memoizing the entire `Rack::Request` instance later, and instead keep the simplicity of your approach here.

I'm against memozing `Rack::Request` instances in `env`. It's been tried before, and removed because of the  problems that resulted (eefbed89c4ece749e889132012d0f67cd87926a8, 0bf8d8e840ae226d6efed8af345feb67a10f6aa6).

> Regarding `rack.headers`, I'd prefer if we focused on the interface rather the object type. The reason is, I'd prefer if I can substitute in my own, already existing, rich object for headers.

I'm against that approach.  The reason for Rack's success IMO is keeping the interface simple, and "unfrozen hash with lowercase string keys" is much simpler than specifying all the methods the object must implement and how they must behave.  We use this type specification for response headers, using it for request headers seems natural.

> If you don't mind, maybe I'll cut a PR for `rack.headers` in the spec, and we can decide if it makes sense or not, or you may have an alternative suggestion.

Certainly you can propose such a change.  However, I'm not sure any change to `env` for request headers is worth the backwards compatibility issues, even my proposal of `rack.headers` as a plain hash.

ioquatix:
> I'm against memozing Rack::Request instances in env. It's been tried before, and removed because of the problems that resulted...

I agree, and I think the same problems will manifest if we try to memoize `rack.request.headers` without extremely careful consideration.

> I'm against that approach. The reason for Rack's success IMO is keeping the interface simple, and "unfrozen hash with lowercase string keys" is much simpler than specifying all the methods the object must implement and how they must behave.

Unfortunately that will still incur the cost of mapping all headers which I'm trying to avoid. Do you have any other ideas to avoid that overhead?

jeremyevans:
> > I'm against memozing Rack::Request instances in env. It's been tried before, and removed because of the problems that resulted...
> 
> I agree, and I think the same problems will manifest if we try to memoize `rack.request.headers` without extremely careful consideration.

Then we should just keep this PR as it is, and only memoize in the Request instance and not in env.

> > I'm against that approach. The reason for Rack's success IMO is keeping the interface simple, and "unfrozen hash with lowercase string keys" is much simpler than specifying all the methods the object must implement and how they must behave.
> 
> Unfortunately that will still incur the cost of mapping all headers which I'm trying to avoid. Do you have any other ideas to avoid that overhead?

Not at present.  However, Rack has always valued interface simplicity over performance, and I don't think that should change.  Note that you'll have to mangle all headers anyway in HTTP/1 to get O(1) case-insensitive access, and almost all Rack usage is with HTTP/1.

ioquatix:
> Rack has always valued interface simplicity over performance

We have "valued simplicity" to the point where doing some extremely basic optimisations and ergonomic improvements is almost impossible and requires years of planning. It's not the kind of simplicity I like.

jeremyevans:
I'm not saying we shouldn't take performance seriously.  However, in terms of priorities, I don't think performance should be first.  I think the priorities for Rack should be:

1. Backwards compatibility
2. Interface simplicity
3. Performance

ioquatix:
I think we should defer this work until after 3.1 is released, maybe we can aim for a 3.2 release with this feature, or something similar.

I'd also like to understand how your proposed interface works with `trailers` (headers received after the request is fully read).

ioquatix:
After considering <https://github.com/puma/puma/issues/3397>, we might want to expose values as an array of strings - especially true for multi-value headers, e.g. `accept`, `accept-encoding` - etc - this is extremely convenient and how Falcon represents headers internally anyway.

lib/rack/request.rb
```diff
@@ -79,6 +79,88 @@ def delete_param(k)
...
+
+      def [](k)
+        @env[header_to_env_key(k)]
+      end
+
+      def []=(k, v)
+        @env[header_to_env_key(k)] = v
+      end
+
+      def add(k, v)
```
thadeu, line 99: Why do you `add` instead `set` name method @jeremyevans ?

jeremyevans: Because this appends and does not overwrite.

samuel-williams-shopify (approved):

samuel-williams-shopify:
Can you add notes to the change log.

PR #1882, opened 2022-04-27
Add methods for HTTP status codes 406 and 408

jgarber623:
This commit adds two new predicate methods to the `Rack::Response` class:

- `not_acceptable?` which returns true on HTTP 406 Not Acceptable
- `request_timeout?` which returns true on HTTP 408 Request Timeout

Links to MDN documentation for each status code:

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/406
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/408

### DIFF
--- a/lib/rack/response.rb
+++ b/lib/rack/response.rb
@@ -190,6 +190,8 @@ module Rack
       def forbidden?;           status == 403;                        end
       def not_found?;           status == 404;                        end
       def method_not_allowed?;  status == 405;                        end
+      def not_acceptable?;      status == 406;                        end
+      def request_timeout?;     status == 408;                        end
       def precondition_failed?; status == 412;                        end
       def unprocessable?;       status == 422;                        end
 
--- a/test/spec_response.rb
+++ b/test/spec_response.rb
@@ -503,6 +503,16 @@ describe Rack::Response do
     res.must_be :client_error?
     res.must_be :method_not_allowed?
 
+    res.status = 406
+    res.wont_be :successful?
+    res.must_be :client_error?
+    res.must_be :not_acceptable?
+
+    res.status = 408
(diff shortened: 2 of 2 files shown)


ioquatix (approved):

simi (approved):

jeremyevans (commented):
Seems like there are many status codes that do not have methods for them.  I'm not in favor of adding methods for additional status codes, but I don't object to this PR.

In terms of a possible alternative to this type of method proliferation, I recommend something like `response.status?(:request_timeout)` and/or `response.status_name # => :request_timeout`.  This way you only need one method instead of one method per status.  We could probably use `Rack::Utils::HTTP_STATUS_CODES` to construct a new hash to store the symbols to return/recognize.

ioquatix:
@jeremyevans I understand your concern and the use case has not been described, so it looks purely for completeness of the existing interface which is why I'm okay with it.

I think there is a performance cost to using symbols, i.e. status code -> symbol -> comparison which is not the case for these explicit predicates. The predicates are also nice for testing. e.g. `expect(response).to be_request_timeout` etc.

As you do not object to this, I'll merge it as I think it's acceptable. However, going forward maybe the `#status_name` interface could be convenient.

jgarber623:
Thanks everyone for the consideration, thoughtful critique, and approval!

> The predicates are also nice for testing. e.g. expect(response).to be_request_timeout etc.

As @ioquatix [noted above](#issuecomment-1111711546), this use case (testing using RSpec's ✨ magic ✨ predicate matchers) is exactly how I stumbled upon these methods _not_ existing in Rack's source. Apologies for not providing a use case in my original PR description.

PR #1884, opened 2022-05-03
Update and rename SECURITY_POLICY.md to SECURITY.md

ioquatix:
Rename the security policy to the recognised name by GitHub. According to https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository#about-security-policies this is what the file should be called.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,4 +1,4 @@
-# Rack maintenance
+# Security Policy
 
 ## Supported versions
 


jeremyevans (approved):
Doesn't look like we reference this file in anything but the ChangeLog, so this seems fine.

ioquatix:
Now it shows up here: https://github.com/rack/rack/security/policy

PR #1885, opened 2022-05-05
Remove dependency on rack-session, and development dependency on minitest-sprint

jeremyevans:


jeremyevans:
Looks like we need to update at least rack-attack to require rack-session as a dependency for the external specs to pass (maybe only development dependency if it is only used in the specs).

ioquatix:
Can you make a PR to rack-attack too? We can then merge them in order so they both pass.

ioquatix (approved):

ioquatix:
https://github.com/rack/rack-attack/pull/581

jeremyevans:
Looks like that rack-attack PR doesn't work. I'm not sure how rack-session interacts with rack 2.2 and below.

I pushed another commit to this PR that modifies the external rack-attack tests to also require rack-session. I think this is a better approach.

jeremyevans:
Looks like the last commit worked, so I'll merge this.

ioquatix:
Ultimately I think external dependencies will need to sniff rack 2.x or 3.x and pull in rack-session as needed.

PR #1886, opened 2022-05-06
Significantly improve test coverage

jeremyevans:
Also fix bugs, simplify code, and remove dead code noticed in coverage testing.
See individual commits for changes to each file.

Before:

```
3311 relevant lines, 3243 lines covered and 68 lines missed. ( 97.95% )
1140 total branches, 1021 branches covered and 119 branches missed. ( 89.56% )
```

After:

```
3286 relevant lines, 3282 lines covered and 4 lines missed. ( 99.88% )
1114 total branches, 1067 branches covered and 47 branches missed. ( 95.78% )
```

I plan to finish adding coverage testing next week, but it's probably best to get
these changes reviewed and committed before then.

### DIFF
--- a/lib/rack/deflater.rb
+++ b/lib/rack/deflater.rb
@@ -57,7 +57,7 @@ module Rack
 
       # Set the Vary HTTP header.
       vary = headers["vary"].to_s.split(",").map(&:strip)
-      unless vary.include?("*") || vary.include?(/accept-encoding/i)
+      unless vary.include?("*") || vary.any?{|v| v.downcase == 'accept-encoding'}
         headers["vary"] = vary.push("Accept-Encoding").join(",")
       end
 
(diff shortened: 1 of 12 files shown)


lib/rack/deflater.rb
```diff
@@ -99,7 +100,7 @@ def each(&block)
         gzip = ::Zlib::GzipWriter.new(self)
         gzip.mtime = @mtime if @mtime
         # @body.each is equivalent to @body.gets (slow)
-        if @body.is_a? ::File
```
ioquatix, line 102: We might have to revisit how this works because it seems like it won't handle streaming bodies correctly, and in any case we probably don't want to handle streaming bodies with the implementation here.

lib/rack/request.rb
```diff
@@ -780,4 +752,6 @@ def x_forwarded_proto_priority
   end
 end
 
+# :nocov:
 require_relative 'multipart' unless defined?(Rack::Multipart)
```
ioquatix, line 756: Do we still need this `defined?`?

ioquatix: In fact can we move this to the top of the file?

ioquatix (approved):
I reviewed the main code changes, and I didn't look too much at the tests since I agree with the implementation and the tests reflect that.

lib/rack/request.rb
jeremyevans: I don't think so.  I believe that will introduce circular requires.

lib/rack/deflater.rb
jeremyevans: Agreed.

lib/rack/request.rb
ioquatix: Multipart no longer depends on request though?

jeremyevans: Ah, in that case, it should be safe to move this to the top and remove the `defined?`

ioquatix: Yes, I think so.

PR #1888, opened 2022-05-10
Reduce Array object allocation in Rack::ContentType's initializer

amatsuda:
Returning multiple values from a method allocates an Array instance, even if the method is the `initialize` method.
Here's a patch fixing that kind of unnecessary object allocation in Rack::ContentType.


ioquatix:
@amatsuda is this the same as https://github.com/rack/rack/pull/1858

PR #1899, opened 2022-05-25
Refactor multipart boundary parsing

jeremyevans:
Remove the use of BOUNDARY_REGEX, as well as the @boundary, @full_boundary,
and @end_boundary instance variables.  Replace them with the existing
@body_regex, which already contains the boundary. Tweak body_regex slightly
so the preceding EOF is required if we aren't at the start of the buffer,
since boundaries must be on their own lines.

When looking for the boundary, scan until using @body_regex (i.e. until you
found the boundary).  Differentiate the full_boundary from the end_boundary
by seeing if the scanned data ends with \r\n (full boundary will,
end_boundary must end with --). If no boundary is found, clear the buffer
and try again, for behavior similar to previous.

For the fast forward state, if the first boundary found is the end boundary,
ignore it and keep scanning.  We could alternatively raise an exception, not
sure if that is better.  The previous behavior of treating the end boundary
as the starting boundary was definitely wrong, though.

Add some comments describing this boundary parsing behavior (which was my
initial intent when working on this).

Speeds up test/spec_multipart.rb more than 6x in my testing.

### DIFF
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -45,8 +45,6 @@ module Rack
         Tempfile.new(["RackMultipart", ::File.extname(filename.gsub("\0", '%00'))])
       }
 
-      BOUNDARY_REGEX = /\A([^\n]*(?:\n|\Z))/
-
       class BoundedIO # :nodoc:
         def initialize(io, content_length)
           @io             = io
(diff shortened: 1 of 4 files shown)


ioquatix (approved):
This looks okay to me but I have not reviewed it thoroughly.

tenderlove (approved):
Nice. It looks great to me!

PR #1901, opened 2022-05-25
Skip a test that is flaky on JRuby

jeremyevans:
It's already skipped on Ruby 2.4 and 2.5, so it's not like we
expect this to pass in all cases anyway.  If it is fixed on JRuby,
we can renable.  But it's better to not generate false negatives
in CI for unrelated changes.


ioquatix (approved):

simi:
nice :partying_face:

PR #1904, opened 2022-05-27
Hex escape unprintable bytes in common logger

jeremyevans:
When using \x prefix, users would expect hex escaping.


Newman1515 (commented):
Approve

ioquatix (approved):

ioquatix:
I feel like there should be a better way of generating this output.

PR #1910, opened 2022-06-06
chore: Included githubactions in the dependabot config

naveensrinivasan:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


jeremyevans:
Rack does not have an runtime dependencies.  Of the 4 development dependencies, 3 are not limited in any way, and we should consider switching `minitest` from `~> 5.0` to `>= 5.0`.

I personally don't think there is any advantage to enabling this, but I'm not going to object if another committer wants to enable it.

Note that this message is either automated or it's being copied and pasted into different repositories.  I got basically the same pull request for Sequel.

ioquatix:
@naveensrinivasan My understanding is this relates to GitHub Actions configuration not packages in the gemfile, is that correct?

naveensrinivasan:
> @naveensrinivasan My understanding is this relates to GitHub Actions configuration not packages in the gemfile, is that correct?

Yes. It is only for GitHub actions.

ioquatix (approved):
Thanks.

PR #1913, opened 2022-06-20
Expect additional optional version segment in version test.

simi:
Last release (2.2.3.1) has added fourth segment of version string which wasn't expected by one test related to `rackup -v`. I have updated test to expect fourth (optional) segment.

CI fail was introduced by https://github.com/rack/rack/commit/f705eaced23649fb29222adbd70d5f658998953b.

### DIFF
--- a/test/spec_server.rb
+++ b/test/spec_server.rb
@@ -162,7 +162,7 @@ describe Rack::Server do
   end
 
   it "support -v option to get version" do
-    test_options_server('-v').must_match(/\ARack \d\.\d \(Release: \d+\.\d+\.\d+\)\nexited\z/)
+    test_options_server('-v').must_match(/\ARack \d\.\d \(Release: \d+\.\d+\.\d+(\.\d+)?\)\nexited\z/)
   end
 
   it "warn for invalid --profile-mode option" do


ioquatix:
@tenderlove just wondering if this was intentional in the sense that from what I can see, there has never been a security release in the past with a 4th version component.

@simi I'm assuming this breaks all stable branches with this test in place?

ioquatix:
@simi what I mean, is all stable 2.0, 2.1, 2.2 etc branches didn't expect a 4-component version?

simi:
@ioquatix this spec was introduced by https://github.com/rack/rack/commit/eab436ef186d4a4163e76396ebad7840237a7052.

by GitHub those are related versions: [v2.2.2](https://github.com/rack/rack/releases/tag/v2.2.2)  [v2.2.1](https://github.com/rack/rack/releases/tag/v2.2.1) [2.2.3.1](https://github.com/rack/rack/releases/tag/2.2.3.1) [2.2.3](https://github.com/rack/rack/releases/tag/2.2.3) [2.2.0](https://github.com/rack/rack/releases/tag/2.2.0)

ioquatix:
I confirmed with @tenderlove that this is normative, so we will need to fix CI.

PR #1918, opened 2022-06-30
Link PR #1748 in CHANGELOG

swrobel:
I had missed this note in there so I figured I might as well add the link to the PR number


ioquatix:
My understanding is we haven't been linking to PRs from the spec changes list. @jeremyevans WDYT?

Also, we might want to have dedicated sections for "new major features" etc.

jeremyevans:
No other changes to the SPEC section have PR numbers.  If we want to add this, we should at least be consistent and add PR numbers to all of the changes in the SPEC section.

swrobel:
It's confusing to me because the linked PR affects more than just the SPEC. Obviously I ended up here because like many others I was curious why Rails 7's `send_stream` wasn't working and I wanted to see where the fix was...

PR #1919, opened 2022-06-30
Better support for case insensitive headers.

ioquatix:
https://github.com/rails/sprockets/issues/746#issuecomment-1168459287


jeremyevans (changes requested):
Looks mostly good.

The `lib/rack/response.rb` change is unrelated.  `CACHE_CONTROL` is already a string, so there is no reason to use `get_header` as opposed to accessing `headers` directly.  Even if you wanted to use `get_header` (for consistency?), it isn't related to case insensitive headers and should be committed separately.

I would mention in CHANGELOG that this change is purely for Rack::ETag, since that is the only middleware with the change.

OK to commit without further review after those changes.

PR #1923, opened 2022-07-07
Add 2.2.4 to changelog on main branch

pariser:
https://github.com/rack/rack/blob/2.2.4/CHANGELOG.md includes an entry from 2.2.4 which is not currently visible on main branch. Copy the changelog contents over from the `CHANGELOG.md` on `2.2.4` tag and submit it as a PR against main branch.


ioquatix (approved):

ioquatix:
Thanks!

PR #1925, opened 2022-07-08
Use "pre-wrap" format and monospace font for exception messages

mame:
This PR makes `Rack::ShowExceptions` friendly to Ruby 3.1's error_highlight. An exception message should be rendered by a monospace font and in "pre" format, i.e., whitespaces are preserved and newlines are respected.

Before:

image

After:

image

### DIFF
--- a/lib/rack/show_exceptions.rb
+++ b/lib/rack/show_exceptions.rb
@@ -163,7 +163,7 @@ module Rack
           div.commands { margin-left: 40px; }
           div.commands a { color:black; text-decoration:none; }
           #summary { background: #ffc; }
-          #summary h2 { font-weight: normal; color: #666; }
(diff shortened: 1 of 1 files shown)


mame:
@jeremyevans Thank you for your review!

> what happens when the exception message is very long? We don't want this to require horizontal scrolling.

It's a fair point. I changed the css to `white-sprace: pre-wrap`, which allows automatic line wrapping. How about this?

jeremyevans (approved):
`pre-wrap` looks like a good approach.

ioquatix:
Thanks! This is better than what we currently have, so I'll merge it. But I'd personally like to see better formatting when dealing with HTML as an output. Seems like we we could do a lot better and I imagine lots of people are consuming error messages this way.

PR #1926, opened 2022-07-08
Use Exception#detailed_message if available

mame:
Ruby 3.2 will provide `Exception#detailed_message` which returns more
informative message including hints for debugging.

https://bugs.ruby-lang.org/issues/18564

The did_you_mean gem and error_highlight gem is planned to use the
method to add their hints in Ruby 3.2. So using `Exception#message` will
not include did_you_mean and error_highlight hints.

This changeset uses `Exception#detailed_message` if available to show
exceptions.

### DIFF
--- a/lib/rack/show_exceptions.rb
+++ b/lib/rack/show_exceptions.rb
@@ -59,7 +59,12 @@ module Rack
     private :accepts_html?
 
     def dump_exception(exception)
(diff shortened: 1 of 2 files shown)


ioquatix:
Thanks! In the future, can we adopt `highlight: true` and if so, what are the implications for generating HTML?

PR #1927, opened 2022-07-09
The response array must always be non-frozen.

ioquatix:


ioquatix:
@jeremyevans going to assume you are happy with this as discussed in https://github.com/rack/rack/pull/1887/#issuecomment-1166209280 - if not just let me know and I'll fix up any issues.

jeremyevans (approved):
Looks good, but please make sure to add this to CHANGELOG.

PR #1929, opened 2022-07-21
Fix typo `interactively`

TonyCTHsu:
Found a typo while reading through doc.

### DIFF
--- a/lib/rack/builder.rb
+++ b/lib/rack/builder.rb
@@ -3,7 +3,7 @@
 require_relative 'urlmap'
 
 module Rack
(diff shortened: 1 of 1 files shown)


ioquatix:
Thanks!

matthewd:
Did this mean "interactively"? The existing use of "iteratively" seems more plausible to me.

ioquatix:
Fair point, let me rewrite it :)

TonyCTHsu:
Wow.... actually that is true. I guess I was blind..

PR #1930, opened 2022-07-25
Update docs and remove a dead link.

ahoglund:
This PR does two things. 

First, it replaces references to the
master branch with main (which seems to be the default now).

It also removes a bad link in a comment as a reference for
not parsing CGI ISINDEX parameters. I could not find a
viable alternative link to defend this assertion. If there is
one I'd be happy to add it.


simi (approved):

ioquatix (approved):
LGTM.

PR #1932, opened 2022-07-29
Add `rack.closed` to Rack::Lint

ioquatix:
See https://github.com/rack/rack/pull/1802 for the previous PR.

### DIFF
--- a/SPEC.rdoc
+++ b/SPEC.rdoc
@@ -141,6 +141,12 @@ There are the following restrictions:
   set. <tt>PATH_INFO</tt> should be <tt>/</tt> if
   <tt>SCRIPT_NAME</tt> is empty.
   <tt>SCRIPT_NAME</tt> never should be <tt>/</tt>, but instead be empty.
+* <tt>rack.closed</tt>:: An array of callbacks that must be
+invoked after a response is completed. This is equivalent to
+<tt>body.close</tt> and the same requirements exist for invoking it.
+The callbacks must take a single argument, <tt>reason = nil</tt>.
+If <tt>reason</tt> is given, it indiciates a failure was detected
+while writing the body to the client.
 
 === The Input Stream
 
--- a/lib/rack/constants.rb
+++ b/lib/rack/constants.rb
@@ -46,6 +46,7 @@ module Rack
   RACK_SESSION_OPTIONS                = 'rack.session.options'
   RACK_SHOWSTATUS_DETAIL              = 'rack.showstatus.detail'
   RACK_URL_SCHEME                     = 'rack.url_scheme'
+  RACK_CLOSED                         = 'rack.closed'
   RACK_HIJACK                         = 'rack.hijack'
   RACK_IS_HIJACK                      = 'rack.hijack?'
   RACK_HIJACK_IO                      = 'rack.hijack_io'
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -374,6 +374,30 @@ module Rack
         unless env[SCRIPT_NAME] != "/"
           raise LintError, "SCRIPT_NAME cannot be '/', make it '' and PATH_INFO '/'"
         end
+
+        ## * <tt>rack.closed</tt>:: An array of callbacks that must be
+        ## invoked after a response is completed. This is equivalent to
+        ## <tt>body.close</tt> and the same requirements exist for invoking it.
+        ## The callbacks must take a single argument, <tt>reason = nil</tt>.
+        ## If <tt>reason</tt> is given, it indiciates a failure was detected
+        ## while writing the body to the client.
+        if callbacks = env[RACK_CLOSED]
+          raise LintError, "rack.closed must be an array of callable objects" unless callbacks.is_a?(Array)
+
+          callbacks.each do |callback|
+            raise LintError, "rack.closed values must respond to call" unless callback.respond_to?(:call)
+
+            arity = if callback.respond_to?(:arity)
+              callback.arity
+            else
+              callback.method(:call).arity
+            end
+
+            raise LintError, "rack.closed values must accept an optional reason argument" unless arity == -1
+          end
+        else
+          raise LintError, "rack.closed must be present"
+        end
       end
 
       ##
--- a/test/spec_lint.rb
+++ b/test/spec_lint.rb
@@ -258,6 +258,31 @@ describe Rack::Lint do
       Rack::Lint.new(nil).call(env("SCRIPT_NAME" => "/"))
     }.must_raise(Rack::Lint::LintError).
       message.must_match(/cannot be .* make it ''/)
+
+    lambda {
+      Rack::Lint.new(nil).call(env("rack.closed" => "not a callable"))
+    }.must_raise(Rack::Lint::LintError).
+    message.must_match(/rack.closed callbacks be an array of callable objects/)
+
+    lambda {
+      Rack::Lint.new(nil).call(env("rack.closed" => ["not a callable"]))
+    }.must_raise(Rack::Lint::LintError).
+    message.must_match(/rack.closed callbacks must respond to call/)
+
+    lambda {
+      Rack::Lint.new(nil).call(env("rack.closed" => [-> () {}]))
+    }.must_raise(Rack::Lint::LintError).
+    message.must_match(/rack.closed callbacks must accept an argument/)
+
+    callable_object = Class.new do
+      def call
+      end
+    end.new
+
+    lambda {
+      Rack::Lint.new(nil).call(env("rack.closed" => [callable_object]))
+    }.must_raise(Rack::Lint::LintError).
+    message.must_match(/rack.closed callbacks must accept an argument/)
   end
 
   it "notice input errors" do
@@ -855,6 +880,17 @@ describe Rack::Lint do
                        env['rack.hijack?'] = true
                        [201, { "content-type" => "text/plain", "content-length" => "0", 'rack.hijack' => lambda {|io| io }, 'rack.hijack_io' => StringIO.new }, []]
                      }).call(env({}))[1]['rack.hijack'].call(StringIO.new).read.must_equal ''
+  end
+
+  it "pass valid rack.closed" do
+    callable_object = Class.new do
+      def call(env)
+      end
+    end.new
+
+    Rack::Lint.new(lambda { |env|
+                     [200, {}, ["foo"]]
+                   }).call(env({ "rack.closed" => [-> (env) {}, lambda { |env| }, callable_object], "content-length" => "3" })).first.must_equal 200
   end
 
 end


ioquatix:
In addition to this PR, I wonder if we should extend `body.close` to take an optional `error` argument. This is a new feature which is also introduced in principle by this PR.

Separately, I wonder if we should consider rejection this feature. We can implement this feature using a middleware:

```ruby
class Rack::AfterReply
  def initialize(app)
    @app = app
  end

  def call(env)
    env['rack.after_reply'] = after_reply = Array.new
    
    begin
      response = @app.call(env)
    rescue error
  end
end
```

jeremyevans:
> Separately, I wonder if we should consider rejection this feature. We can implement this feature using a hypothetical middleware:

I don't think this is correct.  Your middleware calls after_reply after the middleware returns (ensure), and after the body is sent to the client (via BodyProxy). The idea with after_reply is it is only called once, after the reply is sent to the client (or maybe if an exception occurs during processing?), and it doesn't require use of BodyProxy.

That being said, I think it would be a good idea for after_reply/response_finished/whatever to accept an argument for an optional exception.  We already discussed taking additional arguments for the environment and response. Maybe a keyword based API would work best?:

```ruby
after_reply.call(env: , status: , headers: , exception: )
```

I don't think we should pass the body, because once it has been iterated over and closed (if closeable), it shouldn't be reused.  There are some bodies (like array bodies) where it would be safe, and those are probably the majority, but in general you cannot rely on using the body after it is sent to the client.

ioquatix:
> Your middleware calls after_reply after the middleware returns (ensure), and after the body is sent to the client (via BodyProxy).

You are totally right, it was late when I was writing this hypothetical code, here is the correct version.

```ruby
class Rack::AfterReply
  def initialize(app)
    @app = app
  end

  def call(env)
    env['rack.after_reply'] = after_reply = Array.new
    
    begin
      response = @app.call(env)

      callback = proc do |error|
        after_reply.each do |callback|
          callback.call(error) rescue log($!)
        end
      end

      response[2] = ::Rack::BodyProxy.new(response[2], callback)

      return response
    rescue => error
      after_reply.each do |callback|
        callback.call(error) rescue log($!)
      end
    end
  end
end
```

> The idea with after_reply is it is only called once, after the reply is sent to the client (or maybe if an exception occurs during processing?), and it doesn't require use of BodyProxy.

The idea of `after_reply` is to be a semantic replica of `body.close` with less overhead. The problem is, every middleware that wants to add a "after_reply" hook is allocating a body proxy so you potentially end up with N body proxies for N middlewares. My proposal shows it's possible to do it with only one allocation with exactly the same semantics of the current "rack.after_reply" using a middleware. The advantage of this is that applications that need it can insert it, and applications that don't can ignore it and don't have the extra overhead per request.

Someone else called out, that we should probably have a way to do this transparently, i.e.

```ruby
class Rack::AfterReply
  def self.wrap(env, response, callback)
    if after_reply = env['rack.after_reply']
      after_reply << callback
    else
      response[2] = Rack::BodyProxy.new(response[2], callback)
    end
  end
end
```

Thus it becomes totally transparent to the end user which approach is used and it's up to the application server or application middleware to configure which approach is most appropriate - an array allocation / overhead for every request, or extra `BodyProxy` allocations when needed. I believe Rails falls into the former, but a lot of applications don't use any kind of callback so if we opted for this PR, they would be paying a price for every request (small I agree) without any benefit.

jeremyevans:
> > Your middleware calls after_reply after the middleware returns (ensure), and after the body is sent to the client (via BodyProxy).
> 
> You are totally right, it was late when I was writing this hypothetical code, here is the correct version.
> 
> ```ruby
> class Rack::AfterReply
>   def initialize(app)
>     @app = app
>   end
> 
>   def call(env)
>     env['rack.after_reply'] = after_reply = Array.new
>     
>     begin
>       response = @app.call(env)
> 
>       callback = proc do |error|
>         after_reply.each do |callback|
>           callback.call(error) rescue log($!)
>         end
>       end
> 
>       response[2] = ::Rack::BodyProxy.new(response[2], callback)
> 
>       return response
>     rescue => error
>       after_reply.each do |callback|
>         callback.call(error) rescue log($!)
>       end
>     end
>   end
> end
> ```

This is better. You still run into the issue where if an exception is raised after this middleware returns, but before the server handles the response, the after_reply callback is not called.

> > The idea with after_reply is it is only called once, after the reply is sent to the client (or maybe if an exception occurs during processing?), and it doesn't require use of BodyProxy.
> 
> The idea of `after_reply` is to be a semantic replica of `body.close` with less overhead. The problem is, every middleware that wants to add a "after_reply" hook is allocating a body proxy so you potentially end up with N body proxies for N middlewares. My proposal shows it's possible to do it with only one allocation with exactly the same semantics of the current "rack.after_reply" using a middleware. The advantage of this is that applications that need it can insert it, and applications that don't can ignore it and don't have the extra overhead per request.

I like the idea that only people that would use this feature pay the cost for it.  It's certainly possible for servers that support after_reply to operate the same way (only allocating if configured to do so), but using middleware allows the application to control whether it is needed in a server-independent manner.  The main question is whether or not you can safely implement this in middleware.  As mentioned above, I don't think you can do so 100% safely, though you can probably get close to 100% if you ensure Rack::AfterReply is the outer-most middleware.

Considering allocations, I think someone pointed out that if the after_reply callback isn't called with data the application may need, then you at least need a proc allocation or similar per request using after_reply, which is significantly more expensive than an empty array allocation.  So if the idea is to save allocations, unless we break backwards compatibility and start calling body.close with arguments, I don't think this approach makes sense as it ultimately would not save allocations.  If after_reply doesn't accept arguments needed, then the application would allocate a proc per request (using variables in scope).  If after reply does accept arguments needed, the middleware would need to create a proc to pass to BodyProxy calls the passed after_reply with variables in the middleware scope.  In either case, a per-request callable allocation is needed.

If we implement a new after_reply/response_finish callback that accepts arguments for things like env/status/headers/exception, it's likely applications/middleware using after_reply could use a static callback and not one that needs to be allocated per request.  But in that case, you cannot emulate after_reply/response_finished using BodyProxy in a middleware.

> Someone else called out, that we should probably have a way to do this transparently, i.e.
> 
> ```ruby
> class Rack::AfterReply
>   def self.wrap(env, response, callback)
>     if after_reply = env['rack.after_reply']
>       after_reply << callback
>     else
>       response[2] = Rack::BodyProxy.new(response[2], callback)
>     end
>   end
> end
> ```
> 
> Thus it becomes totally transparent to the end user which approach is used and it's up to the application server or application middleware to configure which approach is most appropriate - an array allocation / overhead for every request, or extra `BodyProxy` allocations when needed. I believe Rails falls into the former, but a lot of applications don't use any kind of callback so if we opted for this PR, they would be paying a price for every request (small I agree) without any benefit.

Can you explain what code would use this theoretical `Rack::AfterReply.wrap`?  It seems unlikely servers would use this, because they know whether after_reply is used and wouldn't need a conditional. It seems unlikely applications would use it, because at the time you want to add the after_reply hook, the response probably hasn't been constructed yet.  I suppose it could be used by middleware, but I'm not sure how useful it would be in that case.

ioquatix:
I completely agree that there is some missing specifications around when close should be called, and between the app generating a response (we have the body) and, say, the client disconnecting, and the body being discarded. It should absolutely be the case, that `body.close` in every possible case whereby the client disconnects, or the body is consumed either partially or completely.

The idea of `AfterReply.wrap` is "I don't care how this gets called, just that it should get called after the reply is done". Whether the application opts into an aggregate `after_reply` style callback array, or a nested list of `BodyProxy` instances should be an implementation detail and a trade off the application can decide.

Can you show me examples of what kinds of arguments you think we should be passing? I think we still win on less `BodyProxy` instances even if we still need to allocate the actual callbacks to capture state.

jeremyevans:
> I completely agree that there is some missing specifications around when close should be called, and between the app generating a response (we have the body) and, say, the client disconnecting, and the body being discarded. I strongly believe that `body.close` should be called in every possible case whereby the client disconnects, or the body is consumed either partially or completely (or just ignored/discarded).

Unfortunately, the server cannot close the body if an exception is raised instead of a response being returned, because then there is no body to close.  That's one reason I think it would make sense to support after_reply/response_finished, because once the callback is added, any future exceptions raised should theoretically still result in the callback being called.

> The idea of `AfterReply.wrap` is "I don't care how this gets called, just that it should get called after the reply is done". Whether the application opts into an aggregate `after_reply` style callback array, or a nested list of `BodyProxy` instances should be an implementation detail and a trade off the application can decide.

As I mentioned, I don't think applications would be able to use `AfterReply.wrap`, because in general they would want to add the after_reply callback before the response is prepared, and `AfterReply.wrap` requires the response already be created.

> Can you show me examples of what kinds of arguments you think we should be passing? 

I gave an example in an earlier comment:

```ruby
after_reply.call(env: , status: , headers: , exception: )
```

This omits body because it isn't safe to access the body once it has been iterated over and closed.

> I think we still win on less `BodyProxy` instances even if we still need to allocate the actual callbacks to capture state. Regarding arguments, since I personally believe there should be a symmetry between `body.close` and `after_reply` callbacks, we shouldn't require arguments that don't exist at all times, i.e. someone might write:
> 
> ```ruby
> response = @app.call
> response[2].close # (env: , status: , headers: , exception: ) which of these exist and are meaningful here?
> ```

As originally designed, none are needed, since close was originally designed so you could pass IO-like instances as bodies and have them closed.  However, consider using after_reply callback for logging of responses.  In that case, you probably want all of those arguments for useful logging.

> I'm not yet convinced that a hypothetical `after_reply` callback should be different in this regard from `body.close`, as in, I think it should be possible to implement `after_reply` as my hypothetical middleware above which uses `body.close` to invoke the `after_reply` callbacks.

I explained this in an earlier comment: `You still run into the issue where if an exception is raised after this middleware returns, but before the server handles the response, the after_reply callback is not called.`

Basically, I don't think you can implement this in middleware safely, it has to be done at the server level or there is a chance you miss calling the callbacks.

> If we think they should deviate significantly, I'd like to know more detail about the use case.

I don't think we can change body.close API for backwards compatibility reasons.  I don't think we can emulate after_reply safely via a middleware that uses body.close.  We can keep things as-is with using body.close with BodyProxy, but if we are going to add after_reply as a mandatory or optional server requirement, I think it makes sense to accept additional arguments so that it could support a static callback instead of requiring a callback allocated per request.

ioquatix:
The use case of `rack.after_reply` is different from the callback provided to `BodyProxy` then we are solving a different problem than the one that is originally described, which is ensuring resources are freed after a response is "finished". The problem we are trying to solve is reducing the number of body proxy instances being created. If the problem scope expands to be "handle cases where the middleware throws an exception and the body was not closed correctly", that's a different problem.

As it stands, I agree adding `close(error = nil)` is too hard/incompatible. I tried it but I think the compatibility cost is too high.

I think that if someone has a body, but an exception is raised, it's on them to ensure that body.close is called if the exception propagates up.

I'm open to solving additional semantic issues with `after_reply` style callbacks, but I'm not sure what the use case is. I only consider the use case which is let's reduce the number of `BodyProxy` instances + as an aside how to propagate errors/failures. But honestly, propagating errors back into Rack might not be that important. In `Async::HTTP` it is because we need to handle many more scenarios where the middleware stack terminates at a client, cache, proxy and error propagation is important for correctly handling cases (e.g. the cache didn't get the full data because the client failed).

jeremyevans:
> If the use case of `rack.after_reply` is different from the callback provided to `BodyProxy` then we are solving a different problem than the one that is originally described, which is ensuring resources are freed after a response is "finished". The problem we are trying to solve is reducing the number of body proxy instances being created. If the problem scope expands to be "handle cases where the middleware throws an exception and the body was not closed correctly", that's a different problem.

I don't think the use case is different.  I think after_reply would be a more efficient (fewer allocations) and more reliable (if implemented by the server) for all cases where BodyProxy is currently used.  As it does allocate an additional array per request (if enabled), it could potentially be less efficient in cases where BodyProxy is not used.  I don't like the idea of having to pay the cost for something that isn't used, though I expect the single array allocation is not material in the scope of the request.
 
> As it stands, I agree adding `close(error = nil)` is too hard/incompatible. I tried it but I think the compatibility cost is too high. Actually I apologise because this deviates from the core of the problem we are trying to solve, as does providing extra arguments to `after_reply` callbacks - this is a nice time to think about it, but fundamentally not part of the problem statement and I'm causing the issue to be distracted/derailed.

I think if we are going to standardize an after_reply callback in some form, it makes sense to consider what arguments should be passed to it.

> Regarding your specific point about middleware failing with an exception, I think that if someone has a body, but an exception is raised, it's on them to ensure that `body.close` is called if the exception propagates up. i.e.
> 
> ```
> def call(env)
>   response = @app.call(env)
> 
>   success = true
>   return response
> ensure
>   response[2].close unless success
> end
> ```

This suffers from the same issue, where an outer middleware raises an exception.  I don't think this exception handling problem is solvable with a body.close approach.  With an after_reply approach, it's much less likely to leak (though likely still possible in extreme circumstances).

Honestly, I doubt this is an major issue.  As far as I know, users aren't complaining that their closeable bodies aren't being closed when exceptions are raised.

> If this is not done, there is a chance we will leak File instances or other body types which have underlying resources. I guess GC might eventually pick these up?

This depends on the implementation of the object, but any reasonable object will automatically perform the equivalent of close when being garbage collected (File/IO does).

> I'm open to solving additional semantic issues with `after_reply` style callbacks, but I'm not sure what the use case is. I only consider the use case which is let's reduce the number of `BodyProxy` instances + as an aside how to propagate errors/failures. But honestly, propagating errors back into Rack might not be that important. In `Async::HTTP` it is because we need to handle many more scenarios where the middleware stack terminates at a client, cache, proxy and error propagation is important for correctly handling cases (e.g. the cache didn't get the full data because the client failed).

One use case is logging.  You want to log requests, and include the time of the request. To avoid per-request proc allocation, you could do something like `env['request.start'] = Time.now; after_reply << REQUEST_LOGGER`, where

```
  REQUEST_LOGGER = lambda do |env:, status:, headers:, exception:|
    # either status and headers provided, or exception provided
    env['rack.logger'] << "#{env['REQUEST_METHOD']} #{env['SCRIPT_NAME']}#{env['PATH_INFO']} => #{exception ? exception.message : "#{status} #{headers['CONTENT_LENGTH']}"} #{Time.now - env['request.start']}s"
  end
```

If the after_reply callback doesn't accept arguments, you basically have to allocate a callable per request.

ioquatix:
The logging one is an interesting one, but I think it's better handled by a middleware. The only advantage of your example is that you truly get the time it took to write the response to the client for example, rather than the time it took to generate it.

I can see the advantages of such an interface, but this is new functionality rather than solving the problem of creating lots of `BodyProxy` instances. I agree, we can introduce new functionality, but I'm concerned that it's then fairly hard to do virtual/internal requests with this hypothetical interface.

https://github.com/socketry/utopia/blob/3458234a060d6f7f1610c21d3735356724dfbb9a/lib/utopia/localization.rb#L216-L222

In this code, in theory, I should be allocating a new `after_reply` array for each internal request. Ignoring the various bugs this code has, it's going to be hard to justify it to be written like this:

```ruby
preferred_locales(env) do |localized_env|
	after_reply = localised_env['rack.after_reply'] = Array.new

	response = @app.call(localized_env)

	break unless response[0] >= 400

	begin
		response[2].close
	ensure
		after_reply.each do |callback|
			begin
				callback.call
			rescue => error
				log(error)
			end
		end
	end
end
```

jeremyevans:
> The logging one is an interesting one, but I think it's better handled by a middleware. The only advantage of your example is that you truly get the time it took to write the response to the client for example, rather than the time it took to generate it.

The difference in time is precisely why you would want to implement this via after_reply

> I can see the advantages of such an interface, but this is new functionality rather than solving the problem of creating lots of `BodyProxy` instances. I agree, we can introduce new functionality, but I'm concerned that it's then fairly hard to do virtual/internal requests with this hypothetical interface.
> 
> https://github.com/socketry/utopia/blob/3458234a060d6f7f1610c21d3735356724dfbb9a/lib/utopia/localization.rb#L216-L222
> 
> In this code, in theory, I should be allocating a new `after_reply` array for each internal request. Ignoring the various bugs this code has, it's going to be hard to justify it to be written like this:

I don't see the problem with keeping the same array and having the web server run all the after_reply callbacks for all requests.  Could you explain why that would be problematic?  Why in theory should you allocate a new after_reply array to do internal requests?

ioquatix:
> I don't see the problem with keeping the same array and having the web server run all the after_reply callbacks for all requests. Could you explain why that would be problematic? Why in theory should you allocate a new after_reply array to do internal requests?

Well, if you use your example of the log output, you will end up adding an extra log statement for every internal request.

For Rails, I believe they want to use it to clean up database connections and other resources. I believe it's incorrect to accumulate all the connections and then try to release them all at once if you performed an internal redirect/request.

In don't think we should be mixing the life cycle of multiple requests and we should have a clear model for "This is how you do a request and clean up afterwards" and that shouldn't require an actual server.

jeremyevans:
> > I don't see the problem with keeping the same array and having the web server run all the after_reply callbacks for all requests. Could you explain why that would be problematic? Why in theory should you allocate a new after_reply array to do internal requests?
> 
> Well, if you use your example of the log output, you will end up adding an extra log statement for every internal request.

This seems fine to me.  If you want to differentiate internal from external requests, pass an additional environment variable for internal requests and use it for different handling.  In my opinion, by default, all calls to the rack app (e.g. `app.call(env)`), internal or external, should be treated the same.

> For Rails, I believe they want to use it to clean up database connections and other resources. I believe it's incorrect to accumulate all the connections and then try to release them all at once if you performed an internal redirect/request.

Rails stores connections in a hash keyed by current thread (or optionally, fiber): https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb

`release_connection` appears to be idempotent if called in the same thread, assuming there wasn't a connection checkout between the calls: https://github.com/rails/rails/blob/main/activerecord/lib/active_record/connection_adapters/abstract/connection_pool.rb#L203

So multiple after_reply calls to release the connection should be idempotent unless they were intermixed with after_reply calls that checked out a connection.  Note that that even if you did intermix such calls, while not idempotent, it still should be fine, as the connection checked out by an earlier after_reply call is still checked in by a later after_reply call, even if it is a different connection checked in than a previous after_reply call.

> In don't think we should be mixing the life cycle of multiple requests and we should have a clear model for "This is how you do a request and clean up afterwards".

If you are calling the the app, it is expected you are operating by the Rack SPEC and would do everything  SPEC requires (consuming the body via each/call and close).  It's certainly possible to take shortcuts, and presumably, if you are making internal requests, you should know which shortcuts are acceptable.  If you don't, the you should be building a new hash compatible with SPEC, calling the app with it, and handling the response returned according to SPEC.   The example you gave earlier for utopia doesn't do that.  You could say it is out of SPEC.  Presumably you know what you are doing in utopia is safe.

To be more clear, we already have a clear model for "This is how you do a request and clean up afterwards".  It's called SPEC.  I don't think we need to discuss internal requests and which shortcuts are allowable, as I consider that outside of scope for SPEC.

ioquatix:
What you say is reasonable but:

(1) I'm no longer convinced this is a mandatory feature for Rack.
(2) I'm happy that the vast majority of use cases that we care about can be implemented by a middleware, and this will solve the original problem as stated.
(3) I no longer feel strongly that this is needed for Rack 3.0 so we can punt on it.
(4) We clarified enough of the requirements on Body so that `body.close` is effectively providing the same hooks, and even in Falcon we can emulate `rack.after_reply` with a middleware and it's good enough provided Falcon follows the updated spec precisely (it just won't have the extra arguments).

I've tried to reach out to @tenderlove since he was the original stakeholder but I believe we can satisfy his needs without any changes to the spec.

You are right, if we make this an optional part of the `env`, it won't matter that users don't try to invoke it on internal/virtual requests. It may be an issue, but at least not one which they have to care about since it's not a mandatory part of the SPEC (and I don't believe I will ever support it being a mandatory part of the SPEC). So, even if I disagree with the complexities of the life-cycle as discussed, it no longer matters because I can just ignore it if we aren't requiring it as part of the Rack 3 SPEC.

If we want to go back to adding `rack.after_reply` or some other feature for "The server has finished sending the response to the client", that's an entirely separate feature and I think we can work on it if there is enough interest. As we discussed, we need to be be careful about how it's defined since it's not part of a request per-se but part of the entire server request-response handling rather than the rack request-response (i.e. `@app.call` in an abstract sense).

PR #1933, opened 2022-07-31
Clarify handling of body and add optional `error` argument to `close`.

ioquatix:
This is an attempt to clarify the behaviour of `body.close` and expose any failure to consume said body via `close`.

This general approach has been used in `async-http` for several years and has proven itself to be invaluable to certain cases like streaming and error propagation.

https://github.com/rack/rack/pull/1932

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ All notable changes to this project will be documented in this file. For info on
 - Response header keys can no longer include uppercase characters.
 - Response header values can be an `Array` to handle multiple values (and no longer supports `\n` encoded headers).
 - Response body can now respond to `#call` (streaming body) instead of `#each` (enumerable body), for the equivalent of response hijacking in previous versions.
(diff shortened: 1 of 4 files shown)


jeremyevans:
I'm against this as it breaks passing a IO-like object as the body (where close does not accept an argument).  I would also be against testing arity of body.close at runtime to determine if an exception should passed.

I think we should keep backwards compatibility for body.close, and instead introduce a hook such as after_reply/response_finished that will pass an exception as one of the arguments.

ioquatix:
> I'm against this as it breaks passing a IO-like object as the body (where close does not accept an argument). I would also be against testing arity of body.close at runtime to determine if an exception should passed.

Yes, I also agree. But I wanted to try it.

> I think we should keep backwards compatibility for body.close, and instead introduce a hook such as after_reply/response_finished that will pass an exception as one of the arguments.

I am slightly against this because it introduces a discontinuity in the interface between "body.close" and "response finished". I'm not sure there is a big enough advantage to justify diverging on this interface (and making it impossible to implement "response finished" as a middleware layer).

PR #1935, opened 2022-08-01
Move Rack::MockRequest/Response into dedicated files.

ioquatix:
At some point I think we want to improve the implementation of `Rack::Mock`
in a separate gem. So let's be consistent with naming these files to avoid
clobbering namespace in the future.


jeremyevans:
I don't think the benefits of this outweigh the costs.  Forcing people to change their requires from rack/mock to rack/mock_* results in unnecessary churn and makes it more difficult for gems to support both rack 2 and 3.1.  This also makes it more difficult to follow the history.

ioquatix:
We already discussed this, and we decided we will introduce a gem, but we can't do that if `rack/mock.rb` exists because it will be clobbered by the gem.

ioquatix:
We want to introduce `Rack::Mock::Request` and `Rack::Mock::Response` in https://github.com/rack/rack-mock

ioquatix:
https://github.com/rack/rack/pull/1791#issuecomment-1021680192

ioquatix:
Also, I don't think:

```
begin
  require 'rack/mock_request'
rescue LoadError
  require 'rack/mock'
end
```

is particularly hard if we need backwards compatibility... or even just the autoloads defined in `Rack` which already exist for `Rack::MockRequest` etc.

ioquatix:
I've removed the deprecation warning for now, we can deal with the exact mechanics of how to do this later, I think the rest of the implementation is fine, right?

jeremyevans:
I don't see the point of splitting rack/mock into separate files if there is a plan to remove it later. One could make the argument that this allows you to require MockResponse without MockRequest.  No code actually does that, though.  All internal requires are for MockRequest, which depends on MockResponse.  So this rearranges the deck chairs for not purpose other than consistency.  If the plan is to move this code to a separate rack-mock gem, this should be done after the code is moved, not before.

ioquatix:
The point is that all classes should be in the same file as the class name, which is a very common pattern and used pretty much everywhere else. This is the first step to actually splitting this code out, so it sounds like once I copy this code into the rack mock gem, you agree with merging this PR?

jeremyevans:
> The point is that all classes should be in the same file as the class name, which is a very common pattern and used pretty much everywhere else. This is the first step to actually splitting this code out, so it sounds like once I copy this code into the rack mock gem, you agree with merging this PR?

No, I don't agree with merging this PR in `rack`, as long as the plan is to move it to a `rack-mock` gem.  If you move the existing `rack/mock` into a `rack-mock` gem, then I'd be fine with splitting the file in the `rack-mock` gem.  However, I see no point in splitting the current `rack/mock` file in `rack` if the plan is to move it to `rack-mock` gem.

lib/rack/mock_request.rb
```diff
@@ -0,0 +1,172 @@
+# frozen_string_literal: true
+
+require 'uri'
+require 'stringio'
+require 'cgi/cookie'
```
jeremyevans, line 5: CGI::Cookie is only used in MockResponse, not in MockRequest. URI is only used in MockRequest, not in MockResponse. I'm guessing there are other issues.  If you are going to split these up, please split the tests up as well, and ensure each file only requires the parts it needs.  `SEPARATE=1 ruby test/spec_mock_request.rb` and similar for response should help you figure out which parts are needed.

jeremyevans (changes requested):
As discussed yesterday, I'm fine with the general approach as long as requiring `rack/mock` keeps the current API until Rack 4, to allow dependencies time to move to the new requires.

lib/rack/mock_request.rb
ioquatix: Thanks for this constructive feedback, I agree with your assessment and will update the PR.

jeremyevans (approved):

PR #1939, opened 2022-08-01
Separate and simplify full & partial hijack.

ioquatix:
- Remove `rack.hijack_io` which is at best difficult to use (and at worst odd/buggy).
- Remove partial hijack - replaced by streaming body. Not sure there is any advantage to keep partial hijack - and advantages to remove it (performance, simplicity).

We retain full hijack which is the most useful hijack (but only supported on HTTP/1).

These changes should reduce the server implementation complexity and at best might give a slight performance boost due to

1. `rack.hijack` was the only response header which was rack specific, so servers will no longer have to process the headers in any way.
2. `rack.hijack_io` modified `env` and was hard to use if any request has done `env.dup` so it's practically useless - one less overhead to deal with.

Note that:

1. This doesn't prevent servers from implementing a blended Rack 2 / 3 spec, but we could force that by explicitly checking that `rack.hijack` response header doesn't exist.
2. It encourages applications to use streaming response body which going to be handled by middleware better.
3. It encourages servers to avoid complex/costly optional behaviour.


ioquatix:
By the way, I see there could be an argument to retain partial hijack.However, I don't think we should, because we want middleware to correctly handle streaming responses, and we don't want to have two different mechanisms for picking up the body - let's streamline things for our users.

jeremyevans:
I'm against removing partial hijacking.  First, hijack support is optional, so servers already can drop support if they want to. I think in most cases, for a while servers are going to want to support both Rack 2 and 3, and partial hijacking is the same as streaming, the only difference being where you get the callable.  So it's trivial for servers to support both with very minimal code.  Most servers will keep their existing partial hijack support, but also now have to check for callable bodies.

Applications/frameworks that want to support both Rack 2 and 3 can use partial hijacking.  Applications/frameworks that drop support for Rack 2 and only support Rack 3 can use callable bodies.

> * Remove `rack.hijack_io` which is at best difficult to use (and at worst odd/buggy).
> * Remove partial hijack - replaced by streaming body. Not sure there is any advantage to keep partial hijack - and advantages to remove it (performance, simplicity).

I'm not sure why you are ignoring backwards compatibility as an advantage.  Backwards compatibility is a huge advantage.  The advantages to removing it are minor in comparison.  Can you show a production benchmark showing a significant difference between partial hijacking and callable bodies?  The argument for simplicity doesn't mean much when the code for partial hijacking is already written, and support is already optional.

> These changes should reduce the server implementation complexity and at best might give a slight performance boost due to
> 
> 1. `rack.hijack` was the only response header which was rack specific, so servers will no longer have to process the headers in any way.
> 2. `rack.hijack_io` modified `env` and was hard to use if any request has done `env.dup` so it's practically useless - one less overhead to deal with.

I would be surprised if the performance differences are significant in a production application.

> Note that:
> 
> 1. This doesn't prevent servers from implementing a blended Rack 2 / 3 spec, but we could force that by explicitly checking that `rack.hijack` response header doesn't exist.

We 100% should not do this.  That is user-hostile design.

> 2. It encourages applications to use streaming response body which going to be handled by middleware better.

I doubt that middleware will handle callable bodies different than hijack bodies.

> 3. It encourages servers to avoid complex/costly optional behaviour.

In most cases, the complex/costly behavior is supporting the streaming itself.  Whether it gets the callable from the headers or body probably doesn't matter much.

ioquatix:
@jeremyevans we discussed this on Slack and decided:

- `rack.hijack_io` can go.
- Partial hijack depends only on `env['rack.hijack?']` being truthy.
- Full `rack.hijack` can expect to yield an `IO` instance and only depends on `rack.hijack` being truthy (and callable) in the request `env`.

Can you review the updated SPEC?

lib/rack/handler/webrick.rb
```diff
@@ -80,8 +80,6 @@ def service(req, res)
           RACK_ERRORS       => $stderr,
           RACK_URL_SCHEME   => ["yes", "on", "1"].include?(env[HTTPS]) ? "https" : "http",
           RACK_IS_HIJACK    => true,
-          RACK_HIJACK       => lambda { raise NotImplementedError, "only partial hijack is supported."},
```
ioquatix, line 83: I think the removal of this line is a good reason to support this PR.

lib/rack/lint.rb
```diff
@@ -70,7 +73,7 @@ def response
         check_headers(@headers)
 
         hijack_proc = check_hijack_response(@headers, @env)
-        if hijack_proc && @headers.is_a?(Hash)
```
ioquatix, line 73: `@headers` is now always a hash.

lib/rack/lint.rb
```diff
@@ -513,129 +510,92 @@ def close(*args)
         end
       end
 
-      class HijackWrapper
-        extend Forwardable
-
-        REQUIRED_METHODS = [
-          :read, :write, :read_nonblock, :write_nonblock, :flush, :close,
```
ioquatix, line 520: This has been split into full hijack which we agree should be an `IO` instance, and partial hijack which should follow the same interface as StreamWrapper.

lib/rack/lint.rb
```diff
@@ -513,129 +510,90 @@ def close(*args)
...
-      ## ==== Request (before status)
+      ## The hijacking interfaces provides a means for an application to take
+      ## control of the HTTP connection. There are two distinct hijack
+      ## interfaces: full hijacking where the application takes over the raw
+      ## connection, and partial hijacking where the application takes over
+      ## just the response body stream. In both cases, the application is
+      ## responsible for closing the hijacked stream.
+      ##
+      ## Full hijacking only works with HTTP/1. Partial hijacking has been
+      ## superseeded by streaming responses.
```
jeremyevans, line 524: It would be `superseded`, but that can also have the interpretation that partial hijacking is no longer available. I would reword to `Partial hijacking is functionally equivalent to streaming bodies, and is still optionally supported for backwards compatibility with older Rack versions.`

lib/rack/lint.rb
```diff
@@ -513,129 +510,90 @@ def close(*args)
...
+      ##
+      ## ==== Full Hijack
+      ##
+      ## Full hijack is typically used when the application wants to upgrade an
```
jeremyevans, line 528: I would remove this first sentence. That's not my typical use for full hijack. Alternatively, change this to `Full hijack is used to completely take over an HTTP/1 connection.`

lib/rack/lint.rb
```diff
@@ -513,129 +510,90 @@ def close(*args)
...
-          ## If rack.hijack? is false, then rack.hijack_io should not be set.
-          raise LintError, "rack.hijack? is false, but rack.hijack_io is present" unless env[RACK_HIJACK_IO].nil?
         end
       end
 
       ##
-      ## ==== Response (after headers)
+      ## ==== Partial Hijack
+      ##
+      ## The partial hijack is used for bi-directional streaming of the
```
jeremyevans, line 557: `s/The partial hijack/Partial hijack/` (for consistency with full hijack text).

jeremyevans (changes requested):

lib/rack/lint.rb
ioquatix: Thanks, this is much better.

lib/rack/lint.rb
ioquatix: Updated with the latter. Thanks!

lib/rack/lint.rb
ioquatix: Thanks, this makes total sense.

jeremyevans (approved):

boazsegev:
Question about the `IO` hijack object... is there a way to support an IO "like" object rather than Ruby's IO?

Iodine is implemented in C and the IO design follows the layered model where the TLS layer hidden away (so the same C API is used for TLS and clear-text sockets).

I have no idea how to move the lower layer TLS state to the Ruby IO object, and I doubt it makes sense to try and do that. It would be much easier it I could implement a fake `IO` object that will map to the server's C API.

But the new specification requires that the object be a Ruby IO object... so...?

ioquatix:
The full hijack API does typically expect an instance of `IO`. A sub-class or interface compatible object is probably acceptable in practice.

At the very least, it needs to be something that responds to `filedes` because ActionCable uses `NIO4r` for an event loop which handles multiple hijacked WebSockets via HTTP/1.

Since it is not expected that the server handles the socket after a full hijack, it may be possible to use `IO.for_fd`. Of course, this won't work with TLS. In that case, a compatible interface may work.

An alternative solution which I've tried in the past is to use a pipe, e.g. `Iodine Server Connection + TLS <-> Pipe <-> Full Hijack`.

This is one of the reasons why I wanted to remove full hijack.

boazsegev:
I've been looking into subclassing the IO class and overriding the whole read/write function family ... but this doesn't promise that the `filedes` won't be used for read / write calls directly (not to mention that that's a lot of API to override).

Perhaps if the SPECs stated that `read` / `write` operations MUST be performed using the returned IO object and listed a limited list of methods that the IO object must be implement and may be called by conforming implementations, we could at least minimize the surface area...?

ioquatix:
I think that's reasonable, or we could also revive my proposal to deprecate and remove full hijack. I think since we couldn't agree to do that last time, I basically specified it as narrowly as possible w.r.t. how existing servers were actually working, and what applications were actually doing.

Issue #1843, opened 2022-04-06
Fix `Rack::VERSION` to be more canonical.


ioquatix:
`Rack::VERSION` should be something like `"1.2.3"` but it's an `Array` which is pretty atypical. I'd like to fix this to be more consistent with other gems which define it as a `String`.

jeremyevans:
I don't think the benefits of consistency with other gems outweigh the cost of the loss of backwards compatibility.

simi:
What's the policy for backward compatibility for Rack 3?

jeremyevans:
As in an major version change, backwards compatibility changes are allowed.  However, I have a very strong preference for only breaking backwards compatibility when the benefits significantly outweigh the costs.

rafaelfranca:
We could add a new `Rack.version` method that returns a string? That is what we end up doing in Rails.

tenderlove:
I agree with @jeremyevans and @rafaelfranca.  I'm fine if we add a method, but "consistency with other gems" isn't a goal of the project.

jeremyevans:
We already have a `Rack.version` method.  However, it gives the rack protocol version (currently 1.3), not the gem version (which would be 3.0.0).  The gem version is available via `Rack.release`.  I propose we make no changes.

ioquatix:
Does the rack protocol version have any meaning? Can anyone depend on it for anything useful?

rafaelfranca:
Rails also has [`gem_version`](https://github.com/rails/rails/blob/main/railties/lib/rails/gem_version.rb#L6).

I don't have a strong opinion if we should add a method like that or not, but I'm giving us options.

ioquatix:
The concept of compatibility version or "interface version" or "spec version" I don't think has ever been useful.

My gut feeling is we should remove this. We don't have middleware which change their behaviour based on the version of rack nor should we encourage it.

I don't mind if we need a middle state to get to the ideal, but my ideal state is this:

```rack
module Rack
  VERSION = "3.0.0"
end
```

ioquatix:
```
> bundle gem foobar
Creating gem 'foobar'...
MIT License enabled in config
      create  foobar/Gemfile
      create  foobar/lib/foobar.rb
      create  foobar/lib/foobar/version.rb
      create  foobar/foobar.gemspec
      create  foobar/Rakefile
      create  foobar/README.md
      create  foobar/bin/console
      create  foobar/bin/setup
      create  foobar/.gitignore
      create  foobar/.travis.yml
[...]

PR #1940, opened 2022-08-02
Bump version to "3.0.0".

ioquatix:
I know the change to `VERSION` from Array to String could be a compatibility issue, but I'd like to propose it and then see if anything breaks. I'd rather just be consistent with the standard `GemName::VERSION` which basically every other gem in the world follows.

`Rack::VERSION` has never been a useful constant for any real world situations anyway, so I don't think any real world code would be depending on it. In Rack 2, this was provided by the request `env`, but I guess it's possible some code was checking `Rack::VERSION`.

I don't think we should have "VERSION" and "RELEASE" as separate things. It seems like a snowflake thing to do, added complexity, no value.

### DIFF
--- a/lib/rack/server.rb
+++ b/lib/rack/server.rb
@@ -119,7 +119,7 @@ module Rack
           end
 
           opts.on_tail("--version", "Show version") do
-            puts "Rack #{Rack.version} (Release: #{Rack.release})"
+            puts "Rack #{Rack.release}"
             exit
           end
         end
--- a/lib/rack/version.rb
+++ b/lib/rack/version.rb
@@ -13,17 +13,34 @@
 
 module Rack
   # The Rack protocol version number implemented.
-  VERSION = "3.0.0"
+  VERSION = String.new("3.0.0")
+
+  def VERSION.join(separator)
+    warn "Rack::Version is now a String, use it directly!", uplevel: 1
+    VERSION.split('.').join(separator)
+  end
+
+  def VERSION.[](index)
+    warn "Rack::Version is now a String, use it directly!", uplevel: 1
+    VERSION.split('.')[index]
+  end
+
+  VERSION.freeze
 
   # Return the Rack protocol version as a dotted string.
   def self.version
+    # In a future release, say, 3.1?
+    # warn "Rack::VERSION is now a String, use it directly!", uplevel: 1
     VERSION
   end
 
   RELEASE = VERSION
-
+  deprecate_constant :RELEASE
+  
   # Return the Rack release as a dotted string.
   def self.release
+    # In a future release, say, 3.1?
+    # warn "Rack::VERSION is now a String, use it directly!", uplevel: 1
     VERSION
   end
 end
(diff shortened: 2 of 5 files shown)


ioquatix:
It was discussed in https://github.com/rack/rack/issues/1843 and my assertion/hypothesis would be that there is very little breakage of existing systems with my proposed change, but I'll also go looking for it.

ioquatix:
The other proposed option is to deprecate `VERSION` and `version` in Rack 3 and fix it in Rack 3.1 - I'd be fine with that, @jeremyevans if you don't agree with this PR do you agree with that process (that you proposed?) I'd like to think we only have to go on that path if we can find actual instances where this PR breaks real systems. My guess is I could count it on one hand.

ioquatix:
I scanned the top 10,000 Ruby Gems which by weight of downloads represents a huge proportion of all gems downloaded.

Working around the fact that it's an array (exactly my point, most users want a String):

```
bee/beerdb-1.2.2/lib/beerdb.rb:      "rack":          "#{Rack::RELEASE} (#{Rack::VERSION.join('.')})",     ## note: VERSION is the protocoll version as an array e.g.[1,2]
spo/sportdb-2.3.2/lib/sportdb.rb:      "rack":          "#{Rack::RELEASE} (#{Rack::VERSION.join('.')})",     ## note: VERSION is the protocoll version as an array e.g.[1,2]
veg/vegas-0.1.11/lib/vegas/runner.rb:          puts "rack #{Rack::VERSION.join('.')}"
```

Won't matter much:

```
exc/exception_notification-4.4.3/test/exception_notifier/email_notifier_test.rb:          * rack.version                              : #{Rack::VERSION}
```

Looks like it would need to be updated anyway, to remove version from `env`:

```
fun/functions_framework-1.0.0/lib/functions_framework/testing.rb:          ::Rack::RACK_VERSION    => ::Rack::VERSION,
miz/mizuno-0.6.11/lib/mizuno/rack_handler.rb:            env['rack.version'] = Rack::VERSION
rac/rack-client-0.4.2/lib/rack/client/adapter/base.rb:        env.update 'rack.version'       => Rack::VERSION
res/rest-client-components-1.5.0/lib/restclient/components.rb:        "rack.version" => ::Rack::VERSION,
web/webmock-3.14.0/lib/webmock/rack_response.rb:      env['rack.version']    = Rack::VERSION
```

ioquatix:
For the one use case, `Rack::VERSION.join('.')` we could do:

```
def VERSION.join(separator)
  warn "Rack::Version is now a String, use it directly!"
  VERSION.split('.').join(separator)
end
```

As a compatibility shim for a short time.

jeremyevans:
I think a better approach is deprecating `VERSION`, `VERSION_STRING`, and `version` in Rack 3.0, and leaving `RELEASE` and `release` alone.  After removal of the deprecated `VERSION`, `VERSION_STRING`, and `version` (in Rack 3.1 or 4), I'd be fine adding back `version` as an alias to `release`.

ioquatix:
Okay @jeremyevans I've updated to the way you proposed.

jeremyevans (approved):

PR #1942, opened 2022-08-04
Allow `Rack::Builder#run` to take a block instead of an argument.

ioquatix:
When I was first introduced to rack many years ago, I recalled that this particular part of the `config.ru` caused me a lot of initial confusion.

I think we should allow `run {|env| ...}` to work, it's more canonical Ruby and it's easy for new users to understand since they don't need to learn about multiple concepts (e.g. lambdas).

I can't see any major downsides to accepting this, except it will mean older servers won't be compatible with the new syntax, but I think that's an acceptable trade off given how many other things are also changing.

### DIFF
--- a/lib/rack/builder.rb
+++ b/lib/rack/builder.rb
@@ -157,22 +157,40 @@ module Rack
     ruby2_keywords(:use) if respond_to?(:ruby2_keywords, true)
     # :nocov:
 
-    # Takes an argument that is an object that responds to #call and returns a Rack response.
-    # The simplest form of this is a lambda object:
+    # Takes a block or argument that is an object that responds to #call and
+    # returns a Rack response.
+    #
+    # You can use a block:
+    #
+    #   run do |env|
+    #     [200, { "content-type" => "text/plain" }, ["Hello World!"]]
+    #   end
+    #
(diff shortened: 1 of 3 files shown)


lib/rack/builder.rb
```diff
@@ -157,22 +157,40 @@ def use(middleware, *args, &block)
...
+    #
+    #   module HelloWorld
+    #     def call(env)
+    #      [200, { "content-type" => "text/plain" }, ["Hello World"]]
+    #     end
+    #   end
+    #
+    #   run HelloWorld
+    def run(app = nil, &block)
+      @run = app || block
```
jeremyevans, line 193: I think we should warn or raise if both app and block are passed.  warn and use argument instead of block is the more backwards compatible behavior, so that is probably better.

jeremyevans (changes requested):
One small change requested.  Otherwise looks fine.

lib/rack/builder.rb
ioquatix: Good idea! I like the idea of raising an exception better.

simi (approved):

jeremyevans (approved):

PR #1945, opened 2022-08-06
Add note about status changes to upgrade guide

auroranockert:
This documents what you need to do to be compatible with https://github.com/rack/rack/pull/1662.


ioquatix (approved):
This looks good to me. Thanks!

PR #1948, opened 2022-08-08
Reduce the number of files being distributed with Rack.

ioquatix:


jeremyevans (approved):

ioquatix:
```
-rw-r--r--  1 samuel users 645K Aug  9 17:26 rack-3.0.0.beta1.gem
-rw-r--r--  1 samuel users 105K Aug  9 17:26 rack-3.0.0.beta2.gem
```

ioquatix:
105k ... It still seems quite large. But I think after we remove some of the deprecated code it will get smaller (auth seems big).

PR #1949, opened 2022-08-08
Documentation updates.

ioquatix:


README.md
```diff
@@ -20,10 +20,10 @@ by a [supported web framework](#supported-web-frameworks):
 
 ```bash
 # Install it generally:
-$ gem install rack
+$ gem install rack --version 3.0.0.beta1
```
jeremyevans, line 23: I think we can use `gem install rack --pre` in these cases.  That way we shouldn't need an update for `rc1`.

jeremyevans (changes requested):
One small change recommended.

README.md
ioquatix: Thanks, that makes total sense to me.

jeremyevans (approved):

PR #1950, opened 2022-08-09
Add information on using Rack::Headers to the upgrade guide

jeremyevans:
Minor edits while here.

### DIFF
--- a/UPGRADE-GUIDE.md
+++ b/UPGRADE-GUIDE.md
@@ -246,7 +246,7 @@ each header key used, instead of using a plain hash for headers, you can use
 `Rack::Headers` on Rack 3.
 
 ```ruby
-  headers = Rack.release >= '3' ? Rack::Headers.new : {}
+  headers = defined?(Rack::Headers) ? Rack::Headers.new : {}
 ```
 
 `Rack::Headers` is a subclass of Hash that will automatically lowercase keys:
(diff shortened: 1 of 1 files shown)


UPGRADE-GUIDE.md
```diff
@@ -241,6 +241,23 @@ def call(env)
 end
 ```
 
+If you want your code to work with Rack 3 without having to manually lowercase
+each header key used, instead of using a plain hash for headers, you can use
+`Rack::Headers` on Rack 3.
+
+```ruby
+  headers = Rack.release >= '3' ? Rack::Headers.new : {}
```
ioquatix, line 249: Very few features require version sniffing, and this is a common one.

I could think of two options

1. Backport `Rack::Headers` to Rack 2 as a no-op, e.g. `Rack::Headers = Hash`.
2. Feature detect, e.g. `Rack.const_defined?(:Headers)`.

Do you have any opinion about how we can make this easier?

jeremyevans: `Rack.const_defined?(:Headers)` or `defined?(Rack::Headers)` is fine with me.

jeremyevans: Pushed a change to use `defined?(Rack::Headers)`

ioquatix (approved):
LGTM.

PR #1951, opened 2022-08-21
Do not add lib to load path when running separate tests

jeremyevans:
This is a better check that the internals are correctly using
require_relative and not relying on autoload.


ioquatix (approved):
LGTM.

PR #1952, opened 2022-08-26
Add `rack.response_finished` to `Rack::Lint`.

ioquatix:
This updates Rack::Lint to validate that `rack.response_finished` is an
array of callables when present in the `env`. e.g. procs, lambdas, or
objects that respond to `call`.

This validates that:

- `rack.response_finished` is an array
- The contents of the array all respond to `call`

And specifies that:

- The callback will receive `(env, status, headers, error)` as arguments.

Bringing https://github.com/rack/rack/pull/1802 to completion according to @tenderlove's recommendations/requirements.

### DIFF
--- a/SPEC.rdoc
+++ b/SPEC.rdoc
@@ -132,10 +132,12 @@ There are the following restrictions:
   set. <tt>PATH_INFO</tt> should be <tt>/</tt> if
   <tt>SCRIPT_NAME</tt> is empty.
   <tt>SCRIPT_NAME</tt> never should be <tt>/</tt>, but instead be empty.
-<tt>rack.response_finished</tt>:: An array of callables run after the HTTP response has been finished,
-either normally or with an error (e.g. the client disconnected).
-Invoked with <tt>(env, status, headers, error)</tt> arguments. If there is no error sending the response,
-the error argument will be +nil+, otherwise you can expect an exception object, e.g. +IOError+.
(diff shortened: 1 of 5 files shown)


SPEC.rdoc
```diff
@@ -132,6 +132,10 @@ There are the following restrictions:
   set. <tt>PATH_INFO</tt> should be <tt>/</tt> if
   <tt>SCRIPT_NAME</tt> is empty.
   <tt>SCRIPT_NAME</tt> never should be <tt>/</tt>, but instead be empty.
+<tt>rack.response_finished</tt>:: An array of callables run after the HTTP response has been finished,
```
jeremyevans, line 135: Maybe `s/finished, either normally or with/fully sent to the client, or there was/`.  I'm not sure whether `finished` is descriptive enough.

jeremyevans (commented):
This doesn't check that if `rack.response_finished` is present, each entry in the array, if called, is called with the four arguments.  I'm not sure whether it is not worth wrapping each callable in the array in a wrapper that does such checks, but we do take similar steps in other cases (e.g. InputWrapper, ErrorWrapper, StreamWrapper).

SPEC.rdoc
ioquatix: I'm fine with that, thanks.

ioquatix: I've reworked this to clarify in a more abstract sense.

tenderlove (approved):

Issue #1946, opened 2022-08-07
Consider introducing `rack.protocol` env variable.


ioquatix:
To correctly handle WebSockets, HTTP/1 requires the use of the upgrade headers, while HTTP/2 requires the use of `:protocol` pseudo header.

We don't have any mechanism like we discussed in https://github.com/rack/rack/issues/1878 for putting these pseudo headers into `env`.

Therefore, I've been using `rack.protocol` and mapping it specially.

- In HTTP/1, the `upgrade` header is mapped to `rack.protocol`, can contain multiple values, and the 101 response should include the specific protocol upgrade. Response should include `rack.protocol` to indicate an upgrade.
- In HTTP/2, the `:protocol` pseudo header is mapped to `rack.protocol`, can only contain a single value, and the 200 response indicates that the stream will transport that protocol.

Another option is to map HTTP/2 `:protocol` header to `env["HTTP_UPGRADE"]` although this feels a little ugly, and the same with the response header. HTTP/2 RFCs explicitly call out that this is invalid, that the headers are not the same. However, if we consider rack through the lens of a CGI gateway, this is probably acceptable - except that we don't do this strictly for other features.

If we can discuss this and get agreement on a specification, I'll create a PR. This is not urgent, but it would be nice to have, and to be compatible with Falcon's existing behaviour where possible. If not, it's okay to change.

ioquatix:
Another example of this problem: https://github.com/socketry/falcon/issues/202

ioquatix:
This is implemented by <https://github.com/rack/rack/pull/1954>.

PR #1954, opened 2022-08-28
Introduce `rack.protocol` header for handling version-agnostic protocol upgrades.

ioquatix:
https://github.com/rack/rack/issues/1946 for discussion about the semantics/problem.

### DIFF
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -81,8 +81,9 @@ module Rack
         end
 
         ## and the *body*.
-        check_content_type(@status, @headers)
-        check_content_length(@status, @headers)
+        check_content_type_header(@status, @headers)
+        check_content_length_header(@status, @headers)
+        check_rack_protocol_header(@status, @headers)
         @head_request = @env[REQUEST_METHOD] == HEAD
 
         @lint = (@env['rack.lint'] ||= []) << self
@@ -181,6 +182,16 @@ module Rack
         ## <tt>rack.hijack</tt>:: See below, if present, an object responding
         ##                        to +call+ that is used to perform a full
         ##                        hijack.
+
+        ## <tt>rack.protocol</tt>:: An optional +Array+ of +String+, containing
+        ##                          the protocols advertised by the client in
+        ##                          the +upgrade+ header (HTTP/1) or the
+        ##                          +:protocol+ pseudo-header (HTTP/2).
+        if protocols = @env['rack.protocol']
+          unless protocols.is_a?(Array) && protocols.all?{|protocol| protocol.is_a?(String)}
+            raise LintError, "rack.protocol must be an Array of Strings"
+          end
+        end
 
         ## Additional environment specifications have approved to
         ## standardized middleware APIs. None of these are required to
@@ -723,9 +734,9 @@ module Rack
       end
 
       ##
-      ## === The content-type
+      ## ==== The +content-type+ Header
       ##
-      def check_content_type(status, headers)
+      def check_content_type_header(status, headers)
         headers.each { |key, value|
           ## There must not be a <tt>content-type</tt> header key when the +Status+ is 1xx,
           ## 204, or 304.
@@ -739,9 +750,9 @@ module Rack
       end
 
       ##
-      ## === The content-length
+      ## ==== The +content-length+ Header
       ##
-      def check_content_length(status, headers)
+      def check_content_length_header(status, headers)
         headers.each { |key, value|
           if key == 'content-length'
             ## There must not be a <tt>content-length</tt> header key when the
@@ -766,6 +777,29 @@ module Rack
         end
       end
 
+      ## 
+      ## ==== The +rack.protocol+ Header
+      ##
+      def check_rack_protocol_header(status, headers)
+        ## If the +rack.protocol+ header is present, it must be a +String+, and
+        ## must be one of the values from the +rack.protocol+ array from the
+        ## environment.
+        protocol = headers['rack.protocol']
+
+        if protocol
+          request_protocols = @env['rack.protocol']
+
+          if request_protocols.nil?
+            raise LintError, "rack.protocol header is #{protocol.inspect}, but rack.protocol was not set in request!"
+          elsif !request_protocols.include?(protocol)
+            raise LintError, "rack.protocol header is #{protocol.inspect}, but should be one of #{request_protocols.inspect} from the request!"
+          end
+        end
+      end
+      ##
+      ## Setting this value informs the server that it should perform a
+      ## connection upgrade. In HTTP/1, this is done using the +upgrade+
+      ## header. In HTTP/2, this is done by accepting the request.
       ##
       ## === The Body
       ##
--- a/test/spec_lint.rb
+++ b/test/spec_lint.rb
@@ -948,4 +948,38 @@ describe Rack::Lint do
                      [200, {}, ["foo"]]
                    }).call(env({ "rack.response_finished" => [-> (env) {}, lambda { |env| }, callable_object], "content-length" => "3" })).first.must_equal 200
   end
+
+  it "notices when the response protocol is specified in the response but not in the request" do
+    app = Rack::Lint.new(lambda{|env|
+      [101, {'rack.protocol' => 'websocket'}, ["foo"]]
+    })
+
+    lambda do
+      app.call(env())
+    end
+      .must_raise(Rack::Lint::LintError)
+      .message.must_match(/rack.protocol header is "websocket", but rack.protocol was not set in request/)
+  end
+
+  it "notices when the response protocol is specified in the response but not in the request" do
+    app = Rack::Lint.new(lambda{|env|
+      [101, {'rack.protocol' => 'websocket'}, ["foo"]]
+    })
+
+    lambda do
+      app.call(env('rack.protocol' => ['smtp']))
+    end
+      .must_raise(Rack::Lint::LintError)
+      .message.must_match(/rack.protocol header is "websocket", but should be one of \["smtp"\] from the request!/)
+  end
+
+  it "pass valid rack.protocol" do
+    app = Rack::Lint.new(lambda{|env|
+      [101, {'rack.protocol' => 'websocket'}, ["foo"]]
+    })
+
+    response = app.call(env({'rack.protocol' => ['websocket']}))
+
+    response.first.must_equal 101
+  end
 end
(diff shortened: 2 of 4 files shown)


lib/rack/lint.rb
```diff
@@ -179,6 +180,11 @@ def check_environment(env)
         ##                        to +call+ that is used to perform a full
         ##                        hijack.
 
+        ## <tt>rack.protocol</tt>:: If the request is an HTTP/1 upgrade or
```
jeremyevans, line 183: I'm not in favor of the `rack.protocol` naming.  It seems like it would be which HTTP version is in use (`SERVER_PROTOCOL`) or `http` or `https` (`rack.url_scheme`).

I assume the HTTP/1 upgrade header is already available in `HTTP_UPGRADE`.  Is the HTTP/2 protocol pseudo header exposed anywhere yet?

Absent a better idea, I recommend switching to `rack.upgrade_protocol`, since it looks this this is used for upgrading the protocol used for the connection.

Regardless of what we select, I think we want to explicitly specify that this in an array (even if it can only contain one value in certain HTTP versions), since it looks like we specify that the env key should be an array later.

lib/rack/lint.rb
```diff
@@ -714,6 +720,28 @@ def verify_content_length(size)
         end
       end
 
+      ## 
+      ## ==== The +rack.protocol+ Header
+      ##
+      def check_rack_protocol_header(status, headers)
+        ## If the +rack.protocol+ header is present, it must be a String, and
+        ## must be one of the values from the +rack.protocols+ array from the
```
jeremyevans, line 728: Do we want to name the env key different than the response header?  I'm not against that, but doing so is inconsistent with the changelog and earlier changes in the diff and the code that Lint runs.

lib/rack/lint.rb
```diff
@@ -714,6 +720,28 @@ def verify_content_length(size)
...
+        ## environment.
+        protocol = headers['rack.protocol']
+        if protocol
+          request_protocols = Array(@env['rack.protocol'])
```
jeremyevans, line 732: If the env entry is required to be an array if present, we shouldn't need to use `Array` to turn it into an array.

jeremyevans (commented):
See inline comments for details.

Can you add a discussion of how connection upgrades in a rack app are currently handled without this?  I assume they are handled differently in HTTP/1 and HTTP/2, but before trying to standardize something to unify behavior between HTTP versions, I would like to know what a rack app currently needs to return to initiate a procotol upgrade.  It sounds like HTTP/1 uses the Upgrade request and response headers, but I'm not sure how it is handled currently for HTTP/2, or whether it is impossible in HTTP/2 without this change (or similar change to handle HTTP/2 pseudo headers).

lib/rack/lint.rb
ioquatix: I'm okay with that, however in HTTP/2, the semantics are that it should only contain a single value, i.e. a string, while in HTTP/1, it can have multiple values. I'm not sure if we want to just specify that it can be a string (single value HTTP/2 semantics) or Array (multiple values HTTP/1 semantics) or an Array with a single value (HTTP/2 wrapped in HTTP/1 semantics). It's probably easier to use if we always assume it's an Array, but 99.9999999999% of the time, this will always be a single value, `websocket`. I've never seen anyone with an upgrade header like `upgrade: websocket, smtp, redis`. I think that's why HTTP/2 standardised it as a single-value pseudo header - realistically no one is asking to upgrade to several distinct protocols, and/or protocol versions. I don't have a strong opinion, but I consider the HTTP/1 semantic to be largely "deprecated" based on the directions the standards appear to be going in. I don't have a strong opinion here either way.

lib/rack/lint.rb
ioquatix: No, I made a mistake, it should be `rack.protocol` in both directions.

lib/rack/lint.rb
ioquatix: HTTP/2 literally called this the `:protocol` pseudo header. I actually don't have a strong opinion about the name but I've already used this name for my current implementation of websocket simply because I had to choose something and it seemed like the logical approach.

I also checked https://datatracker.ietf.org/doc/html/rfc3875 and could find no reference about how to handle `connection: upgrade` requests. It doesn't seem like the CGI specification which we follow for our `env` keys has any guidance on the matter.

`rack.upgrade_protocol` - I would suggest we avoid this, or the "upgrade" terminology in general, as HTTP/2 explicitly calls this out in a number of places as being unsupported and I think the word "upgrade" is deprecated in favour of "connect" and "protocol".

- https://www.rfc-editor.org/rfc/rfc9113.html#name-the-upgrade-header-field
- https://www.rfc-editor.org/rfc/rfc8441.html#section-1

The specific introduction of the protocol pseudo header is here: https://www.rfc-editor.org/rfc/rfc8441.html#section-4

I'm fine with specifying that it should be an `Array`.

lib/rack/lint.rb
jeremyevans: If it is going to 99.9999999999% be a string in HTTP/1, and 100% in HTTP/2, we probably should just specify it must be a string.  If a client submits `upgrade: websocket, smtp`, and the app uses the response header `upgrade: websocket`, it wouldn't pass Lint since it wouldn't match, and I think that's OK.  We shouldn't design for the 00.0000000001% use case.

lib/rack/lint.rb
jeremyevans: `rack.connect_protocol` maybe?  `rack.protocol` is too confusing as a name for this, IMO.

I would still like to know how changing the connection protocol is currently handled for HTTP/2 in rack.  Is it impossible with the current design, does it require server-specific code, or is it possible and just implemented a different way?

ioquatix:
> Can you add a discussion of how connection upgrades in a rack app are currently handled without this?

Essentially, they aren't.

I tried to do protocol upgrade with Puma, thinking at least it should work with a hijacked body, but apparently it's unsupported: https://github.com/puma/puma/issues/2914

I don't know about other servers, but I assume they are equally complicated as it's never been fully specified or tested.

> I assume they are handled differently in HTTP/1 and HTTP/2, but before trying to standardize something to unify behavior between HTTP versions, I would like to know what a rack app currently needs to return to initiate a procotol upgrade.

Yes, it's totally different between HTTP/1 and HTTP/2 - HTTP/1 supports `connection: upgrade` while HTTP/2 uses the `CONNECT` method. There is very little similarity between the two methods except that the result is a bi-directional stream of some specific wire format.

The closest thing we can unify is that "I have a list of protocols I'd like you to use for this connection".

- HTTP/1 has multiple protocols and the response must indicate which one was selected and return status 101 and `upgrade: $protocol` to accept.
- HTTP/2 has a single protocol and the response must respond with status 200 to accept.

> It sounds like HTTP/1 uses the Upgrade request and response headers, but I'm not sure how it is handled currently for HTTP/2, or whether it is impossible in HTTP/2 without this change (or similar change to handle HTTP/2 pseudo headers).

Yes, HTTP/1 uses the upgrade mechanism, and no such mechanism exists for HTTP/2.

This might beg the question, why did HTTP/1 use an upgrade mechanism for WebSocket? My guess is because of poorly conforming browsers and servers. Upgrade is like an escape hatch. But it's not needed - HTTP/1.1 supports bi-directional streaming and in theory the same "CONNECT" style could work in HTTP/1.1 just fine. In my opinion, connection upgrade is a poorly designed feature of HTTP/1 and we should not seek to standardise on it, rather following the direction of HTTP/2+ with the protocol pseudo header. Unfortunately, the vast majority of web socket connections use HTTP/1 upgrade mechanism, so it's unavoidable, but I imagine in 5-10 years it will be much less relevant in practice.

The reason for this, is because HTTP/1 upgrade connections consume an entire TCP connection, which is the very reason HTTP/2+ was designed to avoid with stream multiplexing.

My guess is, eventually we won't need WebSockets - full bi-directional streaming will eventually be supported making this entire piece irrelevant, but that's still along way off: https://github.com/whatwg/fetch/issues/1438

lib/rack/lint.rb
ioquatix: `rack.protocol` is the intersection of all the different semantics. `connect_protocol` is too specific to HTTP/2 that it makes the HTTP/1 use case clumsy. It doesn't make a lot of sense to have `rack.connect_protocol` in the response headers map to an HTTP/1 connection upgrade IMHO. I understand the value of being more specific, but this isn't all or nothing, the intersection between HTTP/1 and HTTP/2 semantics is very wide and doesn't invite more specific naming IMHO. In other words, being too specific in the naming is equally confusing.

lib/rack/lint.rb
ioquatix: I'd be okay with this, but `async-http` will internally follow the HTTP semantics and has it as an Array, which is how servers should be checking it. Therefore, I think we should encourage servers to implement the behaviour correctly, essentially:

```
if env['rack.protocol'].find?{|protocol| protocol.casecmp?('websocket')}
```

lib/rack/lint.rb
ioquatix: I probably should have been more specific. There is nothing that ties the `:protocol` pseudo header to the `CONNECT` method, so linking them as `rack.connect_protocol` would be a bit clumsy if later on another method was introduced that also used the `:protocol` pseudo header with a different method.

jeremyevans:
The related WhatWG issue is closed and implemented, but maybe only for half duplex and not full duplex. Is this still needed?  If so, I'm OK with merging with some language changes (server can ignore this if it doesn't support protocol upgrades), as long as @tenderlove does not object.

tenderlove:
> If so, I'm OK with merging with some language changes (server can ignore this if it doesn't support protocol upgrades), as long as @tenderlove does not object.

Yes, I agree.  I think it's fine.

ioquatix:
Thanks @jeremyevans @tenderlove I'll update the language as requested.

test/spec_lint.rb
```diff
@@ -948,4 +948,38 @@ def call(env, status, headers, error)
...
+    })
+
+    lambda do
+      app.call(env())
+    end
+      .must_raise(Rack::Lint::LintError)
+      .message.must_match(/rack.protocol header is "websocket", but rack.protocol was not set in request/)
+  end
+
+  it "notices when the respones protocol is specified in the response but not in the request" do
```
jeremyevans, line 964: This has the same description as previous, and both misspell `responses`

jeremyevans (changes requested):
One minor change related to spec descriptions.  OK to merge after that is fixed.

test/spec_lint.rb
ioquatix: Thanks, fixed!

PR #1956, opened 2022-08-29
Allow calling close on `rack.input`.

ioquatix:
My guess is this is a hang-over from CGI stdin/stdout model. It's advantageous to close the input if you don't need it to free any associated resources as soon as possible. I can't find any reason/justification as to why this is the way it is.

### DIFF
--- a/test/spec_lint.rb
+++ b/test/spec_lint.rb
@@ -742,14 +742,6 @@ describe Rack::Lint do
                      }).call(env("rack.input" => eof_weirdio))
     }.must_raise(Rack::Lint::LintError).
       message.must_match(/read\(nil\) returned nil on EOF/)
-
-    lambda {
-      Rack::Lint.new(lambda { |env|
-                       env["rack.input"].close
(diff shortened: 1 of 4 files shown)


test/spec_lint.rb
```diff
@@ -742,14 +742,6 @@ def each
                      }).call(env("rack.input" => eof_weirdio))
     }.must_raise(Rack::Lint::LintError).
       message.must_match(/read\(nil\) returned nil on EOF/)
-
```
jeremyevans, line 745: This spec should be switched to check for a successful result, not removed.

jeremyevans (changes requested):
I agree with this change in principle.  I think the reason that `close` was not allowed originally is that it would have broken rewindable input.  Now that input need not be rewindable, closing the input should be fine.

See inline note about the specs.

test/spec_lint.rb
ioquatix: Updated, thanks!

jeremyevans (approved):

PR #1959, opened 2022-08-31
The stream argument must implement `#<<`.

ioquatix:
This makes it easier to use the stream argument with template systems like ERB, Trenni, etc for real time template streaming/generation.


jeremyevans (approved):
I originally thought it would best to only add this restriction on callable bodies (i.e. not for hijack), as that approach would not introduce backwards compatibility issues.  However, I guess there is no really issue with also doing so for hijack, and there is an advantage in keeping things consistent between the two (callable bodies operating exactly the same as partial hijack currently).

ioquatix:
I understand. I've deliberately tried to keep rack partial hijack and rack streaming response semantics identical to make it easier for server implementor.

PR #1960, opened 2022-09-05
Fix unclosed strings in `UPGRADE-GUIDE.md`

joshmgross:
If I'm understanding the examples correctly, it looks like these strings are missing a closing `"`


ioquatix (approved):
Thanks, yes, that looks like a typo!

PR #1962, opened 2022-09-07
Remove lobster examples

dentarg:
Does not work since the rackup extraction (#1937):

```
~/src/rack main
arm64 $ ruby -Ilib example/protectedlobster.rb
<internal:/Users/dentarg/.arm64_rubies/3.1.2/lib/ruby/3.1.0/rubygems/core_ext/kernel_require.rb>:85:in `require': cannot load such file -- rack/lobster (LoadError)
	from <internal:/Users/dentarg/.arm64_rubies/3.1.2/lib/ruby/3.1.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
	from example/protectedlobster.rb:4:in `<main>'
```


ioquatix (approved):
LGTM.

PR #1966, opened 2022-09-09
Remove deprecated code.

ioquatix:


test/spec_utils.rb
```diff
@@ -702,17 +658,10 @@ def initialize(*)
...
-    Rack::Utils.make_delete_cookie_header(nil, 'name', {}).
-      must_equal "name=; max-age=0; expires=Thu, 01 Jan 1970 00:00:00 GMT"
-    Rack::Utils.add_remove_cookie_to_header(nil, 'name').
-      must_equal "name=; max-age=0; expires=Thu, 01 Jan 1970 00:00:00 GMT"
-  end
 end
 
 describe Rack::Utils, "get_byte_ranges" do
-  deprecated "pase simple byte ranges from env" do
+  it "pase simple byte ranges from env" do
```
jeremyevans, line 664: Not a bug in the PR, but may as well fix the typo in the description: `s/pase/parse/`

jeremyevans (approved):
Looks good.  See one inline comment, but that can be in a separate PR if you want.

test/spec_utils.rb
ioquatix:
```suggestion
  it "parse simple byte ranges from env" do
```

cklim24 (commented):
gem install sinatra
gem install puma # or any other server

Issue #1975, opened 2022-10-19
question: How to resolve `uninitialized constant Rack::Directory (NameError)`?


yumetodo:
When I install `rackup-0.2.2`, `rack-3.0.0`, `webrick-1.7.0`, `uninitialized constant Rack::Directory (NameError)` was occurred.

I've read these document and no change was found.
 
- https://www.rubydoc.info/gems/rack/Rack/Directory
- https://www.rubydoc.info/gems/rack/2.2.4/Rack/Directory

`config.ru`

```ruby
run Rack::Directory.new '.'
```

```
$gem install webrick 
Fetching webrick-1.7.0.gem
Successfully installed webrick-1.7.0
Parsing documentation for webrick-1.7.0
Installing ri documentation for webrick-1.7.0
Done installing documentation for webrick after 1 seconds
1 gem installed

$gem install rack
Fetching rack-3.0.0.gem
Successfully installed rack-3.0.0
Parsing documentation for rack-3.0.0
Installing ri documentation for rack-3.0.0
Done installing documentation for rack after 5 seconds
1 gem installed

$gem install rackup
Successfully installed rackup-0.2.2
Parsing documentation for rackup-0.2.2
Installing ri documentation for rackup-0.2.2
Done installing documentation for rackup after 0 seconds
1 gem installed

$rackup -p 4000
C:/msys64/home/yumetodo/foo/config.ru:1:in `block in <main>': uninitialized constant Rack::Directory (NameError)
        from C:/msys64/mingw64/lib/ruby/gems/3.1.0/gems/rack-3.0.0/lib/rack/builder.rb:106:in `eval'
        from C:/msys64/mingw64/lib/ruby/gems/3.1.0/gems/rack-3.0.0/lib/rack/builder.rb:106:in `new_from_string'
        from C:/msys64/mingw64/lib/ruby/gems/3.1.0/gems/rack-3.0.0/lib/rack/builder.rb:97:in `load_file'
        from C:/msys64/mingw64/lib/ruby/gems/3.1.0/gems/rack-3.0.0/lib/rack/builder.rb:67:in `parse_file'
        from C:/msys64/mingw64/lib/ruby/gems/3.1.0/gems/rackup-0.2.2/lib/rackup/server.rb:351:in `build_app_and_options_from_config'
        from C:/msys64/mingw64/lib/ruby/gems/3.1.0/gems/rackup-0.2.2/lib/rackup/server.rb:260:in `app'
[...]

PR #1976, opened 2022-10-20
fix: describe module autoloading in config.ru

yumetodo:
It seems to me that described at #1975 is not well documented.

### DIFF
--- a/UPGRADE-GUIDE.md
+++ b/UPGRADE-GUIDE.md
@@ -49,22 +49,6 @@ run do |env|
 end
 ```
 
-### `config.ru` autoloading is disabled unless `require 'rack'`
-
-Previously, rack modules like `rack/directory` was autoloaded. In Rack 3, you will need to write `require 'rack'` or require specific module explicitly.
-
-```diff
-+require 'rack'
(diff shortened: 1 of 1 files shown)


jeremyevans (changes requested):
Note that the same behavior was true in Rack 2 (need to `require 'rack'` to get autoloading), it just wasn't necessarily safe in Rack 2 to require specific components, since the components didn't specify their dependencies (nor were there tests for requiring specific components).  I think that the difference is that the extracted `rackup` gem only loads the parts of rack that it needs, while in Rack 2, `rackup` does `require 'rack'`.

If you are using a separate server (puma/unicorn/falcon/etc.), I don't think there is a difference in behavior. As that is the vast majority of production usage, this documentation should be specific that it only applies to `rackup`, and moved to the section that discusses `rackup`.

yumetodo:
@jeremyevans Thank you for your review. To follow your review, I move module autoloading section into `rackup` section.

jeremyevans (approved):
Looks good, thanks!

ioquatix:
This is really great, thanks for pointing out the issue here and making a PR to improve the documentation.

PR #1978, opened 2022-11-07
Updated simplest run sample

vimutter:
Current version gives:
```ruby
...gems/rack-2.2.4/lib/rack/builder.rb:176:in `run': wrong number of arguments (given 0, expected 1) (ArgumentError)
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,9 +37,9 @@ $ gem install rack-session rackup
 Create a file called `config.ru` with the following contents:
 
 ```ruby
-run do |env|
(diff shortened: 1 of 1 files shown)


jeremyevans:
That's not the current version, you are using the old version (2.2.4 instead of the current 3.0.0).  The code given in the Rack 3.0.0 README should work on Rack 3.0.0, and the code given in the Rack 2.2.4 README should work on Rack 2.2.4.

PR #1980, opened 2022-11-09
Do not look for override method if there is no form/parseable data

jcsjcs:
See discussion at https://github.com/jeremyevans/roda/discussions/291

### DIFF
--- a/lib/rack/method_override.rb
+++ b/lib/rack/method_override.rb
@@ -46,7 +46,7 @@ module Rack
     end
 
     def method_override_param(req)
-      req.POST[METHOD_OVERRIDE_PARAM_KEY]
+      req.POST[METHOD_OVERRIDE_PARAM_KEY] if req.form_data? || req.parseable_data?
     rescue Utils::InvalidParameterError, Utils::ParameterTypeError
       req.get_header(RACK_ERRORS).puts "Invalid or incomplete POST params"
     rescue EOFError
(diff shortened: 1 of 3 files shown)


jeremyevans:
Change looks good, but this needs tests before it can be committed.

jcsjcs:
Sorry, test added.

jeremyevans (approved):

lib/rack/method_override.rb
```diff
@@ -46,7 +46,7 @@ def allowed_methods
     end
 
     def method_override_param(req)
-      req.POST[METHOD_OVERRIDE_PARAM_KEY]
+      req.POST[METHOD_OVERRIDE_PARAM_KEY] if req.form_data? || req.parseable_data?
```
ioquatix, line 49: I will look myself, but if you already know you can help me, there is no significant performance cost of calling these methods right? (i.e. it's not parsing the form body etc).

jeremyevans: The methods just check the request method and headers, they don't deal with the body.

test/spec_method_override.rb
```diff
@@ -112,4 +112,15 @@ def app
 
     env["REQUEST_METHOD"].must_equal "POST"
   end
+
+  it "not set form input when the content type is JSON" do
+    env = Rack::MockRequest.env_for("/",
+            "CONTENT_TYPE" => "application/json",
+            method: "POST",
+            input: '{"_method":"options"}')
+    app.call env
```
ioquatix, line 121:
```suggestion
    env = Rack::MockRequest.env_for("/",
      "CONTENT_TYPE" => "application/json",
      method: "POST",
      input: '{"_method":"options"}')
    app.call env
```

PR #1981, opened 2022-11-11
Fix `respond_to?(:each)` with `Rack::Lint` with streaming bodies.

ioquatix:
This is the last piece required to get webrick passing the conformance test suite with Rack 3.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file. For info on
 ### Fixed
 
 - `MethodOverride` does not look for an override if a request does not include form/parseable data.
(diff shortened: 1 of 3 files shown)


ioquatix:
Confirmed fix in https://github.com/socketry/rack-conform/pull/11

jeremyevans (changes requested):
`to_ary` and `each` handling looks correct.  I can't see where `to_path` is implemented (there is `verify_to_path`, called by `each`).  If you are going to add this, I think you should implement `Wrapper#to_path`, and have it call `@body.to_path`.  Otherwise, if the body responds to `to_path`, `Wrapper` will say it responds to `to_path` when it doesn't, assuming I'm reading the code correctly.

jeremyevans (approved):

PR #1989, opened 2022-12-03
Fix Utils.build_nested_query to URL-encode all query string fields

jdufresne:
In the formatted query string, the field should be fully URL-encoded. The previous implementation left the square brackets as un-escaped.

It was noticed that if params were provided as a nested hash it produced a different result than a hash of string fields. For example, the params

    { 'q[s]' => 'name' }

would result in the query string:

    q%5Bs%5D=name

but

    { q: { s: 'name' } }

would result in:

    q[s]=name

Both forms now produce the same correct result with fully URL-encoded params.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file. For info on
 
 - `MethodOverride` does not look for an override if a request does not include form/parseable data.
 - `Rack::Lint::Wrapper` correctly handles `respond_to?` with `to_ary`, `each`, `call` and `to_path`, forwarding to the body. ([#1981](https://github.com/rack/rack/pull/1981), [@ioquatix])
+- `Utils.build_nested_query` URL-encodes nested field names including the square brackets.
 
 ## [3.0.0] - 2022-09-06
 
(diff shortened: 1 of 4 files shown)


ioquatix:
I'm okay with this but it seems like a potential breaking change.

jeremyevans (approved):
I agree that this could potentially be breaking, but RFC 3986 is clear that `[` and `]` are reserved characters and should be percent encoded.  So I think we should merge this.

AlexWayfer:
Yeah, sadly it has broken tests of my Rack-based framework unclear and non-expected, just by some patch-version release.

ioquatix:
IIUC, one way to avoid this problem in your test is to avoid asserting the encoded representation and actually parse the URL and extract the query parameters. This should work regardless of the encoding and is probably a more robust test in general.

AlexWayfer:
> IIUC, one way to avoid this problem in your test is to avoid asserting the encoded representation and actually parse the URL and extract the query parameters. This should work regardless of the encoding and is probably a more robust test in general.

I can agree. I've checked the tests again. It's simple tests for `path` method, returning String, sometimes with query. So, parsing it with standard tools may be a pain.

Maybe I should add `URI.decode_uri_component`, but, again, not sure how safe it is.

ioquatix:
> Maybe I should add URI.decode_uri_component

I think this is probably the correct approach.

PR #1993, opened 2022-12-05
Allow passing through streaming bodies.

ioquatix:
An alternative to https://github.com/rack/rack/pull/1987 which allows passing through streaming bodies without considering status code (e.g. allowing 101 response with streaming body).

### DIFF
--- a/lib/rack/response.rb
+++ b/lib/rack/response.rb
@@ -96,7 +96,7 @@ module Rack
     end
 
     def no_entity_body?
-      # The response body is an enumerable body and it is not allowed to have an entity body.,
+      # The response body is an enumerable body and it is not allowed to have an entity body.
       @body.respond_to?(:each) && STATUS_WITH_NO_ENTITY_BODY[@status]
     end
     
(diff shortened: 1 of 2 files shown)


lib/rack/response.rb
```diff
@@ -95,11 +95,23 @@ def chunked?
       CHUNKED == get_header(TRANSFER_ENCODING)
     end
 
+    def enumerable_body?
+      @body.respond_to?(:each)
+    end
+
+    def streaming_body?
```
jeremyevans, line 102: This function seems unnecessary.  If you think we should add it, we should have specs for it.

lib/rack/response.rb
```diff
@@ -95,11 +95,23 @@ def chunked?
       CHUNKED == get_header(TRANSFER_ENCODING)
     end
 
+    def enumerable_body?
```
jeremyevans, line 98: Since this is a public method, it should have it's own specs.  Otherwise, it should be a private method.

jeremyevans (changes requested):
I'm OK with the behavior.  See inline comments.

PR #1997, opened 2022-12-16
Better handling of multipart parsing when request is missing body.

### LINKED ISSUE #1994: should not process form data if it is a GET method
I would like to make another suggestion on this issue. #1787

The implementation of Go is a problem. But in reality there are such HTTP clients.

I think rack should ignore them.

sinatra has implemented the following response to this problem.

https://github.com/sinatra/sinatra/pull/1743/files

I think this sinatra change is terrible. However, due to the implementation of rack, sinatra has no choice but to do this.

I propose the following changes to this issue.

https://github.com/rack/rack/blob/a7d56490fd2fb41de8c94ead2f63fbad71c5c489/lib/rack/request.rb#L470-L475

We will change the implementation of this function so that form_data? is false for GET methods. For example:

```diff
-         (meth == POST && type.nil?) || FORM_DATA_MEDIA_TYPES.include?(type)
+         (meth == POST && type.nil?) || (meth != GET && FORM_DATA_MEDIA_TYPES.include?(type))
```

We rarely send form data with the GET method.

However, this implementation would cause multiple rack tests to fail. This should be corrected in the test itself.

### LINKED ISSUE #1995: Make `env['rack.input']` optional if no entity body is given.
See <https://github.com/rack/rack/issues/1994> for more context.

Most requests don't have an entity body.

It's difficult to detect this condition except for reading from the input stream provided.

But this may fail after the fact if there is no entity body.

I feel like it would be more efficient and transparent if we simply did not set `env['rack.input']` if there is no entity body.

See <https://github.com/rack/rack/issues/1994#issuecomment-1352117937> for an example use case.

### LINKED ISSUE #1996: Introduce `Rack::BadRequest` exception.
Something like:

```ruby
class Rack::BadRequest < StandardError
  def initialize(message, status = 400)
    super(message)
    @status = status
  end
  attr :status
end
```

This class would be used to indicate failures within Rack, and servers should be able to catch it and correctly map it to a failed response.

There are places in the multipart parsing code and `Rack::Request` where various conditions can be the result of a bad client HTTP request. It feels like it would be useful to have an exception that indicates that. The current situation right now, is we may raise an `IOError` or some other exception which becomes a 500 Internal Server Error, which while correct in a way (the server basically blew up), doesn't really reflect the nature of the "request being invalid" - e.g. a bad request, preconditions failed, etc.

See <https://github.com/rack/rack/issues/1994#issuecomment-1352117937> for an example use case.

ioquatix:
- Introduce `module Rack::BadRequest` for unified exception handling.
- Make `env['rack.input']` optional.

`rack.input` feels like a hangover from the CGI days when `$stdin` was always present and open. But in modern servers, we know that there is no entity body in most cases where it's relevant. However, we still need to allocate something which fails on the first call to `read` (`EOFError`, `IOError` or something else). There is actually no way to detect that there is no entity body. So, I'd like to fix that, as it makes handling the multipart error more straight forward.

If you think there are any other Rack error classes that should include `BadRequest`, please feel free to suggest it.

Fixes #1994. Fixes #1995. Fixes #1996.

### DIFF
--- a/lib/rack.rb
+++ b/lib/rack.rb
@@ -15,6 +15,7 @@ require_relative 'rack/version'
 require_relative 'rack/constants'
 
 module Rack
+  autoload :BadRequest, "rack/bad_request"
   autoload :Builder, "rack/builder"
   autoload :BodyProxy, "rack/body_proxy"
   autoload :Cascade, "rack/cascade"
--- a/lib/rack/bad_request.rb
+++ b/lib/rack/bad_request.rb
@@ -1,0 +1,4 @@
+module Rack
+  module BadRequest
+  end
+end
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -56,9 +56,6 @@ module Rack
         raise LintError, "No env given" unless @env
         check_environment(@env)
 
-        @env[RACK_INPUT] = InputWrapper.new(@env[RACK_INPUT])
-        @env[RACK_ERRORS] = ErrorWrapper.new(@env[RACK_ERRORS])
-
         ## and returns a non-frozen Array of exactly three values:
         @response = @app.call(@env)
         raise LintError, "response is not an Array, but #{@response.class}" unless @response.kind_of? Array
@@ -265,11 +262,9 @@ module Rack
         ## is reserved for use with the Rack core distribution and other
         ## accepted specifications and must not be used otherwise.
         ##
-
-        %w[REQUEST_METHOD SERVER_NAME QUERY_STRING SERVER_PROTOCOL
-           rack.input rack.errors].each { |header|
+        %w[REQUEST_METHOD SERVER_NAME QUERY_STRING SERVER_PROTOCOL rack.errors].each do |header|
           raise LintError, "env missing required key #{header}" unless env.include? header
-        }
+        end
 
         ## The <tt>SERVER_PORT</tt> must be an Integer if set.
         server_port = env["SERVER_PORT"]
@@ -328,10 +323,17 @@ module Rack
           raise LintError, "rack.url_scheme unknown: #{env[RACK_URL_SCHEME].inspect}"
         end
 
-        ## * There must be a valid input stream in <tt>rack.input</tt>.
-        check_input env[RACK_INPUT]
+        ## * There may be a valid input stream in <tt>rack.input</tt>.
+        if rack_input = env[RACK_INPUT]
+          check_input_stream(rack_input)
+          @env[RACK_INPUT] = InputWrapper.new(rack_input)
+        end
+
         ## * There must be a valid error stream in <tt>rack.errors</tt>.
-        check_error env[RACK_ERRORS]
+        rack_errors = env[RACK_ERRORS]
+        check_error_stream(rack_errors)
+        @env[RACK_ERRORS] = ErrorWrapper.new(rack_errors)
+
         ## * There may be a valid hijack callback in <tt>rack.hijack</tt>
         check_hijack env
 
@@ -384,7 +386,7 @@ module Rack
       ##
       ## The input stream is an IO-like object which contains the raw HTTP
       ## POST data.
-      def check_input(input)
+      def check_input_stream(input)
         ## When applicable, its external encoding must be "ASCII-8BIT" and it
         ## must be opened in binary mode, for Ruby 1.9 compatibility.
         if input.respond_to?(:external_encoding) && input.external_encoding != Encoding::ASCII_8BIT
@@ -488,7 +490,7 @@ module Rack
       ##
       ## === The Error Stream
       ##
-      def check_error(error)
+      def check_error_stream(error)
         ## The error stream must respond to +puts+, +write+ and +flush+.
         [:puts, :write, :flush].each { |method|
           unless error.respond_to? method
--- a/lib/rack/mock_request.rb
+++ b/lib/rack/mock_request.rb
@@ -41,11 +41,6 @@ module Rack
       end
     end
 
-    DEFAULT_ENV = {
-      RACK_INPUT        => StringIO.new,
-      RACK_ERRORS       => StringIO.new,
-    }.freeze
-
     def initialize(app)
       @app = app
     end
@@ -104,7 +99,7 @@ module Rack
       uri = parse_uri_rfc2396(uri)
       uri.path = "/#{uri.path}" unless uri.path[0] == ?/
 
-      env = DEFAULT_ENV.dup
+      env = {}
 
       env[REQUEST_METHOD]  = (opts[:method] ? opts[:method].to_s.upcase : GET).b
       env[SERVER_NAME]     = (uri.host || "example.org").b
@@ -144,20 +139,21 @@ module Rack
         end
       end
 
-      opts[:input] ||= String.new
       if String === opts[:input]
         rack_input = StringIO.new(opts[:input])
       else
         rack_input = opts[:input]
       end
 
-      rack_input.set_encoding(Encoding::BINARY)
-      env[RACK_INPUT] = rack_input
+      if rack_input
+        rack_input.set_encoding(Encoding::BINARY)
+        env[RACK_INPUT] = rack_input
 
-      env["CONTENT_LENGTH"] ||= env[RACK_INPUT].size.to_s if env[RACK_INPUT].respond_to?(:size)
+        env["CONTENT_LENGTH"] ||= env[RACK_INPUT].size.to_s if env[RACK_INPUT].respond_to?(:size)
+      end
 
       opts.each { |field, value|
-        env[field] = value  if String === field
+        env[field] = value if String === field
       }
 
       env
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -3,18 +3,28 @@
 require 'strscan'
 
 require_relative '../utils'
+require_relative '../bad_request'
 
 module Rack
   module Multipart
-    class MultipartPartLimitError < Errno::EMFILE; end
+    class MultipartPartLimitError < Errno::EMFILE
+      include BadRequest
+    end
 
     # Use specific error class when parsing multipart request
     # that ends early.
-    class EmptyContentError < ::EOFError; end
+    class EmptyContentError < ::EOFError
+      include BadRequest
+    end
 
     # Base class for multipart exceptions that do not subclass from
     # other exception classes for backwards compatibility.
-    class Error < StandardError; end
+    class BoundaryTooLongError < StandardError
+      include BadRequest
+    end
+
+    Error = BoundaryTooLongError
+    deprecate_constant :Error
 
     EOL = "\r\n"
     MULTIPART = %r|\Amultipart/.*boundary=\"?([^\";,]+)\"?|ni
@@ -96,7 +106,7 @@ module Rack
         if boundary.length > 70
           # RFC 1521 Section 7.2.1 imposes a 70 character maximum for the boundary.
           # Most clients use no more than 55 characters.
-          raise Error, "multipart boundary size too large (#{boundary.length} characters)"
+          raise BoundaryTooLongError, "multipart boundary size too large (#{boundary.length} characters)"
         end
 
         io = BoundedIO.new(io, content_length) if content_length
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -459,11 +459,6 @@ module Rack
         media_type_params['charset']
       end
 
-      # Whether the request potentially has input.
-      def input?
-        !!get_header(RACK_INPUT)
-      end
-
       # Determine whether the request body contains form-data by checking
       # the request content-type for one of the media-types:
       # "application/x-www-form-urlencoded" or "multipart/form-data". The
@@ -507,13 +502,21 @@ module Rack
         end
 
         begin
-          # If the form input was already memoized:
-          if get_header(RACK_REQUEST_FORM_INPUT) == get_header(RACK_INPUT)
-            return get_header(RACK_REQUEST_FORM_HASH)
+          rack_input = get_header(RACK_INPUT)
+
+          # If the form hash was already memoized:
+          if form_hash = get_header(RACK_REQUEST_FORM_HASH)
+            # And it was memoized from the same input:
+            if get_header(RACK_REQUEST_FORM_INPUT).equal?(rack_input)
+              return form_hash 
+            end
           end
 
           # Otherwise, figure out how to parse the input:
-          if input? && (form_data? || parseable_data?)
+          if rack_input.nil?
+            set_header RACK_REQUEST_FORM_INPUT, nil
+            set_header(RACK_REQUEST_FORM_HASH, {})
+          elsif form_data? || parseable_data?
             unless set_header(RACK_REQUEST_FORM_HASH, parse_multipart)
               form_vars = get_header(RACK_INPUT).read
 
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -459,6 +459,11 @@ module Rack
         media_type_params['charset']
       end
 
+      # Whether the request potentially has input.
+      def input?
+        !!get_header(RACK_INPUT)
+      end
+
       # Determine whether the request body contains form-data by checking
       # the request content-type for one of the media-types:
       # "application/x-www-form-urlencoded" or "multipart/form-data". The
@@ -469,9 +474,10 @@ module Rack
       # content-type header is provided and the request_method is POST.
       def form_data?
         type = media_type
-        meth = get_header(RACK_METHODOVERRIDE_ORIGINAL_METHOD) || get_header(REQUEST_METHOD)
 
-        (meth == POST && type.nil?) || FORM_DATA_MEDIA_TYPES.include?(type)
+        request_method = get_header(RACK_METHODOVERRIDE_ORIGINAL_METHOD) || get_header(REQUEST_METHOD)
+
+        (request_method == POST && type.nil?) || FORM_DATA_MEDIA_TYPES.include?(type)
       end
 
       # Determine whether the request body contains data by checking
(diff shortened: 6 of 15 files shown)


lib/rack/request.rb
```diff
@@ -497,7 +497,8 @@ def GET
       # multipart/form-data.
       def POST
```
ioquatix, line 498: This method seems a little messy, I might try to tidy it up, if we approve of this general (PR) direction.

test/spec_mock_request.rb
```diff
@@ -46,7 +46,7 @@
   end
 
   it "should handle a non-GET request with both :input and :params" do
-    env = Rack::MockRequest.env_for("/", method: :post, input: nil, params: {})
+    env = Rack::MockRequest.env_for("/", method: :post, input: "", params: {})
     env["PATH_INFO"].must_equal "/"
     env.must_be_kind_of Hash
     env['rack.input'].read.must_equal ''
```
ioquatix, line 52: This line fails, which I think is reasonable, if you specify `nil` input.

MSP-Greg:
LGTM

> If you think there are any other Rack error classes that should include BadRequest, please feel free to suggest it.

If that's a possibility, might `Rack::BadRequest` have some way of providing the status, headers & body?  Maybe have it respond to `call`, just like an app?

jeremyevans:
> @MSP-Greg @dentarg @nateberkopec can I please get your feedback on this change as you would be expected to add `rescue Rack::BadRequest` and map that to a `400` error.

I think `would be expected to` is better phrased as `can`.  Neither server, middleware, frameworks, or applications should have to require rack the library to be rack compatible.

Personally, I don't think servers should directly be rescuing the exception.  We should probably ship a middleware that does the handling (maybe `Rack::BadRequest::Middleware`), which servers/frameworks/applications that want to handle this such exceptions automatically can use.

> If that's a possibility, might `Rack::BadRequest` have some way of providing the status, headers & body? Maybe have it respond to `call`, just like an app?

I don't think we should do that.  Anything that could provide the status, headers, and body should just return the expected rack response array instead of raising an exception and relying on something else to do the translation.

ioquatix:
> Personally, I don't think servers should directly be rescuing the exception.

How, in your opinion, should we turn this error into a 400 Bad Request error?

jeremyevans:
> > Personally, I don't think servers should directly be rescuing the exception.
> 
> How, in your opinion, should we turn this error into a 400 Bad Request error?
> 
> The only other option is some kind of middleware, but is that really good enough? 

Directly after the sentence you quoted, I wrote `We should probably ship a middleware that does the handling`, which indicates I considered it good enough.

> As it stands, it feels like some parts of `Rack::Request` would need to signal "This input environment is bad/invalid" and we'd want to communicate that with a 400-status rather than a 500-status.

Code that uses `Rack::Multipart` for parsing (or other parts of Rack where this exception could be used) should have no issues using `Rack::BadRequest::Middleware` to handle it at the middleware level.

MSP-Greg:
From [RFC 9110 15.5 "Client Error 4xx"](https://www.rfc-editor.org/rfc/rfc9110.html#section-15.5):

"the server SHOULD send a representation containing an explanation of the error situation, and whether it is a temporary or permanent condition. These status codes are applicable to any request method. User agents SHOULD display any included representation to the user."

Maybe it's best left to the app, rather than Rack?

nateberkopec:
"No one has to `require rack`" feels like something that's too useful to lose.

ioquatix:
Well, I'm also fine with this just being a 500 error. But that won't solve the original problem.

dentarg:
I think it is okay if Rack just raises some error that web frameworks can rescue and return whatever status code they want for it. It is of course nice if there's a middleware included with Rack that can handle them automatically if you are fine with the responses that middleware makes.

I'll comment on the suggestion in #1996 where `BadRequest` has an attribute `status`. We got this bug report in Sinatra https://github.com/sinatra/sinatra/issues/1518 and indeed the bug was with Sinatra (only checking if certain method existed), but seeing that bug I would be a bit wary of introducing error objects with such common method names. Maybe there are other code bases making the same mistake as Sinatra.

lib/rack/multipart/parser.rb
```diff
@@ -3,18 +3,23 @@
 require 'strscan'
 
 require_relative '../utils'
+require_relative '../bad_request'
 
 module Rack
   module Multipart
-    class MultipartPartLimitError < Errno::EMFILE; end
+    class MultipartPartLimitError < Errno::EMFILE;
```
catatsuy, line 10: you forgot to remove ;

catatsuy:
I think that other language implementations would often ignore this.
But if you don't want them to be processed, a new implementation would be more appropriate than the current one

lib/rack/multipart/parser.rb
ioquatix: Fixed, thanks!

ioquatix:
> Maybe it's best left to the app, rather than Rack?

The problem with this, is that Rack is the one actually generating the errors, because `Rack::Request` is a library level implementation, not part of "rack the interface". We are literally providing application level constructs that are independent of the server interface, and those can signal failures due to bad request data, and we need some way to signal this so that applications can handle it correctly.

ioquatix:
@jeremyevans what do you think of the proposed change?

The middleware `Rack::BadRequest::Middleware` can help for people who want to adopt it.

The only issue is now:

```
    class EmptyContentError < ::EOFError
      include BadRequest
      # "Middleware" as a constant is defined here?
    end
```

Is it acceptable or should we reorganise the modules, e.g.

```
module Rack::BadRequest

class Rack::ErrorHandler
  def call(env)
    @app.call(env)
  rescue Rack::BadRequest
    [400 ...
  rescue
    [500 ...
  end
end
```

etc

lib/rack/bad_request.rb
```diff
@@ -0,0 +1,4 @@
+module Rack
+  module BadRequest
```
jeremyevans, line 2: If this is only used by the multipart parsing, it should be `Rack::Multipart::BadRequest` and moved to `rack/multipart.rb`

lib/rack/error_handler.rb
```diff
@@ -0,0 +1,23 @@
+module Rack
+  class ErrorHandler
+    def initialize(app)
+      @app = app
+    end
+
+    def bad_request(error)
+      [400, {}, ["Bad Request: #{error.message}"]]
```
jeremyevans, line 8: This could potentially leak sensitive information.  We should not include ruby exception messages in response bodies.  Same for `internal_server_error`.

lib/rack/error_handler.rb
```diff
@@ -0,0 +1,23 @@
+module Rack
+  class ErrorHandler
```
jeremyevans, line 2: Similar to `BadRequest`, unless this handles something other than multipart errors, it should be moved under the `Multipart` namespace.

Also, there doesn't appear to be any specs for this.

lib/rack/error_handler.rb
```diff
@@ -0,0 +1,23 @@
...
+
+    def internal_server_error(error)
+      [500, {}, ["Internal Server Error: #{error.message}"]]
+    end
+
+    def call(env)
+      @app.call(env)
+    rescue BadRequest => error
+      return bad_request(error)
+    rescue => error
```
jeremyevans, line 19: May be worth rescuing both ScriptError and StandardError here.

lib/rack/lint.rb
```diff
@@ -328,10 +323,20 @@ def check_environment(env)
...
+          @env[RACK_INPUT] = InputWrapper.new(rack_input)
+        end
+
         ## * There must be a valid error stream in <tt>rack.errors</tt>.
-        check_error env[RACK_ERRORS]
+        if rack_errors = env[RACK_ERRORS]
+          check_error_stream(rack_errors)
+          @env[RACK_ERRORS] = ErrorWrapper.new(rack_errors)
+        else
+          raise LintError, "rack.errors is not set"
```
jeremyevans, line 337: It appears that this is redundant with the `env missing required key` check above.

lib/rack/lint.rb
```diff
@@ -328,10 +323,20 @@ def check_environment(env)
...
-        check_input env[RACK_INPUT]
+        ## * There may be a valid input stream in <tt>rack.input</tt>.
+        if rack_input = env[RACK_INPUT]
+          check_input_stream(rack_input)
+          @env[RACK_INPUT] = InputWrapper.new(rack_input)
+        end
+
         ## * There must be a valid error stream in <tt>rack.errors</tt>.
-        check_error env[RACK_ERRORS]
+        if rack_errors = env[RACK_ERRORS]
```
jeremyevans, line 333: Because of the `env missing required key check` above, this `if` is unnecessary, as `rack.errors` should always be set at this point.

lib/rack/mock_request.rb
```diff
@@ -144,20 +143,25 @@ def self.env_for(uri = "", opts = {})
         end
       end
 
-      opts[:input] ||= String.new
+      unless opts.key?(:input)
```
jeremyevans, line 146: If we are switching from must to may, it seems a better default would be to not set an empty string if `:input` is not provided.

test/spec_mock_request.rb
jeremyevans: Seems like the spec should be changed so that when :input is `nil` or not given, `env['rack.input']` is not present. Or there should be a separate spec for that.

test/spec_multipart.rb
```diff
@@ -42,6 +42,13 @@ def multipart_file(name)
     }.must_raise Rack::Multipart::Error
   end
 
+  it "raises a bad request exception if no body is given" do
```
jeremyevans, line 45: Probably want a more complete description, such as `... if no body is given but content type indicates a multipart body`

jeremyevans (changes requested):
I think this PR should be split.  There should be a separate PRs for the spec change from must to may, for BadRequest, and for ErrorHandler.

lib/rack/bad_request.rb
ioquatix: I think the goal is to use it in other places, but this is the first place which makes the most sense. I could imagine potential usage in `Rack::Request`.

lib/rack/error_handler.rb
ioquatix: What about the error class name?

lib/rack/error_handler.rb
ioquatix: It's currently handling ALL errors which aren't `Rack::BadRequest` and mapping them to status=500.

I haven't written specs because it's just an example of what we can do.

lib/rack/error_handler.rb
ioquatix: What status codes would we map them to?

ioquatix: Okay, all mapping to internal server error.

lib/rack/bad_request.rb
ioquatix: I've added it to `query_parser.rb` too.

lib/rack/lint.rb
ioquatix: Thanks, fixed.

lib/rack/lint.rb
ioquatix: Thanks, fixed.

test/spec_multipart.rb
ioquatix: Done, thanks for the clear test title.

lib/rack/mock_request.rb
ioquatix: I'm fine with that but it breaks compatibility since the default changes. Let me see how tricky it is.

ioquatix: I've set the default to be nil and fixed the failing tests, so you can see the impact as a separate commit.

lib/rack/error_handler.rb
ioquatix: Added some specs.

test/spec_mock_request.rb
ioquatix: There are several other specs which I think test the case you are concerned about.

jeremyevans (changes requested):
Please split this into 3 separate pull requests, one for rack.input, one for BadRequest, and one for ErrorHandler.

lib/rack/error_handler.rb
```diff
@@ -0,0 +1,40 @@
...
+        if error.respond_to?(:full_message)
+          error_stream.puts(error.full_message)
+        else
+          error_stream.puts(error.message)
+        end
+      end
+    end
+
+    def bad_request(error)
+      [400, {}, ["Bad Request: #{error.class}"]]
```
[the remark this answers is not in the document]
ioquatix: Apparently the RFCs require some descriptive information. Do we want to ignore that?

lib/rack/error_handler.rb
```diff
@@ -0,0 +1,40 @@
...
+      @app = app
+    end
+
+    def log_error(env, error)
+      if error_stream = env[RACK_ERRORS]
+        # Exception#full_message was introduced in Ruby 2.5+. If it's not
+        # available, we fall back to the old behavior of just printing the
+        # exception message.
+        if error.respond_to?(:full_message)
+          error_stream.puts(error.full_message)
```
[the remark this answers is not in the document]
ioquatix: That's what `full_message` includes.

lib/rack.rb
```diff
@@ -25,6 +25,7 @@ module Rack
   autoload :ContentLength, "rack/content_length"
   autoload :ContentType, "rack/content_type"
   autoload :ETag, "rack/etag"
+  autoload :ErrorHandler, "rack/error_handler"
```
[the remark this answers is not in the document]
ioquatix: Thanks, good idea, done.

test/spec_lint.rb
```diff
@@ -13,8 +13,12 @@
     [200, { "content-type" => "test/plain", "content-length" => "3" }, ["foo"]]
   end
 
-  def env(*args)
-    Rack::MockRequest.env_for("/", *args)
+  def env(options = {})
+    unless options.key?(:input)
+      options[:input] = String.new
```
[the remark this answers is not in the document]
ioquatix: Lot's of tests fail without it because moist of the tests assume that it's an empty string. If we change that, it will make backporting harder.

lib/rack/mock_request.rb
```diff
@@ -42,8 +42,7 @@ def string
     end
 
     DEFAULT_ENV = {
```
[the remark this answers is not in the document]
ioquatix: That seems reasonable, I've done it.

lib/rack/error_handler.rb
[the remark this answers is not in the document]
jeremyevans: Then the else branch should be modified to get the same behavior if `full_message` is not available.

lib/rack/error_handler.rb
```diff
@@ -0,0 +1,40 @@
+require_relative 'constants'
+
+module Rack
+  class ErrorHandler
+    def initialize(app)
+      @app = app
+    end
+
+    def log_error(env, error)
+      if error_stream = env[RACK_ERRORS]
```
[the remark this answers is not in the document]
ioquatix: I personally think it's good to be defensive here.

An example of where this could go wrong: If a request is missing `rack.errors`, and the `Rack::Lint` raises an exception, the error handler will also blow up.

I don't really have a strong opinion either way, so assuming you don't care about this case, I'll change it.

lib/rack/error_handler.rb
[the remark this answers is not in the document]
jeremyevans: Maybe for BadRequest, we can allow the class name, since we presumably know all classes, and they are all parameter parsing errors and unlikely to provide sensitive information.  For the 500 error, we shouldn't include the class, as it could be any exception.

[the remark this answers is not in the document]
ioquatix: As I don't really have a strong opinion about it, I've removed it. We can always discuss whether we should add it again in the future.

test/spec_lint.rb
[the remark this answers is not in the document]
jeremyevans: That should give you a rough indication of how much code this change will break.  If we aren't willing to modify rack's own tests to fully implement the change, it's hard to argue the benefits of the change are worth the cost.

lib/rack/multipart/parser.rb
```diff
@@ -3,18 +3,28 @@
...
 
     # Base class for multipart exceptions that do not subclass from
     # other exception classes for backwards compatibility.
-    class Error < StandardError; end
+    class BoundaryTooLongError < StandardError
+      include BadRequest
+    end
+
+    Error = BoundaryTooLongError
+    deprecate_constant :Error
```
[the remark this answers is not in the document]
ioquatix: This error AFAIK doesn't exist in Rack 2.x?

It was introduced here: https://github.com/rack/rack/commit/d866dc42f5d886d94b91eacba9f7da8ba72e66ce

Deprecations only show up with warnings enabled.

How would you like to deprecate this?

lib/rack/error_handler.rb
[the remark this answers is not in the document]
jeremyevans: You make a good point regarding Lint.  I'm not sure anyone would use Lint and ErrorHandler together.  Lint is a development (possibly test) tool, while ErrorHandler is more for production (basic error messages, error details not visible to the user).  However, I don't object to exiting early if there is no error stream.

lib/rack/multipart/parser.rb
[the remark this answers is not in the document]
jeremyevans: I think we should keep it until Rack 4.

jeremyevans (changes requested):
I would still like this split into 3 pull requests, as previously indicated.

Also, please do not mark conversations as resolved.  The reviewer should decide when a conversation has been resolved, not the person proposing the change.

lib/rack/multipart/parser.rb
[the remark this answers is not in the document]
ioquatix: My question was when should we deprecate it?

test/spec_lint.rb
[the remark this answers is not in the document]
ioquatix: > If we aren't willing to modify rack's own tests to fully implement the change

Well, we did modify tack's own tests, and it was just in one place since the test fixture uses this and these tests assume an empty string.

ioquatix:
```
      "ActionDispatch::Http::Parameters::ParseError"       => :bad_request,
      "ActionController::BadRequest"                       => :bad_request,
      "ActionController::ParameterMissing"                 => :bad_request,
      "Rack::QueryParser::ParameterTypeError"              => :bad_request,
      "Rack::QueryParser::InvalidParameterError"           => :bad_request
```

Rails is already doing this basically, but in a worse way.

lib/rack/request.rb
ioquatix: Okay, I tidied this up.

tenderlove (changes requested):
I like the idea of having a BadRequest exception that people can rescue, but I agree with @jeremyevans that this PR should be split.  I am having a hard time reasoning about the changes with such a big diff.

ioquatix:
Okay, I will split it.

ioquatix:
See https://github.com/rack/rack/pull/2018 and https://github.com/rack/rack/pull/2019 (depends on 2018).

Issue #2003, opened 2023-01-07
Rack::Server was moved to rackup gem 


aithscel:
in Rack 2 it was possible to start the server with  Rack::Server.start
Now in Rack 3 this does not work  anymore  without the new rackup gem, as Rack::Server was moved there.

Maybe this information should be added in the Rack 3 upgrade guide.

aithscel:
ok, i have create a PR. 
For sake of completeness i have mentioned the move of Handler and Lobster too.

aithscel:
Fixed by #2005

PR #2005, opened 2023-01-08
Document the move of Rack::Server to rackup gem

aithscel:
Add information in upgrade guide about the move of Rack::Server and Handler and Lobster to rackup gem

Proposal for #2003


ioquatix (approved):

ioquatix:
LGTM! Thanks!

PR #2006, opened 2023-01-10
Rack::MethodOverride handle QueryParser::ParamsTooDeepError

casperisfine:
This middleware already handle two types of parsing issues but somehow not this one.

I think this is consistent with https://github.com/rack/rack/pull/835


jeremyevans (approved):

ioquatix:
Thanks!

PR #2008, opened 2023-01-12
Prefer lowercase "samesite" property name for cookies.

ioquatix:
Does it make sense to be consistent with other fields in cookies? We already lowercase `httponly`, `path` etc.


jeremyevans (approved):
The properties are case insensitive, so it shouldn't matter.  Shouldn't be a problem to make it consistent, though I imagine it will break some poorly coded tests.

ioquatix:
Rails already has issues with lower case `httponly`.

Issue #2009, opened 2023-01-16
`MethodOverride` middleware silently eats invalid parameters.


ioquatix:
See also <https://github.com/rack/rack/pull/2006>. cc @byroot @jeremyevans 

The current implementation of `method_override_param` effectively eats invalid parameters.

```ruby
    def method_override_param(req)
      req.POST[METHOD_OVERRIDE_PARAM_KEY] if req.form_data? || req.parseable_data?
    rescue Utils::InvalidParameterError, Utils::ParameterTypeError
      req.get_header(RACK_ERRORS).puts "Invalid or incomplete POST params"
    rescue EOFError
      req.get_header(RACK_ERRORS).puts "Bad request content body"
    end
```

The first time this is called, an exception is raised and caught. However, the params is cached, and so no follow-up call to params doesn't appear to fail with the same error. This can change the behaviour of web applications which expect `Rack::Request#POST` or `#params` to raise an exception on invalid input.

I'm not sure what the ideal behaviour should be. Maybe we need to cache the exception and raise it in subsequent calls?

Example reproduction:

```ruby
#!/usr/bin/env ruby

require 'stringio'
require 'rack/request'

env = {
  'REQUEST_METHOD' => 'POST',
  'PATH_INFO' => '/foo',
  'rack.input' => StringIO.new('invalid=bar&invalid[foo]=bar'),
  'HTTP_CONTENT_TYPE' => "application/x-www-form-urlencoded",
}

2.times do
  begin
    pp params: Rack::Request.new(env).POST
  rescue => error
    pp error: error
  end
end
```

Output:

```
["/home/samuel/.gem/ruby/3.2.0/gems/rack-3.0.3/lib/rack/query_parser.rb:79:in `block in parse_nested_query'",
 "/home/samuel/.gem/ruby/3.2.0/gems/rack-3.0.3/lib/rack/query_parser.rb:76:in `each'",
 "/home/samuel/.gem/ruby/3.2.0/gems/rack-3.0.3/lib/rack/query_parser.rb:76:in `parse_nested_query'",
 "/home/samuel/.gem/ruby/3.2.0/gems/rack-3.0.3/lib/rack/request.rb:648:in `parse_query'",
 "/home/samuel/.gem/ruby/3.2.0/gems/rack-3.0.3/lib/rack/request.rb:512:in `POST'",
 "/home/samuel/Desktop/request.rb:15:in `block in <main>'",
 "/home/samuel/Desktop/request.rb:13:in `times'",
[...]

PR #2010, opened 2023-01-16
`Rack::Request#POST` should consistently raise errors.

ioquatix:
Cache errors that occur when invoking `Rack::Request#POST` so they can be raised again later (consistently).

I propose we backport this to `3-0-stable` as this is a regression on 2.2 behaviour.

Fixes <https://github.com/rack/rack/issues/2009>.

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -497,7 +497,7 @@ module Rack
       # multipart/form-data.
       def POST
         if error = get_header(RACK_REQUEST_FORM_ERROR)
-          raise error
+          raise error.class, error.message, cause: error
         end
 
         begin
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -496,26 +496,35 @@ module Rack
       # This method support both application/x-www-form-urlencoded and
       # multipart/form-data.
       def POST
-        if get_header(RACK_INPUT).nil?
-          raise "Missing rack.input"
-        elsif get_header(RACK_REQUEST_FORM_INPUT) == get_header(RACK_INPUT)
-          get_header(RACK_REQUEST_FORM_HASH)
-        elsif form_data? || parseable_data?
-          unless set_header(RACK_REQUEST_FORM_HASH, parse_multipart)
-            form_vars = get_header(RACK_INPUT).read
+        if error = get_header(RACK_REQUEST_FORM_ERROR)
+          raise error
+        end
 
-            # Fix for Safari Ajax postings that always append \0
-            # form_vars.sub!(/\0\z/, '') # performance replacement:
-            form_vars.slice!(-1) if form_vars.end_with?("\0")
+        begin
+          if get_header(RACK_INPUT).nil?
+            raise "Missing rack.input"
+          elsif get_header(RACK_REQUEST_FORM_INPUT) == get_header(RACK_INPUT)
+            get_header(RACK_REQUEST_FORM_HASH)
+          elsif form_data? || parseable_data?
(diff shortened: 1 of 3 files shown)


lib/rack/request.rb
```diff
@@ -496,26 +496,35 @@ def GET
...
+            set_header RACK_REQUEST_FORM_INPUT, get_header(RACK_INPUT)
+            set_header(RACK_REQUEST_FORM_HASH, {})
           end
-          set_header RACK_REQUEST_FORM_INPUT, get_header(RACK_INPUT)
-          get_header RACK_REQUEST_FORM_HASH
-        else
-          set_header RACK_REQUEST_FORM_INPUT, get_header(RACK_INPUT)
-          set_header(RACK_REQUEST_FORM_HASH, {})
+        rescue => error
+          set_header(RACK_REQUEST_FORM_ERROR, error)
```
casperisfine, line 526: I don't think storing an Exception instance is ideal, because it means we also cache the backtrace:

```ruby
@error = nil

def params
  if @error
    raise @error
  end

  begin
    raise "oops"
  rescue => e
    @error = e
    raise
  end
end

def method_1
  params
end

def method_2
  params
end


method_1 rescue nil
method_2
```

```
/tmp/foo.rb:9:in `params': oops (RuntimeError)
	from /tmp/foo.rb:17:in `method_1'
	from /tmp/foo.rb:25:in `<main>'
```

Note how the backtrace say `method_1` when it should be `method_2`.

I see multiple solutions:

  - Store the error class and message only
  - Raise using `raise error.class, error.message, cause: error`

ioquatix: I don't have a strong opinion but I slightly prefer the latter, because it's still only one thing to store. So let's do that?

casperisfine (approved):
LGTM

lib/rack/request.rb
```diff
@@ -496,26 +496,35 @@ def GET
...
-            form_vars = get_header(RACK_INPUT).read
-
-            # Fix for Safari Ajax postings that always append \0
-            # form_vars.sub!(/\0\z/, '') # performance replacement:
-            form_vars.slice!(-1) if form_vars.end_with?("\0")
-
-            set_header RACK_REQUEST_FORM_VARS, form_vars
-            set_header RACK_REQUEST_FORM_HASH, parse_query(form_vars, '&')
+        if error = get_header(RACK_REQUEST_FORM_ERROR)
+          raise error.class, error.message, cause: error
```
jeremyevans, line 500: I don't think the use of `cause` is appropriate here.  `cause` is used when you are wrapping errors.  In this case, you are storing the original error as an optimization of storing one reference (error instance) instead of two (error class and message).

lib/rack/request.rb
jeremyevans: I agree with @casperisfine that caching the error instance is bad as we don't want to keep the backtrace (and potentially nested exception references).  If you only want to store one thing:

```ruby
# When storing error
set_header(RACK_REQUEST_FORM_ERROR, [error.class, error.message])

# When raising error
raise *error
```

jeremyevans (changes requested):
I agree that we should be consistent.  Please see requested changes about the implementation.

lib/rack/request.rb
ioquatix: Okay what about using `error.cause`?

jeremyevans (approved):

lib/rack/request.rb
ioquatix: Done.

PR #2013, opened 2023-01-17
This behaviour is not the same as Rack 2.x - is that acceptable?

ioquatix:
URI specifications mostly assume UTF-8 encoded URIs.

I don't have a strong opinion, but these tests confirm the current implementation behaviour - i.e. invalid UTF-8 attribute keys and values in the query string is acceptable.

Do we want to be more strict?

### DIFF
--- a/test/spec_request.rb
+++ b/test/spec_request.rb
@@ -572,11 +572,27 @@ class RackRequestTest < Minitest::Spec
   end
 
   it "parse the query string" do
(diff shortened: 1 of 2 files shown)


jeremyevans:
Not sure.  I think some users may want stricter behavior, so requests cannot introduce non-UTF-8 parameter keys/values.  However, forcing a UTF-8-only design seems quite heavy handed.

ioquatix:
I'm fine with that, so then you approve of these tests?

jeremyevans (approved):

PR #2014, opened 2023-01-18
Make RFC2183 work with Ruby 3.2's caching Regexp

jhawthorn:
Ruby 3.2 includes a cache-based regexp optimization, which is detailed in https://bugs.ruby-lang.org/issues/19104. Which can speed up the regex engine on many cases which would previously have resulted in a ReDoS.

One caveat of the implmentation is (quoting the issue):
> A bounded or fixed times repetition nesting in another repetition (e.g. /(a{2,3})*/). It is an implementation issue entirely, but we believe it is hard to support this case correctly.

Because of that limitation the RFC2183 regex was not previously able to use that optimization and was not able to mitigate two recent ReDoS CVEs.

This commit manually expands a `{2}` fixed repetition, which allows Ruby 3.2's optimization to take effect.

Before:

    > Regexp.linear_time?(/([0-9]{2})*/)
    => false
    > Regexp.linear_time?(Rack::Multipart::RFC2183)
    => false

After:

    > Regexp.linear_time?(/([0-9][0-9])*/)
    => true
    > Regexp.linear_time?(Rack::Multipart::RFC2183)
    => true

I want to make this change as additional hardening against possible ReDoS attacks in this regex. At the moment I don't know of any which are unpatched.


tenderlove (approved):

ioquatix:
Should we cut a new release for this or wait until the next release goes out?

jhawthorn:
This isn't a security patch (merely follow up hardening from security work) so I think we can wait for a regular release cycle.

PR #2017, opened 2023-01-19
Introduce mailmap for accurate git log and other commands.

ioquatix:
This is not extensive, but it's the best I could do in the time I had and the priorities for extracting gems. I believe we should continue to improve it.


ioquatix:
cc @leahneukirchen lmk if there are any other changes you would like to make.

tenderlove (approved):

PR #2018, opened 2023-01-19
Make `env['rack.input']` optional.

ioquatix:
Fixes https://github.com/rack/rack/issues/1994. Fixes https://github.com/rack/rack/issues/1995.

PR #2019, opened 2023-01-19
Add general `Rack::BadRequest`.

ioquatix:
Used to communicate a class of exceptions that represent 400 Bad Request semantics.

Fixes https://github.com/rack/rack/issues/1996.


jeremyevans:
I'll review this after #2018 is merged and it no longer requires #2020 (#2021 fixes that issue).

jeremyevans (approved):

PR #2023, opened 2023-01-19
Ignore CI failures for Ruby 2.4 and 2.5

jeremyevans:


.github/workflows/test.yaml
```diff
@@ -34,5 +34,7 @@ jobs:
       with:
         ruby-version: ${{matrix.ruby}}
         bundler-cache: true
+      continue-on-error: ${{ startsWith(matrix.ruby, '2.4') || startsWith(matrix.ruby, '2.5') }}
 
     - run: bundle exec rake
+      continue-on-error: ${{ startsWith(matrix.ruby, '2.4') || startsWith(matrix.ruby, '2.5') }}
```
ioquatix, line 40: It might just be easier to have a `matrix` variable called `unsupported` or `experimental` and write:

```suggestion
      continue-on-error: ${{ matrix.unsupported }}
```

ioquatix:
Thanks for your effort here.

As I said before, I'm on the fence regarding this. I'm slightly against technical debt for the sake of old versions.

.github/workflows/test.yaml
jeremyevans: Unfortunately, I don't know how to set new matrix variables in GitHub Actions, or whether it is possible to do so (where 2.4 and 2.5 could be marked as unsupported, and the other versions not).  If it is possible, can you point me to the documentation for doing so?

tenderlove (approved):

.github/workflows/test.yaml
dentarg: You would remove the ruby versions from the `ruby` list and instead include them in the matrix with `include` (where you would set `unsupported`), example of that at https://github.com/sporkmonger/addressable/blob/1fdd676753fb2f44c9d7e0953c97c3e6becfb36a/.github/workflows/test.yml#L116-L118 (but there it called `allow-failure` instead of `unsupported`)

PR #2025, opened 2023-01-21
Remove outdated contributing guideline

michaelherold:
There hasn't been an "extras" group in the Gemfile since 2020-05-25 so the step that mentions it hasn't been needed for almost two years. Also, since Bundler is deprecating the "without" behavior, modern Rubies issue a warning with their default version of Bundler.


jeremyevans (approved):

ioquatix:
Thanks!

PR #2026, opened 2023-01-22
Make QueryParser::Params a Hash subclass

jeremyevans:
Because of the to_params_hash method, this cannot be Hash itself. We should consider deprecating that method and maybe the Params constant as well.

Most of the spec changes are just changing overrides of Params#initialize.

This tweaks the "not create infinite loops with cycle structures" spec, because it wasn't actually testing for loops, it was checking the string representation was the same between Hash and Params.  This spec could probably be eliminated, but the tweak allows it to pass by checking that the level 1 hash string representation is the same as the level 2 hash string representation.

### DIFF
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -194,59 +194,7 @@ module Rack
       Utils.unescape(s)
     end
 
-    class Params
-      def initialize
-        @size   = 0
-        @params = {}
-      end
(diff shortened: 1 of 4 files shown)


ioquatix (approved):
LGTM, agree with deprecations too.

ioquatix:
Thanks!

PR #2041, opened 2023-02-28
Enhance documentation in `builder.rb`

yoones:
> After requiring, assumes App constant contains Rack application
> After requiring, assumes MyApp constant contains Rack application

~Since we assume these constants contain Rack applications and are not ones, the comments should `run` instances of these classes and not the classes themselves.~

EDIT: The word `contains` is replaced by `is a` for more clarity.

### DIFF
--- a/lib/rack/builder.rb
+++ b/lib/rack/builder.rb
@@ -53,12 +53,12 @@ module Rack
     #   Rack::Builder.parse_file('app.rb')
     #   # requires app.rb, which can be anywhere in Ruby's
     #   # load path. After requiring, assumes App constant
-    #   # contains Rack application
+    #   # is a Rack application
     #
     #   Rack::Builder.parse_file('./my_app.rb')
     #   # requires ./my_app.rb, which should be in the
     #   # process's current directory.  After requiring,
-    #   # assumes MyApp constant contains Rack application
+    #   # assumes MyApp constant is a Rack application
     def self.parse_file(path)
       if path.end_with?('.ru')
         return self.load_file(path)


jeremyevans:
I don't think the current documentation is incorrect, but I'll give you that `contains` may not be the best wording.  It doesn't mean instances are Rack applications, though, it means that the value of the constant is a Rack application.  After all, there is no requirement that either `App` or `MyApp` is a class (it could be any object responding to `call`). I don't think changing the example code is a good idea. If any change should be made, we can change `constant contains` to `is a`.

yoones:
Thanks for your quick feedback.

This is a personal feeling, it might just apply to myself: I feel it could be beneficial to change the documentation to running instances because it would avoid any confusion with the use of middleware.

When I read this part of the documentation:

```ruby
    # Takes a lambda or block that is used to warm-up the application. This block is called
    # before the Rack application is returned by to_app.
    #
    #   warmup do |app|
    #     client = Rack::MockRequest.new(app)
    #     client.get('/')
    #   end
    #
    #   use SomeMiddleware
    #   run MyApp
```

I felt a bit confused and had to test different things to make sure of the behavior of `run`.

```ruby
require_relative 'middleware/ping_pong'
require_relative 'middleware/useless_signature'
require_relative 'application/dummy_rack_application'

app = Rack::Builder.app do
  use PingPongMiddleware
  use UselessSignatureMiddleware

  # Should I run:
  run DummyRackApplication

  # or:
  run DummyRackApplication.new
end

run app
```

Let me know what you think, I can rewrite the diff if you still feel replacing `constant contains` to `is a` would be clearer.

jeremyevans:
All Rack applications I run are either classes or procs.  It's quite common for Rack applications to be classes, so that you have an instance of the class per request (i.e. `call` calls `new`).   In your example, whether you should use `run DummyRackApplication` or `run DummyRackApplication.new` depends entirely on what `DummyRackApplication` is.

I still feel like that replacing `constant contains` to `is a` is a better change.

jeremyevans (approved):

PR #2042, opened 2023-03-03
Avoid rebuilding regex per-request part

jhawthorn:
This was spotted while working on CVE-2023-27530 (https://github.com/rack/rack/commit/8e8869d625e73e16b576b6d31b50208e9ec8002f), though itself not a security issue (we chose not to include it in the patch release).

Previously we would rebuild this regex once per-part. We can instead compile it once per-request, which can save significant processing time on requests with many parts.


ioquatix (approved):

ioquatix:
Thanks!

PR #2043, opened 2023-03-03
Default to & as separator instead of either & or ;

mimosa:
Allowing ; as separator by default can lead to web cache poisoning.

Fixes https://github.com/rack/rack/issues/1732


mimosa:
@rafaelfranca I need to patch on rack2 because I use Rails 6 in the production env and need to pass snyk's security check.

jeremyevans:
I don't think we are planning to backport this change to Rack 2.2, due to the backwards compatibility issues. See #1732 and #1733.

mimosa:
Thanks @jeremyevans, I will reconsider the changes this fix brings to rack 2.2 and our project.

PR #2044, opened 2023-03-07
Handle string reuse by body.each when buffering bodies in Rack::Response

### LINKED ISSUE #1957: Mock::Response returns empty body when body parts use same buffer object
It is reproduced with rack 2.2 and 3.0.beta1

Reproducing script
```ruby
require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  git_source(:github) { |repo| "https://github.com/#{repo}.git" }
  gem "rack", "3.0.beta1"
  # gem "rack", "2.2.4"
end

class FileIteratorWithBuffer
  def initialize(file)
    @file = file
  end

  def each
    buffer = String.new

    while @file.read(5, buffer)
      yield(buffer)
    end
  end
end

file = StringIO.new('Large large file content')
body = FileIteratorWithBuffer.new(file)

puts "Result: " + Rack::MockResponse.new(200, {}, body).body.inspect

# => Result: ""
```

jeremyevans:
An alternative approach would be using a single string inside an array and appending to that.  This approach is more backwards compatible, but results in more memory usage.

Fixes #1957


tenderlove:
I guess if calling `write` on a response actually wrote the string to a socket or output buffer, this wouldn't be as much of a problem. 😅

This seems like a fine fix, but maybe this should just be a "known bug".  Adding the `dup` punishes apps that didn't reuse strings in the caller.

In a case like #1957, should they not be directly assigning the "body" generator to the response object?  I'm not sure why you'd write a "body generator" type class like this and feed the output to an instance of a `Response`.

Judging from #1957, this was in a test, so presumably the actual usecase is trying to buffer the response with a `MockResponse` so that the tests can assert the response body.  Maybe the `MockResponse` class should be changed rather than `Response`?

tenderlove:
Sorry, I should have read the entire issue before comment.  Looks like you already modified MockResponse at #1958.  Regardless, I'm not sure this is a use case we should support.  OP can assign the body directly to the response object, or use a MockResponse for tests.

jeremyevans:
@tenderlove MockResponse is a subclass of Response, so it should be simple to override MockResponse#append,  move the dup call to that, and call super.  I can switch to that approach if you want. Note that this could affect production applications using Response, but it seems a less likely use case, because you have to have a body that is reusing strings (assuming that no buffering is happening), with a call to Response#write (which forces buffering).  This always affects MockResponse, as MockResponse always forces buffering.

tenderlove:
> Note that this could affect production applications

"this" meaning _this_ PR, or chanting `MockResponse`?  Changing `MockResponse` shouldn't impact apps in production, just the test suite right?

I'm not keen on adding the `dup` in `Response`, but it's probably not the end of the world.  Feels like we'd be adding overhead to a common use (passing a not-going-to-be-mutated string to a method) in order to support a rare use.

I *think* what you're saying is "we may as well dup since it's going to be buffered anyway" which seems like a valid point.

ioquatix:
I personally think we should be clear about which model we follow:

1. `yield(chunk)` does not confer ownership, i.e. `chunk` can be reused after yield returns.
2. `yield(chunk)` confers ownership and chunk cannot be reused.

(1) is probably preferable as it's more efficient if `yield(chunk) -> `socket.write(chunk)`. If code wants to buffer the chunks it should call `dup`. It pays the cost of buffering. Internally, some CoW mechanisms might minimise the overhead.

ioquatix:
I'm okay with this PR.

If users don't want dup, they can freeze the string. If they want to mutate it, the implementation should internally dup.

If users want more control over buffering, they should not use `Rack::Response`.

Also, internally, `String#dup` should use CoW IIRC, so the memory overhead should be minimal.

That being said, we could also simply document this method as "taking ownership of the string" which means we don't need to call dup at all.

ioquatix:
```ruby
      def buffered_body!
        if @buffered.nil?
          if @body.is_a?(Array)
            # The user supplied body was an array:
            @body = @body.compact
            @body.each do |part|
              @length += part.to_s.bytesize
            end
          elsif @body.respond_to?(:each)
            # Turn the user supplied body into a buffered array:
            body = @body
            @body = Array.new

            body.each do |part|
              @writer.call(part.to_s)
            end

            body.close if body.respond_to?(:close)

            @buffered = true
          else
            @buffered = false
          end
        end

        return @buffered
      end
```

It might also be an option to modify this code, such that `@writer.call(part.to_s.dup)` is used... eh, I'm not sure about that either.

ioquatix:
There is one other, more extreme option: `Rack::Response` should not be used for generating responses.

i.e. it should not have mutable methods like `append`.

If we want to expose such a "response buffer", we should do so explicitly, e.g. `Rack::Response::Builder` or something.

Because the interface for a response, and the "interim state" of a response, are two different things, and we are considering them to be one thing. Because of that, buffering is kind of weird.

jeremyevans:
> There is one other, more extreme option: `Rack::Response` should not be used for generating responses.
> 
> i.e. it should not have mutable methods like `append`.

The main problem there is `Rack::Response` is already used for generating responses.  We would have to break backwards compatibility for this.

We have to decide what tradeoff to make here:

1. Do nothing, applications should not reuse strings.  This would require anything using `Rack::Response` with buffering or using `Rack::MockResponse` to not reuse strings yielded to `body.each`, which hurts efficiency.
2. Change `Rack::Response#append` to `dup` the string if not frozen.  This fixes the issue and hurts performance for all users, even though few `Response` users would be affected by the issue, even if they are reusing strings (they must reuse strings and also write to the body after initialization to be effected).
3. Change `Rack::MockResponse#append` to `dup` the string if not frozen.  This fixes the issue and hurts performance only for `MockResponse`, but allows reuse of strings generally for more efficient applications. `MockResponse` is generally only used in testing, where performance is less important, and all usage of `MockResponse` is affected by this issue if reusing strings, since `MockResponse` always buffers the body (unlike `Response`).  Users using `Response` with buffering would still have problems if they reused strings.

This PR currently implements option 2, but I'm fine switching it to option 3.

ioquatix (approved):
Any time we retain a string, that is not frozen, we should assume it could change. As we are acting as a buffer, the user code could change that string but with the expectation that it was "written to a socket", such that mutating the string wouldn't matter. This PR introduces a string dup in `append`, which essentially enforces the contract "we are retaining this string in a buffer". My only contention is that there are other parts of the code where this happens.

PR #2046, opened 2023-03-08
Store downcased common headers at class level

amatsuda:
Since all header-downcasing are integrated in headers.rb in Rack 3, and almost all HTTP header keys are basically predictable, we can now make Rack 3 almost zero-allocation with regard to header key conversion per each request.

Here's a patch that stores pre-downcased common set of header key names in Rack::Headers class. The key names are taken from "Standard response fields" and "Common non-standard response fields" sections in "List of HTTP header fields" page on Wikipedia https://en.wikipedia.org/wiki/List_of_HTTP_header_fields
plus, some keys that we see in Rails and Sinatra.

With this patch (together with #2045), I confirmed that I'm seeing no new string allocation on basic Rails app and Sinatra app.

Of course there can be possibilities of receiving some unknown keys, and in that case, Rack::Headers would just fall back to the current behavior.

### DIFF
--- a/lib/rack/headers.rb
+++ b/lib/rack/headers.rb
@@ -6,6 +6,89 @@ module Rack
   # (by using non-lowercase response header keys), automatically handling
   # the downcasing of keys.
   class Headers < Hash
+    KNOWN_HEADERS = {}
+    %w(
+      Accept-CH
+      Accept-Patch
+      Accept-Ranges
+      Access-Control-Allow-Credentials
+      Access-Control-Allow-Headers
+      Access-Control-Allow-Methods
+      Access-Control-Allow-Origin
+      Access-Control-Expose-Headers
+      Access-Control-Max-Age
+      Age
+      Allow
+      Alt-Svc
+      Cache-Control
+      Connection
+      Content-Disposition
+      Content-Encoding
+      Content-Language
+      Content-Length
+      Content-Location
+      Content-MD5
+      Content-Range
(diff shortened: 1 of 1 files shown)


lib/rack/headers.rb
```diff
@@ -4,6 +4,22 @@ module Rack
   # (by using non-lowercase response header keys), automatically handling
   # the downcasing of keys.
   class Headers < Hash
+    KNOWN_HEADERS = {}
+    %w(
+      Accept-CH Access-Control-Allow-Origin Access-Control-Allow-Credentials Access-Control-Expose-Headers Access-Control-Max-Age Access-Control-Allow-Methods Access-Control-Allow-Headers
```
jeremyevans, line 9: Could we please sort these alphabetically and have one header per line?

jeremyevans (dismissed):
Thanks for the patch! I'm fine with the general idea.  Only one requested change for implementation.

lib/rack/headers.rb
amatsuda: Sure! I did so.

ioquatix:
I have another suggestion.

It might be pedantic, so feel free to ignore it.

However, every Rack application is going to, at load time, compute this list, which will add a little latency and computation overhead. But we already know what this mapping should look like.

Maybe we should just hard code the mapping, rather than having the processing step (the last `each`).

ioquatix (approved):

PR #2047, opened 2023-03-09
Fix external tests.

ioquatix:


MSP-Greg:
You could change:
```
    - uses: ruby/setup-ruby@v1
      with:
        ruby-version: ${{matrix.ruby}}
        bundler-cache: true

    - name: Installing packages (ubuntu)
      if: matrix.os == 'ubuntu-latest'
      run: |
        sudo apt-get update
        sudo apt-get install libfcgi-dev libmemcached-dev
    - name: Installing packages (macos)
      if: matrix.os == 'macos-latest'
      run: brew install fcgi libmemcached
```

to:
```
    - uses: ruby/setup-ruby-pkgs@v1
      with:
        ruby-version: ${{matrix.ruby}}
        bundler-cache: true
        apt-get: _update_ libfcgi-dev libmemcached-dev
        brew: fcgi libmemcached
```

Note that `ruby/setup-ruby-pkgs` installs the packages **before** running bundle.

MSP-Greg:
> it looks like _update_ should always be done. I can't imagine any situation where it isn't a good idea.

I'd agree, but some people always complain about anything that may take time.   It also works with Windows (mingw & mswin).

See PR #2048

ioquatix:
If people want that, why not have a separate key, `update: false` or something?

PR #2052, opened 2023-03-13
Add `QueryParser#missing_value` for handling missing values + tests.

ioquatix:
This is required for customizing how `Rack::QueryParser` handles missing values and make it easy for sub-classes to change this behaviour. Based on <https://github.com/rack/rack/pull/2038#issuecomment-1436060688>.

For the purpose of Rails:

```ruby
module ActionDispatch
  class Request
    # ...
    if Rack.release >= "3"
      class QueryParser < ::Rack::QueryParser
        def missing_value
          nil
        end
      end

      def query_parser
        QueryParser
      end
    end
```

A different design option would be to make this an argument, e.g. `QueryParser.new(missing_value: nil)`.

cc @matthewd @tenderlove

### DIFF
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -1,6 +1,7 @@
 # frozen_string_literal: true
 
 require_relative 'bad_request'
+require 'uri'
 
 module Rack
   class QueryParser
@@ -111,6 +112,14 @@ module Rack
       _normalize_params(params, name, v, 0)
     end
 
+    # This value is used by default when a parameter is missing (nil). This
+    # usually happens when a parameter is specified without an `=value` part.
+    # The default value is an empty string, but this can be overridden by
+    # subclasses.
+    def missing_value
+      String.new
+    end
+
     private def _normalize_params(params, name, v, depth)
       raise ParamsTooDeepError if depth >= param_depth_limit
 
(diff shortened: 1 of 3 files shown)


lib/rack/query_parser.rb
```diff
@@ -207,8 +216,8 @@ def params_hash_has_key?(hash, key)
       true
     end
 
-    def unescape(s)
-      Utils.unescape(s)
+    def unescape(string, encoding = Encoding::UTF_8)
+      URI.decode_www_form_component(string, encoding)
```
jeremyevans, line 220: This seems unrelated to the change being made.  I'm not necessary opposed to this, but it shouldn't be part of this PR.  It can be a separate PR with a reason given for the change.  Ditto for the `uri` require at the top.

jeremyevans (changes requested):
I'm fine with this in principle. One inline comment.

lib/rack/query_parser.rb
ioquatix: I couldn't make the separate tests pass without it.

jeremyevans: Fix for this is simple, submitted in #2055.  However, considering this was backported without any discussion, I'm not sure what we want to do here.

PR #2055, opened 2023-03-13
Revert the unapproved QueryParser#unescape change

jeremyevans:
This change was only made because @ioquatix couldn't figure out how to make the tests pass without the change.  This reverts the code and fixes the tests.


ioquatix:
I don't think we should do this, it looks objectively worse to me to use conditional requires. It's fairly easy for this to break. That doesn't mean I don't think what I did was the best or that we shouldn't do something better. But I don't think query parser should depend on utils.

PR #2056, opened 2023-03-15
Backport Rack::Request cache helpers.

ioquatix:
I would characterise this as a bug fix, as without it, the cache is not computed correctly.

https://github.com/rack/rack/pull/2054


lib/rack/request.rb
```diff
@@ -480,25 +480,114 @@ def parseable_data?
...
+        raise
+      end
+
+      # This cache key is used by cached values generated by class_cache_for,
+      # specfically GET and POST. This is to ensure that the cache is not
+      # shared between instances of different classes which have different
+      # behaviour. This includes sub-classes that override query_parser or
+      # expand_params.
+      def cache_key
+        query_parser.class
```
jeremyevans, line 526: I still think this is a mistake, and think the onus should be on the framework to override this appropriately.  If you have to base it on the query parser, `query_parser` would be better than `query_parser.class`, considering that the default is to use a singleton `QueryParser` instance (`Utils.default_query_parser`), and singleton classes allow for the changing of behavior for individual instances.  It would be best to fix that first, and include that in the backport.

jeremyevans (changes requested):
I would consider this a new feature, not a bug fix, and would usually be against backporting.  However, considering what has already been backported was much more invasive/disruptive, I guess there isn't much additional harm in backporting this.  So I'm OK with the idea of backporting this.  See inline comment regarding cache_key (which was changed after I approved the related PR), which I think should be fixed before backporting.

ioquatix:
We decided to give up on this, and change the query parser behaviour back to 2.x where nil values can be exposed.

PR #2057, opened 2023-03-15
Add mjs MIME type

axilleas:
`mjs` is a JavaScript module and has the same MIME type as JavaScript. See https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types.


ioquatix (approved):
LGTM.

axilleas:
@ioquatix thanks for merging! I see this is not included in the new 3.0.7 patch release, was this intended? If yes, do you know when a new release will be out with this change? Thanks!

PS. Some backstory, we're [trying to use](https://gitlab.com/gitlab-org/gitlab-docs/-/merge_requests/3664) mermaid.js with Nanoc's local live webserver ([adsf](https://github.com/denisdefreyne/adsf)) which relies on Rack, and `.mjs` mimetypes are not supported, so the diagrams are now broken.

ioquatix:
I don't have a strong opinion about back porting this or not. We have made a bit of a mess of the 3.0.x release series w.r.t. to back porting stuff. However, I think we'd like to avoid back porting unless strictly needed for security or major bugs.

Regarding this feature, I'd say the earliest we will see it is Rack 3.1. However, that doesn't prevent you from manually adding it in your app. i.e.

```
Rack::Mime::MIME_TYPES['.mjs'] = ... etc
```

axilleas:
@ioquatix that's sounds reasonable, thanks for the explanation 🙂

PR #2059, opened 2023-03-16
Make query parameters without = have nil values

jeremyevans:
This was Rack's historical behavior.  While it doesn't match
URL spec section 5.1.3.3, keeping the historical behavior avoids
all of the complexity required to support the URL spec standard
by default, but also support frameworks that want to be backwards
compatible.

This reverts the four recent commits that added a lot of additional
complexity just to get the backwards compatible behavior. While
I think the URL spec behavior is better, I think it's better to have
the backwards compatible behavior without the complexity than
the URL spec behavior with all of the complexity required to also
allow frameworks to get the backwards compatible behavior.

This keeps as much of the specs added by the recently reverted
commits that make sense.


ioquatix (approved):
LGTM.

ioquatix:
I think what we've identified here is that the way `Rack::Request` works, has a lot of shortcomings, and trying to make it more standards compliant revealed that we had a ton of assumptions which don't hold true (order dependent) when sub-classes change the behaviour. The level of complexity required to make this work was insane and we've simply designed to roll it all back. Given that the previous behaviour was a superset of Rack 3, that isn't a compatibility issue (although internal implementation and some public methods are changed as part of the roll back).

tenderlove:
> I think the URL spec behavior is better, I think it's better to have
the backwards compatible behavior without the complexity than
the URL spec behavior with all of the complexity required to also
allow frameworks to get the backwards compatible behavior.

💯💯💯

PR #2065, opened 2023-03-27
Update MIME types associated to font extensions `.ttf`, `.woff`, `.woff2`, and `.otf`

davidstosik:
## What?

This PR updates the MIME types associated to font file extensions `.ttf`, `.woff`, `.woff2`, and `.otf` to the more recent types prefixed with `font/`.

## Why?

Working on deploying a simple Rails application to Fly.io, I noticed that even though it was serving most of my assets gzipped (`content-encoding: gzip`), the TTF font that it serves was desperately uncompressed (and heavy!).
It turns out that, as a precompiled asset, it gets served with the header `content-type: application/octet-stream`, and I believe this is due to Rack's MIME types.
I believe this content-type might be excluded from getting gzipped by Fly.io's reverse proxy (and maybe other hosting solutions too?).

Looking into it, I noticed that `.woff` and `.woff2` were using deprecated MIME types and decided to update them too.

## How?

Simply updated the values in Rack's hash that associates file extensions to MIME types.

## Anything Else?

Resources on the IANA website:

- [`font/woff`](https://www.iana.org/assignments/media-types/font/woff):
  > File extension(s):  woff
  > (...)
  >  Deprecated Alias:  The existing registration "application/font-woff" is deprecated in favor of "font/woff".
- [`font/woff2`](https://www.iana.org/assignments/media-types/font/woff2):
  > File extension(s):  woff2
- [`font/ttf`](https://www.iana.org/assignments/media-types/font/ttf):
  > File extension(s):  Font file extensions used for TrueType / OFF / OpenType fonts: .ttf and .otf

Note that it might be worth revisiting the `.otf` too, but I was unable to find clear information about this extension. ([`font/otf`](https://www.iana.org/assignments/media-types/font/otf) exists, but the `.otf` extension is also associated with [`application/vnd.oasis.opendocument.formula-template`](https://www.iana.org/assignments/media-types/application/vnd.oasis.opendocument.formula-template).)

https://github.com/rack/rack/blob/b172805895fc0490bf81e942487233e6771f4643/lib/rack/mime.rb#L413

For reference, I added this to my Rails application's `config/initializers/assets.rb`, as a temporary workaround:

```rb
Rack::Mime::MIME_TYPES[".ttf"] = "font/ttf"
```


jeremyevans (approved):

ioquatix:
Regarding `otf`, what does the Ruby mime-types gem say about it? I find that is a good source of truth.

davidstosik:
@ioquatix [`.otf` seems to be associated to the font type](https://github.com/mime-types/mime-types-data/blob/7109b079da2c87924066c5c3e7eb9b65ffcec9e8/types/font.yaml#L13-L23):
```yaml
- !ruby/object:MIME::Type
  content-type: font/otf
  encoding: base64
  extensions:
  - otf
  xrefs:
    rfc:
    - rfc8081
    template:
    - font/otf
  registered: true
```

simi:
To make (web)fonts support complete, what about `eot`? Not sure what's the adoption today, but it was popular earlier due to IE.

ioquatix:
I would be fine for `.otf` -> `font/otf`. That seems reasonable to me.

I don't have any opinion about `eot`. I suppose in order to keep the list "short", we try to only include relevant, modern, mime types. So, I guess on that bases, if `eot` is no longer in common usage, I'd suggest we don't add it.

ioquatix:
@davidstosik will you update the PR and add `.otf`?

davidstosik:
Sure!

`.eot` is in there already, mapped to [`application/vnd.ms-fontobject`](https://www.iana.org/assignments/media-types/application/vnd.ms-fontobject). `font/eot` does not seem to be a thing, so there's nothing to change for this one.

ioquatix (approved):

ioquatix:
@davidstosik Do you mind adding an entry to the changelog, thanks!

davidstosik:
Done! Am I supposed to add my GitHub handle, or is that reserved to maintainers?

ioquatix:
You can add it.

Issue #1968, opened 2022-09-22
Errno::ENAMETOOLONG in multipart parser.


ioquatix:
This issue cropped up again in our production environment.

I think you are right, the file name on disk should not be part of the "content type" or any other user supplied data = security risk. Let's have a think about how to deal with this. I like your proposal.

_Originally posted by @ioquatix in https://github.com/rack/rack/issues/1874#issuecomment-1254436186_

dentarg:
Okay, so I too had to rescue `ENAMETOOLONG` in my apps today.

Posting a simple repro to provide some more context for anyone new to this issue.

```ruby
# config.ru
run ->(env) do
  [200, {}, ["params=#{Rack::Request.new(env).params}"]]
end
```
```shell
# This triggers the crash
curl --form 'file=foo;filename=a.'\
'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'\
'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'\
'aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa'\
'aaaaaaaaaaaaaaaaaaaaaaaa' \
http://127.0.0.1:5000
```

dentarg:
@simi what was your alternative ideas? https://github.com/rack/rack/pull/1874#issuecomment-1110322846

dentarg:
Make Rack return `BadRequest` on anything that causes `ENAMETOOLONG`? (Can't be backported to v2 though, as `BadRequest` doesn't exist there)

dentarg:
> @simi what was your alternative ideas? [#1874 (comment)](https://github.com/rack/rack/pull/1874#issuecomment-1110322846)

Perhaps it was

> If we would like to fix this still in Rack, I would go for random tempfile names and keeping original name around to be able to do checks (mime type?) on that instead of tempfile name.

jeremyevans:
> I think you are right, the file name on disk should not be part of the "content type" or any other user supplied data = security risk. Let's have a think about how to deal with this. I like your proposal.

[...]

PR #2069, opened 2023-04-18
Limit file extension length of multipart tempfiles

dentarg:
Close https://github.com/rack/rack/issues/1968

### DIFF
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -59,7 +59,9 @@ module Rack
       BUFSIZE = 1_048_576
       TEXT_PLAIN = "text/plain"
       TEMPFILE_FACTORY = lambda { |filename, content_type|
-        Tempfile.new(["RackMultipart", ::File.extname(filename.gsub("\0", '%00'))])
+        extension = ::File.extname(filename.gsub("\0", '%00'))[0, 129]
+
+        Tempfile.new(["RackMultipart", extension])
       }
 
       class BoundedIO # :nodoc:
--- a/test/spec_multipart.rb
+++ b/test/spec_multipart.rb
@@ -780,6 +780,28 @@ contents\r
     params["file"][:filename].must_equal 'long' * 100
   end
 
+  it "limits very long file name extensions in multipart tempfiles" do
+    data = <<-EOF
+--AaB03x\r
+content-type: text/plain\r
+content-disposition: attachment; name=file; filename=foo.#{'a' * 1000}\r
+\r
+contents\r
+--AaB03x--\r
+    EOF
+
+    options = {
+      "CONTENT_TYPE" => "multipart/form-data; boundary=AaB03x",
+      "CONTENT_LENGTH" => data.length.to_s,
+      :input => StringIO.new(data)
+    }
+    env = Rack::MockRequest.env_for("/", options)
+    params = Rack::Multipart.parse_multipart(env)
+
+    params["file"][:filename].must_equal "foo.#{'a' * 1000}"
+    File.extname(env["rack.tempfiles"][0]).must_equal ".#{'a' * 128}"
+  end
+
   it "parse unquoted parameter values at end of line" do
     data = <<-EOF
 --AaB03x\r


ioquatix (approved):

lib/rack/multipart/parser.rb
```diff
@@ -59,7 +59,9 @@ class Parser
       BUFSIZE = 1_048_576
       TEXT_PLAIN = "text/plain"
       TEMPFILE_FACTORY = lambda { |filename, content_type|
-        Tempfile.new(["RackMultipart", ::File.extname(filename.gsub("\0", '%00'))])
+        extension = ::File.extname(filename.gsub("\0", '%00'))[0..128]
```
ioquatix, line 62: This behaviour is kind of complex, and I think we should test it in isolation.

Maybe introduce a method like `tempfile_name` which takes the filename and content type and returns a new file name?

I wonder if we should filter out anything except for ASCII characters from the extension (i.e. be more defensive).

jeremyevans: Looks like we don't use the filename, only the extension, and I cannot think of a reasonable extension that uses non-ASCII characters. However, that gsub is probably a bad idea, as we don't want to use `%00` for a replacement and we need to handle `.`.  Also, because this is only for extensions, I don't think we need to make the maximum length configurable.

For the limit, `[0, 129]` is faster than `[0..128]`.  And we can use `gsub!` and not `gsub` to save a string allocation.  Maybe something like:

```rb
ext = ::File.extname(filename)
ext.gsub!(/[^.a-zA-Z0-9]/, '')
ext = ext[0, 129]
```

This could result in ext being `.` in pathological cases, but `Dir::Tmpname`  handles that fine.

Regarding `# Does this need to be an array?`, yes, because that's what `Dir::Tmpname` handles for a prefix and suffix.

In terms of `self.tempfile_name`, if we are going to make this a method, I'd rather the method return the `Tempfile`, and we do: `TEMPFILE_FACTORY = method(:tempfile_factory)` instead of a separate lambda.  I'm OK with adding both `tempfile_name` and `tempfile_factory`, and having `tempfile_factory` call `tempfile_name` to get the name to use if you want to test it separately.

dentarg: I don't like the idea of adding methods for this, I think it makes the solution more complex than it needs to be. To do `TEMPFILE_FACTORY = method(:tempfile_factory)` the class methods needs to be defined before `TEMPFILE_FACTORY` is defined. IMHO makes the Ruby code looks a bit weird. Also, do we want these methods part of the interface?

Without `gsub("\0", '%00')`, with this code

> ```ruby
> ext = ::File.extname(filename)
> ext.gsub!(/[^.a-zA-Z0-9]/, '')
> ext = ext[0, 129]
> ```

this spec failed

```
  1) Error:
Rack::Multipart#test_0033_parse multipart form with a null byte in the filename:
ArgumentError: path name contains null byte
    /Users/dentarg/src/rack/lib/rack/multipart/parser.rb:73:in `extname'
    /Users/dentarg/src/rack/lib/rack/multipart/parser.rb:73:in `tempfile_name'
    /Users/dentarg/src/rack/lib/rack/multipart/parser.rb:69:in `tempfile_factory'
    /Users/dentarg/src/rack/lib/rack/multipart/parser.rb:189:in `call'
    /Users/dentarg/src/rack/lib/rack/multipart/parser.rb:189:in `on_mime_head'
    /Users/dentarg/src/rack/lib/rack/multipart/parser.rb:346:in `handle_mime_head'
    /Users/dentarg/src/rack/lib/rack/multipart/parser.rb:262:in `block in parse'
    /Users/dentarg/src/rack/lib/rack/multipart/parser.rb:254:in `loop'
    /Users/dentarg/src/rack/lib/rack/multipart/parser.rb:254:in `parse'
    /Users/dentarg/src/rack/lib/rack/multipart/parser.rb:136:in `parse'
    /Users/dentarg/src/rack/lib/rack/multipart.rb:37:in `parse_multipart'
    test/spec_multipart.rb:389:in `block (2 levels) in <main>'
```

ioquatix: I'm fine with not extracting methods provided we can test it easily.

ioquatix: I'm basically okay with it, I already approved this PR. But I did wonder if we should be a bit more strict about extensions allowed, e.g. we allow non-printable characters as extensions?

dentarg: Great. I think we should try to do one thing at once. Seems like non-printable characters hasn't been a problem so far? It could be a separate contribution in the future.

ioquatix: If @jeremyevans is fine with this PR, I am too.

I'm not really sure under what circumstances non-printable characters could be an issue but it seems like a super weird thing to try and write to the file-system.

simi:
This makes me wonder why https://github.com/rack/rack/pull/1874/ wasn't accepted.

My next plan was to move handling files to "user land" to make this exception (or even custom one) easy to handle in rails base controller for example.

ioquatix:
@simi sorry about past PRs where the outcome might have been different. Everyone's contributions are valued and we do our best as maintainers to ensure that we keep moving forward. If there are additions or improvements in #1874 that you think we should apply, please consider opening a new PR and we appreciate the time you invest in such endevours.

PR #2071, opened 2023-04-23
Support underscore in host names for Rack 2.2 (Fixes #2070)

### LINKED ISSUE #2070: HTTP_HOST not parsing port correctly at 2.X
Hello, I think we might have a regression from this PR https://github.com/rack/rack/pull/1606.

With Rack 2.1.4.3 (Rack 3.X also works as expected)

```ruby
irb(main):001:0> r = Rack::Request.new({"HTTP_HOST" => "some_service:3001"})
irb(main):002:0> r.host
=> "some_service"
irb(main):003:0> r.port
=> 3001
```

With Rack 2.2.6.4 (But any version at 2.2.X will behave the same way)
```ruby
irb(main):001:0> r = Rack::Request.new({ "HTTP_HOST" => "some_service:3001" })
irb(main):002:0> r.host
=> "http://some_service:3001"
irb(main):003:0> r.port
=> nil
```

Am I missing something here?

Ruby: 2.7.8
Rails 7.0.4

jeremyevans:
This makes Rack 2.2 behavior similar to Rack 2.1 and Rack 3.0 in regards to underscore in host names.

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -608,7 +608,7 @@ module Rack
           (?<ip4>[\d\.]+)
           |
           # A hostname:
-          (?<name>[a-zA-Z0-9\.\-]+)
+          (?<name>[a-zA-Z0-9\.\-_]+)
         )
         # The optional port:
         (:(?<port>\d+))?
--- a/test/spec_request.rb
+++ b/test/spec_request.rb
@@ -122,6 +122,11 @@ class RackRequestTest < Minitest::Spec
     req.hostname.must_equal "123foo.example.com"
 
     req = make_request \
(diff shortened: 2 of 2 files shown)


ioquatix (approved):
This seems okay to me for consistency.

The current Rack 3 implementation for reference: https://github.com/rack/rack/blob/f5666bc8cb13b8d731ea0222fbd3ada670f2cd55/lib/rack/request.rb#L703-L714

ioquatix:
@jeremyevans do we need to add tests to Rack main?

jeremyevans:
I think it would be a good idea.

ioquatix:
If you have time, can you make a follow-up PR.

PR #2077, opened 2023-04-28
Add Content-Disposition parameter parser

### LINKED ISSUE #2076: Parsing uploaded file parameter name from Content-Disposition header fails 2.2.6.3 update
There's a fix for ReDos vulnerability in
https://github.com/rack/rack/commit/ee25ab9a7ee981d7578f559701085b0cf39bde77

which breaks the Content-Disposition parameter name parsing in case header contains extra attributes like modification-date.

I.e. this works, parameter name is considered "file"
Content-Disposition: form-data; name="file"; filename="sample.sql"
 
However, in this case it fails to parse parameter name and file name itself is used as parameter name:
Content-Disposition: form-data; filename="sample.sql"; modification-date="Wed, 26 Apr 2023 11:01:34 GMT"; size=24; name="file"

jeremyevans:
The ReDoS fix in ee25ab9a7ee981d7578f559701085b0cf39bde77 breaks valid requests, because colons are valid inside parameter values.  You cannot use a regexp scan and ensure correct behavior, since values inside parameters can be escaped.  Issues like this are the reason for the famous "now they have two problems" quote regarding regexps.

Add a basic parser for parameters in Content-Disposition.  This parser is based purely on String#{index,slice!,[],==}, usually with string arguments for #index (though one case uses a simple regexp).  There are two loops (one nested in the other), but the use of slice! ensures that forward progress is always made on each loop iteration.

In addition to fixing the bug introduced by the security fix, this removes multiple separate passes over the mime head, one pass to get the parameter name for Content-Disposition, and a separate pass to get the filename. It removes the get_filename method, though some of the code is kept in a smaller normalize_filename method.

This removes 18 separate regexp contents that were previously used just for the separate parse to find the filename for the content disposition.

This is obviously a major change and definitely needs thorough review and ideally testing with production traffic to ensure it handles cases as expected and doesn't break things.

Currently, the parser isn't as strict as it could be.  We could have it:

* fail for empty parameter names (instead of skipping them)
* fail for quoted parameter values not ending with a semicolon (currently semicolon after end of quoted parameter value is optional)
* fail for duplicate parameter names (currently, last one wins)

Currently, for `filename` parameters, the handling of escaped characters inside quoted parameter values is to include the backslash in the value, as that is necessary to keep tests passing for old IE which submits full paths in `filename`.  We may want to consider dropping that support, and always removing escape characters.

Certainly open to changes regarding how the parser works, but I am convinced we should move to a parsing approach and not keep using a regexp approach for this.

This seems risky to backport to 2.2 and 3.0, but as the security fix was backported, I think we'll have to unless a fix for the regexp approach can be developed.

Fixes #2076

### DIFF
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -32,28 +32,9 @@ module Rack
 
     EOL = "\r\n"
     MULTIPART = %r|\Amultipart/.*boundary=\"?([^\";,]+)\"?|ni
-    TOKEN = /[^\s()<>,;:\\"\/\[\]?=]+/
-    CONDISP = /Content-Disposition:\s*#{TOKEN}\s*/i
-    VALUE = /"(?:\\"|[^"])*"|#{TOKEN}/
-    BROKEN = /^#{CONDISP}.*;\s*filename=(#{VALUE})/i
     MULTIPART_CONTENT_TYPE = /Content-Type: (.*)#{EOL}/ni
-    MULTIPART_CONTENT_DISPOSITION = /Content-Disposition:[^:]*;\s*name=(#{VALUE})/ni
+    MULTIPART_CONTENT_DISPOSITION = /Content-Disposition:(.*)(?=#{EOL}(\S|\z))/ni
     MULTIPART_CONTENT_ID = /Content-ID:\s*([^#{EOL}]*)/ni
-    # Updated definitions from RFC 2231
-    ATTRIBUTE_CHAR = %r{[^ \x00-\x1f\x7f)(><@,;:\\"/\[\]?='*%]}
-    ATTRIBUTE = /#{ATTRIBUTE_CHAR}+/
-    SECTION = /\*[0-9]+/
-    REGULAR_PARAMETER_NAME = /#{ATTRIBUTE}#{SECTION}?/
-    REGULAR_PARAMETER = /(#{REGULAR_PARAMETER_NAME})=(#{VALUE})/
-    EXTENDED_OTHER_NAME = /#{ATTRIBUTE}\*[1-9][0-9]*\*/
-    EXTENDED_OTHER_VALUE = /%[0-9a-fA-F][0-9a-fA-F]|#{ATTRIBUTE_CHAR}/
-    EXTENDED_OTHER_PARAMETER = /(#{EXTENDED_OTHER_NAME})=(#{EXTENDED_OTHER_VALUE}*)/
-    EXTENDED_INITIAL_NAME = /#{ATTRIBUTE}(?:\*0)?\*/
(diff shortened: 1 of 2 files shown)


tenderlove:
+1000 for changing this to a non regexp based parser.  Maybe we could allow people to opt out of the ReDoS fix on the maintenance branches?  I'm hesitant to backport this, but you're right that we need to fix it.

ioquatix:
I basically agree with your position, high risk, high reward, nice to clean up.

I personally would like to see us adopt formally verified parsers e.g. using Ragel.

lib/rack/multipart/parser.rb
```diff
@@ -316,13 +297,89 @@ def handle_mime_head
...
+            end
           else
             name = head[MULTIPART_CONTENT_ID, 1]
           end
 
-          filename = get_filename(head)
+          if filename_star
+            encoding, _, filename = filename_star.split("'", 3)
+            filename = normalize_filename(filename || '')
+            filename.force_encoding(::Encoding.find(encoding))
```
tenderlove, line 378: Do we need to handle the case of someone sending an encoding that doesn't exist?

tenderlove (approved):
It looks good to me.  I left a comment wrt filename encoding, but this patch seems to have the same behavior as before so no need to address the comment here.

lib/rack/multipart/parser.rb
jeremyevans: I think it currently raises an ArgumentError in that case.  I notice in `tag_multipart_encoding`, we rescue and use binary if it fails.  I'm open to doing that here if you want to make that change.

ioquatix:
I haven't spent a lot of time thinking about this PR so here are some quick thoughts.

- I'm basically fine to merge it.
- I think we should adopt a more robust set of parsers in the future using formal verification.
- I think we should consider moving this kind of behaviour into a separate gem so it can be versioned independently of other functionality.

ioquatix (approved):

lib/rack/multipart/parser.rb
tenderlove: Ya, why don't we do that here.  It seems better to do that kind of thing very close to where we expect to see the exception (then we're not accidentally rescuing other `ArgumentError` exceptions).

tenderlove:
> So by default, you get the regexp approach with the ReDoS fix in 2.2/3.0, but you can switch to this approach via a flag.

This is acceptable to me.

> I think we should adopt a more robust set of parsers in the future using formal verification.

Lets discuss this some other time.  I'm not sold on using any parser generators.  IME hand rolled recursive descent turn out to be faster (certainly we can JIT them better) and have better error handling support.

> I think we should consider moving this kind of behaviour into a separate gem so it can be versioned independently of other functionality.

This maybe makes sense (I'm on the fence), but again I think it's outside the scope of this PR.

PR #2079, opened 2023-04-29
Limit max size and number of parameters parsed for Content-Disposition

jeremyevans:
Not strictly necessary, but this limits the damage in pathological cases.  These limits are probably already too generous, we could probably get by with 8 params and 1024 bytes.  One of tests uses more than 1024 bytes, though.  Still, it seems unlikely any legitimate requests would exceed these limits.  We could make the limits configurable via an accessor method, if desired.

### DIFF
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -293,18 +293,27 @@ module Rack
         end
       end
 
+      CONTENT_DISPOSITION_MAX_PARAMS = 16
+      CONTENT_DISPOSITION_MAX_BYTES = 1536
       def handle_mime_head
         if @sbuf.scan_until(@head_regex)
           head = @sbuf[1]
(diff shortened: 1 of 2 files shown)


ioquatix (approved):

lib/rack/multipart/parser.rb
```diff
@@ -293,18 +293,27 @@ def handle_consume_token
...
+
             # ignore actual content-disposition value (should always be form-data)
             i = disposition.index(';')
             disposition.slice!(0, i+1)
             param = nil
+            num_params = 0
 
             # Parse parameter list
             while i = disposition.index('=')
+              # Only parse up to 32 parameters, to avoid potential denial of service
```
natematykiewicz, line 313: This should be 16 not 32, right? @jeremyevans

jeremyevans: Yep, thanks for the heads up.  I'll push a fix.

Issue #2080, opened 2023-05-04
Handle CloudFront-Forwarded-Proto as a lower priority X-Forwarded-sheme option


tomharvey:
Instead of X-Forwarded-scheme headers, cloudfront sends CloudFront-Forwarded-Proto with http or https as the value.  

see: https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/adding-cloudfront-headers.html#cloudfront-headers-other

My current setup is

cloudfront -> ALB -> puma (on ECS)

It’s common to add headers to Nginx or Apache to ensure request origin matches the base. But I have no nginx here.

I could modify headers at the cloudfront distribution by running some JavaScript on each request :(

But, the (non standard) header with the information I require is there. So, I’m considering adding middleware to my rails app in line with this suggestion to handle CloudFlare’s similar non standard header. https://github.com/rails/rails/issues/22965#issuecomment-568655496

Alternatively, as the Cloudfront header is fairly simple (the cloudflare one is Json while this is just text), is there appetite to add the handling of this cloudfront header to Rack?

perhaps after checking forwarded, [then x-forwarded headers](https://github.com/rack/rack/blob/main/lib/rack/request.rb#L44) it could check third party CDN headers? 

If there is appetite I would be happy to contribute (this is permission request, not a labor request)

jeremyevans:
I don't think it is appropriate to do this by default in Rack (doing so would unfairly favor a single vendor, and results in a slippery slope).  However, I would be OK shipping a middleware with Rack or adding a feature to Rack::Request that allowed users to customize which headers to treat as as x-forwarded headers.

tomharvey:
Added a PR for the middleware

jeremyevans:
This was fixed by #2089

PR #2082, opened 2023-06-10
adding middleware to use CloudFront-Forwarded-Proto in CSRF

tomharvey:
In response to https://github.com/rack/rack/issues/2080 is this the kind of thing you would find acceptable?


tomharvey:
Closing for now. Until I do some testing in a live env

jeremyevans:
As mentioned in #2080, this approach unfairly favors a single vendor and would therefore not be acceptable. Please make any future pull request vendor agnostic, so that users can configure it, something like:

```ruby
  use Rack::SetXForwardedProtoHeader, 'CloudFront-Forwarded-Proto'
```

PR #2088, opened 2023-06-22
When parsing a multi-part POST, retain original pairs

matthewd:
The input is not guaranteed to be rewindable, and the default parameter expansion loses details of the parameter names.

For form-encoded data, we (already) retain the full string: it contains only simple key/value pairs, and will be of a manageable size.

Multipart data may contain uploaded files, so we don't wish to retain the full input string -- that's why we dropped the rewindability requirement on the input stream. Instead, we retain an array of two-element arrays representing the "raw" key-value pairs as we parsed the stream. This minimizes additional memory use, because the values are the same objects we use in the params hash, while still allowing an interested consumer to apply their own logic to parameter name interpretation.

---

@jeremyevans @ioquatix I believe this is the minimal version of what we discussed at Kaigi. Apologies for the delay.

For a consumer to use it in this state, they'd need to call `Rack::Request#POST`, relying on its side-effect, then manually pull `rack.request.form_pairs`. Neither of those steps feels like a great API, so I think it would make sense to provide a `body_param_list`-like method as in #2038, but it might be easier to discuss that in a separate follow-up PR, if this seems independently reasonable as the direct implementation / extra retained values?

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -516,7 +516,10 @@ module Rack
             set_header RACK_REQUEST_FORM_INPUT, nil
             set_header(RACK_REQUEST_FORM_HASH, {})
           elsif form_data? || parseable_data?
-            unless set_header(RACK_REQUEST_FORM_HASH, parse_multipart)
+            if pairs = Rack::Multipart.parse_multipart(env, Rack::Multipart::ParamList)
+              set_header RACK_REQUEST_FORM_PAIRS, pairs
+              set_header RACK_REQUEST_FORM_HASH, expand_param_pairs(pairs)
+            else
               form_vars = get_header(RACK_INPUT).read
 
               # Fix for Safari Ajax postings that always append \0
@@ -670,6 +673,16 @@ module Rack
 
       def parse_multipart
         Rack::Multipart.extract_multipart(self, query_parser)
+      end
+
+      def expand_param_pairs(pairs, query_parser = query_parser())
+        params = query_parser.make_params
+
+        pairs.each do |k, v|
+          query_parser.normalize_params(params, k, v)
+        end
+
+        params.to_params_hash
       end
 
       def split_header(value)
--- a/lib/rack/constants.rb
+++ b/lib/rack/constants.rb
@@ -54,6 +54,7 @@ module Rack
   RACK_RESPONSE_FINISHED              = 'rack.response_finished'
   RACK_REQUEST_FORM_INPUT             = 'rack.request.form_input'
   RACK_REQUEST_FORM_HASH              = 'rack.request.form_hash'
+  RACK_REQUEST_FORM_PAIRS             = 'rack.request.form_pairs'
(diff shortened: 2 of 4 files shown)


lib/rack/request.rb
```diff
@@ -516,7 +516,10 @@ def POST
             set_header RACK_REQUEST_FORM_INPUT, nil
             set_header(RACK_REQUEST_FORM_HASH, {})
           elsif form_data? || parseable_data?
-            unless set_header(RACK_REQUEST_FORM_HASH, parse_multipart)
+            if pairs = Rack::Multipart.parse_multipart(env, Rack::Multipart::ParamList)
```
jeremyevans, line 519: I originally thought that we should check if `RACK_REQUEST_FORM_PAIRS` is already set here, and if so, use it instead of calling `Rack::Multipart.parse_multipart`.   However, we don't do that for `RACK_REQUEST_FORM_VARS`.  However, since we can no longer reparse the input, do you think we should make the change for both?

jeremyevans (approved):
Thank you for the discussion regarding this at RubyKaigi, it helped me better understand the issue.

This looks good to me.  I have one inline comment, but it can be done after this is merged if you and @ioquatix agree.

lib/rack/request.rb
matthewd: I initially wrote it that way, then realised it was wrong: `_HASH` and `_PAIRS` get set at the same time, so if we get here when `_PAIRS` is set, that means we decided we need to re-parse, so we must have a new input: https://github.com/rack/rack/blob/8f5c885f7e0427b489174a55e6d88463173f22d2/lib/rack/request.rb#L506-L510

... which does mean I've arguably propagated an existing misbehaviour here: everywhere that we set `_HASH` should really be clearing `_VARS` [and now `_PAIRS`], for instances where the new input has a different format from the old.

jeremyevans: Consider:

```ruby
request.POST
request.delete_header(RACK_REQUEST_FORM_HASH)
request.POST
```

No change in input in this case.  I guess you could argue that deleting that header means all bets are off, but it seems that using `RACK_REQUEST_FORM_VARS` or `RACK_REQUEST_FORM_PAIRS` would be better in this case.

This idea probably makes more sense if you consider that @tenderlove, @ioquatix, and I all want to remove the automatic cache invalidation currently implemented (#2073 being one approach for that).

ioquatix (approved):

ioquatix:
I discussed this with Matthew.

There is an expected follow up PR, creating a method, like `form_pairs` with the following possible implementation:

```ruby
def form_pairs
  if form_data? || parseable_data?
    if pairs = Rack::Multipart.parse_multipart(env, Rack::Multipart::ParamList)
      set_header RACK_REQUEST_FORM_PAIRS, pairs
    end
  end

  get_header RACK_REQUEST_FORM_PAIRS # cached value.
end
```

> And then in theory POST is form_pairs |> expand_param_pairs, + hash caching

PR #2089, opened 2023-06-24
Cloudfront forwarded proto header

tomharvey:
Ok - taking another swing at solving https://github.com/rack/rack/issues/2080

This time where we can pass any header in and the value of that will be copied to the value of `HTTP_X_FORWARDED_PROTO`

This allows the user to rely on the value in `CloudFront-Forwarded-Proto` provided by AWS Cloudfront. If other vendors have custom headers for this, then this should also provide an option.

### DIFF
--- a/lib/rack/set_x_forwarded_proto_header.rb
+++ b/lib/rack/set_x_forwarded_proto_header.rb
@@ -1,0 +1,32 @@
+# frozen_string_literal: true
+
+module Rack
+
+  # Middleware to set the X-Forwarded-Proto header to the value
+  # of another header.
+  #
+  # This header can be used to ensure the scheme matches when comparing
+  # request.origin and request.base_url for CSRF checking, but Rack
+  # expects that value to be in the X_FORWARDED_PROTO header.
+  #
+  # Example Rails usage:
+  # If you use a vendor managed proxy or CDN which sends the proto in a header add 
+  #`config.middleware.use Rack::SetXForwardedProtoHeader, 'Vendor-Forwarded-Proto-Header'`
+  # to your application.rb file
+  
+  class SetXForwardedProtoHeader
+    def initialize(app, vendor_forwarded_header)
+      @app = app
+      # Rack expects to see UPPER_UNDERSCORED_HEADERS, never SnakeCased-Dashed-Headers
+      @vendor_forwarded_header = "HTTP_#{vendor_forwarded_header.upcase.gsub "-", "_"}"
+    end
+
+    def call(env)
+      if value = env[@vendor_forwarded_header]
+        env["HTTP_X_FORWARDED_PROTO"] = value
+      end
+      @app.call(env)
+    end
+
+  end
+end
--- a/test/spec_set_x_forwarded_proto_header.rb
+++ b/test/spec_set_x_forwarded_proto_header.rb
@@ -1,0 +1,52 @@
+# frozen_string_literal: true
+
+require_relative 'helper'
+
+separate_testing do
+  require_relative '../lib/rack/set_x_forwarded_proto_header'
+  require_relative '../lib/rack/lint'
+  require_relative '../lib/rack/mock_request'
+end
+
+
+describe Rack::SetXForwardedProtoHeader do
+  response = lambda {|e| [200, {}, []]  }
+
+  it "leaves the value of X_FORWARDED_PROTO intact if there is no vendor header passed in the request" do
+    vendor_forwarded_header = "not passed in the request"
+    env = Rack::MockRequest.env_for("/", "HTTP_X_FORWARDED_PROTO" => "http")
+
+    Rack::Lint.new(Rack::SetXForwardedProtoHeader.new(response, vendor_forwarded_header)).call env
+
+    env["HTTP_X_FORWARDED_PROTO"].must_equal "http"
+  end
+
+  it "does not set X-Forwarded-Proto when there is no vendor header passed in the request" do
+    vendor_forwarded_header = "not passed in the request"
+    env = Rack::MockRequest.env_for("/", "FOO" => "bar")
+
+    Rack::Lint.new(Rack::SetXForwardedProtoHeader.new(response, vendor_forwarded_header)).call env
+
+    env["FOO"].must_equal "bar"
+    assert_nil(env["HTTP_X_FORWARDED_PROTO"])
+  end
+
+
+  it "copies the value of the header to X-Forwarded-Proto" do
+    env = Rack::MockRequest.env_for("/", "HTTP_VENDOR_FORWARDED_PROTO_HEADER" => "https")
+
+    Rack::Lint.new(Rack::SetXForwardedProtoHeader.new(response, "Vendor-Forwarded-Proto-Header")).call env
+
+    env["HTTP_X_FORWARDED_PROTO"].must_equal "https"
+  end
+
+  it "copies the value of the header to X-Forwarded-Proto overwriting an existing X-Forwarded-Proto" do
+    env = Rack::MockRequest.env_for("/", "HTTP_VENDOR_FORWARDED_PROTO_HEADER" => "https", "HTTP_X_FORWARDED_PROTO" => "http")
+
+    Rack::Lint.new(Rack::SetXForwardedProtoHeader.new(response, "Vendor-Forwarded-Proto-Header")).call env
+
+    env["HTTP_X_FORWARDED_PROTO"].must_equal "https"
+  end
+
+  
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ All notable changes to this project will be documented in this file. For info on
 - Add fallback lookup and deprecation warning for obsolete status symbols. ([#2137](https://github.com/rack/rack/pull/2137), [@wtn])
 - In `Rack::Files`, ignore the `Range` header if served file is 0 bytes. ([#2159](https://github.com/rack/rack/pull/2159), [@zarqman])
 - rack.early_hints is now officially supported as an optional feature (already implemented by Unicorn, Puma, and Falcon). ([#1831](https://github.com/rack/rack/pull/1831), [@casperisfine, @jeremyevans])
+- - Add `Rack::SetXForwardedProtoHeader` middleware ([#2089](https://github.com/rack/rack/pull/2089), [@tomharvey])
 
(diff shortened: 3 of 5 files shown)


lib/rack/set_x_forwarded_proto_header.rb
```diff
@@ -0,0 +1,44 @@
...
+  #
+  # Example Rails usage with AWS Cloudfront:
+  # In this AWS example, add 
+  #`config.middleware.use Rack::SetXForwardedProtoHeader, 'CloudFront-Forwarded-Proto'`
+  # to your application.rb file
+  
+  class SetXForwardedProtoHeader
+    def initialize(app, vendor_forwarded_header)
+      @app = app
+      @vendor_forwarded_header = standard_header vendor_forwarded_header
```
jeremyevans, line 23: Let's inline the code here:

```ruby
@vendor_forwarded_header = "HTTP_#{vendor_forwarded_header.upcase.gsub "-", "_"}"
```

lib/rack/set_x_forwarded_proto_header.rb
```diff
@@ -0,0 +1,44 @@
...
+    end
+
+    def call(env)
+      return @app.call(env) unless env[@vendor_forwarded_header]
```
jeremyevans, line 27: Let's simplify:

```ruby
  if value = env[@vendor_forwarded_header]
    env["HTTP_X_FORWARDED_PROTO"] = value
  end
  @app.call(env)
```

test/spec_set_x_forwarded_proto_header.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+
+require_relative 'helper'
+
+
+describe Rack::SetXForwardedProtoHeader do
+  response = lambda {|e| [200, {}, []]  }
```
jeremyevans, line 7: I would switch to:

```ruby
  response = lambda {|e| [200, {}, [e['HTTP_X_FORWARDED_PROTO']]]  }
```

Then you can more easily test expected behavior by just checking for the result in the body.

test/spec_set_x_forwarded_proto_header.rb
```diff
@@ -0,0 +1,43 @@
...
+
+
+  it "copies the value of the header to X-Forwarded-Proto" do
+    env = Rack::MockRequest.env_for("/", "HTTP_CLOUDFRONT_FORWARDED_PROTO" => "https")
+
+    Rack::Lint.new(Rack::SetXForwardedProtoHeader.new(response, "CloudFront-Forwarded-Proto")).call env
+
+    env["HTTP_X_FORWARDED_PROTO"].must_equal "https"
+  end
+
```
jeremyevans, line 41: Should add a spec to test when both the vendor header and `HTTP_X_FORWARDED_PROTO` are set, testing that the vendor header overwrites an existing `HTTP_X_FORWARDED_PROTO` header.

lib/rack/set_x_forwarded_proto_header.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  # Middleware to set the X-Forwarded-Proto header to the value
+  # of another header.
+  #
+  # This header can be used to ensure the scheme matches when comparing
+  # request.origin and request.base_url for CSRF checking, but Rack
+  # expects that value to be in the X_FORWARDED_PROTO header.
+  #
+  # For example, AWS Cloudfront sets a CloudFront-Forwarded-Proto header
+  # https://docs.aws.amazon.com/AmazonCloudFront/latest/DeveloperGuide/adding-cloudfront-headers.html#cloudfront-headers-other
```
jeremyevans, line 13: I would prefer if the documentation and specs removed all examples with CloudFront.  Please switch to a generic key so that we are not giving a preference to a specific vendor.

jeremyevans (changes requested):
Thanks for the pull request.  Please see the requested changes.

test/spec_set_x_forwarded_proto_header.rb
tomharvey: ✅

test/spec_set_x_forwarded_proto_header.rb
tomharvey: I'm unclear on why this makes things easier to test.

From a readability point of view, as the code stands in my PR, I can see a benefit in having tests that show that we're asserting that the header value has changed. If we were to return that header value in the body, we would have tests that assert some value of the body, abstracting the intention by one step.

But, open to feedback, this is your project.

jeremyevans: One advantage of returning the value in the body is it allows for a more functional approach.  With your current approach, you are relying on a side effect of changing `env`.  However, it's probably fine to leave as is.

test/spec_set_x_forwarded_proto_header.rb
```diff
@@ -0,0 +1,51 @@
...
+    env["HTTP_X_FORWARDED_PROTO"].must_equal "http"
+  end
+
+  it "returns early when there is no vendor header passed in the request" do
+    vendor_forwarded_header = "not passed in the request"
+    env = Rack::MockRequest.env_for("/", "FOO" => "bar")
+
+    header_middleware = Rack::SetXForwardedProtoHeader.new(response, vendor_forwarded_header)
+    # Patch to ensure we return early and do not call `copy_header_value`
+    def header_middleware.copy_header_value
```
jeremyevans, line 24: This method no longer exists.  Probably want to change the spec to test that `HTTP_X_FORWARDED_PROTO` is not set, with the spec description updated accordingly.

jeremyevans (changes requested):
Looks good.  One more minor spec change requested, then I'm fine with this.

test/spec_set_x_forwarded_proto_header.rb
tomharvey: Doh! Sorry about that.

jeremyevans (approved):
Looks good, thanks!

tomharvey:
> Looks good, thanks!

Great. Thanks for the speedy and clear feedback on the review

ioquatix:
Do you mind checking why the tests are failing?

> DEPRECATED: Use assert_nil if expecting nil from /home/runner/work/rack/rack/test/spec_set_x_forwarded_proto_header.rb:25. This will fail in Minitest 6.

tomharvey:
will have a look tomorrow. 

> Do you mind checking why the tests are failing?

jeremyevans:
I'm OK with merging this if the conflicts and test issues are fixed.

tenderlove:
> I'm OK with merging this if the conflicts and test issues are fixed.

+1

Seems like there's a merge conflict committed.

@tomharvey after you fix the conflicts would you mind squashing this PR?  Thank you!

tomharvey:
> > I'm OK with merging this if the conflicts and test issues are fixed.
> 
> +1
> 
> Seems like there's a merge conflict committed.
> 
> @tomharvey after you fix the conflicts would you mind squashing this PR? Thank you!

Hey hey - will do.

Sorry, thought I'd switch it over to a draft while I sorted out my nonsense. Sorry for the noise.

tomharvey:
> > > I'm OK with merging this if the conflicts and test issues are fixed.
> > 
> > 
> > +1
> > Seems like there's a merge conflict committed.
> > @tomharvey after you fix the conflicts would you mind squashing this PR? Thank you!
> 
> Hey hey - will do.
> 
> Sorry, thought I'd switch it over to a draft while I sorted out my nonsense. Sorry for the noise.

Fixed my obvious doofus errors.

Squashed

Tests run locally in ruby 3.2.2 - can I have some workflows to check the other versions please?

ioquatix:
I understand the problem trying to be solved with this middleware, but after consideration, I disagree with merging this PR into `rack`.

1. As discussed, it seems extremely vendor specific and doesn't align with the goals of "rack the interface". As such, I think this would be a suitable candidate for `rack-contrib` instead.
2. `X-Forwarded-Proto` is a de-facto standard at best, and the actual standard is the [`Forwarded` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded). I think we should be encouraging use of the `Forwarded` header and we support this in `Rack::Request`.

As such, I'd like to revert this PR and suggest moving it to `rack-contrib`, while additionally considering whether we can follow the published standards (e.g. the `forwarded` header).

tomharvey:
> I understand the problem trying to be solved with this middleware, but after consideration, I disagree with merging this PR into `rack`.
> 
> 1. As discussed, it seems extremely vendor specific and doesn't align with the goals of "rack the interface". As such, I think this would be a suitable candidate for `rack-contrib` instead. By vendor specific, I mean that a normal HTTP server handling traffic from a browser would not need this middleware - only vendor-specific load balancers that aren't following the standards require such a mapping.
> 2. `X-Forwarded-Proto` is a de-facto standard at best, and the actual standard is the [`Forwarded` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Forwarded). I think we should be encouraging use of the `Forwarded` header and we support this in `Rack::Request`. It would make more sense to me to have a mapping for `X-Forwarded-*` -> `Forwarded` and as a feature, we could specify the prefix, e.g. `Foo-Forwarded-*`.
> 
> As such, I'd like to revert this PR and suggest moving it to `rack-contrib`, while additionally considering whether we can follow the published standards (e.g. the `Forwarded` header).

I’m unclear if the above is a proposal for the team to consider or a decision on the way forward. Can you clarify how the rack team makes these kinds of decisions?

If rack-contrib is the way forward; I’d appreciate the opportunity to move the code to rack-contrib myself.

ioquatix:
> I’m unclear if the above is a proposal for the team to consider or a decision on the way forward. Can you clarify how the rack team makes these kinds of decisions?

It's now a decision on the way forward. Apologies for being opaque, but sufficient consensus was reached within the @rack/release-managers team.

> If rack-contrib is the way forward; I’d appreciate the opportunity to move the code to rack-contrib myself.

Please do - [either continue the work from here](https://github.com/rack/rack-contrib/pull/192) (a copy of this PR) or open a new one. I'd be delighted if you could continue the work.

tomharvey:
I'd prefer to open a new one, with apologies to @mpalmer that the PR will lose his valuable feedback.

My motivation is based on an opportunity to contribute a useful feature to a project that has given the community so much. But, I do still expect that my contributions would be recognised as my own instead of a PR from @ioquatix

PR #2094, opened 2023-07-15
Add support for specifying options to builder and thus `config.ru`.

ioquatix:
Rails 7.1 will introduce support for Rack 3.

Falcon supports Rack 2 and Rack 3, but the concurrency model is per-fiber, not per-thread or per-process as in other servers.

While in theory, the safest option for all code is to behave per-fiber (the smallest unit of simultaneous execution), this can break existing applications that depend on Rails state to be shared between fibers. However, this behaviour is not acceptable for request-per-fiber servers and can cause bugs due to unintentional sharing of, e.g. database transactions or per-request state.

Rails introduced `IsolatedExecutionState` in https://github.com/rails/rails/pull/43596 however this must be manually configured. I am concerned that this raises the bar for adoption of Falcon so I'd like to simplify it a bit.

This PR introduces `Rack::Builder#options` with a proposed `isolation` option which can one of `:process`, `:thread` or `:fiber`. We can discuss the exact semantics but I think we can leave the exact options to be server-specific metadata that is known at "build time".

By introducing this option, Falcon can publish `isolation => fiber` and Rails can use this to automatically configure itself correctly.

In the `config.ru` someone can just write:

```ruby
# config.ru

if isolation = options[:isolation]
  Rails.application.config.active_support.isolation_level = isolation
end
```

As this feature is primarily aimed at Rails 7.1, I'd like to propose we backport this to 3.0. Alternatively, maybe we can cut a 3.1 release of Rack from 3.0 stable... but I'm less enthusiastic about that. As a last resort, we could try to see whether Rails 7.1 is compatible with the Rack HEAD (i.e. 3.1).

cc @byroot @matthewd

### DIFF
--- a/lib/rack/builder.rb
+++ b/lib/rack/builder.rb
@@ -2,6 +2,9 @@
 
 require_relative 'urlmap'
 
+module Rack; end
+Rack::BUILDER_TOPLEVEL_BINDING = ->(builder){builder.instance_eval{binding}}
+
 module Rack
   # Rack::Builder provides a domain-specific language (DSL) to construct Rack
   # applications. It is primarily used to parse +config.ru+ files which
@@ -59,9 +62,9 @@ module Rack
     #   # requires ./my_app.rb, which should be in the
     #   # process's current directory.  After requiring,
     #   # assumes MyApp constant is a Rack application
-    def self.parse_file(path)
+    def self.parse_file(path, **options)
       if path.end_with?('.ru')
-        return self.load_file(path)
+        return self.load_file(path, **options)
       else
         require path
         return Object.const_get(::File.basename(path, '.rb').split('_').map(&:capitalize).join(''))
@@ -81,7 +84,7 @@ module Rack
     #   use Rack::ContentLength
     #   require './app.rb'
     #   run App
-    def self.load_file(path)
+    def self.load_file(path, **options)
       config = ::File.read(path)
       config.slice!(/\A#{UTF_8_BOM}/) if config.encoding == Encoding::UTF_8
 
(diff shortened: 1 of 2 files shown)


ioquatix:
@headius it looks like JRuby has some major issues with deflate?

jeremyevans (approved):
Approved for merging into master, though I think we should wait a while for @tenderlove to review.  However, I'm not in favor of merging this into 3.0 (in principle, we should not backport new features, only bug fixes). However, if @tenderlove or other Rails maintainers would like this merged into 3.0, I don't object.

jeremyevans:
> @headius it looks like JRuby has some major issues with deflate? Anything I can do to help?

This has been the case since JRuby 9.4.3.0, I think due to the `jzlib` changes.

byroot:
I see how this is convenient, but my concern is that flipping `config.active_support.isolation_level` can break the application code (e.g. code spawning a fiber and expecting to have some request state available).

So if you are running Falcon in production, you should set the isolation level to `fiber` in all environments, including `test`, as well as in your background job worker (e.g. sidekiq) etc.

ioquatix:
I understand, but just because we can’t solve all the problems doesn’t mean we shouldn’t make it a bit easier. A background job server may not have the same concurrency model so not sure it applies to sidekiq?

byroot:
> just because we can’t solve all the problems doesn’t mean we shouldn’t make it a bit easier.

My point is that in some way this is worse. Because if you don't manually set `isolation_level` your code will behave differently on test vs production, so you might feel like you are covered by tests when in fact you aren't.

> A background job server may not have the same concurrency model so not sure it applies to sidekiq?

You want the same isolation level for all your environments, otherwise some code may work in some environments and not in some others.

This is not to say this feature isn't interesting, I think we could use it to show a warning or error out but I don't think Rails should switch the isolation level based on that.

ioquatix:
I'm fine if it's only a warning. We can leave that decision up to Rails.

headius:
I'll look into the zlib issues.

headius:
https://github.com/jruby/jruby/issues/7879

Issue #2096, opened 2023-07-17
HTML escape of slash is not recommended by OWASP


JunichiIto:
`Rack::Utils.escape_html` escapes `/` to `&#x2F;`, but OWASP doesn't mention it now:

https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html

So `Rack::Utils.escape_html` should escape `&<>'"` only.

## Motivation

There is no difference in the code's safety whether you leave this specification as it is or make changes. However, I spent much time investigating why `Rack::Utils.escape_html` escapes slashes and whether there is any significance to escaping them. Unnecessary escaping can confuse users like me.

Additionally, I found that `CGI.escapeHTML` only escapes the characters `'&"<>`. It also made me confused and investigate why the two similar methods have different specifications.

https://docs.ruby-lang.org/en/3.2/CGI/Util.html#method-i-escapeHTML

## Related pages

- https://github.com/rack/rack/issues/27
- https://github.com/rack/rack/commit/37ca2539c423f29337486cf0469dd49a019cfea3
- https://security.stackexchange.com/a/249423
- https://github.com/OWASP/CheatSheetSeries/commit/d7a6823db92af82da3f9538b23b03851d7556d78

ioquatix:
I agree the security should not be impacted by this change, and making it consistent with `CGI.escapeHTML` should allow us to use that implementation for improved performance.

PR #2097, opened 2023-07-17
updated escape_html not to escape forward slash

JunichiIto:
fix: https://github.com/rack/rack/issues/2096

NOTE: This PR will break backwards compatibility.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file. For info on
 ### SPEC Changes
 
 - `rack.input` is now optional. ([#1997](https://github.com/rack/rack/pull/1997), [@ioquatix])
+- `Rack::Utils.escape_html` doesn't escape forward slash (`/`) now. ([#2097](https://github.com/rack/rack/pull/2097), [@JunichiIto])
 
 ### Changed
 
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -179,15 +179,14 @@ module Rack
       "<" => "&lt;",
       ">" => "&gt;",
       "'" => "&#x27;",
-      '"' => "&quot;",
-      "/" => "&#x2F;"
+      '"' => "&quot;"
     }
 
     ESCAPE_HTML_PATTERN = Regexp.union(*ESCAPE_HTML.keys)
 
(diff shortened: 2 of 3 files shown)


jeremyevans:
Can you switch this to use `CGI.escapeHTML` if available for improved performance?

ioquatix:
> Can you switch this to use CGI.escapeHTML if available for improved performance?

I think we should do that, but I'm fine if it's a separate PR.

ioquatix (approved):
LGTM.

JunichiIto:
@ioquatix 
Thank you for your approve.

@jeremyevans 

> Can you switch this to use CGI.escapeHTML if available for improved performance?

I have not measured yet, but I guess `Rack::Utils.escape_html` would be faster because `CGI.escapeHTML` has more lines:

https://github.com/ruby/cgi/blob/v0.3.6/lib/cgi/util.rb#L74-L91

And some behaviors are different:

- Rack escapes `'` to `&#x27;` but CGI escapes to `&#39;`. (hexadecimal vs decimal)
- Rack cannot escape `\300<` (raises "invalid byte sequence in UTF-8 (ArgumentError)"), but CGI can escape it because CGI converts encoding to ASCII-8BIT before calling `gsub!`.

I"m not sure if it can be replaced with `CGI.escapeHTML`.

### Related commits

- https://github.com/rack/rack/commit/37ca2539c423f29337486cf0469dd49a019cfea3#diff-85960cbd609e7f51a9881d1876b82d748f4ba8312a31e3329a5bb9bf791c60beR135
- https://github.com/rack/rack/commit/5a942ebd79e147479047221e571bc86d9da24a33
- https://github.com/ruby/cgi/commit/2900336e945d7792627c343c382d7268284c54fd

jeremyevans:
`CGI.escapeHTML` is implemented in C in recent versions of the `cgi` gem, so it should definitely be faster (https://github.com/ruby/cgi/blob/master/ext/cgi/escape/escape.c#L458C6-L458C6).  Basically, you require `cgi/escape`, if it succeeds you use it, and if it fails, then you can fallback to the existing implementation.

I don't think the behavior differences you mention are blockers to merging, though they may require test updates.  CGI behavior using decimal is better as it saves a character :), and supporting non-UTF8 is also a useful feature.

PR #2098, opened 2023-07-19
Make query parameters without = have empty string values

jeremyevans:
This was reverted in 77cf0579ddd53da6e1a449a26850464284a75aec so that Rails could have historical Rack behavior without the complexity of splitting form/query parsing.

Now that splitting form/query parsing has been added back in 3855d1d2638af92193c0636107247bc31c98b2f1, this changes the behavior to what originally shipped in Rack 3.

This matches URL spec section 5.1.3.3.  Frameworks that want Rack's historical behavior of using nil values instead of empty string values can reparse QUERY_STRING and use rack.request.form_pairs to get the behavior they want.

### DIFF
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -129,6 +129,8 @@ module Rack
 
       return if k.empty?
 
+      v ||= String.new
+
       if after == ''
         if k == '[]' && depth != 0
           return [v]
--- a/test/spec_query_parser.rb
+++ b/test/spec_query_parser.rb
@@ -12,6 +12,6 @@ describe Rack::QueryParser do
   it "can normalize values with missing values" do
     query_parser.parse_nested_query("a=a").must_equal({"a" => "a"})
     query_parser.parse_nested_query("a=").must_equal({"a" => ""})
-    query_parser.parse_nested_query("a").must_equal({"a" => nil})
+    query_parser.parse_nested_query("a").must_equal({"a" => ""})
   end
 end
--- a/test/spec_request.rb
+++ b/test/spec_request.rb
@@ -583,9 +583,9 @@ class RackRequestTest < Minitest::Spec
   it "parse the query string" do
     request = make_request(Rack::MockRequest.env_for("/?foo=bar&quux=bla&nothing&empty="))
     request.query_string.must_equal "foo=bar&quux=bla&nothing&empty="
-    request.GET.must_equal "foo" => "bar", "quux" => "bla", "nothing" => nil, "empty" => ""
+    request.GET.must_equal "foo" => "bar", "quux" => "bla", "nothing" => "", "empty" => ""
     request.POST.must_be :empty?
-    request.params.must_equal "foo" => "bar", "quux" => "bla", "nothing" => nil, "empty" => ""
+    request.params.must_equal "foo" => "bar", "quux" => "bla", "nothing" => "", "empty" => ""
   end
 
   it "handles invalid unicode in query string value" do
--- a/test/spec_utils.rb
+++ b/test/spec_utils.rb
@@ -141,7 +141,7 @@ describe Rack::Utils do
 
   it "parse nested query strings correctly" do
     Rack::Utils.parse_nested_query("foo").
-      must_equal "foo" => nil
+      must_equal "foo" => ""
     Rack::Utils.parse_nested_query("foo=").
       must_equal "foo" => ""
     Rack::Utils.parse_nested_query("foo=bar").
@@ -158,7 +158,7 @@ describe Rack::Utils do
     Rack::Utils.parse_nested_query("&foo=1&&bar=2").
       must_equal "foo" => "1", "bar" => "2"
     Rack::Utils.parse_nested_query("foo&bar=").
-      must_equal "foo" => nil, "bar" => ""
+      must_equal "foo" => "", "bar" => ""
     Rack::Utils.parse_nested_query("foo=bar&baz=").
       must_equal "foo" => "bar", "baz" => ""
     Rack::Utils.parse_nested_query("my+weird+field=q1%212%22%27w%245%267%2Fz8%29%3F").
@@ -168,19 +168,19 @@ describe Rack::Utils do
       must_equal "pid=1234" => "1023", "a" => "b"
 
     Rack::Utils.parse_nested_query("foo[]").
-      must_equal "foo" => [nil]
+      must_equal "foo" => [""]
     Rack::Utils.parse_nested_query("foo[]=").
       must_equal "foo" => [""]
     Rack::Utils.parse_nested_query("foo[]=bar").
       must_equal "foo" => ["bar"]
     Rack::Utils.parse_nested_query("foo[]=bar&foo").
-      must_equal "foo" => nil
+      must_equal "foo" => ""
     Rack::Utils.parse_nested_query("foo[]=bar&foo[").
-      must_equal "foo" => ["bar"], "foo[" => nil
+      must_equal "foo" => ["bar"], "foo[" => ""
     Rack::Utils.parse_nested_query("foo[]=bar&foo[=baz").
       must_equal "foo" => ["bar"], "foo[" => "baz"
     Rack::Utils.parse_nested_query("foo[]=bar&foo[]").
-      must_equal "foo" => ["bar", nil]
+      must_equal "foo" => ["bar", ""]
     Rack::Utils.parse_nested_query("foo[]=bar&foo[]=").
       must_equal "foo" => ["bar", ""]
 
@@ -192,7 +192,7 @@ describe Rack::Utils do
       must_equal "foo" => ["bar"], "baz" => ["1", "2", "3"]
 
     Rack::Utils.parse_nested_query("x[y][z]").
-      must_equal "x" => { "y" => { "z" => nil } }
+      must_equal "x" => { "y" => { "z" => ""} }
     Rack::Utils.parse_nested_query("x[y][z]=1").
       must_equal "x" => { "y" => { "z" => "1" } }
     Rack::Utils.parse_nested_query("x[y][z][]=1").


dentarg:
Is this going to ship with Rack 3.1?

ioquatix:
I suggest we consider this for Rack 3.2+.

jeremyevans:
I would like this behavior in 3.1.  This was the original Rack 3.0.0 behavior. We compromised our integrity to change the behavior to be the same as 2.2 in response to complaints from Rails.  Then we merged the form_pairs support, also in response to complaints from Rails, to give Rails the control it wants.  Now that the form_pairs commit has been merged, we can change back to the behavior desired and shipped in 3.0.0, and Rails can use form_pairs to implement the behavior it wants.

tenderlove:
I think this is good, but I'd like input from @matthewd

ioquatix:
@jeremyevans sorry to impose, but do you mind giving a brief summary of the advantages of this change again?

lib/rack/query_parser.rb
```diff
@@ -129,6 +129,8 @@ def normalize_params(params, name, v, _depth=nil)
 
       return if k.empty?
 
+      v ||= String.new
```
ioquatix, line 132: At one point, I think we made this configurable using a method - is that still worth considering? I liked that approach.

jeremyevans: I don't think so.  There isn't a strong need for it now that form_pairs can be used for custom parameter parsing.

jeremyevans:
> @jeremyevans sorry to impose, but do you mind giving a brief summary of the advantages of this change again?

Compliance with the URL spec, section 5.1.3.3.  You agreed that this behavior is correct: https://github.com/rack/rack/issues/1696#issuecomment-1020779676 .  This is how the browser Javascript methods handle things: https://github.com/rack/rack/issues/1696#issuecomment-667295351

ioquatix:
While I agree it's correct according to the spec, I'm worried about things like <https://github.com/rack/rack/issues/1696>.

ioquatix:
I am okay with this PR, but would like @matthewd to confirm that this would be reasonable for Rack 3.1. I'd also like @tenderlove to approve it too if possible. This one has a bigger chance for backwards compatibility issues, so I'd like everyone to be clear on that.

ioquatix (approved):
@matthewd can you please also review this.

matthewd:
So, my obvious concern is going to be that while Rails now has access to APIs to do its own parameter parsing, no existing releases are actually doing that.

I can fix that for the incipient 7.2, but I do worry that people might upgrade existing 7.1 applications to the latest Rack without much concern for edge cases outside their test coverage.

... and I now realise that 7.1 has a completely open-ended dependency on Rack, so that's somewhat of a permanent problem [arguably of our own making]. Given an overabundance of druthers, I might hope we could have a medium-sized [or at least non-zero] window where "latest-available Rails does its own back-compat param parsing" precedes "latest-available Rack diverges from previous behaviour"... but I don't think that's really a reasonable thing for me to ask.

I agree that making it configurable is unnecessary given we have a more holistic solution available... one option would be to just extract it to a method, though -- to at least allow any affected applications (Rails or otherwise) to monkey-patch their way back to historical compatibility without having to reproduce all of `_normalize_params` until they move to a newer framework version that can handle it for them. 🤷🏻‍♂️

(I still think it's weird for us to make this change given that all our array/hash parsing is, by definition & design, wildly non-conformant -- but I accept that we're past the question of _whether_ it should happen.)

ioquatix:
If that's the case we should probably punt this to a future release of Rack.

jeremyevans:
> If that's the case we should probably punt this to a future release of Rack.

OK with me.

ioquatix:
@matthewd can you please advise us when Rails has got a fix in place, and then at the next major release of Rails, we have the option to adopt this PR. IOW, we will wait at least one major Rails release with the fix, before releasing a major release of Rack with this change.

ioquatix:
@jeremyevans To avoid string allocations, instead of `v ||= String.new` what about using a frozen string sentinel?

jeremyevans:
> @jeremyevans To avoid string allocations, instead of `v ||= String.new` what about using a frozen string sentinel?

I don't think it is worth it.  This is an edge case, almost nothing sends query parameters without equals, so there is no point in trying to optimize it.

ioquatix:
Given that your bar for compatibility going forward is much higher, is this still a change you want to move forward with?

jeremyevans:
There are a couple significant differences between this and #2253:

* This reduces the possible parameter types (removing `nil`, leaving `String`, `Array`, `Hash`), while 2253 increases the possible `env["QUERY_STRING"]` types (changing `String` to `String` or `nil`).  So this reduces complexity, while #2253 increases complexity.
* This increases standards compliance with URL spec, whereas #2253 is unrelated to standards compliance.

That being said, I agree that this would not meet the bar.  The backwards compatibility issues are minor, but the benefits are also minor.  

I'm happy to close this if there is a general agreement that we will avoid backwards incompatible changes going forward unless the benefits greatly exceed the backwards compatibility impact.  In other words:

* no backwards incompatible changes for minor usability improvements
* accepting minor backwards incompatibility changes for major usability improvements
* no major backwards incompatible changes

jeremyevans:
> I think your criteria is too strict.

Fair enough.  Your opinion regarding this doesn't surprise me. :)

> https://www.hyrumslaw.com/

I can't tell if you are using this as an argument for or against your position, so you may want to explain. IMO, it would be an argument for my position, because if even minor changes are likely to break things you do not expect, the bar should be high.  Do you disagree with that, not care about breakage, or care but consider breakage of only minor importance?

ioquatix:
I see Hyrum's Law as a valuable thought experiment for understanding the challenges of change management.

Unless we declare Rack "complete" and commit to never making another change, every modification carries the risk of breaking something downstream. However, even the idea of considering Rack "complete" is unrealistic, as we’re tied to the ongoing evolution of Ruby itself — changes like moving stdlib to "not installed by default" gems have already created challenges for us. In other words, we have to accept that some degree of disruption is inevitable.

The real challenge is figuring out how to manage changes in a way that minimizes compatibility issues. I'm not looking to say, "No, you can't merge this PR because it doesn't meet some arbitrary bar (that we decided on)." You believe the PR has value, and I can't see any technical reason not to implement it. But it's also important to acknowledge that merging it will have downstream impact, and we need to be mindful of how we handle that (as you are).

At present, the current mechanism we have for managing changes is major and minor versions. Maybe it's insufficient - or maybe we need to use semantic versioning to communicate breaking changes more easily. Or maybe we need better validations (e.g. `rack-conform` is one way in which I've tried to do that). `Rack::Lint` is okay but doesn't validate applications against every possible input.

Another option we've toyed with is allowing `config.ru` to specify options - e.g. an application could specify whether it's capable of dealing with `rack.input=nil` or `query_string=nil`. In the case of this PR, we could make it a non-default option, until a major version bump, e.g.

```ruby
# config.ru

options[:default_query_parser] = Rack::QueryParser.new(default_empty: "")
```

We've also got "global" options within `Rack` itself, e.g. `Rack::Utils.default_query_parser`. But we found out with Rails that it wasn't really sufficient.

I'm not advising on any specific approach, but what I think we should do is figure out a way to adopt changes like this in a way which allows us to move forward without breaking things if possible. Then, maybe a strategy about adopting those changes as defaults in major version bumps might be more acceptable?

In other words, we allow changes like this, provided they are gated by a feature flag, and feature flags could only be changed in a major release?

jeremyevans:
> I see Hyrum's Law as a valuable thought experiment for understanding the challenges of change management.
> 
> Unless we declare Rack "complete" and commit to never making another change, every modification carries the risk of breaking something downstream. However, even the idea of considering Rack "complete" is unrealistic, as we’re tied to the ongoing evolution of Ruby itself — changes like moving stdlib to "not installed by default" gems have already created challenges for us. In other words, we have to accept that some degree of disruption is inevitable.

Some degree of disruption is inevitable.  However, I think we have a duty to our users to minimize the amount of disruption, and only accept unnecessary disruption in cases where the disruption is minimal compared to the benefit provided.

> The real challenge is figuring out how to manage changes in a way that minimizes compatibility issues. I'm not looking to say, "No, you can't merge this PR because it doesn't meet some arbitrary bar (that we decided on)." You believe the PR has value, and I can't see any technical reason not to implement it. But it's also important to acknowledge that merging it will have downstream impact, and we need to be mindful of how we handle that (as you are).

The best way to minimize compatibility issues is to reject changes that result in compatibility issues, unless those changes have a much more significant benefit.

I believe the answer to the question "Should we break backwards compatibility" is not "Never", but it is "Never without a very good reason".

> At present, the current mechanism we have for managing changes is major and minor versions. Maybe it's insufficient - or maybe we need to use semantic versioning to communicate breaking changes more clearly. Or maybe we need better validations (e.g. `rack-conform` is one way in which I've tried to do that). `Rack::Lint` is okay but doesn't validate applications against every possible input.

Or maybe we need to reject changes that result in compatibility issues, unless there is a much more significant benefit.

> Another option we've toyed with is allowing `config.ru` to specify options - e.g. an application could specify whether it's capable of dealing with `rack.input=nil` or `query_string=nil`. In the case of this PR, we could make it a non-default option, until a major version bump, e.g.
> 
> ```ruby
> # config.ru
> 
> options[:default_query_parser] = Rack::QueryParser.new(default_empty: "")
> ```

I'm not necessarily opposed to an opt-in approach for `config.ru`, as long as existing `config.ru` files work in a backwards compatible manner.

However, I do not think such an approach is appropriate for `SPEC`.

> I'm not advising on any specific approach, but what I think we should do is figure out a way to adopt changes like this in a way which allows us to move forward without breaking things if possible. Then, maybe a strategy about adopting those changes as defaults in major version bumps might be more acceptable?

I am advising a general approach of only breaking backwards compatibility when the benefits are much greater than the backwards compatibility lost.  Not in minor versions, not in major versions, not ever, unless the benefits are much greater than the backwards compatibility lost.

While this bar is much higher than our past practice, I don't think this is an unreasonably high bar.  I think it is reasonable, because successful projects like SQLite and PostgreSQL appear to have similarly high bar for backwards incompatible changes. There are successful projects with even higher bars than what I am proposing, such as curl (https://everything.curl.dev/libcurl/api.html) 

> In other words, we allow changes like this, provided they are gated by a feature flag, and feature flags could only be changed in a major release? (I also acknowledge this approach has significant maintenance overhead).

There are times when it makes sense to break backwards compatibility on a regular basis or with limited concern.  Newly developed libraries figuring out what approaches will work best, and libraries without significant usage are a couple of examples.   That does not describe rack, a library that is used by almost all web applications and which has been around for a long time.

IMO, we would serve the greater good with my proposal for handling backwards incompatible changes, compared to your proposal.

This does not mean we have to stop improving Rack.  However, we should attempt to ensure our improvements to Rack are backwards compatible.  SQLite, PostgreSQL, and curl show that you can have a high bar for backwards incompatible changes, and still continue to improve over time.

ioquatix:
> The best way to minimize compatibility issues is to reject changes that result in compatibility issues, unless those changes have a much more significant benefit.

**The best way to minimize compatibility issues is to reject changes that result in compatibility issues** bar none. In other words, only additive changes are acceptable. I should note that I find this position to be too extreme and I'm okay with something that allows more subjectivity.

> I am advising a general approach of only breaking backwards compatibility when the benefits are much greater than the backwards compatibility lost.

It's a fair point, but I don't think it's possible to be objective about this. One way to make it more objective is to test the downstream impact (which we are doing with the external tests). Adding more external tests and using that as a benchmark for "compatibility" might be one way to put a line in the sand.

If we don't want to break backwards compatibility, the only way to move forward with this PR is to have some kind of feature flag or equivalent option.

`Rack::Builder#options` was already designed for that purpose (per-app configuration), so it might not be a bad idea to use that. Given that seems to be the direction we are heading in, are you able to explore how that might work?

jeremyevans:
> The best way to minimize compatibility issues is to reject changes **bar none**. In other words, only independent "additions" are acceptable. I should note that I find this position to be too extreme and I'm okay with something that allows more subjectivity.

Note that I am not advocating for a position that extreme.  You didn't directly imply I was, but I want to be clear that I'm not.

> > I am advising a general approach of only breaking backwards compatibility when the benefits are much greater than the backwards compatibility lost.
> 
> It's a fair point, but I don't think it's possible to be objective about this. One way to make it less subjective is to test the downstream impact (which we are doing with the external tests). Adding more external tests and using that as a benchmark for "compatibility" might be one way to put a line in the sand.

I agree, there is no objective approach for this.  Adding more external tests is good to a point, but the point of those external tests is to alert us to potential backwards compatibility issues that we may not be aware of.  We should not be making the assumption that changes that we know will break backwards compatibility are OK if there is minimal or no breakage in our external tests.  There will be easily at least 10000x more code relying on rack that we wouldn't be testing.

Note that there is nothing wrong with subjectivity here.  I am the one who proposed this pull request, and I can subjectively state that I think the benefits of this pull request are higher than the backwards compatibility costs, but not high enough that it comes close to meeting the bar I am proposing.  The benefits should be much higher than the backwards compatibility costs, ideally an order of magnitude higher.

FWIW, I think that #2253 and #2239 are even further from the bar than this PR.

> If we don't want to break backwards compatibility, the only way to move forward with this PR is to have some kind of feature flag or equivalent option.

As I said, I'd be fine with closing/changing this if we could come to an agreement on how to handle backwards compatibility.

I don't think it makes sense to close/change this if we would consider merging #2253 or #2239, as the benefit from those seems lower.

Feature flags have costs, and I don't think the benefit of supporting this with a feature flag is necessarily worth it either (likewise for #2253 or #2239)

> `Rack::Builder#options` was already designed for that purpose (per-app configuration), so it might not be a bad idea to use that. Given that seems to be the direction we are heading in, are you able to explore how that might work? Rails takes this approach with its configuration, e.g. loading defaults for version X of Rails as part of the application configuration.

For middleware/app configuration, I think such an approach could work.  For things that affect `SPEC`, I don't think such an approach would work.

For this PR, the backwards compatible approach would be to do `v ||= some_method`, where `some_method` defaulted to `nil` (for backwards compatibility), but which apps could redefine if they wanted to opt in to the new behavior.  That doesn't require a feature flag, merely offering an extension point.

ioquatix:
> Note that I am not advocating for a position that extreme. You didn't directly imply I was, but I want to be clear that I'm not.

Understood.

I don't want to be a barrier to you adopting something you consider a good change. I've already approved this PR and I think fundamentally the compatibility issues can be mitigated by documentation / release notes / etc. However as we've directly experienced, this change breaks downstream libraries and applications. I agree with you that we should hold ourselves to a high bar. But my only counter point is that I think we should have an escape hatch for moving things forward (e.g. feature flags / options / or something else).

Regarding this specific change and the proposal to make it a method that can be overridden. In essence I'm fine with that. However, essentially global configuration. We considered using this with Rails to configure the query parser behaviour, and it was (very  understandably) unacceptable to the Rails team to use a global configuration within Rack as it would potentially affect any library/application that loaded Rails.

There are lots of things in Rack which are currently configurable global state. However, I think it would be vastly superior if this could be configured in a per-app (i.e. per `Rack::Builder`) way. Even if you chose to add an extension point here, it would be vastly superior if we could then write:

```ruby
# config.ru

options[:rack_query_parser] = Rack::QueryParser.new(...)

```

Or something to that effect. IMHO.

matthewd:
> I think it would be vastly superior if this could be configured in a per-app (i.e. per Rack::Builder) way

I think the trouble (in my view) is that for compatibility, this sort of thing needs to be configured in a per-_`app`_ way -- with each involved middleware able to experience the env in the way it expects to.

(I agree, but don't feel I have much fresh insight to add, with the assertion that a breaking change to SPEC necessarily warrants a higher compatibility bar... although in a post-no-rewind world, it's also true that everyone is effectively forced to use Rack-the-gem [and not implement their own direct `env` access following Rack-the-SPEC], as soon as they want a glimpse of the params.)

jeremyevans:
After further considering this and discussing it with @ioquatix, I'm closing this.  I believe the benefits of this exceed the backwards compatibility cost, but the benefits of this are not much greater than the backwards compatibility cost.

The bar for backwards incompatible changes going forward is the benefits should be at least an order of magnitude (10x) greater than the costs.  As benefits and costs are subjective, this bar is open to interpretation, but the benefits have to be much, much higher than the costs to break backwards compatibility.

ioquatix:
> (I agree, but don't feel I have much fresh insight to add, with the assertion that a breaking change to SPEC necessarily warrants a higher compatibility bar... although in a post-no-rewind world, it's also true that everyone is effectively forced to use Rack-the-gem [and not implement their own direct env access following Rack-the-SPEC], as soon as they want a glimpse of the params.)

Regarding this, one way is to document the keys that `Rack::Request` uses to parse the parameters. That way, we can standardise it in a way that doesn't depend on the implementation of the gem.

Issue #39, opened 2010-07-19
Improvements to the Session Store


josevalim:
- Provide :cookie_only as option and true as default. If false, allows SID to be retrieved from GET/POST params;
- Do not send the cookie back to the client if session id did not change (improves both server and client side perf);
- Make Abstract::ID implementation more modular, allowing Cookie implementation 
  to be simpler by inheriting from it;

All here:

http://github.com/josevalim/rack/commit/5dac45a38c38003f6608ca94fdf40d1d6f4e68ab

This is a giant step forward to make Rails depends on Rack::Session store instead of reimplementing everything from scratch.

PR #2099, opened 2023-07-22
Delegate `Rack::Utils.escape_html` to `CGI.escapeHTML`

JunichiIto:
This PR improves the performance of `Rack::Utils.escape_html`.

## Background
https://github.com/rack/rack/pull/2097#issuecomment-1639050856

## Benchmark

I defined two methods:

```ruby
    def escape_html(string)
      CGI.escapeHTML(string)
    end

    def escape_html_old(string)
      string.to_s.gsub(ESCAPE_HTML_PATTERN, ESCAPE_HTML)
    end
```

I wrote benchmark script:

```ruby
require 'benchmark'
require 'rack/utils'

# Generated by ChatGPT
HTML_EXAMPLES = <<HTML.lines(chomp: true)
<p>It's a beautiful day.</p>
<a href="https://example.com/page?id=123&category=5">Link</a>
<button onclick="alert('Hello, world!')">Click me</button>
<input type="text" value="It's my text">
<div class='container'>Content goes here</div>
<span>5 &times; 10 = 50</span>
<img src='image.jpg' alt="Nature's beauty">
<p style='color: red;'>This text is red.</p>
<a href='javascript:alert("XSS attack!");'>Click here</a>
<input type='text' placeholder='Enter your name & email'>
HTML
targets = (HTML_EXAMPLES * 10).shuffle

num_iteration = 10_000

Benchmark.bm 5 do |r|
  r.report "new" do
    num_iteration.times do
      targets.each do |string|
        Rack::Utils.escape_html(string)
      end
    end
  end

  r.report "old" do
    num_iteration.times do
      targets.each do |string|
        Rack::Utils.escape_html_old(string)
      end
    end
  end
end
```

Here is the result on my local machine:

```
$ bundle exec ruby benchmark.rb
            user     system      total        real
new     0.257449   0.001930   0.259379 (  0.259876)
old     1.706600   0.004664   1.711264 (  1.711273)
```

The new version is 6.6 times faster than old one.

## Requirements
Requires Ruby 2.3.0 or higher since it depends on 'cgi/escape' library.

> CGI.escapeHTML is optimized with C extention. https://github.com/ruby/ruby/pull/1164
>
> https://github.com/ruby/ruby/blob/v2_3_0/NEWS

I'm not sure if rack should support Ruby 2.2.x or older.

## Backward compatibility
`'` is now escaped to `#39;` instead of `#x27;` (decimal vs hexadecimal)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ All notable changes to this project will be documented in this file. For info on
 ### SPEC Changes
 
 - `rack.input` is now optional. ([#1997](https://github.com/rack/rack/pull/1997), [@ioquatix])
-- `Rack::Utils.escape_html` doesn't escape forward slash (`/`) now. ([#2097](https://github.com/rack/rack/pull/2097), [@JunichiIto])
+- `Rack::Utils.escape_html` is now delegated to `CGI.escapeHTML`. `'` is escaped to `#39;` instead of `#x27;`. (decimal vs hexadecimal) ([#2099](https://github.com/rack/rack/pull/2099), [@JunichiIto])
 
 ### Changed
 
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -6,6 +6,7 @@ require 'fileutils'
 require 'set'
 require 'tempfile'
 require 'time'
+require 'cgi/escape'
 
 require_relative 'query_parser'
 require_relative 'mime'
@@ -174,20 +175,8 @@ module Rack
       matches&.first
     end
 
-    ESCAPE_HTML = {
-      "&" => "&amp;",
-      "<" => "&lt;",
-      ">" => "&gt;",
-      "'" => "&#x27;",
-      '"' => "&quot;"
-    }
-
-    ESCAPE_HTML_PATTERN = Regexp.union(*ESCAPE_HTML.keys)
-
     # Escape ampersands, brackets and quotes to their HTML/XML entities.
-    def escape_html(string)
-      string.to_s.gsub(ESCAPE_HTML_PATTERN, ESCAPE_HTML)
-    end
+    define_method(:escape_html, CGI.method(:escapeHTML))
 
     def select_best_encoding(available_encodings, accept_encoding)
       # http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html
--- a/test/spec_utils.rb
+++ b/test/spec_utils.rb
@@ -475,17 +475,10 @@ describe Rack::Utils do
     Rack::Utils.escape_html("f&o").must_equal "f&amp;o"
     Rack::Utils.escape_html("f<o").must_equal "f&lt;o"
     Rack::Utils.escape_html("f>o").must_equal "f&gt;o"
-    Rack::Utils.escape_html("f'o").must_equal "f&#x27;o"
+    Rack::Utils.escape_html("f'o").must_equal "f&#39;o"
     Rack::Utils.escape_html('f"o').must_equal "f&quot;o"
     Rack::Utils.escape_html("<foo></foo>").must_equal "&lt;foo&gt;&lt;/foo&gt;"
-  end
-
-  it "escape html entities even on MRI when it's bugged" do
-    test_escape = lambda do
-      Rack::Utils.escape_html("\300<").must_equal "\300&lt;"
-    end
-
-    test_escape.must_raise ArgumentError
+    Rack::Utils.escape_html("\300<").must_equal "\300&lt;"
   end
 
   it "escape html entities in unicode strings" do


ioquatix:
```
define_method(:escape_html, CGI.method(:escape_html))
```

What about this? It should avoid 1 layer of indirection per call :)

JunichiIto:
@ioquatix Thank you for your suggestion. I applied it.

JunichiIto:
@ioquatix truffleruby test failed. Can you guess why??
https://github.com/rack/rack/actions/runs/5630066289/job/15255702711

```
  1) Error:
Rack::Utils#test_0028_escape html entities [&><'"/]:
NoMethodError: undefined method `escape_html' for Rack::Utils:Module
Did you mean?  escape
    /home/runner/work/rack/rack/test/spec_utils.rb:474:in `test_0028_escape html entities [&><'"/]'
```

ioquatix:
cc @eregon do you have any opinion on the best way to proxy this method and why my suggestion failed?

eregon:
I'll take a look.
I thought maybe it's `module_function` + `define_method` but that seems to work fine:
```
$ ruby -ve 'module M; module_function; define_method(:foo) { 42 }; end; p M.foo'
truffleruby 23.1.0-dev-b64cd1d4, like ruby 3.1.3, GraalVM CE Native [x86_64-linux]
42
```

eregon:
```
$ ruby -ve 'def Enumerable.bar; end; module M; module_function; define_method(:foo, Enumerable.method(:bar)); end; p M.foo'
ruby 3.1.3p185 (2022-11-24 revision 1a6b16756e) [x86_64-linux]
-e:1:in `define_method': can't bind singleton method to a different class (TypeError)
	from -e:1:in `<module:M>'
	from -e:1:in `<main>'
```
So it shouldn't be allowed even on CRuby it seems?

```
irb(main):001:0> CGI.method(:escapeHTML)
=> #<Method: #<Class:CGI>(CGI::Escape)#escapeHTML(_)>
```
Ah so because it's defined on CGI::Escape then it's allowed.

Probably the difference is related to https://github.com/oracle/truffleruby/issues/3134
Will try to repro next week.

jeremyevans (changes requested):
You cannot be sure that `cgi/escape` exists.  You need to guard the `require`, and if it raises `LoadError`, fallback to the current implementation.  You should also change the current implementation so it uses the same escape for `'`.

JunichiIto:
@eregon 
Thank you for your support.

@jeremyevans 
I updated the code per your request, but I don't know how to test the fallback case. Do you have any idea?

> You cannot be sure that `cgi/escape` exists. 

I think `cgi/escape` always exists because cgi is a default gem. Could you tell me when it happens?

```
$ ruby -v
ruby 3.2.2 (2023-03-30 revision e51014f9c0) [arm64-darwin22]
$ gem list cgi

*** LOCAL GEMS ***

cgi (default: 0.3.6)
$ gem uninstall cgi
Gem cgi-0.3.6 cannot be uninstalled because it is a default gem
```

jeremyevans:
> @eregon Thank you for your support.
> 
> @jeremyevans I updated the code per your request, but I don't know how to test the fallback case. Do you have any idea?
> 
> > You cannot be sure that `cgi/escape` exists.
> 
> I think `cgi/escape` always exists because cgi is a default gem. Could you tell me when it happens?
> 
> ```
> $ ruby -v
> ruby 3.2.2 (2023-03-30 revision e51014f9c0) [arm64-darwin22]
> $ gem list cgi
> 
> *** LOCAL GEMS ***
> 
> cgi (default: 0.3.6)
> $ gem uninstall cgi
> Gem cgi-0.3.6 cannot be uninstalled because it is a default gem
> ```

Apologies, cgi/escape is supported in stdlib cgi in all versions of Ruby that Rack supports (it was added in Ruby 2.3).  I thought it was added more recently, but I was mistaken.  So your original version was fine.  I'm sorry I didn't check that before requesting the changes and caused you to do unnecessary work.

JunichiIto:
@jeremyevans No problem! I reverted the code😊

JunichiIto:
FYI
I ran benchmark for define_method and def:

```ruby
define_method(:escape_html, CGI.method(:escapeHTML))

def escape_html_by_def(string)
  CGI.escapeHTML(string)
end
```

```ruby
require 'benchmark'
require 'rack/utils'

HTML_EXAMPLES = <<HTML.lines(chomp: true)
<p>It's a beautiful day.</p>
<a href="https://example.com/page?id=123&category=5">Link</a>
<button onclick="alert('Hello, world!')">Click me</button>
<input type="text" value="It's my text">
<div class='container'>Content goes here</div>
<span>5 &times; 10 = 50</span>
<img src='image.jpg' alt="Nature's beauty">
<p style='color: red;'>This text is red.</p>
<a href='javascript:alert("XSS attack!");'>Click here</a>
<input type='text' placeholder='Enter your name & email'>
HTML
targets = (HTML_EXAMPLES * 10).shuffle

num_iteration = 10_000

Benchmark.bm 15 do |r|
  r.report "by define_method" do
    num_iteration.times do
      targets.each do |string|
        Rack::Utils.escape_html(string)
      end
    end
  end

  r.report "by def" do
    num_iteration.times do
      targets.each do |string|
        Rack::Utils.escape_html_by_def(string)
      end
    end
  end
end
```

Here is the result:

```
$ ruby -v
ruby 3.2.2 (2023-03-30 revision e51014f9c0) [arm64-darwin22]

$ bundle exec ruby benchmark.rb
                      user     system      total        real
by define_method  0.122637   0.000460   0.123097 (  0.123102)
by def            0.133945   0.000311   0.134256 (  0.134256)
```

Certainly, using `define_method` is faster, but after executing it 1 million times (10 examples * 10 repeats * 10,000 iteration), the difference was only 11 milliseconds.

jeremyevans (approved):

JunichiIto:
@ioquatix I see. Thank you for your explanation!

eregon:
OK I found the bug: https://github.com/oracle/truffleruby/issues/3181
An easy workaround is to use:
```ruby
    define_method(:escape_html, CGI.method(:escapeHTML))
    module_function :escape_html
```
Or you can wait the truffleruby fix, up to you.

BTW there is also a `module_function :forwarded_values` in that file which is not needed.

eregon:
FWIW a disadvantage of `define_method(:escape_html, CGI.method(:escapeHTML))` is it's pretty hard to find the definition of it:
```
$ ruby -v -Ilib -e 'require "rack/utils"; p Rack::Utils.escape_html Object.new'
ruby 3.1.3p185 (2022-11-24 revision 1a6b16756e) [x86_64-linux]
-e:1:in `escapeHTML': no implicit conversion of Object into String (TypeError)
	from -e:1:in `<main>'
```
i.e. that backtrace does not include anything about `rack/utils.rb`.

vs if defined in the usual way:
```
ruby -v -Ilib -e 'require "rack/utils"; p Rack::Utils.escape_html Object.new'
ruby 3.1.3p185 (2022-11-24 revision 1a6b16756e) [x86_64-linux]
/home/eregon/code/rack/lib/rack/utils.rb:183:in `escapeHTML': no implicit conversion of Object into String (TypeError)
	from /home/eregon/code/rack/lib/rack/utils.rb:183:in `escape_html'
	from -e:1:in `<main>'
```

ioquatix:
I'm personally fine with either option. To be honest, I don't think I've actually ever seen code using `Rack` method, most are directly using `CGI`.

eregon:
FWIW that bug (https://github.com/oracle/truffleruby/issues/3181) is fixed in truffleruby-head now.

I would recommend to keep it simple, i.e., no `define_method` but explicit and clear delegation.

JunichiIto:
At this point, we have three implementation patterns. Which one should we choose? I like pattern 3 because it's very straightforward. Additionally, I think we should avoid unkind backtrace because it can easily confuse developers.

## pattern 1

```ruby
define_method(:escape_html, CGI.method(:escapeHTML))
```

- pros: fast
- cons: [requires TruffleRuby 23.1.0 (not released yet)](https://github.com/oracle/truffleruby/issues/3181#event-9945686531), unkind backtrace

## pattern 2

```ruby
define_method(:escape_html, CGI.method(:escapeHTML))
module_function :escape_html
```

- pros: fast, executable on current TruffleRuby
- cons: unkind backtrace

## pattern 3

```ruby
def escape_html(string)
  CGI.escapeHTML(string)
end
```

- pros: kind backtrace, executable on current TruffleRuby
- cons: a little bit slower

jeremyevans:
I vote for option 1 (i.e. commit as it is).

ioquatix:
@JunichiIto thanks so much for presenting all the options so comprehensively. I really appreciate your effort.

I'm fine with either option 1 or 3.

Since Jeremy also agrees option 1, let's merge it.

ioquatix (approved):

mperham:
This change breaks Sidekiq (and I imagine a lot of other code) because it makes one very large semantic change: nil is no longer a valid argument. Is this appropriate for a minor release?

ioquatix:
@mperham thanks for your report.

Unfortunately there were never any tests for non-`String` usage. It's a case of <https://www.hyrumslaw.com>.

I'm okay with changing the method to accept non-`String` arguments, restoring previous behaviour - there is a slight performance cost, the linked PR looks good to me and adds tests to cover this usage.

PR #2101, opened 2023-07-30
Fix inefficient assert pattern in Rack::Lint [2-2-stable]

skipkayhil:
Partial backport of 1970771c7e01d54cb631dae0bc7618e2561ad1c7

Differences from original commit:
- `#assert` is not deprecated
- `include Assertion` is not removed
- The assertion in `check_status` is the 2-2-stable version since the SPEC [change][1] to require statuses to be integers was merged before this
- `#dump` was added to the `REQUEST_METHOD` assertion [after][2] the original commit, so it was added here as well

This fixes an issue encountered when adding `Rack::Lint` to the Rails test suite. Rails puts a lazily evaluated class inside the request env that has the potential to raise when converted to a string. Since this assertion in `Rack::Lint` calls `#inspect` on `env` whether or not `env` is a `Hash`, the lazily evaluated class would raise at that point and prevent the rest of `Rack::Lint` from validating the conformity of the request.

By backporting this change, `#inspect` is now only called when the `LintError` would be raised which avoids the problem.

[1]: https://github.com/skipkayhil/rack/commit/ba25ade98fd34c873429e7a7dfa52df739962e4d
[2]: https://github.com/skipkayhil/rack/commit/b426cc224908ec6ed6eb8729325392b048215d88

---

~I started with the smallest change needed to fix the problem in the Rails test suite, but I'm happy to add more based on feedback~

### DIFF
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -40,7 +40,7 @@ module Rack
 
     def _call(env)
       ## It takes exactly one argument, the *environment*
-      assert("No env given") { env }
+      raise LintError, "No env given" unless env
       check_env env
 
       env[RACK_INPUT] = InputWrapper.new(env[RACK_INPUT])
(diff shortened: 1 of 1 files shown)


tenderlove (approved):

ioquatix:
@skipkayhil I've talked with everyone and we agree that if you want to implement this for all of `Rack::Lint` (backport or rework), we'd happily accept it. Otherwise, if it's just this one causing problems, that's also okay.

LMK.

skipkayhil:
Awesome! I'll go ahead and do them all

jeremyevans (approved):

Issue #2103, opened 2023-08-01
Is a boundary delimiter-only body an invalid request?


alpaca-tc:
Sending the following request in a chrome browser will generate a boundary request with only delimiter-line.

```javascript
const formData = new FormData();
const request = new Request('http://127.0.0.1:8080/', {
  method: 'POST',
  body: formData,
});
const response = fetch(request);
```

```
curl 'http://127.0.0.1:8080/' -H 'Content-Type: multipart/form-data; boundary=----WebKitFormBoundaryR1LC4tR6ayskIXJm' --data-raw $'------WebKitFormBoundaryR1LC4tR6ayskIXJm--\r\n'
```

This request causes `Rack::Multipart::EmptyContentError` in rack application.

As @jeremyevans mentions, it doesn't seem to match the RFC2046 specification.
https://github.com/rack/rack/issues/1603#issuecomment-683124204

> The body must then contain one or more body parts, each preceded by a boundary delimiter line, and the last one followed by a closing boundary delimiter line.

Unlike https://github.com/rack/rack/issues/1603#issuecomment-683101436, the body described in this issue is not completely empty and contains a delimiter.
I am creating this issue because I think there is a possibility that delimiter-only requests have not been considered before.

But as far as I read RFC2046, is it a bug in chromium?

jeremyevans:
For RFC 2046, it appears invalid:

> The body must then contain
> one or more body parts, each preceded by a boundary delimiter line,

If you consider the updated RFC (RFC 7578), it still appears invalid because RFC 7578 references section 5.1 of RFC 2046:

> The media type multipart/form-data follows the model of multipart
> MIME data streams as specified in [Section 5.1 of [RFC2046]](https://datatracker.ietf.org/doc/html/rfc2046#section-5.1); changes
> are noted in this document.

RFC 7578  does not state that zero parts are allowed as a change from RFC2046.

That being said, it doesn't appear to be just an issue with Chrome, I see identical behavior with Firefox.  So while not blessed by the RFCs, as a practical matter, I'm open to supporting this and having the multipart parser return an empty value (e.g. `{}`).

PR #2104, opened 2023-08-02
Return empty when parsing a multi-part POST with only one end delimiter.

alpaca-tc:
Fixed: #2103

Sending the following request in a browser generates a request with with only one end delimiter.

```javascript
const formData = new FormData();
const request = new Request('http://127.0.0.1:8080/', {
  method: 'POST',
  body: formData,
});
const response = fetch(request);
```

```
curl 'http://127.0.0.1:8080/' \
  -H 'Content-Type: multipart/form-data; boundary=----WebKitFormBoundaryR1LC4tR6ayskIXJm' \
  --data-raw $'------WebKitFormBoundaryR1LC4tR6ayskIXJm--\r\n'
```

This request is not compliant RFC7578, but is generated by major browsers such as FireFox and Chrome.
Supporting this request will cause the multipart parser to return an empty value.

### DIFF
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -209,6 +209,7 @@ module Rack
         @sbuf = StringScanner.new("".dup)
         @body_regex = /(?:#{EOL}|\A)--#{Regexp.quote(boundary)}(?:#{EOL}|--)/m
         @body_regex_at_end = /#{@body_regex}\z/m
+        @end_boundary_size = boundary.bytesize + 4 # (-- at start, -- at finish)
         @rx_max_size = boundary.bytesize + 6 # (\r\n-- at start, either \r\n or -- at finish)
         @head_regex = /(.*?#{EOL})#{EOL}/m
       end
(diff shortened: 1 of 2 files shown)


jeremyevans (approved):
Looks good, thanks for working on this!

ioquatix:
Can you please review and fix the failing test case:

```
  1) Failure:
Rack::Multipart#test_0007_raises for invalid data preceding the boundary [/home/runner/work/rack/rack/test/spec_multipart.rb:76]:
Rack::Multipart::EmptyContentError expected but nothing was raised.
```

Odd that it passes on Ruby <= 2.5 but fails after.

ioquatix:
Sorry, looks like another CI failure. Do you mind checking?

alpaca-tc:
I think the error of passing a String to `StringScanner#match?` has been fixed, I'm waiting for CI to re-run it. 🕺

ioquatix:
Thanks for your contribution and working diligently to resolve the CI issues.

PR #2109, opened 2023-08-15
Add `base64` to runtime dependency

koic:
This PR adds `base64` to runtime dependency to suppress the following Ruby 3.3's warning:

```console
$ ruby -wve 'require "rack/auth/basic"'
ruby 3.3.0dev (2023-08-14T15:48:39Z master 52837fcec2) [x86_64-darwin22]
/Users/koic/.rbenv/versions/3.3.0-dev/lib/ruby/gems/3.3.0+0/gems/rack-3.0.8/lib/rack/auth/basic.rb:5:
warning: base64 which will be not part of the default gems since Ruby 3.4.0
```

cf: https://github.com/rails/rails/pull/48907

### DIFF
--- a/rack.gemspec
+++ b/rack.gemspec
@@ -34,6 +34,8 @@ Gem::Specification.new do |s|
     "source_code_uri"   => "https://github.com/rack/rack"
   }
 
+  s.add_dependency 'base64'
+
   s.add_development_dependency 'minitest', "~> 5.0"
   s.add_development_dependency 'minitest-global_expectations'
 


ioquatix (approved):

jeremyevans (commented):
Looks like the base64 gem is supported back to Ruby 2.3, so this shouldn't break backwards compatibility.

However, I think it would be better to drop the base64 dependency completely, and switch to String#unpack1.  I'll submit a pull request for that.

ioquatix:
@koic thanks for your contribution. We decided to adopt https://github.com/rack/rack/pull/2110 instead.

PR #2110, opened 2023-08-15
Remove base64 dependency

jeremyevans:
base64 will be removed from the default gems in Ruby 3.4. Switch to using String#unpack1 instead.

### DIFF
--- a/lib/rack/auth/basic.rb
+++ b/lib/rack/auth/basic.rb
@@ -2,7 +2,6 @@
 
 require_relative 'abstract/handler'
 require_relative 'abstract/request'
-require 'base64'
 
 module Rack
   module Auth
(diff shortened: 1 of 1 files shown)


dentarg (approved):

ioquatix:
Any performance implications?

ioquatix (approved):

dentarg:
> Any performance implications?

Should be none? Maybe even faster? https://github.com/ruby/ruby/blob/efd611ceb661a230f121797079376cdb4e831951/lib/base64.rb#L61-L63

Nakilon:
I wish this stdlib was less cryptic. Why `m` vs `m0`? Why `unpack` vs `unpack1`? When I need base64 I should not try remembering these magick pack/unpack symbols. `Base64` was making it clearer.

PR #2113, opened 2023-08-24
Remove autoloads for constants no longer shipped with rack

jeremyevans:
Removes the following autoloads:

* File
* Handler
* Server
* Auth::Digest

Sort the remaining autoloads alphabetically.

Fixes #2112


ioquatix (approved):

p8:
Should the autoload of Handler be added to rackup? Or should that be removed from rackup altogether?

jeremyevans:
I think it is a good idea to remove `lib/rack/{handler,server}.rb` from rackup.

ioquatix:
They are compatibility shims but basically I agree with you.

p8:
I've created a PR for rackup: https://github.com/rack/rackup/pull/10

PR #2114, opened 2023-08-27
Fixed `Rack::Lint` incorrectly flagging `OPTIONS *`

doriantaylor:
`Rack::Lint` was incorrectly complaining about `PATH_INFO` not starting with a `/` pursuant to an `OPTIONS *` request (because it wouldn't). This patch fixes that.

### DIFF
--- a/test/spec_lint.rb
+++ b/test/spec_lint.rb
@@ -225,10 +225,13 @@ describe Rack::Lint do
     }.must_raise(Rack::Lint::LintError).
       message.must_match(/must start with/)
 
-    # XXX not actually sure what *should* happen in an OPTIONS * request.
-    lambda {
-      Rack::Lint.new(nil).call(env("REQUEST_METHOD" => "OPTIONS", "PATH_INFO" => ?*))
-    }.must_be_kind_of Proc
+    # A non-empty PATH_INFO starting with something other than / has
+    # implications for Rack::Request#path and methods downstream from
+    # it. Note that RFC3875 does not actually anticipate dealing with
+    # `OPTIONS *`; that should be considered a bug in the spec.
+    Rack::Lint.new(
+      lambda { |_| [200, {}, []] }
+    ).call(env("REQUEST_METHOD" => "OPTIONS", "PATH_INFO" => ?*)).first.must_equal 200
 
     lambda {
       Rack::Lint.new(nil).call(env("CONTENT_LENGTH" => "xcii"))
--- a/test/spec_lint.rb
+++ b/test/spec_lint.rb
@@ -225,6 +225,11 @@ describe Rack::Lint do
     }.must_raise(Rack::Lint::LintError).
       message.must_match(/must start with/)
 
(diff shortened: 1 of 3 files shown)


test/spec_lint.rb
```diff
@@ -225,6 +225,11 @@ def obj.fatal(*) end
     }.must_raise(Rack::Lint::LintError).
       message.must_match(/must start with/)
 
+    # XXX not actually sure what *should* happen in an OPTIONS * request.
```
ioquatix, line 228: What should we do here?

doriantaylor: If you do an `OPTIONS *` it seems to set `PATH_INFO` to `*` and not touch `SCRIPT_NAME` which I suppose is fine but there might be implications e.g. for things like `Rack::Request#path`.

test/spec_lint.rb
```diff
@@ -225,6 +225,11 @@ def obj.fatal(*) end
...
+    lambda {
+      Rack::Lint.new(nil).call(env("REQUEST_METHOD" => "OPTIONS", "PATH_INFO" => ?*))
+    }.must_be_kind_of Proc
```
jeremyevans, line 231: `must_be_kind_of Proc` does not actually test anything.  You don't need to wrap this in a lambda, just execute the code and check the result.  You do have to pass a valid Rack app to `Rack::Lint.new`, though.

jeremyevans (changes requested):
I'm OK with the code change, but the spec needs to be fixed, see inline comment.

test/spec_lint.rb
doriantaylor: Fair; it was late and I'm not familiar with minitest.

doriantaylor: New test is in place.

jeremyevans (approved):

ioquatix:
@doriantaylor can you please run `rake spec` to update the `SPEC.rdoc`.

ioquatix:
@doriantaylor can you please run rake spec to update the SPEC.rdoc.

@jeremyevans A lot of code will write path = File.join(env['PATH_INFO'], @root). This now has the potential to generate what I'd imagine was fairly unexpected paths. Especially considering * can be interpreted as a wildcard in some contexts. I wonder if it would be safer to put this into a different value, e.g. env['OPTIONS']. Not saying we should do it - just thinking out loud.

doriantaylor:
> @doriantaylor can you please run `rake spec` to update the `SPEC.rdoc`.

I did not touch `SPEC.rdoc`, I think @jeremyevans meant the test case in `spec_lint.rb`.

ioquatix:
@doriantaylor please run `rake spec` and it will update `SPEC.rdoc`, please add that file to the PR.

ioquatix:
On the basis that `PATH_INFO` can already have a value of `*` if `Rack::Lint` is not used, we can accept this PR. However, I still think we should have a follow up discussion about the right behavior here.

PR #2118, opened 2023-08-28
Better error handling in `MockRequest.env_for` w.r.t. input encodings.

ioquatix:
Previously, we would try to invoke `#set_encoding` on all input bodies. Let's be more specific and issue more valuable errors when dealing with non-binary input bodies.

Fixes <https://github.com/rack/rack/pull/2115>.

See also <https://github.com/rack/rack/pull/2116>.

### DIFF
--- a/lib/rack/mock_request.rb
+++ b/lib/rack/mock_request.rb
@@ -139,11 +139,15 @@ module Rack
         end
       end
 
-      if String === opts[:input]
-        rack_input = StringIO.new(opts[:input])
+      input = opts[:input]
+      if String === input
+        rack_input = StringIO.new(input)
         rack_input.set_encoding(Encoding::BINARY)
       else
-        rack_input = opts[:input]
+        if input.respond_to?(:encoding) && input.encoding != Encoding::BINARY
+          raise ArgumentError, "input must be binary"
+        end
+        rack_input = input
       end
 
       if rack_input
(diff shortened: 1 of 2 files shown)


lib/rack/mock_request.rb
```diff
@@ -139,14 +139,18 @@ def self.env_for(uri = "", opts = {})
...
-      if String === opts[:input]
-        rack_input = StringIO.new(opts[:input])
+      input = opts[:input]
+      if String === input
+        rack_input = StringIO.new(input)
+        rack_input.set_encoding(Encoding::BINARY)
       else
-        rack_input = opts[:input]
+        if input.respond_to?(:encoding) && input.encoding != Encoding::BINARY
+          raise ArgumentError, "input must be binary"
```
tenderlove, line 148: Should we raise an exception, or just emit a warning and still call `set_encoding`?  That way we're backwards compatible and folks can fix their code.

ioquatix: People often ignore warnings, and we won't release this until Rack 3.1 so I personally vote for error.

However, I don't have a strong opinion one way or the other. I think the amount of external test code failing because of this will be fairly small.

That being said, it is a breaking change. A warning wouldn't be. We'd still need to write:

```
warn "input encoding not binary", uplevel: 1
if input.respond_to?(:set_encoding)
  input.set_encoding(Encoding::BINARY)
else
  raise ArgumentError, "could not coerce input to binary encoding"
end
```

tenderlove (approved):
LGTM!

PR #2121, opened 2023-09-12
backport #2104

alpaca-tc:
Bug fix #2104 backported to 2-2stable because rack3.0 is not available in Rails 7.0.

Since the implementation of `lib/rack/multipart/parser.rb` is very different between rack 2.0 and 3.0, the backported changes have been corrected a bit.


ioquatix:
Can you please add a note to the changelog? Thanks!

alpaca-tc:
@ioquatix Added changelog 👍

PR #2131, opened 2023-10-21
Support for the `Partitioned` cookie attribute

flavio-b:
Adds the option to include the `Partitioned` cookie attribute, required by Chrome in some contexts.

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -289,11 +289,7 @@ module Rack
           else
             raise ArgumentError, "Invalid :same_site value: #{value[:same_site].inspect}"
           end
-        partitioned =
-          if value[:partitioned]
-            raise ArgumentError, 'Partitioned cookies must be set with `secure`' unless value[:secure]
-            "; partitioned"
-          end
+        partitioned = "; partitioned" if value[:partitioned]
         value = value[:value]
       else
         key = escape(key)
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -289,6 +289,11 @@ module Rack
           else
             raise ArgumentError, "Invalid :same_site value: #{value[:same_site].inspect}"
           end
+        partitioned =
+          if value[:partitioned]
+            raise ArgumentError, 'Partitioned cookies must be set with `secure`' unless value[:secure]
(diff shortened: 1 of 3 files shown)


ioquatix:
Thanks! Do you mind writing some tests for this?

flavio-b:
Just added a couple of tests. I hope this is good. I'm not too familiar with the repo.

lib/rack/utils.rb
```diff
@@ -289,6 +289,11 @@ def set_cookie_header(key, value)
           else
             raise ArgumentError, "Invalid :same_site value: #{value[:same_site].inspect}"
           end
+        partitioned =
+          if value[:partitioned]
+            raise ArgumentError, 'Partitioned cookies must be set with `secure`' unless value[:secure]
```
jeremyevans, line 294: `samesite=none` also requires secure cookies, but we don't have the same check.  We should be consistent in these cases.  Additionally, we should decide whether it is better to error, or to implicitly mark the cookie secure.

flavio-b: Good point. Seeing that browsers have some variability for cookie requirements, perhaps it's better to remove this check altogether and let the user have full control.

At second glance, `set_cookie_header` does read (and work) more like a lower level method that just puts together the string, so I would prefer **not** to set `secure` implicitly either. Maybe, requirement checks could fit in a future, optional, higher level class that calls `set_cookie_header` internally (i.e. `Cookie.new(**please_check_these_args).header`), but that's probably for another time.

Do you want me to remove the checks from the method and from the test?

jeremyevans: Probably best to remove the checks now.  If we decide to change the behavior both for Partitioned and SameSite=None, we can do that for both at some later point.

flavio-b:
Is there anything else I can do to help with this PR? I marked the previous conversation resolved but not sure if there's some other step.

mikegee:
Hi @flavio-b, Could you edit the pull request description since the error part changed?

flavio-b:
@mikegee, done!

ioquatix:
Can you please add a note to the change log and then we can merge.

ioquatix (approved):

PR #2132, opened 2023-10-24
Add missing headers constants, and replace strings with them

davidalejandroaguilar:
On https://github.com/rack/rack/issues/1592, it was decided to lowercase all response headers. Subsequently, constants for headers were defined on the `Rack` module, but there were some headers throughout the codebase that were still using hard-coded values.

This PR:

* Adds missing header constants
    * ACCEPT_ENCODING
    * ACCEPT_RANGES
    * ALLOW
    * CONTENT_ENCODING
    * CONTENT_RANGE
    * CONTENT_DISPOSITION
    * LAST_MODIFIED
    * LOCATION
    * VARY
    * X_CASCADE

* Uses these in places where they were hard-coded.


lib/rack/constants.rb
```diff
@@ -16,13 +16,22 @@ module Rack
...
+  CONTENT_LENGTH      = 'content-length'
+  CONTENT_RANGE       = 'content-range'
+  CONTENT_TYPE        = 'content-type'
+  ETAG                = 'etag'
+  EXPIRES             = 'expires'
+  LAST_MODIFIED       = 'last-modified'
+  SET_COOKIE          = 'set-cookie'
+  TRANSFER_ENCODING   = 'transfer-encoding'
+  VARY                = 'vary'
+  X_CASCADE           = 'x-cascade'
```
davidalejandroaguilar, line 34: Adding new ones in alphabetical order, but not reordering what was already out of order.

jeremyevans:
Can you explain the benefit of adding these constants? The cost is increased complexity and somewhat reduced performance.  Not much performance decrease in the constant reference vs. literal string case, but more in cases where there was no interpolation previously, and now there is interpolation.

davidalejandroaguilar:
Hi @jeremyevans, thank you for your time checking out this PR.

I'm just following the existing pattern I saw. i.e. I noticed there were existing constants for referencing these headers, and noticed usage of them on some files, and no usage on others.

If interpolation performance is a concern, perhaps we could remove referencing the constants on interpolation.

Regarding increased complexity, I actually think it's the opposite: it brings clarity/readability, because you're seeing uppercased text when reading the code. The usual benefit of using constants is also present: prevent typos and have a single place to change them (not that HTTP headers will change, but still).

If you rather think we should not use constants at all. Perhaps this PR should be about removing those constants and their usage throughout the codebase.

jeremyevans:
The issue with removing constants is that they are public API, so you cannot remove them without breaking backwards compatibility.  It was helpful to have constants when we changed from mixed case to lower case.  However, since it is very unlikely we will change the constant values again, adding new constants is unlikely to result in a significant benefit.

Consistency is generally a good thing, but certainly not the most important thing.

Regarding complexity, I think "#{HEADER}: bar" is more complex than "header: bar", but that is subjective.

Did you notice/fix any typos in the PR?  If not, doing this on the basis of preventing typos does not seem like a good reason.

davidalejandroaguilar:
@jeremyevans Fair enough. You're right about the existing constants being public API, because I was actually going to reference these new constants in another project.

If there's no desire of having more consistency, then I'll just close this PR, thanks again for your time!

PR #2133, opened 2023-11-01
Fix minor typo in SPEC.rdoc

mctaylorpants:
Hello! I found a super minor typo while reading the spec. There's a period `.` that should be a comma `,`.

### DIFF
--- a/SPEC.rdoc
+++ b/SPEC.rdoc
@@ -301,7 +301,7 @@ the body.
 
 The Enumerable Body must respond to +each+.
 It must only be called once.
-It must not be called after being closed.
+It must not be called after being closed,
 and must only yield String values.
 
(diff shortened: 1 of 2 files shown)


jeremyevans:
Thanks for the patch.  This looks good, but should be made to lib/rack/lint.rb.  SPEC.rdoc is generated from that file using `rake spec`.

mctaylorpants:
Ah cool, thanks for pointing that out 👍 @jeremyevans I've updated it in `lint.rb`, then ran `rake spec` locally to update the RDOC output.

jeremyevans (approved):

ioquatix (approved):

ioquatix:
Thanks :)

PR #2140, opened 2023-12-31
2.2.8

majioa:
- [x] fixed by backporting mocking rack request 
- [x] fixed rack response if buffered response isn't array

### DIFF
--- a/lib/rack/mock.rb
+++ b/lib/rack/mock.rb
@@ -216,11 +216,11 @@ module Rack
       body = body.body if body.respond_to?(:body)
 
       if body.respond_to?(:each)
-         body.each do |chunk|
-            buffer << chunk
-         end
+        body.each do |chunk|
+          buffer << chunk
+        end
       else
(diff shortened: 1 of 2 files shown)


lib/rack/mock.rb
```diff
@@ -212,8 +212,15 @@ def body
...
-        buffer << chunk
+      body = super
+      body = body.body if body.respond_to?(:body)
+
+      if body.respond_to?(:each)
+         body.each do |chunk|
+            buffer << chunk
+         end
+      else
+         buffer = body.to_s
```
mikegee, line 223: There's an extra space of indentation here.

```suggestion
      if body.respond_to?(:each)
        body.each do |chunk|
          buffer << chunk
        end
      else
        buffer = body.to_s
```

ioquatix:
Hello, do you mind clarifying what you are trying to fix here?

jeremyevans:
We should close this, none of the requested backport commits are security related.

PR #2145, opened 2024-01-21
Add an initializer to Rack::Headers to make it behave like Rack::Utils::HeaderHash

stefansundin:
Hello,

I am trying to update a middleware to rack 3, and I got this warning:

```
warning: Rack::Utils::HeaderHash is deprecated and will be removed in Rack 3.1, switch to Rack::Headers
```

This middleware initializes the headers by passing it in the initializer, like this:

```ruby
headers = Rack::Utils::HeaderHash.new(headers)
```

However, just swapping out `Rack::Utils::HeaderHash` with `Rack::Headers` like the warning suggests does not work, since the initializer is not there so it inherits the `Hash` initializer.

It may be intended but I figured I'd open this PR and see what everyone thinks. I basically just copied the initializer from `Rack::Utils::HeaderHash` except for one line.

Thanks!

### DIFF
--- a/lib/rack/headers.rb
+++ b/lib/rack/headers.rb
@@ -88,6 +88,11 @@ module Rack
       KNOWN_HEADERS[str] = KNOWN_HEADERS[downcased] = downcased
     end
 
+    def initialize(hash = {})
+      super()
+      hash.each { |k, v| self[k] = v }
+    end
+
     def self.[](*items)
       if items.length % 2 != 0
         if items.length == 1 && items.first.is_a?(Hash)


ioquatix:
@jeremyevans was this a deliberate design choice?

jeremyevans:
The current behavior was a deliberate decision choice.  This change would break compatibility with code that expects Rack::Headers.new to operate as Hash.new, and I do not think that is acceptable.

ioquatix:
Just FYI, IIRC, you can write:

```ruby
headers = Rack::Headers[headers]
```

stefansundin:
Wow, thanks! I didn't know about that syntax 🤯

Issue #2148, opened 2024-01-30
Multiple writes to Rack::Response result in incorrect content-length header in Rack 3


mattbrictson:
## Problem

After upgrading to Rack 3, I found that `Rack::Response` generates an incorrect content-length header when:

- it is constructed with an empty array for the body, and
- `write` is called multiple times

Here is a simple reproduction script:

```ruby
require "rack"

status = 200
headers = {}
response = Rack::Response.new([], status, headers)

# Write 11 total bytes
response.write "hello"
response.write "world"
response.write "!"

# Resulting content-length should be 11, but is 26
response.headers
# => {"content-length"=>"26"}
```

## Underlying cause

When `Rack::Response` is constructed with an empty array, it sets this internal state:

https://github.com/rack/rack/blob/0cd4d40a759f3e68f13c0467091cf836d086e9a1/lib/rack/response.rb#L82

This causes the value of `@length` to accumulate on every call to `Rack::Response#write`:

https://github.com/rack/rack/blob/0cd4d40a759f3e68f13c0467091cf836d086e9a1/lib/rack/response.rb#L317-L323

When `chunked?` is `false`, `@length` is incremented yet again, and then emitted as a content-length header:

https://github.com/rack/rack/blob/0cd4d40a759f3e68f13c0467091cf836d086e9a1/lib/rack/response.rb#L348-L350

The result is that the content-length value is too large, and the error gets larger on every call to `write`.

ioquatix:
Without digging into it, it appears that the latter `@length += chunk.bytesize` is incorrect. Do you think removing it is the correct course of action?

mattbrictson:
I'm actually not sure I understand the intention of the existing code, in particular why `@buffered` is set to `nil` in the constructor and lazily initialized later. Since I may be missing the point of `@buffered`, take what follows with a grain of salt.

The problem I see is that there is a missing assignment of `@buffered` in one of the conditional branches within `buffered_body!`.

Abbreviated, the code currently looks like this:

```ruby
if @buffered.nil?
  if @body.is_a?(Array)
    # ...
    # ⚠️ @buffered is not set in this branch
  elsif @body.respond_to?(:each)
    # ...
    @buffered = true
  else
    @buffered = false
  end
end
```

It seems like the point of this code was to lazily init `@buffered` to true or false, but currently that isn't happening in the first branch. So I think the fix is to add `@buffered = true`, like this:

```diff
--- lib/rack/response.rb
+++ lib/rack/response.rb
[...]

PR #2149, opened 2024-01-30
Clarify use of `@buffered` and only update `content-length` when `#finish`.

ioquatix:
See <https://github.com/rack/rack/issues/2148> for background context.

This change adjusts `Rack::Response` so that the internal `@length` variable is used until the request is finished, at which point `@headers['content-length']` is added if appropriate. This simplifies the internal logic and avoids any potential synchronisation issues.

### DIFF
--- a/lib/rack/response.rb
+++ b/lib/rack/response.rb
@@ -99,7 +99,7 @@ module Rack
       # The response body is an enumerable body and it is not allowed to have an entity body.
       @body.respond_to?(:each) && STATUS_WITH_NO_ENTITY_BODY[@status]
     end
-    
+
     # Generate a response array consistent with the requirements of the SPEC.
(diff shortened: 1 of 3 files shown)


SirJosh1987 (commented):
Stop making this so difficult for older generations

jeremyevans:
I think this can be closed. It was a bug fix for #2148, which was fixed by #2150

ioquatix:
I think we should merge this. It's a simplification to the way the response length is handled. Instead of updating `@length` and `@headers['content-length']` and trying to keep them in sync, this PR only updates `@length` and adds `@headers['content-length']` if appropriate when "finishing" the response, which makes the most sense to me. Repeatedly updating `@headers['content-length']` also seems pretty inefficient to me.

PR #2150, opened 2024-01-30
Fix content-length calcuation in Rack:Response#write

### LINKED ISSUE #2148: Multiple writes to Rack::Response result in incorrect content-length header in Rack 3

mattbrictson:
When `Rack::Response` is initialized with an Array, it properly sets its internal `@length` value and emitted content-length header on the first use of `write`. But every subsequent time `write` is called, it increments the length incorrectly, resulting in an incorrect content-length header.

This commit fixes the accumulation bug, and fixes/adds specs to properly test the scenario where `write` is used multiple times.

Fixes #2148

### DIFF
--- a/test/spec_response.rb
+++ b/test/spec_response.rb
@@ -408,7 +408,7 @@ describe Rack::Response do
     status.must_equal 404
   end
 
-  it "correctly updates content-type when writing when not initialized with body" do
+  it "correctly updates content-length when writing when initialized without body" do
     r = Rack::Response.new
     r.write('foo')
     r.write('bar')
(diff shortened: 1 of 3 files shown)


test/spec_response.rb
```diff
@@ -408,7 +408,7 @@ def object_with_each.each
     status.must_equal 404
   end
 
-  it "correctly updates content-type when writing when not initialized with body" do
+  it "correctly updates content-length when writing when initialized without body" do
```
mattbrictson, line 411: 🗒️ These specs incorrectly referred to `content-type`, but are actually testing `content-length`.

test/spec_response.rb
```diff
@@ -419,20 +419,24 @@ def object_with_each.each
     header['content-length'].must_equal '9'
   end
 
-  it "correctly updates content-type when writing when initialized with body" do
-    obj = Object.new
-    def obj.each
-      yield 'foo'
-      yield 'bar'
-    end
-    ["foobar", ["foo", "bar"], obj].each do
```
mattbrictson, line 428: 🗒️ This `.each` doesn't appear to be doing anything useful. There is no block param, so each iteration is identical. I rewrote this spec to simplify it.

test/spec_response.rb
```diff
@@ -419,20 +419,24 @@ def object_with_each.each
...
+    r = Rack::Response.new(["foo"])
+    r.write('bar')
+    r.write('baz')
+    _, header, body = r.finish
+    str = "".dup; body.each { |part| str << part }
+    str.must_equal "foobarbaz"
+    header['content-length'].must_equal '9'
+  end
+
+  it "correctly updates content-length when writing when initialized with String body" do
```
mattbrictson, line 432: 🗒️ Now there are 3 specs, covering all three branching conditions of the `Rack::Response` initializer:

1. `nil` body
2. Array body
3. String body

The Array case is what is fixed by this PR.

mattbrictson:
The jruby CI failures seem unrelated?

test/spec_response.rb
ioquatix: What about "Object that responds to `#each` body"?

mattbrictson: Good catch. I've added a test for that (see 578243c8) and then squashed to one commit.

ioquatix:
Thanks for your contribution. @jeremyevans as this is a bug fix should we consider back porting to 3.0?

jeremyevans:
Sure, this seems reasonable to backport.

PR #2159, opened 2024-02-25
Ignore Range header if served file is 0 bytes

zarqman:
Normally `Rack::Files` truncates a byte range to fit a file's contents--as long as at least some bytes are within range. However, an empty (0 byte) file is a special case since there are no requested bytes that can be within range. 

Previously, 0 byte files requested with a Range header always resulted in a 416 error. This PR changes 0 byte files to ignore the Range header and return the empty file.

Ignoring the Range header is allowed per [RFC 9110 §14.2](https://www.rfc-editor.org/rfc/rfc9110#section-14.2-7).

This improves compatibility with clients that speculatively request files with byte ranges. For example, Nginx's slice / slice_range feature causes all requests to utilize Range. 

Slightly ironically, this also results in a smaller response. (On this basis, it could even be argued to ignore Range for files below 50-100 bytes.)

#### Before
```bash
$ curl -i -H 'Range: bytes=0-1023' http://localhost:3000/favicon.ico
HTTP/1.1 416 Range Not Satisfiable
content-type: image/vnd.microsoft.icon
x-cascade: pass
content-range: bytes */0
Content-Length: 25

Byte range unsatisfiable
```

#### After
```bash
$ curl -i -H 'Range: bytes=0-1023' http://localhost:3000/favicon.ico
HTTP/1.1 200 OK
last-modified: Tue, 15 Jan 2019 21:51:01 GMT
content-type: image/vnd.microsoft.icon
Content-Length: 0
```

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -388,6 +388,8 @@ module Rack
 
     def get_byte_ranges(http_range, size)
       # See <http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.35>
+      # Ignore Range when file size is 0 to avoid a 416 error.
+      return nil if size.zero?
       return nil unless http_range && http_range =~ /bytes=([^;]+)/
       ranges = []
       $1.split(/,\s*/).each do |range_spec|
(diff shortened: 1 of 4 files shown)


ioquatix:
This seems reasonable to me, @jeremyevans @tenderlove do you have any thoughts or potential security issues?

jeremyevans (approved):
Seems fine to me.

lib/rack/utils.rb
```diff
@@ -388,6 +388,7 @@ def byte_ranges(env, size)
 
     def get_byte_ranges(http_range, size)
       # See <http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.35>
+      return nil if size.zero?
```
ioquatix, line 391: I would like to prefix this line with a comment that explains the rationale of ignoring the size if it is zero.

ioquatix (changes requested):
This looks good to me. Can you please:

- [ ] Add an entry to the changelog.

Thanks!

zarqman:
Changelog and comment both added.

ioquatix (approved):

Issue #2163, opened 2024-03-13
Request for release - EOFError on empty multipart request 


JoeDupuis:
Hi, 
I was investigating a [Rails issue](https://github.com/rails/rails/issues/51292). 

Turns out there is already a fix for it on both `2-2-stable` and `main` (but not `3-0-stable`)
https://github.com/rack/rack/pull/2104 (main)
https://github.com/rack/rack/pull/2121 (2-2-stable)

These were excluded from the last releases.  I was wondering if we could cut a release of both 2-2 and 3-0 including this fix and/or what is the policy around releases. Apologies if this is indicated somewhere and I missed it.

As a temporary fix for the Rails issue, I'll tell the user to point their Gemfile to the Rack Github repo. 

I can open a PR to port the #2104 to `3-0-stable` if needed. 
Thank you!

ioquatix:
Can you please open a PR to 3-0-stable and I'll handle the releases once merged.

JoeDupuis:
Awesome! Thank you! Here's the PR #2164

ioquatix:
This is already released as `v2.2.8.1` and I've just released `v3.0.10` which includes the backport.

JoeDupuis:
@ioquatix `v2.2.8.1` was released after the backport, but it looks like it excluded fix. I tried to compare the files from a fresh rubygems download to the [2-0-stable backport PR](https://github.com/rack/rack/pull/2121) and it doesn't look like it made the release.

ioquatix:
Thanks for that clarification. Yeah that's super weird because the commit history says otherwise. Okay, I'll cut a new release.

ioquatix:
Okay, I just released v2.2.9 with this change.

PR #2164, opened 2024-03-20
Backport 2104 3-0-stable

JoeDupuis:
Bug fix #2104 backported to 3-0-stable (to cut a release).

Follow up from #2163

### DIFF
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -213,6 +213,7 @@ module Rack
 
         @sbuf = StringScanner.new("".dup)
         @body_regex = /(?:#{EOL}|\A)--#{Regexp.quote(boundary)}(?:#{EOL}|--)/m
+        @end_boundary_size = boundary.bytesize + 4 # (-- at start, -- at finish)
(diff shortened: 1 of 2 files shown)


ioquatix (approved):

lib/rack/multipart/parser.rb
```diff
@@ -213,6 +213,8 @@ def initialize(boundary, tempfile, bufsize, query_parser)
 
         @sbuf = StringScanner.new("".dup)
         @body_regex = /(?:#{EOL}|\A)--#{Regexp.quote(boundary)}(?:#{EOL}|--)/m
+        @body_regex_at_end = /#{@body_regex}\z/m
```
jeremyevans, line 216: I don't think `@body_regex_at_end` is used by this code, so it should be added.

jeremyevans (changes requested):
One minor change requested, then this should be good.

lib/rack/multipart/parser.rb
JoeDupuis: Oops, it came with the cherry pick by accident. I removed it.

jeremyevans (approved):

ioquatix:
Thanks everyone!

PR #2165, opened 2024-04-01
Support parsing multiple files for one form field

timcraft:
This implements support for parsing multipart requests that contain multiple files with the same name.

From [RFC7578 Section 4.3](https://www.rfc-editor.org/rfc/rfc7578#section-4.3):

> To match widely deployed implementations, multiple files MUST be sent by supplying each file in a separate part but all with the same "name" parameter.

This isn't strictly backwards compatible because it affects `Rack::Utils.parse_nested_query`, however I would argue the existing behaviour there is a bug on the basis that a) it's not a *feature* to have data silently dropped/overwritten so not something many are likely to be relying upon, and b) `parse_nested_query` doesn't currently match the behaviour of `parse_query` which does the correct thing:

    $ irb -r rack
    irb(main):001> Rack::Utils.parse_query('foo=1&foo=2')
    => {"foo"=>["1", "2"]}
    irb(main):002> Rack::Utils.parse_nested_query('foo=1&foo=2')
    => {"foo"=>"2"}

You can test out uploading multiple files from browsers using something like this:

```ruby
require 'sinatra'
require 'json'

get '/' do
  <<-HTML
    <form action="/" method="post" enctype="multipart/form-data">
      <input type="file" name="field" multiple>

      <button type="submit">Upload</button>
    </form>
  HTML
end

post '/' do
  <<-HTML
    <pre>#{JSON.pretty_generate(request.POST)}</pre>
  HTML
end
```

### DIFF
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -132,6 +132,12 @@ module Rack
       if after == ''
         if k == '[]' && depth != 0
           return [v]
+        elsif !params[k].nil? && !v.nil? && depth == 0
+          if params[k].is_a?(Array)
+            params[k] << v
+          else
+            params[k] = [params[k], v]
+          end
         else
           params[k] = v
         end
--- a/test/multipart/multiple_files
+++ b/test/multipart/multiple_files
@@ -1,0 +1,16 @@
+--AaB03x
+content-disposition: form-data; name="files"; filename="file1.txt"
+content-type: text/plain
+
+foo
+--AaB03x
+content-disposition: form-data; name="files"; filename="file2.txt"
+content-type: text/plain
+
+bar
+--AaB03x
+content-disposition: form-data; name="files"; filename="file3.txt"
+content-type: text/plain
+
+quux
+--AaB03x--
--- a/test/spec_multipart.rb
+++ b/test/spec_multipart.rb
@@ -518,6 +518,18 @@ describe Rack::Multipart do
     params["files"].size.must_equal 252
   end
 
+  it "parses multiple files with same name" do
+    env = Rack::MockRequest.env_for("/", multipart_fixture(:multiple_files))
+    params = Rack::Multipart.parse_multipart(env)
+    params["files"].must_be_instance_of Array
+    params["files"][0][:filename].must_equal "file1.txt"
+    params["files"][0][:tempfile].read.must_equal "foo"
+    params["files"][1][:filename].must_equal "file2.txt"
+    params["files"][1][:tempfile].read.must_equal "bar"
+    params["files"][2][:filename].must_equal "file3.txt"
+    params["files"][2][:tempfile].read.must_equal "quux"
+  end
+
   it "parses IE multipart upload and cleans up the filename" do
     env = Rack::MockRequest.env_for("/", multipart_fixture(:ie))
     params = Rack::Multipart.parse_multipart(env)
--- a/test/spec_utils.rb
+++ b/test/spec_utils.rb
@@ -150,7 +150,7 @@ describe Rack::Utils do
       must_equal "foo" => "\"bar\""
 
     Rack::Utils.parse_nested_query("foo=bar&foo=quux").
-      must_equal "foo" => "quux"
+      must_equal "foo" => ["bar", "quux"]
     Rack::Utils.parse_nested_query("foo&foo=").
       must_equal "foo" => ""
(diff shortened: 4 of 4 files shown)


jeremyevans:
Doesn't Rack already support this if the name ends with `[]`?  Then Rack will have the parameter store an array of uploaded files.  My advice would be to use `field[]` as the name, then you can use that name to upload as many files as you want.

Personally, the `parse_query` behavior seems wrong to me.  If we are going to change things in a backwards-incompatible manner, we should make `parse_query` behave like `parse_nested_query`, not vice-versa.  However, I don't think it's worth breaking backwards compatibility here.

timcraft:
> Doesn't Rack already support this if the name ends with `[]`?

Yes but that's a Rails invention, not part of the spec.

> My advice would be to use `field[]` as the name

What about when it's not possible to change the name of the field on the client side, or you need to support field names with and without the `[]` suffix?

> Personally, the `parse_query` behavior seems wrong to me.

Browsers, the multipart-post gem, and the cgi gem all support fields with the same name. For example:

    $ irb -r cgi
    irb(main):001> CGI.parse('foo=1&foo=2')
    => {"foo"=>["1", "2"]}

It's also implemented that way in other languages. For example:

    $ python3
    ...
    >>> from urllib.parse import parse_qs
    >>> parse_qs('foo=1&foo=2')
    {'foo': ['1', '2']}

    $ bun repl
    ...
    > querystring.parse('foo=1&foo=2')
    Object <[Object: null prototype] {}> { foo: [ '1', '2' ] }

jeremyevans:
> > Doesn't Rack already support this if the name ends with `[]`?
> 
> Yes but that's a Rails invention, not part of the spec.

While it originally came from Rails, the behavior has been part of Rack for I think over 15 years.

> > My advice would be to use `field[]` as the name
> 
> What about when it's not possible to change the name of the field on the client side, or you need to support field names with and without the `[]` suffix?

The main advantage of the current behavior is that if you specify `[]` on the client side, you are sure the server will interpret as an array, and if you do not specify `[]`, you are sure the server will interpret as a string/file.  If you try to automatically convert from string to array at first duplicate key, then all code that deals with the parameter must be prepared to handle both string/file and array.  That's a large backwards compatibility breakage across the entire ecosystem, which would affect almost all applications that use `parse_nested_query` for parsing (which I would guess is over 99% of Ruby web applications).

If we were going to support this behavior, it would have to be opt-in and off by default.  Considering that the inability to control the client side parameter name is fairly rare, I'm not sure it is worth supporting at all.

matthewd:
> What about when it's not possible to change the name of the field on the client side, or you need to support field names with and without the [] suffix?

Then you don't want the [custom, built-atop-the-underlying-protocol] Rack nested params parsing algorithm; you want raw access to the parameters as supplied. That's something Rack provides at a couple of levels, including via `parse_query` -- but `parse_nested_query` is a different thing.

timcraft:
> Then you don't want the [custom, built-atop-the-underlying-protocol] Rack nested params parsing algorithm; you want raw access to the parameters as supplied. That's something Rack provides at a couple of levels, including via `parse_query` -- but `parse_nested_query` is a different thing.

Yes... I don't want the nested params parsing algorithm but that's what `Rack::Multipart.parse_multipart` and `Rack::Multipart::Parser.parse` use internally. Maybe I'm looking in the wrong place but I can't see how to get raw access to the parameters without the nested params parsing algorithm? Using `Utils.parse_query` or `Utils.parse_nested_query` directly wouldn't work, because those aren't for multipart/form-data.

jeremyevans:
I'm against this change.  I think you can use `Rack::Multipart.parse_multipart(env, Rack::Multipart::ParamList)` to get a list of pairs, and use that to implement the behavior you want.

tenderlove:
Yes, I agree we shouldn't change this method.

@timcraft if @jeremyevans's suggestion doesn't work for you, I'm open to adding new API that will allow you to do what you need. I don't think we should mess with the existing parameter parsing code though.

timcraft:
@jeremyevans @tenderlove Looks like `Rack::Multipart.parse_multipart(env, Rack::Multipart::ParamList)` works nicely, as long as that's officially supported I can depend on that once it's available so no need to add a new API.

Thank you 🙏

tenderlove:
@timcraft yep, it's officially supported! Please feel free to rely on that API 😄

PR #2166, opened 2024-04-02
Remove unused `ostruct` require

Earlopain:
OpenStruct usage was removed in #2004 but this was missed.

In Ruby 3.4 this is likely to trigger a warning, see https://github.com/ruby/ruby/pull/10428 / https://bugs.ruby-lang.org/issues/20309


ioquatix (approved):

ioquatix:
Thanks!

PR #2167, opened 2024-04-16
Support for the Partitioned cookie attribute

dchitragupt-conga:
MR

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -259,12 +259,13 @@ module Rack
           else
             raise ArgumentError, "Invalid SameSite value: #{value[:same_site].inspect}"
           end
+        partitioned = "; partitioned" if value[:partitioned]  
         value = value[:value]
       end
       value = [value] unless Array === value
 
       cookie = "#{escape(key)}=#{value.map { |v| escape v }.join('&')}#{domain}" \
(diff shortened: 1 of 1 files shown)


jeremyevans:
I'm not in favor of backporting this feature, but if we do, we certainly shouldn't backport to 2.0-stable, which is unmaintained. 2.2-stable is in security maintenance only mode, so it wouldn't make sense to backport to that either.  Considering we've made breaking changes to 3.0-stable, I suppose you could argue for backporting to that, but that's as far as I would go, assuming we plan on backporting the feature at all.

dchitragupt-conga:
We are using older versions of rack and can’t update to latest version. Due to this we are failing features.


Get Outlook for iOS<https://aka.ms/o0ukef>

Increase revenue lifecycle certainty
Leave a review
|Support|Twitter|
LinkedIn
________________________________

mikegee:
> We are using older versions of rack and can’t update to latest version.

That is certainly unfortunate. Please understand that your particular application cannot dictate how past versions of Rack are maintained. See https://github.com/rack/rack/blob/main/SECURITY.md for Rack's maintenance policy.

jeremyevans:
I'm closing this.  There is no way we are going to apply patches to 2.0-stable at this point.

PR #2174, opened 2024-04-24
Remove deprecated base64 in favour of pack/unpack (2-2-stable)

adam12:
The base64 library is external as of Ruby 3.4 and is no longer available by default in Ruby. The pack/unpack methods support encoding as base64 and can be used in its place.

String#unpack.first was preferred here to support Ruby 2.3, which didn't yet have String#unpack1 introduced in Ruby 2.4.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@ All notable changes to this project will be documented in this file. For info on
 
 ## Unreleased
 
+- Remove `Base64` use in favour of `pack`/`unpack` ([#2174](https://github.com/rack/rack/pull/2174), [@adam12](https://github.com/adam12))
+
 ## [2.2.9] - 2023-03-21
 
 - Return empty when parsing a multi-part POST with only one end delimiter. ([#2104](https://github.com/rack/rack/pull/2104), [@alpaca-tc])
--- a/lib/rack/auth/basic.rb
+++ b/lib/rack/auth/basic.rb
@@ -2,7 +2,6 @@
 
 require_relative 'abstract/handler'
 require_relative 'abstract/request'
-require 'base64'
 
 module Rack
   module Auth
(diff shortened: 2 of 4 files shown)


ioquatix (approved):

jeremyevans:
I don't think we should backport this to 2.2 before backporting it to 3.0.  It's also not a security issue, so it shouldn't be considered for backport to 2.2, IMO.

adam12:
To preface: I'm OK if this merges or if it doesn't.

I actually hesitated to submit a PR to an older release branch, but I looked for the official stance on backports and didn't see anything (maybe I missed it?). If it's not acceptable for backport, we can go ahead and close the PR.

I have 2 applications with a very sticky Rack 2 dependency that I am in the process of upgrading, but it's been a bit of a challenge and I still want to continue with Ruby upgrades. I'll end up maintaining my own fork of 2-2-stable if I have to.

simi (approved):

adam12:
I just noticed that the `base64` dependency was added to `rack-session` gemspec, which could be done in this case too (if we wanted parity).

ioquatix:
I suppose we either need to do one of those options, or this release will simply stop working with the next major release of Ruby, no?

jeremyevans:
It won't stop working, it will just require that users of Ruby 3.4+ and Rack 2.2 list base64 explicitly in their Gemfile.

ioquatix:
Right, so there is a workaround, which is users would need to explicitly add the missing gem to their application level gemfile. i.e. the Rack gem has an implicit dependency and the user can compensate for that.

jeremyevans:
I'm still against backporting this to 2.2, as it is not a security issue, and users can work around it by adding base64 to Gemfile.

adam12:
I was just thinking about this again yesterday. I'm _assuming_ this isn't the first time that this will come up, so perhaps less intrusively we just update the README?

jeremyevans:
I not opposed to adding this to the README.  The README could stand to be updated, as it still states "Rack v3.0.0 was recently released", and that was over 18 months ago.

tenderlove:
> I'll submit a PR for README changes and then we can close this.

+1

Thank you for your work!

PR #2175, opened 2024-04-25
Add 3.0.9.1 to CHANGELOG

dentarg:
Noticed it was missing.

I did not take on to backfill all the other 2.x releases fixing this, as there are other regular 2.x releases missing. I may (or may not) take a stab at that later.


ioquatix (approved):

ioquatix:
Thanks!

PR #2179, opened 2024-05-10
Remove Ruby 1.9 references from the spec

Earlopain:
A string doesn't respond to each anymore so that comment is redundant with the one a bit above.


jeremyevans (approved):

ioquatix:
Thanks - good to keep things relevant!

PR #2181, opened 2024-05-30
Fine tune `PATH_INFO` validation in `Rack::Lint`.

### LINKED ISSUE #2117: How to handle (legal vis-à-vis HTTP) `OPTIONS *` requests
> On the basis that `PATH_INFO` can already have a value of `*` if `Rack::Lint` is not used, we can accept this PR. However, I still think we should have a follow up discussion about the right behavior here.

_Originally posted by @ioquatix in https://github.com/rack/rack/issues/2114#issuecomment-1694867816_

This is a thorny issue because while [RFC9110](https://datatracker.ietf.org/doc/html/rfc9110#name-options)  asserts (along with its predecessors dating all the way back to [RFC2068](https://www.rfc-editor.org/rfc/rfc2068#section-9.2)) that an `OPTIONS *` request is a perfectly valid mechanism for obtaining global information about a Web server's capabilities. However, the CGI spec, [RFC3875](https://datatracker.ietf.org/doc/html/rfc3875#section-4.1.5), upon which the [Rack specification](/rack/rack/blob/main/SPEC.rdoc) is based, does not seem to acknowledge this validity, and thus does not specify how it should be implemented. An informal poll of Web server implementations suggests that the prevailing behaviour is to put the `*` of `OPTIONS *` into the `PATH_INFO` variable, while leaving `SCRIPT_NAME` untouched. What makes this issue thorny is that Rack applications (that are running without `Rack::Lint`) that do _not_ check the request method could be in for a rude surprise if they try to use the `PATH_INFO` of an `OPTIONS *` under the assumption that it it is either empty or starts with a `/`. A particularly obvious hazard is if the `*` character gets treated as a [`glob(3)`](https://linux.die.net/man/3/glob) pattern.

As it stands for Rack, the method `Rack::Request#path` in particular does not account for the possibility that `PATH_INFO` may contain anything other than an empty string or one that starts with `/`, per RFC3875 (again, it is my opinion that RFC3875 itself is in error). This situation propagates to other methods in `Rack::Request`, such as `#fullpath`, `#url`, and everything downstream from there.

One solution (within `Rack::Request` at least) would be, in the call to `#path`, to check if the request method was `OPTIONS` and the `PATH_INFO` was `*` and faking up a result in that case (e.g. using `/` in lieu). That would mitigate the potential for harm at least for those applications using `Rack::Request`. Those who do not would have to be warned to test the request method before dispatching any URI behaviour (a sensible practice in any case).

I should also footnote that the `OPTIONS` method is not the only one where the request-URI is something other than that which starts with a `/`; the `CONNECT` method for one (which one could conceivably do something useful with in Rack using connection hijacking) and the `PRI *` request for upgrading to HTTP/2 are other specimens, off the top of my head. The point here is it's not _just_ the special case of `OPTIONS *`.

ioquatix:
This makes `Rack::Lint` more specific to the actual HTTP semantics, as described in <https://www.rfc-editor.org/rfc/rfc9112.html#name-request-target>.

Fixes #2117.

### DIFF
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -14,7 +14,7 @@ module Rack
     REQUEST_PATH_ORIGIN_FORM = /\A\/[^?]*\??[^#]*\z/
     REQUEST_PATH_ABSOLUTE_FORM = /\A#{URI::regexp}\z/
     REQUEST_PATH_AUTHORITY_FORM = /\A(.*?)(:\d*)\z/
-    REQUEST_PATH_ASTERISK_FORM = /\A\*\z/
+    REQUEST_PATH_ASTERISK_FORM = '*'
 
     def initialize(app)
       @app = app
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -354,22 +354,22 @@ module Rack
         if env.include?(PATH_INFO)
           case env[PATH_INFO]
           when REQUEST_PATH_ASTERISK_FORM
-            ##   - Only <tt>OPTIONS</tt> requests may have <tt>PATH_INFO</tt> set to * (asterisk-form).
+            ##   * Only <tt>OPTIONS</tt> requests may have <tt>PATH_INFO</tt> set to <tt>*</tt> (asterisk-form).
             unless env[REQUEST_METHOD] == OPTIONS
               raise LintError, "Only OPTIONS requests may have PATH_INFO set to * (asterisk-form)"
             end
           when REQUEST_PATH_AUTHORITY_FORM
-            ##   - <tt>CONNECT</tt> requests must have <tt>PATH_INFO</tt> set to an authority (authority-form).
+            ##   * Only <tt>CONNECT</tt> requests may have <tt>PATH_INFO</tt> set to an authority (authority-form).
             unless env[REQUEST_METHOD] == CONNECT
               raise LintError, "Only CONNECT requests may have PATH_INFO set to an authority (authority-form)"
             end
           when REQUEST_PATH_ABSOLUTE_FORM
-            ##   - <tt>CONNECT</tt> and <tt>OPTIONS</tt> requests must not have <tt>PATH_INFO</tt> set to a URI (absolute-form).
+            ##   * <tt>CONNECT</tt> and <tt>OPTIONS</tt> requests must not have <tt>PATH_INFO</tt> set to a URI (absolute-form).
             if env[REQUEST_METHOD] == CONNECT || env[REQUEST_METHOD] == OPTIONS
               raise LintError, "CONNECT and OPTIONS requests must not have PATH_INFO set to a URI (absolute-form)"
             end
           when REQUEST_PATH_ORIGIN_FORM
-            ##   - Otherwise, <tt>PATH_INFO</tt> must start with a <tt>/</tt> (origin-form).
+            ##   * Otherwise, <tt>PATH_INFO</tt> must start with a <tt>/</tt> (origin-form).
             unless env[PATH_INFO].start_with?("/")
               raise LintError, "PATH_INFO must start with a '/' (origin-form)"
             end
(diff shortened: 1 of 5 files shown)


lib/rack/lint.rb
```diff
@@ -10,6 +11,11 @@ module Rack
   # responses according to the Rack spec.
 
   class Lint
+    REQUEST_PATH_ORIGIN_FORM = /\A\/[^?]*\??[^#]*\z/
+    REQUEST_PATH_ABSOLUTE_FORM = /\A#{URI::regexp}\z/
+    REQUEST_PATH_AUTHORITY_FORM = /\A(.*?)(:\d*)\z/
+    REQUEST_PATH_ASTERISK_FORM = /\A\*\z/
```
jeremyevans, line 17: Can we use `'*'` here?

jeremyevans (commented):
I'm OK with the general idea of validating path info.

Do the exists tests fully cover the new Lint code?  If not, can you please add covering tests?

tenderlove (approved):

lib/rack/lint.rb
ioquatix: Sure.

ioquatix:
I'm going to merge this, but accept there is a bit of risk, as while it's only the linter, it's also going to be a lot more picky about the request target handling. The updated logic may reject requests that were previously being accepted (and vice versa). While all the tests are passing (with minimal reorganisation), I'm still a little concerned that there is a moderate amount of risk if someone was using `Rack::Lint` in production.

If we see any issues, I'll be happy to update the logic to suit.

Bo98:
The SPEC currently states that `PATH_INFO` can be an empty string: 
https://github.com/rack/rack/blob/38b178e9ce3b30453cd5340a92093128bfb4fa36/SPEC.rdoc#L37-L45

AFAICT this can now no longer be the case so does the SPEC need updating?

ioquatix:
AFAIK, "if the request URL targets  the application root and does not have a trailing slash" is not a valid request. It would be `GET  HTTP/1.1` which is invalid.

If we can cite the RFC, it would be a stronger argument to remove "This may be an empty string" - e.g. the CGI RFC might have guidance.

So, probably the SPEC language should be updated.

Bo98:
> AFAIK, "if the request URL targets the application root and does not have a trailing slash" is not a valid request. It would be `GET  HTTP/1.1` which is invalid.

Yeah, I agree that makes sense.

> the CGI RFC might have guidance.

It seems the SPEC is copied word-for-word from PEP 333 which in turn is based from the CGI RFC (3875) which have that empty-string clause.

```
      PATH_INFO = "" | ( "/" path )
      path      = lsegment *( "/" lsegment )
      lsegment  = *lchar
      lchar     = <any TEXT or CTL except "/">
```

But then goes on to say:

```
   [...] The server MAY
   impose restrictions and limitations on what values it permits for
   PATH_INFO, and MAY reject the request with an error if it encounters
   any values considered objectionable. [...]
```

So in the end I guess anything goes?

ioquatix:
More context <https://github.com/sinatra/sinatra/issues/2014>.

Bo98:
Yeah I was testing Sinatra with Rack 3.1 and noticed that. Seems like it could be test-specific behaviour though.

I had a read of CGI specs this afternoon to understand the reasoning for all this a bit more and I think you are right. My understanding is indeed that `PATH_INFO` is allowed to be an empty string because a complete request could be contained within `SCRIPT_NAME`. So at least one of `SCRIPT_NAME` and `PATH_INFO` must be non-empty but never both.

I'm by no means an expert of how Rack actually handles all this though so I'm not sure how the SPEC etc should be worded. CGI is not something I have thought about for many years now so it took a while to even remember what I'm talking about.

ioquatix:
A reasonable change would just be to allow an empty string. Later on, we could consider concatenating `path = env[SCRIPT_NAME] + env[PATH_INFO]` and validating `path` (assuming it's semantically the full request path).

PR #2182, opened 2024-05-30
Skip Rack::Deflater tests on JRuby

jeremyevans:
This starting breaking in JRuby 9.4.3.0.  It's been almost a year since then and the problem is not fixed.  Let's skip the tests on JRuby until JRuby fixes them.

### DIFF
--- a/test/spec_deflater.rb
+++ b/test/spec_deflater.rb
@@ -11,6 +11,10 @@ separate_testing do
 end
 
 describe Rack::Deflater do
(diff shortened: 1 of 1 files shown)


ioquatix:
@headius is there any ETA on when this will be fixed?

ioquatix (approved):

headius:
Oh sorry, I didn't realize this was still broken.

headius:
Fixed by https://github.com/jruby/jruby/pull/8281.

Issue #1879, opened 2022-04-26
Pass a request object to middleware rather than a hash


tenderlove:
One problem I see in many middleware is that they "cast" the env hash to a Rack request object, do some manipulations, then pass then cast it back to a hash and pass it on.

For example:

```ruby
def call(env)
  req = Rack::Request.new(env)

  # do stuff with req

  @app.call(req.env)
end
```

I would really like it if we could stop allocating request objects all the time, but since every middleware is passed a hash it's something that happens quite frequently.

I would like to do 2 things:

1. Make a middleware that casts the env hash to a request object, then passes the request object to the next middleware
2. Make the request object quack enough like a hash that existing legacy middleware can use the request object as if it was the env hash

Step 1 could be as simple as this:

```ruby
class CastsToRequest
  def initialize(app)
    @app = app
  end
  def call(env); @app.call(Rack::Request.new(env)); end
end
```

If some middleware want to take advantage of this "pre allocated" request object, then they would require that `CastsToRequest` is in the middleware stack above them.  Those middleware would no longer have to cast back and forth between hash and request object.

If the `Rack::Request` object quacks enough like a hash, then users could add legacy middleware without changing anything, though they could be upgraded to take advantage of the request object.

I think this idea could work because it should be OK for any middleware to say "yes I work with any Rack compatible web server, but I must be run *after* the `CastsToRequest` middleware.  I think this would also unlock the ability for web servers to start passing their own request objects to middleware (as long as they quack the same as `Rack::Request`).

jeremyevans:
As I mentioned in https://github.com/rack/rack/issues/1878#issuecomment-1110034922, I'm against this proposal. Completely breaking backwards compatibility at this stage in rack's lifecycle makes is a very bad trade off, IMO.

If given a choice between a request object and an h2 style `env`, I think the request object approach is better, as it offers advantages and doesn't just rearrange the deck chairs. That being said, if we want to use a request object, we should specify (in SPEC) each method the request object should respond to, and how each method should operate.  We should not specify that the request object should operate like `Rack::Request`.  For one, the  `Rack::Request` API footprint is gigantic.  Second, one of Rack's basic principals is that you do not need to require rack the library in order to have a rack compliant server or application.

tenderlove:
@jeremyevans what part of this proposal are you against?  The contract with the webserver doesn't change, it just lets downstream rack applications avoid allocating more request objects (should they want to opt-in, and they must opt-in by having the `CastsToRequest` app upstream)

[...]

PR #2183, opened 2024-05-31
Remove deprecated methods and mark `values_at` as deprecated.

ioquatix:
See <https://github.com/rack/rack/issues/1879> for context.


jeremyevans (approved):
I disagree with your reasons for doing this, as I've explained in #1879.  However, I think the change itself is fine.  `values_at` is not a good name because there is no indication it operates on params (I would expect it to operate on env given the name).

PR #2184, opened 2024-05-31
Don't use deprecated `URI.regexp`

Earlopain:
Followup to #2181

> rack/lib/rack/lint.rb:15: warning: URI.regexp is obsolete


ioquatix (approved):

ioquatix:
Thanks, I did't realise it was deprecated!

PR #2186, opened 2024-06-04
Revert "Skip Rack::Deflater tests on JRuby"

headius:
Reverts rack/rack#2182

This will be green once jruby/jruby#8281 has been merged into jruby-head builds.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -21,7 +21,7 @@ jobs:
           - '3.1'
           - '3.2'
           - '3.3'
-          - jruby
+          - jruby-head
           - truffleruby-head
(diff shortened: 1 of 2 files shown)


ioquatix:
Thanks @headius should this be passing now?

headius:
@ioquatix It will be when the mentioned PR is in a jruby-head build retrieved by setup-ruby. There should have been a snapshot build of JRuby last night, but I'm not sure when it will get picked up.

I don't have privileges to restart CI runs so I can't give it a try here.

ioquatix (approved):

ioquatix:
Thanks!

Issue #1865, opened 2022-04-09
Does round-tripping cookie names work as expected?


ioquatix:
While working through the cookie utils methods adding documentation and examples, I noticed the following:

```
Rack::Utils.set_cookie_header("a&b", ":)")
# => "a%26b=%3A%29"

cookies = Rack::Utils.parse_cookies_header("a%26b=%3A%29")
# => {"a%26b"=>":)"}

Rack::Utils.set_cookie_header("a%26b", ":)")
# => "a%2526b=%3A%29"
```

URL encoding the cookie name on the way to the client but not on the way back from the client means that if a server application tries to set the same cookie based on the incoming name, it will have it's percent escapes encoded again, never reaching a fixed point/symmetry.

Not sure if this is the desired behaviour or not. I know there is a security issue with decoding cookies which are URL encoded. Just wondering if we've made the right trade off here.

Looking at the relevant RFCs:

```
         CTL            =  %x00-1F / %x7F
                                ; controls
```

```
       token          = 1*<any CHAR except CTLs or separators>
       separators     = "(" | ")" | "<" | ">" | "@"
                      | "," | ";" | ":" | "\" | <">
                      | "/" | "[" | "]" | "?" | "="
                      | "{" | "}" | SP | HT
```

```
 set-cookie-header = "Set-Cookie:" SP set-cookie-string
 set-cookie-string = cookie-pair *( ";" SP cookie-av )
 cookie-pair       = cookie-name "=" cookie-value
 cookie-name       = token
```

Maybe rather than general URL encoded cookie key names, we should either:

1. Reject any cookie key that doesn't match the `token` pattern, or
2. Have a `token` specific encoded form which encodes and decodes **only** `CTLs or separators` as defined by the parser.

Sorry to get pedantic about this but I'm a little concerned that URL encoding cookie keys is probably not the right behaviour and maybe we should follow the RFCs here more closely.

ioquatix:
So thinking this out a bit more, the following are valid cookie key characters:

```
!#$%&'*+-.^_`|~
```

And should not be encoded, but they are:

```
Rack::Utils.set_cookie_header("!#$%&'*+-.^_`|~", ":)")
# => "%21%23%24%25%26%27*%2B-.%5E_%60%7C%7E=%3A%29"
```

This means that if someone writes:

```
headers = {'set-cookie' => '~=~'}
Rack::Utils.set_cookie_header!(headers, '~', '~')
=> ["~=~", "%7E=%7E"]
```

[...]

PR #2188, opened 2024-06-05
Symmetric escape/unescape cookie keys except when it would be a security issue.

ioquatix:
Possible fix for <https://github.com/rack/rack/issues/1865>.

Other possible options:

- Only escape/unescape a subset of characters that cause problems, e.g. "( ) < > @ , ; : \ " / [ ] ? = { }" (i.e. nothing that would cause ambiguity w.r.t. `__Host-` or `__Secure-`).
- Don't escape/unescape keys at all, and instead throw an error when they contain invalid characters.

I personally vote for the latter, but recognise that this PR is more compatible. I think escaping cookie keys is likely to cause confusion as in JavaScript (`document.cookies`) they won't show up in an expected way.

Therefore, I suggest we adopt this PR as a first step, and I'll create a follow up PR which removes escape/unescape entirely - which maybe we can release in a later version of Rack.

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -211,6 +211,24 @@ module Rack
       (encoding_candidates & available_encodings)[0]
     end
 
+    # A <cookie-name> can be any US-ASCII characters, except control characters, spaces, or tabs. It also must not contain a separator character like the following: ( ) < > @ , ; : \ " / [ ] ? = { }.
+    # Cookies names starting with __Secure- (dash is part of the prefix) must be set with the secure flag from a secure page (HTTPS).
+    # Cookies with names starting with __Host- must be set with the secure flag, must be from a secure page (HTTPS), must not have a domain specified (and therefore aren't sent to subdomains) and the path must be /
+    PARSE_COOKIES_SECURE_PREFIX = /\A(__Secure-)|(__Host)/i.freeze
+
+    def unescape_cookie_key(key)
+      if key =~ PARSE_COOKIES_SECURE_PREFIX
+        return unescape(key)
+      else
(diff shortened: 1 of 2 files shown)


lib/rack/utils.rb
```diff
@@ -211,6 +211,24 @@ def select_best_encoding(available_encodings, accept_encoding)
       (encoding_candidates & available_encodings)[0]
     end
 
+    # A <cookie-name> can be any US-ASCII characters, except control characters, spaces, or tabs. It also must not contain a separator character like the following: ( ) < > @ , ; : \ " / [ ] ? = { }.
+    # Cookies names starting with __Secure- (dash is part of the prefix) must be set with the secure flag from a secure page (HTTPS).
+    # Cookies with names starting with __Host- must be set with the secure flag, must be from a secure page (HTTPS), must not have a domain specified (and therefore aren't sent to subdomains) and the path must be /
```
jeremyevans, line 216: It's odd to have this documentation, but not actually check that these things are true.

jeremyevans (commented):
I'll leave the security review of this to @tenderlove.  It does seem odd to me to only have special handling for `__Secure`/`__Host` cookies, though.

ioquatix:
> It does seem odd to me to only have special handling for __Secure/__Host cookies, though.

Those are the cookies that require special handling due to our unescaping logic.

lib/rack/utils.rb
ioquatix: We do a little more validation in the follow up PR.

ioquatix:
Closing in favour of <https://github.com/rack/rack/pull/2191>.

Issue #1674, opened 2020-06-17
2.2.3 breaks for cookies with spaces in their names


granth:
I think the short answer is, "don't use spaces in your cookie names", but for _historical reasons_ that's something I ran into.

When a cookie with a space in its name is stored, the name is escaped in classic CGI style, with the space (` `) changed to a plus (`+`). In 2.2.2, the plus was unescaped back to a space when the header was parsed. In 2.2.3 that is no longer true (5ccca4722668083732ea2d35c56565fcc25312f8).

This is especially awkward when using `cookies.signed` in Rails. I got around it by copying `cookies[name_with_plus]` to `cookies[name_with_space]` before calling `cookies.signed[name_with_space]` as before.

Anyway, I'm not sure if this is worth fixing. Hopefully this issue will help anyone else who might run into the same thing.

fschwahn:
I'm seeing this issue too - it not only affects cookies with spaces in the name, but also symbols which are escaped, eg. colon:

```ruby
# in controller
cookies.signed["Test::Cookie"] = "ABC"

# later
cookies.signed["Test::Cookie"] # => nil
cookies.signed["Test%3A%3ACookie"] # => "ABC
```

After regressing rack to 2.2.2 it works as expected.

jeremyevans:
This looks like a Rails issue, not a Rack issue. There was a change in Rack 2.2.3 to fix a security issue, and not decoding the cookie names is necessary to fix the security issue.  `cookies.signed` is not part of the Rack API, so I'm guessing the bug is in Rails and should be fixed there.  @tenderlove @eileencodes @matthewd @rafaelfranca any thoughts?

granth:
@jeremyevans The problem surfaced in Rails, but Rack is what actually changed. Bumping the `rack` gem one patch version broke things.

I think the question is whether the change in behaviour I saw is part of the security fix or an inadvertent side effect.

jeremyevans:
It's part of the security fix as far as I can tell.  @tenderlove should be able to confirm.

tenderlove:
Yes, it's part of the security fix. We can't do URI decoding on keys because people can use it to abuse middleware that think they have a secure cookie.  I don't want to revert it because it is a security fix.  I think it would be OK for Rails to encode and decode spaces as a work-around, but I'd rather we don't do any encoding / decoding for these.

I'm really sorry this broke things for you. 🙇🏻‍♂️

granth:
👍 Thanks for confirming!

[...]

PR #2189, opened 2024-06-05
Don't escape/unescape cookie keys.

ioquatix:
This is a follow up to <https://github.com/rack/rack/pull/2188> which removes all cookie key escaping/unescaping.

This is better aligned with how JavaScript's `document.cookies` interface works, and I think will reduce complexity and potential confusion. It may introduce some compatibility issues, but the scenario we are breaking is already explicitly documented here <https://github.com/rack/rack/blob/8d7c029ed11372b3ae56a76f1fb7c9bbccd2f6c2/lib/rack/utils.rb#L250-L251> and users who hit this limitation are already operating in a grey area (not following the requirements of the documentation, and the cookies being set won't have the same name when read back).

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -211,6 +211,24 @@ module Rack
       (encoding_candidates & available_encodings)[0]
     end
 
+    # A <cookie-name> can be any US-ASCII characters, except control characters, spaces, or tabs. It also must not contain a separator character like the following: ( ) < > @ , ; : \ " / [ ] ? = { }.
+    # Cookies names starting with __Secure- (dash is part of the prefix) must be set with the secure flag from a secure page (HTTPS).
+    # Cookies with names starting with __Host- must be set with the secure flag, must be from a secure page (HTTPS), must not have a domain specified (and therefore aren't sent to subdomains) and the path must be /
+    PARSE_COOKIES_SECURE_PREFIX = /\A(__Secure-)|(__Host)/i.freeze
+
+    def unescape_cookie_key(key)
+      if key =~ PARSE_COOKIES_SECURE_PREFIX
+        return unescape(key)
+      else
+        unescaped_key = unescape(key)
+        if unescaped_key =~ PARSE_COOKIES_SECURE_PREFIX
+          return key
+        else
+          return unescaped_key
+        end
+      end
+    end
+
     # :call-seq:
     #   parse_cookies_header(value) -> hash
     #
@@ -227,6 +245,8 @@ module Rack
       value.split(/; */n).each_with_object({}) do |cookie, cookies|
         next if cookie.empty?
         key, value = cookie.split('=', 2)
+        
+        key = unescape_cookie_key(key)
         cookies[key] = (unescape(value) rescue value) unless cookies.key?(key)
       end
     end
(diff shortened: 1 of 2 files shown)


jeremyevans (commented):
I recommend @tenderlove review for security.

Assuming no security issues, due to backwards compatibility issues, this should probably wait until Rack 4.

The commit message mentions unescaping, but nothing in the patch removes unescaping.

ioquatix:
@jeremyevans I'm not sure I completely follow your feedback.

#2188 which is not merged, implements symmetric escape/unescape except where it's a known security issue.

This PR adds a 2nd commit on top of that, which removes escape/unescape entirely, for the reasons outlined. I don't think the compatibility issue will be huge (no more than others we accept in minor releases), but I am okay with merging #2188 and as a follow up we can consider merging this PR.

jeremyevans:
I guess I expect more compatibility issues than you do.  My expectation is there are users relying on the current behavior, and rejecting cookie keys previously not rejected will break their apps.  When we made the change to not to unescape cookies due to the security issues, we broke a significant amount of apps, and this change seems as likely to cause issues.

Maybe the PR description makes sense as a commit message to the second commit.  But in terms of the whole PR, this doesn't deal with unescaping.

ioquatix:
> I guess I expect more compatibility issues than you do. My expectation is there are users relying on the current behavior, and rejecting cookie keys previously not rejected will break their apps. When we made the change to not to unescape cookies due to the security issues, we broke a significant amount of apps, and this change seems as likely to cause issues.

I agree, it may. Do you have any reference to the previous issues we encountered?

jeremyevans:
Looks like cookie keys have always been escaped, dating to the initial import (7ed819ad7).

Unescaping of keys was the cause of CVE-2020-8184 (1f5763de6a9fe515ff84992b343d63c88104654c).

I think that's why we are in the situation we are currently in, where keys are escaped when serializing but not unescaped when deserializing. Users do try to use cookie key names that require escaping (e.g. #1674). Since Rack's introduction, that worked fine because Rack did the escaping. If Rack stops doing escaping, and instead raises an exception, such code will break.

Note that the validation regexp used here I don't think is complete.  It doesn't validate `%` escapes, for example, so it will treat invalid escapes in cookie keys as valid.

ioquatix:
Escaping cookie keys is a rack invention, it's not standard and there is no standard interpretation of percent escapes for the cookie key except for the raw strings. The escaped cookie keys will show up in `document.cookies` in the escaped form. According to TOKEN, `%` is not a special character, so there is no reason it can't be part of the cookie key.

Example server:

```ruby
run do |env|
	[200, {'set-cookie' => '%66oo=bar'}, ["Hello World!"]]
end
```

Browser shows:

image

And on subsequent requests, we see:

```
 "HTTP_COOKIE"=>"%66oo=bar",
```

Rack should never have escaped cookie keys - it's non-standard and the un-escaped name won't match subsequent `cookie` header keys.

jeremyevans:
Thanks for that information.  I'm OK with a deprecation warning for keys that need escaping in Rack 3.1, and removal in Rack 3.2.

ioquatix:
That seems reasonable to me.

ioquatix:
Closing in favour of <https://github.com/rack/rack/pull/2191>.

PR #2191, opened 2024-06-06
Warn on invalid cookie keys.

ioquatix:
As discussed in <https://github.com/rack/rack/pull/2189>, we will warn on invalid cookie keys, and mark the escaping behaviour as deprecated.


jeremyevans:
Note that we should update the changelog for this, but it can be done in a later commit if you want.

ioquatix:
I'll update the PR tomorrow and merge it :)

PR #2198, opened 2024-06-11
Bump minor version.

ioquatix:
See <https://github.com/rack/rack/milestone/26?closed=1> and the change log for more details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to this project will be documented in this file. For info on how to format all future additions to this file please reference [Keep A Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## Unreleased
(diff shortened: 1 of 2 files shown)


Earlopain:
Did someone have a chance to check out https://github.com/rails/rails/issues/52066#issuecomment-2156427472? I don't think an issue was made.

ioquatix:
Nope, this is the first time we saw it.

IIUC, application code should not be specifying `transfer-encoding: chunked`. If your intention is to force streaming, either use a streaming response OR don't implement `to_ary` on the response body.

Earlopain:
I'll just link your reply there, I'm not qualified to make any sort of assessment on this. Thanks (:

ioquatix:
Appreciate it!

ioquatix:
@Earlopain appreciate your attention to detail and cross-references!

PR #2199, opened 2024-06-11
Introduce `Rack::VERSION`.

ioquatix:


Earlopain:
Is it worth to deprecate `Rack::RELEASE` already? It would be nice to keep both around for a while to give time for applications to drop older rack versions so they don't also have to case-when around this in addition to everything else.

ioquatix:
I'm okay with considering different options (e.g. do nothing, or not deprecating it), however it will be between 6-12 months before this is actually released, and then there will be about 6-12 months for another release which actually removes it.

jeremyevans (changes requested):
I agree with @Earlopain , that we don't need to deprecate `RELEASE`.

test/spec_version.rb
```diff
@@ -9,7 +9,7 @@
 describe Rack do
   describe 'version' do
     it 'is a version string' do
-      Rack::RELEASE.must_match(/\d+\.\d+\.\d+/)
+      Rack::VERSION.must_match(/\d+\.\d+\.\d+/)
```
jeremyevans, line 12: Please keep both constants here.  I would also add `Rack.release`

jeremyevans (changes requested):
One small change requested, then this should be good to merge.

test/spec_version.rb
ioquatix: Good idea.

jeremyevans (approved):

PR #2202, opened 2024-06-11
Handle `nil` in `Rack::Utils.escape_html`

Earlopain:
This previously worked but unintentionally changed in #2099. It previously called `to_s` which made this accept almost anything (but especially `nil`)

This is of course a bit slower than the current implementation. Here's a benchmark:

```rb
require 'benchmark/ips'
require 'cgi'

# Generated by ChatGPT
HTML_EXAMPLES = <<HTML.lines(chomp: true)
<p>It's a beautiful day.</p>
<a href="https://example.com/page?id=123&category=5">Link</a>
<button onclick="alert('Hello, world!')">Click me</button>
<input type="text" value="It's my text">
<div class='container'>Content goes here</div>
<span>5 &times; 10 = 50</span>
<img src='image.jpg' alt="Nature's beauty">
<p style='color: red;'>This text is red.</p>
<a href='javascript:alert("XSS attack!");'>Click here</a>
<input type='text' placeholder='Enter your name & email'>
HTML
targets = (HTML_EXAMPLES * 10).shuffle

num_iteration = 10_000

module Utils
  module_function

  ESCAPE_HTML = {
    "&" => "&amp;",
    "<" => "&lt;",
    ">" => "&gt;",
    "'" => "&#x27;",
    '"' => "&quot;"
  }

  ESCAPE_HTML_PATTERN = Regexp.union(*ESCAPE_HTML.keys)

  def escape_html_old(string)
    string.to_s.gsub(ESCAPE_HTML_PATTERN, ESCAPE_HTML)
  end

  define_method(:escape_html_define_method, CGI.method(:escapeHTML))

  def escape_html_or(string)
    CGI.escapeHTML(string || "")
  end

  def escape_html_to_s(string)
    CGI.escapeHTML(string.to_s)
  end
end

Benchmark.ips do |x|
  x.report "old" do
    num_iteration.times do
      targets.each do |string|
        Utils.escape_html_old(string)
      end
    end
  end

  x.report "define_method" do
    num_iteration.times do
      targets.each do |string|
        Utils.escape_html_define_method(string)
      end
    end
  end

  x.report "to_s" do
    num_iteration.times do
      targets.each do |string|
        Utils.escape_html_to_s(string)
      end
    end
  end

  x.report "or" do
    num_iteration.times do
      targets.each do |string|
        Utils.escape_html_or(string)
      end
    end
  end

  x.compare!
end
```

Results for my machine:
```
ruby 3.3.1 (2024-04-23 revision c56cd86388) [x86_64-linux]
                 old     1.000 i/100ms
       define_method     1.000 i/100ms
                to_s     1.000 i/100ms
                  or     1.000 i/100ms
                 old      0.353 (± 0.0%) i/s -      2.000 in   5.664835s
       define_method      8.733 (± 0.0%) i/s -     44.000 in   5.040108s
                to_s      6.946 (± 0.0%) i/s -     35.000 in   5.040787s
                  or      7.438 (± 0.0%) i/s -     38.000 in   5.111461s

Comparison:
       define_method:        8.7 i/s
                  or:        7.4 i/s - 1.17x  slower
                to_s:        6.9 i/s - 1.26x  slower
                 old:        0.4 i/s - 24.74x  slower
```

With YJIT:
```
ruby 3.3.1 (2024-04-23 revision c56cd86388) +YJIT [x86_64-linux]
                 old     1.000 i/100ms
       define_method     1.000 i/100ms
                to_s     1.000 i/100ms
                  or     1.000 i/100ms
                 old      0.364 (± 0.0%) i/s -      2.000 in   5.494063s
       define_method     10.271 (± 0.0%) i/s -     52.000 in   5.064240s
                to_s      9.982 (± 0.0%) i/s -     50.000 in   5.009911s
                  or     10.031 (± 0.0%) i/s -     51.000 in   5.087295s

Comparison:
       define_method:       10.3 i/s
                  or:       10.0 i/s - 1.02x  slower
                to_s:       10.0 i/s - 1.03x  slower
                 old:        0.4 i/s - 28.21x  slower
```

With YJIT there is barely any difference between these variants which I think is pretty great. `| ""` would be an option and it does handle `nil` but not non-strings.


ioquatix (approved):
I'm okay with this, but I'd like @jeremyevans's opinion on it.

lib/rack/utils.rb
```diff
@@ -177,7 +177,9 @@ def best_q_match(q_value_header, available_mimes)
     end
 
     # Escape ampersands, brackets and quotes to their HTML/XML entities.
-    define_method(:escape_html, CGI.method(:escapeHTML))
+    def escape_html(string)
```
jeremyevans, line 180: We could switch from `CGI.escapeHTML` to `ERB::Escape.html_escape` if available and use define_method.  That should be faster.

jeremyevans (changes requested):
I would like to avoid the slowdown, and I think I have a way (at least for some Ruby versions).

lib/rack/utils.rb
Earlopain: That's a great idea and does seem to be faster. It doesn't let me do `define_method(:escape_html_define_method_erb, ERB::Escape.method(:html_escape))` like that though:
> `define_method': can't bind singleton method to a different class (TypeError)

Would you know a different incantation to make this work? Definiting a delegation method incurs the ~1.01 performance penalty from the original PR.

Earlopain: I got it, will push after doing some benchmarking

ioquatix:
Are you happy that `ERB::Escape` is in the same ball-park performance-wise as our current `CGI.escape_html`?

lib/rack/utils.rb
```diff
@@ -7,6 +7,7 @@
 require 'tempfile'
 require 'time'
 require 'cgi/escape'
+require 'erb'
```
jeremyevans, line 10: No need to require `cgi/escape` if `ERB::Escape.html_escape` is available.

jeremyevans (changes requested):
Looks good, one small change requested.

lib/rack/utils.rb
mperham, line 10: You'll likely need to declare a dep on `erb` in the gemspec.

lib/rack/utils.rb
```diff
@@ -176,8 +177,16 @@ def best_q_match(q_value_header, available_mimes)
       matches&.first
     end
 
-    # Escape ampersands, brackets and quotes to their HTML/XML entities.
-    define_method(:escape_html, CGI.method(:escapeHTML))
+    # Introduced in ERB 4.0. ERB::Escape is an alias for ERB::Utils which
+    # doesn't get monkey-patched by rails
+    if defined?(ERB::Escape) && ERB::Escape.instance_method(:html_escape)
```
mperham, line 182: This rats nest of code to shave a % of runtime seems like a bad tradeoff. It makes the code less readable and maintainable.

Earlopain:
I was about to post some numbers (:

`ERB::Escape` is of course still slower than plain `CGI` but it gets some nice wins when the text doesn't need escaping.
Both codeblocks contain two benchmark runs, one where the input is html, and one where the input is just boring text. I think, the numbers for this are pretty nice. I also included `to_s` (my initial proposal) and the current `define_method` version.

<details><summary>Benchmark script</summary>
<p>

```rb
require 'benchmark/ips'
require 'cgi'
require 'erb'

# Generated by ChatGPT
HTML_EXAMPLES = <<HTML.lines(chomp: true)
<p>It's a beautiful day.</p>
<a href="https://example.com/page?id=123&category=5">Link</a>
<button onclick="alert('Hello, world!')">Click me</button>
<input type="text" value="It's my text">
<div class='container'>Content goes here</div>
<span>5 &times; 10 = 50</span>
<img src='image.jpg' alt="Nature's beauty">
<p style='color: red;'>This text is red.</p>
<a href='javascript:alert("XSS attack!");'>Click here</a>
<input type='text' placeholder='Enter your name & email'>
HTML
HTML_NO_ESCAPING = <<HTML.lines(chomp: true)
Lorem ipsum dolor sit amet, consectetur adipiscing elit,
sed do eiusmod tempor incididunt ut labore et dolore magna
aliqua. Ut enim ad minim veniam, quis nostrud exercitation
ullamco laboris nisi ut aliquip ex ea commodo consequat.
Duis aute irure dolor in reprehenderit in voluptate velit
esse cillum dolore eu fugiat nulla pariatur. Excepteur sint
occaecat cupidatat non proident, sunt in culpa qui officia
deserunt mollit anim id est laborum. Lorem ipsum dolor sit
amet, consectetur adipiscing elit, sed do eiusmod tempor
incididunt ut labore et dolore magna aliqua. Ut enim ad
HTML
targets = (HTML_EXAMPLES * 10).shuffle
target_no_escaping = (HTML_NO_ESCAPING * 10).shuffle

num_iteration = 10_000

module Utils
  module_function

  define_method(:escape_html_current, CGI.method(:escapeHTML))
  define_method(:escape_html_new, ERB::Escape.instance_method(:html_escape))

  def escape_html_new_fallback(string)
    CGI.escapeHTML(string.to_s)
  end

  def escape_html_to_s(string)
    CGI.escapeHTML(string.to_s)
  end

end

Benchmark.ips do |x|
  x.report "current" do
    num_iteration.times do
      targets.each do |string|
        Utils.escape_html_current(string)
      end
    end
  end

  x.report "to_s" do
    num_iteration.times do
      targets.each do |string|
        Utils.escape_html_to_s(string)
      end
    end
  end

  x.report "erb" do
    num_iteration.times do
      targets.each do |string|
        Utils.escape_html_new(string)
      end
    end
  end

  x.report "erb fallback" do
    num_iteration.times do
      targets.each do |string|
        Utils.escape_html_new_fallback(string)
      end
    end
  end

  x.compare!
end;


puts "TEXT ONLY"

Benchmark.ips do |x|
  x.report "current" do
    num_iteration.times do
      target_no_escaping.each do |string|
        Utils.escape_html_current(string)
      end
    end
  end

  x.report "to_s" do
    num_iteration.times do
      target_no_escaping.each do |string|
        Utils.escape_html_to_s(string)
      end
    end
  end

  x.report "erb" do
    num_iteration.times do
      target_no_escaping.each do |string|
        Utils.escape_html_new(string)
      end
    end
  end

  x.report "erb fallback" do
    num_iteration.times do
      target_no_escaping.each do |string|
        Utils.escape_html_new_fallback(string)
      end
    end
  end

  x.compare!
end
```

</p>
</details> 

```
ruby 3.3.1 (2024-04-23 revision c56cd86388) [x86_64-linux]
             current     1.000 i/100ms
                to_s     1.000 i/100ms
                 erb     1.000 i/100ms
        erb fallback     1.000 i/100ms
             current      8.741 (± 0.0%) i/s -     44.000 in   5.034073s
                to_s      7.041 (± 0.0%) i/s -     36.000 in   5.112778s
                 erb      8.751 (± 0.0%) i/s -     44.000 in   5.027805s
        erb fallback      7.040 (± 0.0%) i/s -     36.000 in   5.114978s

Comparison:
                 erb:        8.8 i/s
             current:        8.7 i/s - 1.00x  slower
                to_s:        7.0 i/s - 1.24x  slower
        erb fallback:        7.0 i/s - 1.24x  slower

TEXT ONLY
ruby 3.3.1 (2024-04-23 revision c56cd86388) [x86_64-linux]
             current     1.000 i/100ms
                to_s     1.000 i/100ms
                 erb     1.000 i/100ms
        erb fallback     1.000 i/100ms
             current     10.404 (± 0.0%) i/s -     53.000 in   5.094110s
                to_s      8.340 (± 0.0%) i/s -     42.000 in   5.035849s
                 erb     14.559 (± 0.0%) i/s -     73.000 in   5.014386s
        erb fallback      8.326 (± 0.0%) i/s -     42.000 in   5.044305s

Comparison:
                 erb:       14.6 i/s
             current:       10.4 i/s - 1.40x  slower
                to_s:        8.3 i/s - 1.75x  slower
        erb fallback:        8.3 i/s - 1.75x  slower
```

## YJIT

```
ruby 3.3.1 (2024-04-23 revision c56cd86388) +YJIT [x86_64-linux]
             current     1.000 i/100ms
                to_s     1.000 i/100ms
                 erb     1.000 i/100ms
        erb fallback     1.000 i/100ms
             current     10.327 (± 0.0%) i/s -     52.000 in   5.036097s
                to_s      9.741 (± 0.0%) i/s -     49.000 in   5.031537s
                 erb     10.289 (± 0.0%) i/s -     52.000 in   5.057790s
        erb fallback     10.077 (± 0.0%) i/s -     51.000 in   5.061175s

Comparison:
             current:       10.3 i/s
                 erb:       10.3 i/s - 1.00x  slower
        erb fallback:       10.1 i/s - 1.02x  slower
                to_s:        9.7 i/s - 1.06x  slower

TEXT ONLY
ruby 3.3.1 (2024-04-23 revision c56cd86388) +YJIT [x86_64-linux]
             current     1.000 i/100ms
                to_s     1.000 i/100ms
                 erb     1.000 i/100ms
        erb fallback     1.000 i/100ms
             current     12.746 (± 0.0%) i/s -     64.000 in   5.021159s
                to_s     12.025 (± 0.0%) i/s -     61.000 in   5.072948s
                 erb     18.742 (± 0.0%) i/s -     94.000 in   5.016059s
        erb fallback     12.225 (± 0.0%) i/s -     62.000 in   5.071884s

Comparison:
                 erb:       18.7 i/s
             current:       12.7 i/s - 1.47x  slower
        erb fallback:       12.2 i/s - 1.53x  slower
                to_s:       12.0 i/s - 1.56x  slower
```

lib/rack/utils.rb
Earlopain: `erb` is (for now) stdlib. Two other places here already require it.

jeremyevans: I do not believe that is necessary, as `erb` is stdlib and has not yet been proposed to move to bundled gems.

lib/rack/utils.rb
mperham: Sorry for the unkind characterization. "Convoluted" would be better.

Earlopain: I don't particularly see the importance of using `define_method` for only 1% improvement but `ERB::Escape` turns out to be about 50% faster when the text doesn't need escaping. So I would say using `ERB::Escape` is nice, though I would prefer to write it with plain `def` myself.

lib/rack/utils.rb
Earlopain: Good point 👍

ioquatix:
It looks pretty good to me.

jeremyevans (commented):
One difference between `ERB::Escape#html_escape` and `CGI::Escape#escapeHTML` is that if no changes are needed, `ERB::Escape#html_escape` does not allocate, it returns the argument.  Are we OK with that change in behavior?

ioquatix:
That seems reasonable to me.

Earlopain:
> The only problem is that which implementation we choose now depends on load order. If ERB is not loaded before rack, and is loaded AFTER rack, the behaviour may change. That doesn't bother me but maybe worth documenting it?

erb is required at the top of the file so if the required method is available (i.e. recent version of ruby, or erb in the gemfile) it should be deterministic. I don't think the situation you are describing is possible, or am I missing something here?

jeremyevans (approved):

Earlopain:
One thing that is additionally nice here is that this function is now again documented. https://www.rubydoc.info/gems/rack/3.1.3/Rack%2FUtils:escape_html

It didn't cope with `define_method`. I'm sure there would have been annotations to make that work though

PR #2204, opened 2024-06-11
Failing test with Rack::MockResponse

tenderlove:
[This PR](https://github.com/rack/rack/pull/2195) removed checking for `transfer-encoding` from the `Rack::Response` object when [deciding whether or not to set the content-length header](https://github.com/rack/rack/pull/2195/files#diff-cfcc6d2645b6f77c5ec05acd448f280ce3004d3dba85b8aef3b3c81faf6d1373L113-L115).  This had an impact on the `Rack::MockResponse` subclass as the mock response subclass will try to buffer the response immediately [inside the initialize method](https://github.com/rack/rack/blob/0700a9ca85f748c1d21eb4544d23ffa4df2348c9/lib/rack/mock_response.rb#L36).  When the response is buffered, the `@length` instance variable is set, and this causes the superclass to set the content-length header even though it's a streaming response.

I'm unsure how to fix this situation, so this PR is just a failing test that demonstrates the problematic behavior.  Even though the chunked body object _doesn't_ respond to `to_ary`, the mock response object decides to buffer it anyway, causing the "content length" header to be set.  I tried returning early from `buffered_body!`, but this causes other issues surrounding when the body should be "closed" (it turns tests will try to read the response body twice, but the second time it is closed).

Ref: https://github.com/rails/rails/issues/52066
Ref: https://github.com/rails/rails/pull/52092

### DIFF
--- a/lib/rack/constants.rb
+++ b/lib/rack/constants.rb
@@ -22,6 +22,7 @@ module Rack
   ETAG              = 'etag'
   EXPIRES           = 'expires'
   SET_COOKIE        = 'set-cookie'
+  TRANSFER_ENCODING = 'transfer-encoding'
 
   # HTTP method verbs
   GET     = 'GET'
--- a/lib/rack/content_length.rb
+++ b/lib/rack/content_length.rb
@@ -21,6 +21,7 @@ module Rack
 
       if !STATUS_WITH_NO_ENTITY_BODY.key?(status.to_i) &&
          !headers[CONTENT_LENGTH] &&
+         !headers[TRANSFER_ENCODING] &&
          body.respond_to?(:to_ary)
 
         response[2] = body = body.to_ary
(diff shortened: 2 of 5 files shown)


jeremyevans:
I propose reverting #2195 to fix the issue. When I approved it, I was not aware of what was depending on the previous behavior.

tenderlove:
> I propose reverting #2195 to fix the issue. When I approved it, I was not aware of what was depending on the previous behavior.

Ya, I think we should.  I agree that theoretically #2195 is the right thing to do, but it seems like we need to take more situations in to account before we can do it.  "Testing + streaming responses" seems like perilous territory and we should figure out what to do there before we redo #2195

ioquatix:
> and this causes the superclass to set the content-length header even though it's a streaming response.

By the time `Rack::MockResponse` gets the body, and processes it, it's no longer streaming, because `Rack::MockResponse` behaves like a client and reads all the chunks into a single buffer. I'm not sure if this is the correct behaviour or not, but it seems reasonable to me. However, as we can see, it's also problematic.

PR #2208, opened 2024-06-12
Add rails smoke tests to external tests.

ioquatix:
We've had a couple of challenges relating to what essentially amount to compatibility issues between Rails and Rack, e.g. <https://github.com/rails/rails/issues/52066>. Let's ensure that Rails' smoke tests all pass as part of Rack's CI. This should give earlier indications about potential issues.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -19,6 +19,6 @@ group :test do
   gem "webrick"
 
   unless ENV['CI'] == 'spec'
-    gem 'bake-test-external', '~> 0.1.3'
+    gem 'bake-test-external'
   end
 end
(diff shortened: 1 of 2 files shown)


jeremyevans (approved):
May want to consider adding Sinatra as an external test as well. Sinatra and Rails probably cover 90% of Rack usage.  But that can be done later.

dentarg:
I'm testing rack-head in the Sinatra repo. Currently there is one known failure: https://github.com/sinatra/sinatra/issues/2014

ioquatix:
@dentarg are you able to create a PR to add Sinatra to the external tests?

dentarg:
I have too little time right now

simi:
> I have too little time right now

I have some. https://github.com/rack/rack/pull/2210

PR #2210, opened 2024-06-12
Add Sinatra to external test.

simi:
- bump bake-test-external
- add external directory to gitignore

### DIFF
--- a/config/external.yaml
+++ b/config/external.yaml
@@ -11,3 +11,6 @@ roda:
 grape:
   url: https://github.com/ruby-grape/grape
   command: bundle exec rspec --exclude-pattern=spec/integration/**/*_spec.rb
+sinatra:
+  url: https://github.com/sinatra/sinatra
+  command: bundle exec rake test
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,4 @@ doc
 /.bundle
 /.yardoc
 /coverage
+/external


config/external.yaml
```diff
@@ -1,3 +1,6 @@
 protocol-rack:
   url: https://github.com/socketry/protocol-rack
   command: bundle exec bake test
+sinatra:
+  url: https://github.com/sinatra/sinatra
+  command: bundle exec rake test
```
dentarg, line 6: There is also https://github.com/sinatra/sinatra/tree/main/sinatra-contrib and https://github.com/sinatra/sinatra/tree/main/rack-protection – they have their own tests. I don't know if they should be tested here too, just letting you know.

ioquatix: It's fine to add more external tests as long as they can run without a huge amount of extra scaffolding.

ioquatix:
@simi thanks for your great PR, do you mind rebasing it? Once it's green we can merge.

We may need to sort out <https://github.com/rack/rack/pull/2181> too.

config/external.yaml
simi: AFAIK `bake-test-external` doesn't support `Gemfile` in subfolder yet.

simi: ok, I'll check

ioquatix: There was a bug, and I've fixed it and released in v0.5.0

ioquatix:
Should be fixed by <https://github.com/rack/rack/pull/2214>.

config/external.yaml
simi: I'll open separate PR to onboard those as well.

Gemfile
```diff
@@ -19,6 +19,6 @@ group :test do
   gem "webrick"
 
   unless ENV['CI'] == 'spec'
-    gem 'bake-test-external'
+    gem 'bake-test-external', '~> 0.3.3'
```
ioquatix, line 22:
```suggestion
    gem 'bake-test-external'
```

simi:
looks finally green @ioquatix

jeremyevans (approved):

dentarg:
I don't think this should be green?

Both https://github.com/sinatra/sinatra/issues/2018 and https://github.com/sinatra/sinatra/issues/2019 (https://github.com/rack/rack/issues/2218) should fail them

dentarg:
[Looks](https://github.com/rack/rack/actions/runs/9581914215/job/26419753804#step:4:4484) like it is the rack/logger issue that isn't picked up as a failure

PR #2214, opened 2024-06-17
Allow empty PATH_INFO.

ioquatix:
From <https://datatracker.ietf.org/doc/html/rfc3875#section-4.1.5>:

The PATH_INFO variable specifies a path to be interpreted by the CGI
   script.  It identifies the resource or sub-resource to be returned by
   the CGI script, and is derived from the portion of the URI path
   hierarchy following the part that identifies the script itself.
   Unlike a URI path, the PATH_INFO is not URL-encoded, and cannot
   contain path-segment parameters.  A PATH_INFO of "/" represents a
   single void path segment.

```
      PATH_INFO = "" | ( "/" path )
      path      = lsegment *( "/" lsegment )
      lsegment  = *lchar
      lchar     = <any TEXT or CTL except "/">
```

Our interpretation of `PATH_INFO` is more extensive than the CGI specification to allow for things like `OPTIONS *` and `CONNECT host:port` style requests.

### DIFF
--- a/SPEC.rdoc
+++ b/SPEC.rdoc
@@ -130,7 +130,7 @@ There are the following restrictions:
 * There may be a valid early hints callback in <tt>rack.early_hints</tt>
 * The <tt>REQUEST_METHOD</tt> must be a valid token.
 * The <tt>SCRIPT_NAME</tt>, if non-empty, must start with <tt>/</tt>
-* The <tt>PATH_INFO</tt>, if provided, must be a valid request target.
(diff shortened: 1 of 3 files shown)


ioquatix:
@jeremyevans @simi I think once this is merged, <https://github.com/rack/rack/pull/2210> will be green.

ioquatix:
@jeremyevans if this is breaking Sinatra, I'd like to backport, as our interpretation of the RFC was incorrect, so this should be a bug.

jeremyevans:
Backporting to 3.1 is fine.

Issue #2218, opened 2024-06-19
Rack 3.1 is missing content-length on 0 byte responses


zarqman:
In a Rails 7.1 app, empty responses are now missing the `content-length` header when used with Rack 3.1. On Rack 3.0, the header is present with the expected value of `0`.

I'm not positive, but this might be a byproduct of #2149 since that deferred setting `content-length`. If there's no body to iterate, perhaps `content-length` gets missed?

I don't have a thorough grasp on how Rails and Rack interact in the case of an empty response, so it's possible this issue belongs to Rails instead of Rack. However, I'm starting here since changing the Rack version alone is enough to trigger the behavior difference.

##### Why this matters

Some HTTP servers revert to chunked encoding when the payload size is unknown. This is inefficient.

Additionally, some HTTP caches use `content-length` to decide what is and isn't cacheable and its absence can cause the entire request to become uncacheable, causing excess load on app servers.

<details>
<summary>Reproduction example</summary>

```ruby
# frozen_string_literal: true

require "bundler/inline"

gemfile(true) do
  source "https://rubygems.org"

  gem "rails"
  # gem "rack", "~> 3.0.0"
  gem "rack", "~> 3.1.0"
end

require "action_controller/railtie"

class TestApp < Rails::Application
  config.root = __dir__
  config.hosts << "www.example.com"
  config.secret_key_base = "secret_key_base"
  config.action_dispatch.show_exceptions = :rescuable
  config.logger = Logger.new($stdout)
  Rails.logger  = config.logger

  routes.draw do
    get "/head" => "test#one"
    get "/empty" => "test#two"
    get "/body" => "test#three"
  end
end

class TestController < ActionController::Base
  def one
    head 200
  end
  def two
    render plain: ''
  end
  def three
    render plain: 'a', layout: false
  end
end

require "minitest/autorun"

class TestControllerTest < ActionDispatch::IntegrationTest

  # works on rack 3.0
  # content-length key is missing on rack 3.1
  def test_head
    get "/head"
    assert_response 200
    assert_equal '0', headers['content-length']
    # assert headers.key?('content-length')
  end

  # works on rack 3.0
  # content-length key is missing on rack 3.1
  def test_empty_body
    get "/empty"
    assert_response 200
    assert_equal '0', headers['content-length']
    # assert headers.key?('content-length')
  end

  # works on rack 3.0 and 3.1
  def test_with_body
    get "/body"
    assert_response 200
    assert_equal '1', headers['content-length']
  end

  private
[...]

PR #2219, opened 2024-06-21
`Rack::Response` should not generate invalid `content-length` header.

ioquatix:
I'm experimenting with potential fixes for <https://github.com/rack/rack/issues/2218>.

### DIFF
--- a/lib/rack/response.rb
+++ b/lib/rack/response.rb
@@ -72,7 +72,8 @@ module Rack
       if body.nil?
         @body = []
         @buffered = true
-        @length = 0
+        # Body is unspecified - it may be a buffered response, or it may be a HEAD response.
+        @length = nil
       elsif body.respond_to?(:to_str)
         @body = [body]
         @buffered = true
@@ -80,7 +81,7 @@ module Rack
       else
         @body = body
         @buffered = nil # undetermined as of yet.
-        @length = 0
+        @length = nil
       end
 
       yield self if block_given?
@@ -110,14 +111,15 @@ module Rack
         close
         return [@status, @headers, []]
       else
-        if @length && @length > 0 && !chunked?
-          set_header CONTENT_LENGTH, @length.to_s
-        end
-
         if block_given?
+          # We don't add the content-length here as the user has provided a block that can #write additional chunks to the body.
           @block = block
           return [@status, @headers, self]
         else
+          # If we know the length of the body, set the content-length header... except if we are chunked? which is a legacy special case where the body might already be encoded and thus the actual encoded body length and the content-length are likely to be different.
+          if @length && !chunked?
+            @headers[CONTENT_LENGTH] = @length.to_s
+          end
           return [@status, @headers, @body]
         end
       end
@@ -135,7 +137,9 @@ module Rack
       end
     end
 
-    # Append to body and update content-length.
+    # Append a chunk to the response body.
+    #
+    # Converts the response into a buffered response if it wasn't already.
     #
     # NOTE: Do not mix #write and direct #body access!
     #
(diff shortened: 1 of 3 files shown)


ioquatix:
The one Sinatra test failure that is affected by this part of the code:

```
  1) Failure:
StaticTest#test_handles_valid_byte_ranges_correctly_0 [test/static_test.rb:147]:
Incorrect Content-Length for bytes=0-0.
--- expected
+++ actual
@@ -1,3 +1 @@
-# encoding: US-ASCII
-#    valid: true
-"1"
+nil
```

However, the reason why this header is missing is because it is deleted by this code:

https://github.com/sinatra/sinatra/blob/main/lib/sinatra/base.rb#L298

It was being inadvertently added back in by `Response#finish`. In other words, Sinatra's test suite is testing the wrong thing, OR the test is indeed detecting a failure to have the content length header.

ioquatix:
The one Sinatra test failure that is affected by this part of the code:

```
  1) Failure:
StaticTest#test_handles_valid_byte_ranges_correctly_0 [test/static_test.rb:147]:
Incorrect Content-Length for bytes=0-0.
--- expected
+++ actual
@@ -1,3 +1 @@
-# encoding: US-ASCII
-#    valid: true
-"1"
+nil
```

However, the reason why this header is missing is because it is deleted by Sinatra's own code:

https://github.com/sinatra/sinatra/blob/main/lib/sinatra/base.rb#L298

It was being inadvertently added back in by `Response#buffered_body!` (invoked by `MockResponse#initialize`). In other words, it appears that the test is wrong, or Sinatra's code is wrong, and this wrongness was hidden by the previous implementation of `MockResponse` inadvertently adding the `content-length` header when none existed in the response itself.

cc @dentarg

ioquatix:
I'm removing the Sinatra tests from this PR, as I think it's almost good to go. I might have a go at adding some tests for the behaviour of `buffered_body!` not adding content-length if it's not present (would fail on 3-0-stable for example).

test/spec_response.rb
```diff
@@ -78,7 +78,7 @@
     body.each { |part| parts << part }
 
     parts.must_equal ["foo", "bar", "baz"]
-    h['content-length'].must_equal '6'
+    h['content-length'].must_be_nil
```
ioquatix, line 81: Fixed assertion for invalid behaviour (9 bytes in body, content length set to 6???)

lib/rack/response.rb
```diff
@@ -110,14 +111,14 @@ def finish(&block)
...
       else
-        if @length && @length > 0 && !chunked?
-          set_header CONTENT_LENGTH, @length.to_s
-        end
-
         if block_given?
           @block = block
           return [@status, @headers, self]
         else
+          # If we know the length of the body, set the content-length header... except if we are chunked? which is a legacy special case where the body might already be encoded and thus the actual length and the content-length are likely to be different.
```
jeremyevans, line 118: I think the reason we aren't potentially setting the content-length if a block is given is because the block could write to the body, and that would result in an invalid content length? If that is correct, can you add a comment explaining that?

jeremyevans (approved):

lib/rack/response.rb
ioquatix: Yes, that's correct, sure I can explain that too.

ioquatix:
I did a final check with Sinatra external tests:

```
  1) Failure:
HelpersTest::TestLogger#test_logging_works_when_logging_is_enabled_0 [test/helpers_test.rb:1903]:
Expected false to be truthy.

  2) Failure:
StaticTest#test_handles_valid_byte_ranges_correctly_0 [test/static_test.rb:147]:
Incorrect Content-Length for bytes=0-0.
--- expected
+++ actual
@@ -1,3 +1 @@
-# encoding: US-ASCII
-#    valid: true
-"1"
+nil
```

These are expected failures, and should be fixed in Sinatra.

PR #2221, opened 2024-06-22
Rack logger compatibility

ioquatix:
I'd like to consider restoring this, at least temporarily, until we fix Sinatra downstream.

Even if we don't merge, it's helpful to track external test failures.

I still think we should probably remove `Rack::Logger`, along with `rack.errors` and `rack.logger` from the SPEC. I don't think logging should be a concern of Rack.

Sinatra should introduce their own `Sinatra::Logger` and they can easily add a dependency on `logger` gem.

### DIFF
--- a/lib/rack/logger.rb
+++ b/lib/rack/logger.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+require_relative 'constants'
+
+warn "Rack::Logger is deprecated and will be removed in Rack 3.2.", uplevel: 1
+
+module Rack
+  class Logger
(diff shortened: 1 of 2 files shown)


Earlopain:
This previously took a log level as argument, now it takes a logger instead. Logleves are just numbers so you don't need a dependency for that (though it also accepts symbols like `:info`). Info was default so debug output was not included.

ioquatix:
Yeah, we could probably support a basic log level feature.

ioquatix:
We validated the content-length changes, this branch is no longer needed.

Issue #2225, opened 2024-07-02
NoMethodError for "attribute.strip!" with v3.1.5


ledermann:
After updating a Rails application to Rack `v3.1.5`, I get exceptions in https://github.com/rack/rack/blob/v3.1.5/lib/rack/request.rb#L647 when [Updatetime robot](https://uptimerobot.com) comes by:

```
NoMethodError: undefined method `strip!' for nil attribute.strip! 
```

![Honeybadger report](https://github.com/rack/rack/assets/3557/8d441c82-1392-4d13-ab12-6f1ee7eb9888)

Uptime Robots does a GET request with these headers:

```
{
  "User-Agent": "Mozilla/5.0+(compatible; UptimeRobot/2.0; http://www.uptimerobot.com/)",
  "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8",
  "Accept-Language": "en-US,en;q=0.8",
  "Connection": "close",
  "cache-control": "no-cache",
  "Referer": "https://my.example.com",
  "Accept-encoding": ""
}
```

It seems a Safe Navigation Operator (&.) is missing.

Belongs to https://github.com/rack/rack/security/advisories/GHSA-cj83-2ww7-mvq7
References https://github.com/rack/rack/commit/97952ac5ea99976b75469a19191ee71bde0ebb0f

ioquatix:
Damn, thanks for the report, let me take a look.

ioquatix:
@ledermann can you please tell me what version of Ruby you are running on, and what server too?

Earlopain:
No header from the reproducer crashes but passing `", "` to `parse_http_accept_header` does

ledermann:
I'm using Ruby 3.3.3, Puma 6.4.2, Rails 7.1.3.4.

ioquatix:
Please check the linked PR.

ledermann:
To complete my report, the invalid header field sent by @uptimerobot that causes the exception is as follows:

```
"HTTP_ACCEPT_ENCODING" => ", gzip"
```

So, your fix in #2226 handles this nicely.

ioquatix:
I'm curious how you ended up with such a header?

ledermann:
@ioquatix My Rails app is monitored by UptimeRobot, which sends a request to my app every 5 minutes. I have no control over the exact request. After updating my app to Rack 3.1.5, Honeybadger was notified about the exception, so the request headers can be inspected there. 
[...]

PR #2226, opened 2024-07-02
Improved handling of invalid `accept` headers.

ioquatix:
Possible fix for <https://github.com/rack/rack/issues/2225>.

A few thoughts:

- How bad is `split(/\s*,\s*/)` on recent versions of Ruby? Could we return to using it in the future if the performance isn't bad?
- `filter_map` is only available in Ruby 2.7+ - I know we decided to support old Rubies for as long as possible and I don't think we should change that right now, but I just wanted to call out that there is (probably) a slight loss of performance due to backwards compatibility.

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -635,7 +635,8 @@ module Rack
       end
 
       def parse_http_accept_header(header)
-        header.to_s.split(',').filter_map do |part|
+        # It would be nice to use filter_map here, but it's Ruby 2.7+
+        header.to_s.split(',').map do |part|
           part.strip!
           next if part.empty?
 
@@ -647,7 +648,7 @@ module Rack
             quality = $1.to_f
           end
           [attribute, quality]
-        end
+        end.compact
       end
 
       # Get an array of values set in the RFC 7239 `Forwarded` request header.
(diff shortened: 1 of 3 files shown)


lib/rack/request.rb
```diff
@@ -644,7 +648,7 @@ def parse_http_accept_header(header)
             quality = $1.to_f
           end
           [attribute, quality]
-        end
+        end.compact
```
jeremyevans, line 651: I think we should use a temporary variable and `map!` and `compact!` to avoid unnecessary array allocations.

jeremyevans (approved):
Even in recent Ruby versions, this new approach is much faster than the previous approach.  If you can use strings instead of regexps, it's usually faster to use strings.

lib/rack/request.rb
```diff
@@ -645,6 +651,8 @@ def parse_http_accept_header(header)
           end
           [attribute, quality]
         end
+
+        parts.compact
```
jeremyevans, line 655: `parts.compact!; parts` to save another array allocation.

jeremyevans (approved):

lib/rack/request.rb
ioquatix: Sorry, I missed this but intended to do that. Thanks for calling it out.

jeremyevans (approved):

ioquatix:
Thanks @jeremyevans for your quick and excellent reviews.

PR #2227, opened 2024-07-03
Fix encoding setting for non-binary IO-like objects in MockRequest#env_for

jeremyevans:
No core class implements both encoding and set_encoding.

Classes implementing encoding:

* Regexp
* Symbol
* String

Classes implementing set_encoding:

* File
* ARGF.class
* IO

This code was added in 64610db2d2e7e910c4d3e17874dc316ed3eb227a. Previously, set_encoding was always called.
64610db2d2e7e910c4d3e17874dc316ed3eb227a actually broke encoding setting to binary for IO objects, but because there were no tests for the case, it wasn't caught.

I don't see a point of warning of Regexp and Symbol use, and raising for objects that support encoding but not set_encoding doesn't make sense.  I'm guessing the intention was to handle IO-like objects, and then internal_encoding could be checked, but it's simpler to just call set_encoding in that case, as that was the previous behavior.

An alternative to this code would be reverting
64610db2d2e7e910c4d3e17874dc316ed3eb227a, but that is no longer backwards compatible with 3.1.0, which allows objects that do not respond to set_encoding.

Found by coverage testing.

### DIFF
--- a/lib/rack/mock_request.rb
+++ b/lib/rack/mock_request.rb
@@ -139,23 +139,13 @@ module Rack
         end
       end
 
-      input = opts[:input]
-      if String === input
-        rack_input = StringIO.new(input)
-        rack_input.set_encoding(Encoding::BINARY)
-      else
-        if input.respond_to?(:encoding) && input.encoding != Encoding::BINARY
-          warn "input encoding not binary", uplevel: 1
-          if input.respond_to?(:set_encoding)
-            input.set_encoding(Encoding::BINARY)
-          else
-            raise ArgumentError, "could not coerce input to binary encoding"
-          end
-        end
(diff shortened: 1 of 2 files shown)


ioquatix (approved):
LGTM - appreciate you fixing my mess.

ioquatix:
Is this a bug that needs to be back ported?

jeremyevans:
Probably, as 3.0 used to set the encoding to binary and 3.1 doesn't.

ioquatix:
Okay, so we should backport to 3.1 only?

jeremyevans:
I think so.  3.0 didn't have this issue, it always called set_encoding.

PR #2231, opened 2024-07-05
Remove Sinatra as an external test until the issues are fixed

jeremyevans:
There is no point in having CI complain about every commit when there is nothing Rack can do to fix things.

Sinatra was only added because it was thought to be already passing, due to a bug in bake-test-external that has since been fixed.  As soon as Sinatra is fixed and works with Rack main, this commit can be reverted.


ioquatix:
This makes to me.

ioquatix (approved):

PR #2233, opened 2024-07-05
Only treat space and \t as allowed whitespace in headers

jeremyevans:
RFC 7230 (HTTP RFC) Section 3.2.3 is clear that those are the only two allowed characters for required/optional whitespace in headers.

RFC 2046 (MIME/Multipart RFC) refers to RFC 822, and RFC 822 Section 3.3 also specifies that linear white space is made up of only space and tab characters.

While here, I noticed the content-type parser in the multipart code was requiring a space, when whitespace is optional. Make the whitespace optional, and adjust a test where a content-type was ignored because it didn't use whitespace.

The request and utils changes here are operating on header values, and generally other whitespace is allowed there, but I'm not sure we should split on whitespace other than space and \t, so I made the change there as well.

There are a lot of places in Rack where we are using strip/strip! were we should potentially change to be more strict and only strip space and \t, but this commit does not attempt to address strip/strip! usage.

### DIFF
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -32,9 +32,9 @@ module Rack
 
     EOL = "\r\n"
     MULTIPART = %r|\Amultipart/.*boundary=\"?([^\";,]+)\"?|ni
-    MULTIPART_CONTENT_TYPE = /Content-Type: (.*)#{EOL}/ni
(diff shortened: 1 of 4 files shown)


ioquatix (approved):

ioquatix:
I agree with being a little more strict.

What other whitespace does `\s` match that `[ \t]` doesn't - in practice is it going to make any difference?

jeremyevans:
`["\n", "\v", "\f", "\r"]` are the other whitespace characters matched by `\s`.

PR #2235, opened 2024-07-10
Note "Allow empty PATH_INFO" in CHANGELOG

dentarg:
I did not see https://github.com/rack/rack/pull/2214 mentioned in the changelog


jeremyevans (commented):
If we are going to backport this change to 3-1-stable, it makes sense to list it under the backported version and not in the unreleased section, so maybe we can add it there after it is backported?

CHANGELOG is missing a bunch of recent stuff I worked on (purposely to not introduce conflicts).  I'll work on a pull request to update it shortly.

ioquatix (approved):
Thanks for this, I'll sort out the merge conflicts with Jeremy's PR (if any) - and move the position on the log when I do a 3.1.x release today.

PR #2239, opened 2024-07-11
Allow env to be `Hash or Hash-like instance`.

ioquatix:
Implements the proposal described in <https://github.com/rack/rack/issues/1879>:

- Make a middleware that casts the env hash to a request object, then passes the request object to the next middleware
- Make the request object quack enough like a hash that existing legacy middleware can use the request object as if it was the env hash

### DIFF
--- a/SPEC.rdoc
+++ b/SPEC.rdoc
@@ -16,9 +16,7 @@ and the *body*.
 
 == The Environment
 
-The environment must be an unfrozen instance of Hash that includes
-CGI-like headers. The Rack application is free to modify the
-environment.
+The environment must be an unfrozen Hash or Hash-like instance which implements the following methods with equivalent behaviour to the methods of a Hash instance: #[], #[]=, #keys, #key?, #include?, #fetch, #delete, #clear, #to_hash, #each, #each_key, #each_value, #each_pair, #merge, #update, #frozen?, #inspect.
 
 The environment is required to include these variables
 (adopted from {PEP 333}[https://peps.python.org/pep-0333/]), except when they'd be empty, but see
--- a/lib/rack.rb
+++ b/lib/rack.rb
@@ -46,6 +46,7 @@ module Rack
   autoload :Recursive, "rack/recursive"
   autoload :Reloader, "rack/reloader"
   autoload :Request, "rack/request"
+  autoload :RequestWrapper, "rack/request_wrapper"
   autoload :Response, "rack/response"
   autoload :RewindableInput, "rack/rewindable_input"
   autoload :Runtime, "rack/runtime"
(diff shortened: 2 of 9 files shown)


jeremyevans:
As I mentioned in #1879, I'm strongly against this approach due to the backwards compatibility issues.  

I doubt you will be able to get the Roda specs passing with this.  Roda's middleware plugin uses `is_a?(Hash)` to determine whether to treat a passed argument as a request environment or not. You can debate the merits of this approach, but I doubt it is the only code in the ecosystem that operates this way, considering that's been Rack's SPEC for 17 years.  Even from the first commit for the spec (df0c8d01f69d63b474f86afe1eaf44bc0be5169f): `The environment must be an true instance of Hash (no subclassing allowed)`.

ioquatix:
Do you mind pointing me at the code that does this:

> I doubt you will be able to get the Roda specs passing with this. Roda's middleware plugin uses is_a?(Hash) to determine whether to treat a passed argument as a request environment or not.

jeremyevans:
> Do you mind pointing me at the code that does this:

https://github.com/jeremyevans/roda/blob/master/lib/roda/plugins/middleware.rb#L182

ioquatix:
All tests pass locally if I write it like this:

```ruby
          if app.respond_to?(:to_hash)
            super
          else
            Forwarder.new(self, app, *args, &block)
          end
```

jeremyevans:
> All tests pass locally if I write it like this:
> 
> ```ruby
>           if app.respond_to?(:to_hash)
>             super
>           else
>             Forwarder.new(self, app, *args, &block)
>           end
> ```

That will break creating middleware that responds to `to_hash`.  Technically, the current code doesn't work for a middleware that is a hash instance, but since Hash#call is not defined (by default), that's not a problem.

ioquatix:
> That will break creating middleware that responds to to_hash.

Okay, so this is a hypothetical problem, but probably not a real one?

jeremyevans:
> > That will break creating middleware that responds to to_hash.
> 
> Okay, so this is a hypothetical problem, but probably not a real one?

There are no restrictions on what methods middleware objects cannot implement.  I would guess that not many middleware implement `to_hash`, but it would not be surprised if there are middleware in use that do implement it.  Any use of `respond_to?` would be problematic, since valid middleware can theoretically implement any method.

Please focus on the greater point, which is that this will break code, even if the code can be modified to deal with it in most cases.  This breakage is only an example of such breakage that will likely be found elsewhere as well.

`Hash#new` is not `Hash.new` (`#` means instance method.)

ioquatix:
> Please focus on the greater point, which is that this will break code, even if the code can be modified to deal with it in most cases. This breakage is only an example of such breakage that will likely be found elsewhere as well.

I understand that, but I'd still like to explore the proposal and see what the actual impact is and whether we can mitigate it.

ioquatix:
Seems like this might be enough for compatibility's sake:

```
        def is_a?(klass)
          klass == Hash || super
        end
```

ioquatix:
After considering this and the long term compatibility, I decide against introducing this feature at the moment.

I'm not negative on it, but I'm also not strongly positive on it without support from other maintainers and downstream users. I think it can be nice if instead of `env` being a hash, it could be an instance of `Rack::Request` which can behave like a hash - there are clear advantages but also some disadvantages (a more complex definition of `env` itself).

PR #2241, opened 2024-08-22
Remove rack/handler reference

robertomiranda:
Remove the reference to `Rack::Handler` since it has been moved to the [rackup](https://github.com/rack/rackup) gem as part of #1937


jeremyevans (approved):

ioquatix:
Thanks!

PR #2242, opened 2024-08-28
Resolve deprecation warnings about uri `DEFAULT_PARSER`

Earlopain:
Ruby switches the default parser from RFC2396 to RFC3986. This parser contains all methods from the old parser but delegates to RFC2396 for a few and warns. Namely `extract`, `make_regexp`, `escape`,and `unescape`.

Since the RFC2396 alias is not available for old rubies, do a defined check and select the appropriate one. This will work for all versions without warnings.

Additional info:
* https://github.com/ruby/uri/pull/107
* https://bugs.ruby-lang.org/issues/19266

While this removes the warning and passes tests, I can't describe the possible impact for continuing to use the older parser like this. I'll leave that up to someone with more domain knowledge, consider this a notice about the warning if that turns out to be important.

### DIFF
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -12,7 +12,7 @@ module Rack
 
   class Lint
     REQUEST_PATH_ORIGIN_FORM = /\A\/[^#]*\z/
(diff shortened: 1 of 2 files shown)


jeremyevans (approved):

ioquatix:
Thanks for your contribution and the great commit message.

PR #2243, opened 2024-08-28
Add `logger` to the gemfile

Earlopain:
While rack itself doesn't use it anymore, one test does. It checks that `CommonLogger` works with the std logger, so can't exactly change that test without outright removing it. Just add it to the gemfile.

https://github.com/rack/rack/blob/0ed580bbe3858ffe5d530adf1bdad9ef9c03407c/test/spec_common_logger.rb#L47-L53

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,6 +16,7 @@ unless ENV['CI']
 end
 
 group :test do
+  gem "logger"
   gem "webrick"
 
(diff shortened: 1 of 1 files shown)


jeremyevans (approved):

ioquatix:
Thanks for your contribution.

Issue #2245, opened 2024-09-02
 Support for ISO-2022-JP in Multipart Requests


nappa:
There is an issue where an Encoding::CompatibilityError is raised when the request body is multipart and the Content-Type is a text format (e.g., text/html or text/plain) with the charset set to ISO-2022-JP. ISO-2022-JP is currently treated as a dummy encoding in Ruby, meaning it cannot be interpreted or processed by Ruby.

Other dummy encodings include UTF-7 and IBM037, which are rarely used today. However, ISO-2022-JP, though considered a legacy encoding, is still used in emails in Japan due to historical reasons. As a result, some services that relay emails over HTTP may still use ISO-2022-JP.

For example, SendGrid, a service that receives emails via an SMTP Gateway, represents these emails as multipart data and sends them to webhooks via HTTP POST. In this process, the email body, encoded in ISO-2022-JP, is sent as a part of the request in its original encoding.

In our environment, we use Rack to handle requests from SendGrid, but we are encountering an issue where these requests cannot be processed due to the ISO-2022-JP encoding.

matthewd:
For comparison / context, which place(s) raise the `Encoding::CompatibilityError`? 

Re-encoding seems like it might be reasonable, but the other option might be to fix any Rack internals to allow dummy-encoded values to get to the application, and allow the app to decide what to do. 🤔

nappa:
To provide some context, the Encoding::CompatibilityError we're encountering occurs at the following location in our environment: https://github.com/rack/rack/blob/b1deebdc0a4f61cc141cece5a911917ff1e4b901/lib/rack/query_parser.rb#L90
(We are using rack-2.2 in our production environment, so this is the location where the error occurs. Although query_parser.rb was significantly modified in rack-3.0, the same issue persists there.)

I had a similar thought about passing dummy-encoded values directly to the application, but unfortunately, it wasn't feasible. Here's why:

Since ISO-2022-JP is a dummy encoding in Ruby, it doesn't support character-level splitting. To handle it within Rack, we'd need to either re-encode the data to something like UTF-8 (this is my approach) or implement custom handling to parse through ISO-2022-JP's state-transition-based encoding to detect multipart boundary markers. The latter approach would require not only changes in query_parser.rb but likely many other parts of the codebase as well, due to its complexity.

While I agree that allowing the application to handle ISO-2022-JP would be ideal, the complexity of the code required to achieve this doesn't seem to justify the benefits.

dentarg:
[...]

PR #2246, opened 2024-09-02
Add support for ISO-2022-JP encoding in multipart form data handling(#2245)

nappa:
Hi, I added support for ISO-2022-JP encoding in multipart form data handling, which solves #2245 .

- Starting with Ruby 3.4, nkf will be separated into a gem, so this code does not assume its presence. If nkf is not available, the code will behave as it did before.
- This code is written to behave like other encodings if Encoding::ISO_2022_JP is no longer treated as a dummy encoding in the future.
- Introduced `handle_dummy_encoding` method to properly convert ISO-2022-JP encoded parameters and bodies to UTF-8 using the `nkf` library, if available.
- Wrapped `nkf` library loading in a `begin-rescue` block to gracefully handle environments where the library is not available.
- Updated multipart parser to use the new method for handling legacy encodings, ensuring proper parameter normalization.
- Added test case to verify correct handling of multipart data with ISO-2022-JP encoding.

I hope it helps.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ All notable changes to this project will be documented in this file. For info on
 ### Added
 
 - Introduce `Rack::VERSION` constant. ([#2199](https://github.com/rack/rack/pull/2199), [@ioquatix])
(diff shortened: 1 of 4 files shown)


nappa:
I found String#encode is best way to convert encoding, let me close this pull request and write again!

PR #2247, opened 2024-09-02
Add support for ISO-2022-JP encoding in multipart form data handling(#2245)

nappa:
Hi, I introduced simple workaround for ISO-2022-JP encoding in multipart form data handling, which solves #2245 .
Also, I added test case to verify correct handling of multipart data with ISO-2022-JP encoding.

I hope it helps.

### DIFF
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -487,15 +487,21 @@ module Rack
         Encoding::BINARY
       end
 
+      REENCODE_DUMMY_ENCODINGS = {
+        # ISO-2022-JP is a legacy but still widely used encoding in Japan
+        # Here we convert ISO-2022-JP to UTF-8 so that it can be handled.
+        Encoding::ISO_2022_JP => true
+
+        # Other dummy encodings are rarely used and have not been supported yet.
+        # Adding support for them will require careful considerations.
+      }
+
       def handle_dummy_encoding(name, body)
-        # 'dummy' encodings are not properly implemented in Ruby MRI
-        if name.encoding.dummy?
-          # ISO-2022-JP is a legacy but still widely used encoding in Japan
-          # Here we properly convert ISO-2022-JP to UTF-8
-          if name.encoding == Encoding::ISO_2022_JP
-            name = name.encode(Encoding::UTF_8)
-            body = body.encode(Encoding::UTF_8)
-          end
+        # A string object with a 'dummy' encoding does not have full functionality and can cause errors.
+        # So here we covert it to UTF-8 so that it can be handled properly.
+        if name.encoding.dummy? && REENCODE_DUMMY_ENCODINGS[name.encoding]
+          name = name.encode(Encoding::UTF_8)
+          body = body.encode(Encoding::UTF_8)
         end
         return name, body
       end
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -241,7 +241,8 @@ module Rack
         @collector.each do |part|
           part.get_data do |data|
             tag_multipart_encoding(part.filename, part.content_type, part.name, data)
-            @query_parser.normalize_params(@params, part.name, data)
+            name, data = handle_dummy_encoding(part.name, data)
+            @query_parser.normalize_params(@params, name, data)
           end
         end
         MultipartInfo.new @params.to_params_hash, @collector.find_all(&:file?).map(&:body)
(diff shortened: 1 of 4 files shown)


lib/rack/multipart/parser.rb
```diff
@@ -486,6 +487,19 @@ def find_encoding(enc)
...
 
+      def handle_dummy_encoding(name, body)
+        # 'dummy' encodings are not properly implemented in Ruby MRI
+        if name.encoding.dummy?
+          # ISO-2022-JP is a legacy but still widely used encoding in Japan
+          # Here we properly convert ISO-2022-JP to UTF-8
+          if name.encoding == Encoding::ISO_2022_JP
+            name = name.encode(Encoding::UTF_8)
+            body = body.encode(Encoding::UTF_8)
+          end
```
ioquatix, line 498: Why not handle all dummy encodings thsi way, e.g.

```suggestion
          name = name.encode(Encoding::UTF_8)
          body = body.encode(Encoding::UTF_8)
```

ioquatix:
For reference:

```
irb(main):002> Encoding.list.select{|encoding| encoding.dummy?}
=> 
[#<Encoding:UTF-16 (dummy)>,
 #<Encoding:UTF-32 (dummy)>,
 #<Encoding:IBM037 (dummy)>,
 #<Encoding:ISO-2022-JP (dummy)>,
 #<Encoding:ISO-2022-JP-2 (dummy)>,
 #<Encoding:CP50220 (dummy)>,
 #<Encoding:CP50221 (dummy)>,
 #<Encoding:UTF-7 (dummy)>,
 #<Encoding:ISO-2022-JP-KDDI (dummy)>]
```

jeremyevans:
I don't know much about dummy encodings, but if they aren't really usable, encoding to UTF-8 seems reasonable.  It seems more consistent to do this for all dummy encodings, so @ioquatix's suggestion makes sense.  CHANGELOG should be updated when using @ioquatix's suggestion.

Security-wise, this would expose us to any issues in Ruby implementation transcoders from dummy encodings to UTF-8, as it would allow transcoding attacker-supplied data from an attacker-supplied dummy encoding to UTF-8.  Not knowing much about dummy encodings, I cannot estimate the actual risk.

ioquatix:
@duerst are you able to help us review this PR, and can you consider Jeremy's concerns/feedback?

nappa:
Thank you so much for taking the time to view this!

The reason I chose to exclude dummy encodings other than ISO-2022-JP in the `if` statement is that I considered whether they are actually used on the internet. Let me explain the situation with dummy encodings other than ISO-2022-JP:

Among the dummy encodings, CP50220, CP50221, ISO-2022-JP-KDDI, and ISO-2022-JP-2 are all variants of ISO-2022-JP.

- **CP50220**: This is an extension of ISO-2022-JP by Microsoft. It is registered in the IANA Registry. The registration was requested by one of the Ruby developers. [Source](https://data.iana.org/archive/ietf-charsets/msg01882.html) It is used only by a very limited number of applications and is now rarely seen on the internet.  
- **CP50221**: This is also an extension of ISO-2022-JP by Microsoft, but it is *not* registered in the IANA Registry. It is now rarely seen on the internet.  
- **ISO-2022-JP-2**: This is defined in RFC 1554 and is registered in the IANA Registry. However, it is now rarely seen on the internet.  
- **ISO-2022-JP-KDDI**: This is used by older devices from the Japanese mobile carrier KDDI. It is *not* registered in the IANA Registry. Since ISO-2022-JP is not specified in the `Content-Type`, it is necessary to infer that the device is a KDDI mobile phone based on information like the `User-Agent` and then explicitly specify ISO-2022-JP-KDDI for conversion.

Other encodings are as follows:  

- **UTF-16, UTF-32**: To convert these to UTF-8, determining the byte order is necessary. If a BOM (Byte Order Mark) is included at the beginning of the text, it can be processed correctly. However, if a BOM is not included, an `Encoding::InvalidByteSequenceError` will be raised. Attempting conversion in such cases may lead to other issues.  
- **IBM037**: I have never seen this encoding used on the internet, and I believe it is highly unlikely to appear.

Frankly speaking, the considerations for encodings other than ISO-2022-JP are far too numerous, and they also carry a significant potential for causing additional issues. Therefore, narrowing the scope to only ISO-2022-JP is not a bad choice from a practical standpoint.

duerst:
I agree with the direction this is going. At the moment, indeed Encoding::ISO_2022_JP seems to be the only encoding where this special treatment is needed and appropriate.  As for security issues, I think the main concern would be about smuggling some special syntax in the iso-2022-jp that then triggers something like Bobby Tables (https://xkcd.com/327/). That shouldn't happen because in iso-2022-jp, all ASCII characters appear literally. It also shouldn't happen because the security check (or other security measure, such as additional quoting/escaping) should be applied after the encoding step, not before. The multipart/form-data may contain other kinds of encoding on top of iso-2022-jp (e.g. quoted-printable or base64) that hide things further, but that's also possible for all the non-dummy encodings.

jeremyevans (approved):
Based on @nappa's and @duerst's feedback, I think we should accept this as-is.  I recommend we update the CHANGELOG to specify that ISO-2022-JP is transcoded to UTF-8, and I'd be OK with switching to a hash-table approach for more user control, as recommended by @ioquatix.

nappa:
Thank you so much, everyone!
Based on @ioquatix's feedback, I added a Hash table and included comments. I also updated the CHANGELOG.md following @jeremyevans's suggestions. Please let me know if there's anything else I should work on!

jeremyevans (approved):
Looks good.  Thank you for the quick fix!

ioquatix:
I’ll review this tonight in a few hours.

ioquatix:
Thanks everyone.

nappa:
@ioquatix @jeremyevans @duerst I'm happy to have been able to support this project. Thank you so much!

PR #2248, opened 2024-09-09
[2.2-stable] Fix compatibility issues with Ruby 3.4.0dev

casperisfine:
I know `2.2` is marked a security patches only, but given `3.x` contains breaking changes, and a large part of the community still haven't been able to upgrade, I think it would be sensible to also consider simple compatibility patches, to make it work on Ruby 3.4.

The two issues I found where:

  - `base64` warnings, I went the same way as https://github.com/rack/rack/commit/696ed9e8f48053683a0a19fc68eb49f094c0efcb, using `pack/unpack`.
  - `uri` warnings, just applied a simple fix.

NB: Test suite is failing on ruby-head because webrick needs an update: https://github.com/ruby/webrick/pull/144

cc @tenderlove @jeremyevans

### DIFF
--- a/.github/workflows/development.yml
+++ b/.github/workflows/development.yml
@@ -8,7 +8,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
-        ruby: [2.3, 2.4, 2.5, 2.6, 2.7, '3.0', 3.1, 3.2]
+        ruby: [2.3, 2.4, 2.5, 2.6, 2.7, '3.0', 3.1, 3.2, 3.3, 3.4]
     runs-on: ${{matrix.os}}
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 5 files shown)


jeremyevans (approved):
Considering this is just for compatibility with newer Ruby versions, I'm OK backporting these changes.

ioquatix:
@casperisfine would you like me to cut a release from this or are there additional compatibility issues you'd like to sort out first?

byroot:
I'm not aware of any more issues. A release would be much appreciated, but we can also wait a bit see if there is any more changes in 3.4, I just fear we might forget.

ioquatix:
Okay, I will put a note in my calendar for Thursday next week to cut a release. Let me know if anything else comes up before then :)

ioquatix:
I have released v2.2.10 with this (and only this) change.

Thanks for your contribution.

PR #2252, opened 2024-10-23
Opt-in for MFA requirement explicitly

tagliala:
As a pupular gem, `rack` implicitly requires that all privileged operations by any of the owners require OTP.

However, by explicitly setting `rubygems_mfa_required` metadata, the gem will show "NEW VERSIONS REQUIRE MFA" and
"VERSION PUBLISHED WITH MFA" in the sidebar at
https://rubygems.org/gems/rack

Ref:
- https://blog.rubygems.org/2022/08/15/requiring-mfa-on-popular-gems.html
- https://guides.rubygems.org/mfa-requirement-opt-in/

### DIFF
--- a/rack.gemspec
+++ b/rack.gemspec
@@ -31,7 +31,8 @@ Gem::Specification.new do |s|
     "bug_tracker_uri" => "https://github.com/rack/rack/issues",
     "changelog_uri" => "https://github.com/rack/rack/blob/main/CHANGELOG.md",
     "documentation_uri" => "https://rubydoc.info/github/rack/rack",
-    "source_code_uri"   => "https://github.com/rack/rack"
+    "source_code_uri"   => "https://github.com/rack/rack",
+    "rubygems_mfa_required" => "true"
   }
 
   s.add_development_dependency 'minitest', "~> 5.0"


ioquatix (approved):

simi (approved):

tagliala:
According to 

https://guides.rubygems.org/mfa-requirement-opt-in/#disabling-mfa-requirement

> You can disable the MFA requirement by setting **rubygems_mfa_required** to **"false"** or any [ActiveRecord::Type::Boolean::FALSE_VALUES](https://api.rubyonrails.org/classes/ActiveModel/Type/Boolean.html).
>
> Note: We will enforce the MFA requirement on the version being published. MFA requirement will be disabled after you have successfully published a gem with rubygems_mfa_required set to false.

As far as I understand, it should not be necessary to port this change to all the other supported branches, as far as an explicit gem is being published

But it if you prefer, I can backport the commit to 2.2, 2.1, and 2.0

ioquatix:
I'm slightly against back porting this as it's not a security fix or bug fix and it has no practical impact on those releases. I think we are both in agreement (perhaps for different reasons)?

ioquatix:
FYI, one thing that I have been thinking about is doing signed releases. I feel like it's far more valuable - do you have any thoughts on this?

PR #2253, opened 2024-10-28
Make `env[QUERY_STRING]` optional.

ioquatix:
At present, there is no way to differentiate between these two URLs:

1. `http://example.com/foo/bar`
2. `http://example.com/foo/bar?`

Ruby's `uri` implementation does differentiate between these, using `nil` for the former and empty string for the latter:

```
irb(main):002> URI.parse("http://foo.com/bar").query
=> nil
irb(main):003> URI.parse("http://foo.com/bar?").query
=> ""
irb(main):004> URI.parse("http://foo.com/bar?x=y").query
=> "x=y"
```

### DIFF
--- a/SPEC.rdoc
+++ b/SPEC.rdoc
@@ -44,8 +44,8 @@ below.
                      percent-encoded when originating from
                      a URL.
 <tt>QUERY_STRING</tt>:: The portion of the request URL that
-                        follows the <tt>?</tt>, if any. May be
-                        empty, but is always required!
+                        follows the <tt>?</tt>, if any. If no <tt>?</tt>
+                        is present, this value is <tt>nil</tt>.
 <tt>SERVER_NAME</tt>:: When combined with <tt>SCRIPT_NAME</tt> and
                        <tt>PATH_INFO</tt>, these variables can be
                        used to complete the URL. Note, however,
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -134,8 +134,8 @@ module Rack
         ##                      a URL.
 
         ## <tt>QUERY_STRING</tt>:: The portion of the request URL that
-        ##                         follows the <tt>?</tt>, if any. May be
-        ##                         empty, but is always required!
+        ##                         follows the <tt>?</tt>, if any. If no <tt>?</tt>
+        ##                         is present, this value is <tt>nil</tt>.
 
         ## <tt>SERVER_NAME</tt>:: When combined with <tt>SCRIPT_NAME</tt> and
         ##                        <tt>PATH_INFO</tt>, these variables can be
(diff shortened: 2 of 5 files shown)


jeremyevans:
I don't think this change is worth the backwards compatibility breakage.  The way SPEC is written, using a frozen empty string should be allowed, so requiring `QUERY_STRING` does not require an allocation, though it does require an extra hash entry.  I'd be OK with SPEC being more explicit that a frozen value is allowed.

ioquatix:
What compatibility breakage are you concerned about?

jeremyevans:
> What compatibility breakage are you concerned about?

At the risk of being obvious: `env["QUERY_STRING"].some_string_method`.

Is there any reason other than performance that you want this change?  You give an example of what it allows, but not why you would want it.  I don't think differentiating between missing and empty query strings is a good idea.

jeremyevans:
> In addition, we've accepted PRs with more compatibility concerns than this in the 3.x life cycle (for better or for worse).

You are correct, but I disagree that is evidence that we should accept the compatibility concerns in this case. We've received significant pushback for the amount of breakage in Rack 3.x, a fair amount of which is deserved.

I believe our focus going forward should be on improving Rack in a backwards compatible manner, not on making further backwards incompatible changes.  I'm not necessarily against all backwards incompatible changes, but the bar on what I'll approve is significantly higher than it was previously.  There needs to be very large advantages to offset a backwards incompatible change, and I'm not seeing that in this case.

The lack of breakage in external framework tests is not an indication of usage of in the wild.  Not sure if you realize it, but this breaks an example you have in Falcon: https://github.com/socketry/falcon/blob/7c7c3c1fbfe1b3d57c2290570d9a10f5c30081fb/examples/headers/config.ru#L8 (`CGI.parse` does not accept `nil`).

Looking at https://github.com/search?q=env%5B%22QUERY_STRING%22%5D+language%3ARuby&type=code , there looks to be many more things that would either fail or change behavior, including rage-rb and rack-mini-profiler.  And that's just the publicly available code.  You should probably assume 10-100x more breakage in non-open source repositories.

ioquatix:
I basically agree with you, however, it's also true that it's currently impossible to recreate the original URL. It bothers me slightly that this is not possible and I'm concerned there could be subtle security implications.

ioquatix:
Here is a use case I ran into that I felt should work:

```
count = env.fetch('QUERY_STRING', 99).to_i
```

jeremyevans:
> Here is a use case I ran into that I felt should work:
> 
> ```
> count = env.fetch('QUERY_STRING', 99).to_i
> ```

That's a strange example. While the query string is not required to contain key/value parameters, its almost always used that way.  Do you routinely deal with requests like `https://domain/path?123`?

ioquatix:
After considering this and discussing it with @jeremyevans I do not think meets the bar for compatibility.

PR #2256, opened 2024-10-29
Improve `lint.rb` documentation.

ioquatix:
- Generally improve the structure of the document.
- Consolidate documentation for CGI variables and Rack-specific variables.
- Ensure the Table of Contents is structured in a way that is easy to use.
- Improve the readability of `SPEC.rdoc` by merging paragraph lines.


ioquatix:
I've confirmed the formatting and readability is significantly improved.

image

lib/rack/lint.rb
```diff
@@ -21,20 +21,18 @@ class Lint
...
     # should not be removed. The whitespace creates paragraphs in the RDoc
     # output.
     #
-    ## This specification aims to formalize the Rack protocol. You
-    ## can (and should) use Rack::Lint to enforce it.
+    ## = Rack Specification
     ##
-    ## When you develop middleware, be sure to add a Lint before and
-    ## after to catch all mistakes.
+    ## This specification aims to formalize the Rack protocol. You can (and should) use +Rack::Lint+ to enforce it. When you develop middleware, be sure to add a lint wrapper to catch all mistakes.
```
jeremyevans, line 31:
```suggestion
    ## This specification aims to formalize the Rack protocol. When you
    ## develop libraries, frameworks, and/or middleware, be sure to use
    ## +Rack::Lint+ when testing to ensure compatibility with the rack specification.
```

lib/rack/lint.rb
```diff
@@ -99,327 +98,368 @@ def response
...
+
+          next if value.encoding == Encoding::ASCII_8BIT
+
+          unless value.b !~ /[\x80-\xff]/n
+            raise LintError, "env variable #{key} has value containing non-ASCII characters and has non-ASCII-8BIT encoding #{value.inspect} encoding: #{value.encoding}"
+          end
+        end
+
+        ##
+        ## The server and application can store their own data in the environment, too. The keys must contain at least one dot, and should be prefixed uniquely. The prefix <tt>rack.</tt> is reserved for use with the Rack specification.
```
jeremyevans, line 149:
```suggestion
        ## The server and application can store their own data in the environment, too. The keys must contain at least one dot, and should be prefixed uniquely. The prefix <tt>rack.</tt> is reserved for use with the Rack specification and the classes that ship with Rack.
```

jeremyevans (commented):
I started reviewing this, but there are just too many changes, and the changes mix code and SPEC text.  Can you please break this into reviewable commits, and do not mix code changes and SPEC text changes in the same commit?

ioquatix:
@jeremyevans its not particularly easy to break up this PR without lots of mechanical work. All the tests for the lint still pass so we are still testing all the same stuff, but the organisation is far better. On that basis, I think the best way to review this is to check the rdoc output for readability and structure. The mechanics of the implementation are mostly the same except for the introduction of two new helpers which correspond to the documentation (assert required and optional).

dentarg:
The commit says `Improved formatting and flow, clarifications + structure.`

I think it would be good (easier to review) only seeing the 'clarifications'. I also don't know what 'flow' refers to.

ioquatix:
> also don't know what 'flow' refers to.

Flow refers to the structure of the documentation and how one part leads to another. Having documentation for a single field split up into several places is bad flow as it's jarring and hard to find all the information pertaining to one part of the specification.

lib/rack/lint.rb
ioquatix: I don't want to hard wrap the comments in this file because:

1. Hard wrapping to a fixed length in this file causes the rdoc text to be wonky depending on the indentation of the comment = hard to read.
2. Edit amplification when editing the documentation if it requires a reflow of the text is harder to review.

It's far easier in every way to ensure your editor soft wraps the text at whatever width you prefer and it looks good everywhere - the code - the rdoc text - and the output HTML.

Also, IIRC, hard wrapping the rdoc text seemed to cause odd line breaks in the HTML output.

ioquatix: One option that I'd consider is if we can reflow the rdoc file after generating it. However, I'm not aware of any tooling that can do that (I know it exists for markdown for example). If you have any other ideas that don't make this document difficult to edit and difficult to read, that would be great.

jeremyevans: I'm not in favor of removing the hard wrapping, but I don't object to that if you want to do it.  However, I think if you want to do it, you should do so in a separate commit, in a way that results in no changes to the rendered HTML.  Then in other commits, you can make substantive changes.  You can test this by doing:

```
tilt SPEC.rdoc > SPEC.html.before
# make changes to lint.rb, regen SPEC.rdoc
tilt SPEC.rdoc > SPEC.html.after
diff -u SPEC.html.{before,after}
```

With no diff, that proves the change is purely mechanical.  You don't have to use `tilt`, if there is something else that will render a single rdoc file to html (`rdoc` itself seems to require outputting to a directory).

In terms of odd line breaks in the HTML, I'm not seeing any on https://github.com/rack/rack/blob/main/SPEC.rdoc from a brief review.  Can you be specific about where you are seeing them?

jeremyevans: Looks like `rdoc` can do it: `rdoc -p < SPEC.rdoc > SPEC.html`

ioquatix: Okay, I will take this course of action.

ioquatix:
Okay, this PR is updated and rebased on top of changes to the SPEC as discussed:

- https://github.com/rack/rack/pull/2290
- https://github.com/rack/rack/pull/2291
- https://github.com/rack/rack/pull/2296
- https://github.com/rack/rack/pull/2298
- https://github.com/rack/rack/pull/2299

As such, it's now mostly just reorganizes and rewrites existing documentation as discussed in the summary.

SPEC.rdoc
```diff
@@ -135,80 +197,71 @@ The special response header +rack.hijack+ must only be set if the request +env+
...
 
-If the +rack.protocol+ header is present, it must be a +String+, and must be one of the values from the +rack.protocol+ array from the environment.
+If the <tt>rack.protocol</tt> header is present, it must be a +String+, and must be one of the values from the +rack.protocol+ array from the environment.
 
 Setting this value informs the server that it should perform a connection upgrade. In HTTP/1, this is done using the +upgrade+ header. In HTTP/2, this is done by accepting the request.
 
 === The Body
 
-The Body is typically an +Array+ of +String+ instances, an enumerable that yields +String+ instances, a +Proc+ instance, or a File-like object.
+The Body is typically an +Array+ of +String+ values, an enumerable that yields +String+ values, a +Proc+ instance, or a File-like object.
```
jeremyevans, line 236: I think `instances` is better than `values` here, so I would revert this change.

lib/rack/lint.rb
```diff
@@ -885,12 +950,16 @@ def initialize(stream)
...
 ##
 ## == Thanks
-## Some parts of this specification are adopted from {PEP 333 – Python Web Server Gateway Interface v1.0}[https://peps.python.org/pep-0333/] I'd like to thank everyone involved in that effort.
+##
+## We'd like to thank everyone who has contributed to the Rack project over the years. Your work has made this specification possible. That includes everyone who has contributed code, documentation, bug reports, and feedback. We'd also like to thank the authors of the various web servers, frameworks, and libraries that have implemented the Rack specification. Your work has helped to make the web a better place.
+##
+## Some parts of this specification are adapted from {PEP 333 – Python Web Server Gateway Interface v1.0}[https://peps.python.org/pep-0333/]. We'd like to thank everyone involved in that effort.
+##
+
+# :startdoc:
```
jeremyevans, line 965:
```suggestion
```

AFAIK, startdoc only applies to the rest of the file, so it has no affect at the end of the file.  However, maybe there is a reason this was added?

.rdoc_options
```diff
@@ -0,0 +1,26 @@
+---
```
jeremyevans, line 1: Can we please limit this file to only what we feel is important?  For example:

```
autolink_excluded_words:
 - Rack
```

Seems like a good idea.  However, I don't think `tab_width: 8` is, as `CHANGELOG.md` seems like the only non-binary file that has tab characters, and even if we did want to treat tabs specially, typical Ruby tab width (indentation level) is 2, not 8.

jeremyevans (commented):
Apologies for the long delay in review.  In general, I think this is significant improvement.  Please see the inline comments.  Note that I made most of them to SPEC.rdoc as that was easiest to review, even though any changes need to be made to lint.rb.  If you find that complicates your review/implementation of the requested changes, please let me know and I can put together a PR that builds on this one and adds the changes I'm recommending.

.rdoc_options
ioquatix: I generated the default file using Doc's own command, and that's what it produced. I am happy to scope this down.

ioquatix:
@jeremyevans thanks for your excellent and detailed feedback.

lib/rack/lint.rb
ioquatix: I agree with you, but I also found this: https://github.com/ruby/rdoc/issues/398

SPEC.rdoc
```diff
@@ -1,101 +1,171 @@
...
+A Rack application is a Ruby object that responds to +call+. It takes exactly one argument, the +environment+ and returns a non-frozen +Array+ of exactly three values: the +status+, the +headers+, and the +body+.
+
+A Rack application is typically invoked by a Rack compatible server, which constructs an environment from the incoming HTTP request and calls the application with the environment. The application then processes the request and returns a response. The server translates the response into an HTTP response and sends it back to the client.
+
+  run do |env|
+    [200, {'content-type' => 'text/plain'}, ['Hello Rack']]
+  end
+
+Conceptually, the Rack application functions as an HTTP proxy: it accepts an HTTP request, forwards it to an application, and then returns the application's response to the client.
+
```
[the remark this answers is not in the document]
ioquatix: I will remove this, but I don't fully agree with removing it.

I understand your goal to be concise, but in technical writing, sometimes repeating the same thing in slightly different ways or angles is helpful to scaffold people's understanding of what something is and why it's important, etc. Therefore, this part of the documentation was deliberately designed to repeat what is already said, and to provide an extremely high level overview of what is then discussed (repeated) in detail later in the document.

SPEC.rdoc
```diff
@@ -1,101 +1,171 @@
...
+== The Request
```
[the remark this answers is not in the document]
ioquatix: Most people conceptualise HTTP as a "Request" and "Response". The documentation should clearly reflect that. The word environment is from old school CGI `getenv` and is not relevant to the modern semantics of HTTP. I understand Rack still models this conceptually using an environment (hash map of strings), but I don't think this is a useful modern taxonomy. I'm sort of okay with "The Request Environment" but it's pretty wordy and this section is about the entire request handling, not just the environment. So, I'd prefer to leave this as just "The Request".

SPEC.rdoc
```diff
@@ -1,101 +1,171 @@
...
 
-  If +length+ is given and not nil, then this method reads at most +length+ bytes from the input stream.
+For compatibility with the CGI specifiction, the environment must not contain the keys <tt>HTTP_CONTENT_TYPE</tt> or <tt>HTTP_CONTENT_LENGTH</tt>. Instead, the keys <tt>CONTENT_TYPE</tt> and <tt>CONTENT_LENGTH</tt> must be used.
 
-  If +length+ is not given or nil, then this method reads all data until EOF.
+=== Rack-Specific Variables
 
-  When EOF is reached, this method returns nil if +length+ is given and not nil, or "" if +length+ is not given or is nil.
+In addition to CGI variables, the Rack environment includes Rack-specific variables. These variables are prefixed with <tt>rack.</tt> and are reserved for use with the Rack specification.
```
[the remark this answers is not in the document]
ioquatix: The previous wording was:

```
         ## The server or the application can store their own data in the environment, too.  The keys must contain at least one dot, and should be prefixed uniquely.  The prefix <tt>rack.</tt> is reserved for use with the Rack core distribution and other accepted specifications and must not be used otherwise.
```

How about:

```
        ## In addition to CGI variables, the Rack environment includes Rack-specific variables. These variables are prefixed with <tt>rack.</tt> and are reserved for use with the Rack core distribution and other accepted specifications and must not be used otherwise.
```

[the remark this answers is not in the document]
ioquatix: Agreed.

SPEC.rdoc
```diff
@@ -1,101 +1,171 @@
...
 
-  If +buffer+ is given, then the read data will be placed into +buffer+ instead of a newly created String object.
+==== +rack.url_scheme+
```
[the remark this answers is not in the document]
ioquatix: I'm basically okay with that. However, let's do it in a follow up PR.

SPEC.rdoc
```diff
@@ -1,101 +1,171 @@
...
+
+If present, the input stream. See the section below on the input stream for more information.
+
+==== +rack.errors+
+
+The error stream. See the section below on the error stream for more information.
+
+==== +rack.response_finished+
+
+If present, an array of callables that will be run by the server after the response has been processed. This would typically be invoked after sending the response to the client, but it could also be invoked if an error occurs while generating the response or sending the response; in that case, the error argument will be a subclass of +Exception+. The callables are invoked with <tt>environment, status, headers, error</tt> arguments and should not raise any exceptions. They should be invoked in reverse order of registration.
```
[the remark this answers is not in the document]
ioquatix: It is way better.

SPEC.rdoc
```diff
@@ -135,80 +197,71 @@ The special response header +rack.hijack+ must only be set if the request +env+
...
 
 ==== The +content-length+ Header
 
-There must not be a <tt>content-length</tt> header key when the +Status+ is 1xx, 204, or 304.
+There must not be a <tt>content-length</tt> header key when the status is 1xx, 204, or 304.
 
 ==== The +rack.protocol+ Header
 
-If the +rack.protocol+ header is present, it must be a +String+, and must be one of the values from the +rack.protocol+ array from the environment.
+If the <tt>rack.protocol</tt> header is present, it must be a +String+, and must be one of the values from the +rack.protocol+ array from the environment.
```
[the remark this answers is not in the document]
ioquatix: I believe `+` is better for Ruby code and `<tt>` is for plain "use teletype". I've been trying to be more consistent with it.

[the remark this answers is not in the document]
ioquatix: Actually, now that I look at the document, I see most of the constants are using `+` so I'm not sure any more. Do you have any advice?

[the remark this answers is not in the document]
ioquatix: Okay, I've changed instances of `+rack.whatever+` to be `<tt>rack.whatever</tt>` consistently.

SPEC.rdoc
ioquatix: I'd just like to be consistent, as I think whatever we do here, as long as it's consistent.

SPEC.rdoc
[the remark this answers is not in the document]
jeremyevans: This section is about the environment's required and optional keys and their values. If you are sort of okay with "The Request Environment", then please change it to that.

SPEC.rdoc
```diff
@@ -1,103 +1,165 @@
...
+
+If present, the input stream. See the section below on the input stream for more information.
+
+==== <tt>rack.errors</tt>
+
+The error stream. See the section below on the error stream for more information.
+
+==== <tt>rack.response_finished</tt>
+
+If present, an array of callables that will be run by the server after the response has been processed. The callables are called with <tt>environment, status, headers, error</tt> arguments and should not raise any exceptions. The callables would typically be called after sending the response to the client, but it could also be called if an error occurs while generating the response or sending the response (in that case, the +error+ argument will be a kind of +Exception+). The callables will be invoked in reverse order.
```
jeremyevans, line 143:
```suggestion
If present, an array of callables that will be run by the server after the response has been processed. The callables are called with <tt>environment, status, headers, error</tt> arguments and should not raise any exceptions. The callables would typically be called after sending the response to the client, but it could also be called if an error occurs while generating the response or sending the response (in that case, the +error+ argument will be a kind of +Exception+). The callables will be called in reverse order.
```

SPEC.rdoc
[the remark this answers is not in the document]
jeremyevans: I believe `+` is preferred in cases where it works.  `<tt>` should be used in cases where `+` does not generate the appropriate markup.

SPEC.rdoc
jeremyevans: It's still like this to change back to `instances`.

jeremyevans (commented):
Only a few minor comments left.

SPEC.rdoc
ioquatix: The original documentation used the word invoked and I am okay with not changing it as the word invoked is used extensively throughout the document. In addition, I find "The callable will be called" clumsy. If you want to change this, I suggest you follow up with a PR and make it consistent everywhere.

ioquatix:
> It's still like this to change back to instances.

"+String+ value" is used in 9 places, while after this PR, there are no instances of "+String+ instance". So I think we should standardise on "+String+ value" - or in general, value refers to a primitive data type. If you feel strongly about it, why not open a new PR after this one to standardise it one way or another?

ioquatix:
> I believe + is preferred in cases where it works.  <tt> should be used in cases where + does not generate the appropriate markup.

I experimented with this and I found `+...+` hard to reason about. It seems like if there are spaces, it doesn't work? But there are also some patterns like `+<<+` that don't work either. The adhoc usage of `+` and `<tt>` bothers me slightly.

jeremyevans (approved):

ioquatix:
Thanks @jeremyevans - as discussed, I hope we follow up with additional PRs to improve `lint.rb`. I'm happy we could get this over the finish line, but I also don't see it as the end state - let's keep on improving it, and a lot of your feedback above is good.

SPEC.rdoc
```diff
@@ -1,101 +1,171 @@
...
 
-The input stream is an IO-like object which contains the raw HTTP POST data.
-When applicable, its external encoding must be "ASCII-8BIT" and it must be opened in binary mode.
-The input stream must respond to +gets+, +each+, and +read+.
-* +gets+ must be called without arguments and return a string, or +nil+ on EOF.
-* +read+ behaves like <tt>IO#read</tt>.
-  Its signature is <tt>read([length, [buffer]])</tt>.
+== The Application
+
+A Rack application is a Ruby object that responds to +call+. It takes exactly one argument, the +environment+ and returns a non-frozen +Array+ of exactly three values: the +status+, the +headers+, and the +body+.
```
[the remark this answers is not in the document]
ioquatix: Done.

SPEC.rdoc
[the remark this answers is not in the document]
ioquatix: Okay, I might take a 2nd pass at this in a follow up PR.

ioquatix:
I am going to open some issues for the follow up discussions.

PR #2257, opened 2024-10-30
consistent clean_path_info separator

akostadinov:
It is inconsistent to prepend `/` while earlier `clean.join(::File::SEPARATOR)`.

Additionally request `PATH_INFO` should always start with a slash so we
can remove the conditional prepending and just always prepend.

This can also prevent path traversal attacks in case some code checks
the request path beginning using string comparison or regular expression
but the underlying server implementation allowed a request path without
a leading slash through.

I can say that at least puma fails fast on request paths without a
leading slash but I see no harm in being extra sure.

P.S. actually I'd be even happier to make it `join("/")` instead of `clean.join(::File::SEPARATOR)`. Is there any arch that would need a different separator than `/`?

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -607,9 +607,7 @@ module Rack
         part == '..' ? clean.pop : clean << part
       end
 
-      clean_path = clean.join(::File::SEPARATOR)
-      clean_path.prepend("/") if parts.empty? || parts.first.empty?
-      clean_path
+      "#{::File::SEPARATOR}#{clean.join(::File::SEPARATOR)}"
     end
 
     NULL_BYTE = "\0"


lib/rack/utils.rb
```diff
@@ -607,9 +607,7 @@ def clean_path_info(path_info)
         part == '..' ? clean.pop : clean << part
       end
 
-      clean_path = clean.join(::File::SEPARATOR)
-      clean_path.prepend("/") if parts.empty? || parts.first.empty?
-      clean_path
+      "#{::File::SEPARATOR}#{clean.join(::File::SEPARATOR)}"
```
jeremyevans, line 610: This results in different behavior for relative paths (e.g. `"a"`, `"b/a"`).  Previously, `clean_path_info` did not prepend a slash, and now it would. For that reason, I doubt it would be acceptable.

Looking at the history, it appears to be called with `PATH_INFO`, which generally should be empty or start with a slash.  However, the specs have tests with relative paths.  Those tests happen to work with this, but that seems like an accident and not deliberate.  For example, there is a test for `"test/.."`, but not a test for `"test"` by itself.

As we cannot be sure how users are using `clean_path_info`, I don't think we should accept the behavior change.

akostadinov: Could you say in what case `PATH_INFO` would be empty or relative? What valid HTTP request will make it empty?

I think it is important for this method to be as safe as possible for the intended purpose - to make the PATH_INFO predictable to compare or use regular expression on. Rather than some unintended use cases.

Presently just figuring out what would be the actual path of a request is rather hard. And it should not be hard to do it right.

This means, as I mentioned to prevent https://owasp.org/www-community/attacks/Path_Traversal

akostadinov: HTTP 1.1 supports only (https://datatracker.ietf.org/doc/html/rfc9112#name-request-target) 
* absolute path
* full URL (in which case server sets PATH_INFO to an absolute path)
* host:port (in CONNECT requests but we are not a proxy)
* `*` (single star for server global OPTIONS request, but we can handle this as an individual check)

So basically I see no valid use cases where the result should be a relative path.

ioquatix: Just to clarify, we do check those four scenarios here: https://github.com/rack/rack/blob/fc541f3f443ed170b9b1daed4d16c6407a12573d/lib/rack/lint.rb#L364-L389

akostadinov: @ioquatix , I wonder why empty string is accepted. RFC clearly reads:
>  If the target URI's path component is empty, the client MUST send "/" as the path within the origin-form of request-target.

akostadinov: Additionally the RFC states:
> Recipients of an invalid request-line SHOULD respond with either a 400 (Bad Request) error or a 301 (Moved Permanently) redirect with the request-target properly encoded. A recipient SHOULD NOT attempt to autocorrect and then process the request without a redirect, since the invalid request-line might be deliberately crafted to bypass security filters along the request chain.

Which is to a big extend my concern.

ioquatix:
We can check the reason why this method exists: https://github.com/rack/rack/commit/9a74ba3b04f2dabe4741d2d82eae723d440c3aa2

The point is to convert a `PATH_INFO` to a file path.

The usage is something like this:

```ruby
@path = F.join(@root, clean_path_info)
```

There are some issues IMHO:

1. `SEPS = Regexp.union(*[::File::SEPARATOR, ::File::ALT_SEPARATOR].compact)` seems extremely odd for `PATH_INFO` as it would definitely follow the URL specification. I'd be concerned that `PATH_INFO` interpreted with `ALT_SEPARATOR` can be used for cache poisoning attacks.
2. It was originally `::File.join(*clean)` and was changed here to `Array#join` due to performance: https://github.com/rack/rack/commit/15ba6c7c223d6f581002d38af64c668a71c32678 but in all cases had logic to the effect of `clean.unshift '/' if parts.empty? || parts.first.empty?`.

I find it odd that it returns what effectively looks like an absolute path, yet it's treated as a relative path. I think this is error prone, e.g.

```ruby
File.join("/srv/http/mysite/public", "/background.png")
# => "/srv/http/mysite/public/background.png"

File.expand_path("/background.png", "/srv/http/mysite/public")
# => "/background.png"
```

In other words, it looks easy to make a mistake and get the wrong behaviour, IOW, I think the way the method works is inherently unsafe.

I think the safe solution is to pass in the full root, e.g.

```ruby
Rack::Utils.full_path(root, env["PATH_INFO"]) -> file path
```

For security reasons, we should assert that the result still has `root` as a prefix. I don't have a strong opinion about the naming conventions. And I could be convinced that the current behaviour is okay.

ioquatix:
I'm going to close this PR in favour of maybe doing something in https://github.com/rack/rack/pull/2302

PR #2259, opened 2024-11-08
Avoid wrapping nil input

tt:
Rack ships with the `Rack::RewindableInput::Middleware` middleware for compatibility with older apps.

However, this middleware wraps `rack.input` unconditionally meaning that a nil input will also be wrapped potentially leading to exceptions.

Concretely, I have an older app with a middleware that's trying to `req.body&.read` and `req.body&.rewind` which for `GET` and `HEAD` requests now cause the following error:

```
NoMethodError: undefined method `read' for nil
   ~/.gem/ruby/3.3.6/gems/rack-3.1.8/lib/rack/rewindable_input.rb:96:in `make_rewindable'
   ~/.gem/ruby/3.3.6/gems/rack-3.1.8/lib/rack/rewindable_input.rb:41:in `read'
```

This middleware should probably interrogate `req.request_method`.

Even so, it seems that for backwards compatibility (and to avoid the possibility of runtime errors overall), it would be great to never wrap a nil value.

### DIFF
--- a/lib/rack/rewindable_input.rb
+++ b/lib/rack/rewindable_input.rb
@@ -21,7 +21,10 @@ module Rack
       end
 
       def call(env)
-        env[RACK_INPUT] = RewindableInput.new(env[RACK_INPUT])
+        if (input = env[RACK_INPUT])
+          env[RACK_INPUT] = RewindableInput.new(input)
+        end
+
         @app.call(env)
       end
     end
--- a/test/spec_rewindable_input.rb
+++ b/test/spec_rewindable_input.rb
@@ -172,4 +172,11 @@ describe Rack::RewindableInput::Middleware do
     app = Rack::RewindableInput::Middleware.new(app)
     app.call('rack.input'=>StringIO.new(''))[2].must_equal ['Rack::RewindableInput']
   end
+
+  it "preserves a nil rack.input" do
+    app = proc{|env| [200, {}, [env['rack.input'].class.to_s]]}
+    app.call('rack.input'=>nil)[2].must_equal ['NilClass']
+    app = Rack::RewindableInput::Middleware.new(app)
+    app.call('rack.input'=>nil)[2].must_equal ['NilClass']
+  end
 end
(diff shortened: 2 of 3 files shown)


test/spec_rewindable_input.rb
```diff
@@ -167,9 +167,16 @@ def @io.rewind
 
 describe Rack::RewindableInput::Middleware do
   it "wraps rack.input in RewindableInput" do
-    app = proc{|env| [200, {}, [env['rack.input'].class.to_s]]}
-    app.call('rack.input'=>StringIO.new(''))[2].must_equal ['StringIO']
+    app = proc { |env| [200, {}, [env['rack.input'].class.to_s]] }
+    app.call('rack.input' => StringIO.new(''))[2].must_equal ['StringIO']
     app = Rack::RewindableInput::Middleware.new(app)
-    app.call('rack.input'=>StringIO.new(''))[2].must_equal ['Rack::RewindableInput']
+    app.call('rack.input' => StringIO.new(''))[2].must_equal ['Rack::RewindableInput']
```
tt, line 173: Happy to undo these changes if desired and potentially also drop the spaces in the new test. It looked like newer code preferred this syntax.

lib/rack/rewindable_input.rb
```diff
@@ -21,7 +21,7 @@ def initialize(app)
       end
 
       def call(env)
-        env[RACK_INPUT] = RewindableInput.new(env[RACK_INPUT])
+        env[RACK_INPUT] = RewindableInput.new(env[RACK_INPUT]) unless env[RACK_INPUT].nil?
```
ioquatix, line 24:
```suggestion
        if input = env[RACK_INPUT]
          env[RACK_INPUT] = RewindableInput.new(input)
        end
```

How about avoiding extra hash lookups?

test/spec_rewindable_input.rb
ioquatix: For the sake of back porting and review, please don't include unrelated changes.

tt: @ioquatix, want me to drop spaces in the new test for consistency inside this file then?

tt: I removed the spaces. It does add a couple new RuboCop offenses so let me know if you prefer it to be different.

ioquatix (approved):
This looks okay to me.

jeremyevans (approved):

ioquatix:
Thanks for your contribution. I hope we can release this in Rack 3.2 sometime before the end of this year.

Issue #2250, opened 2024-10-13
Headers with parameters which lack `=` raise a NoMethodError


AllyMarthaJ:
Hiya,

We're encountering an issue where we're receiving an unexpected error for a malformed header that we'd otherwise expect to default to something sensible. Take, for example, the following request:

```curl
curl -X POST https://somerackserver/rack -H "Content-Type:application/json;charset=" -d "{}" -vv
```

**ETA** we later realised down below the header being sent had `charset` not `charset=`

This is definitively malformed, but I'd have expected either `charset` to be `nil` or an empty string. 

In particular, the error comes from:
```
[
  "/usr/local/lib64/ruby/gems/3.3.0/gems/rack-3.1.7/lib/rack/media_type.rb:44:in `strip_doublequotes'",
  "/usr/local/lib64/ruby/gems/3.3.0/gems/rack-3.1.7/lib/rack/media_type.rb:37:in `block in params'",
  "/usr/local/lib64/ruby/gems/3.3.0/gems/rack-3.1.7/lib/rack/media_type.rb:33:in `each'",
  "/usr/local/lib64/ruby/gems/3.3.0/gems/rack-3.1.7/lib/rack/media_type.rb:33:in `each_with_object'",
  "/usr/local/lib64/ruby/gems/3.3.0/gems/rack-3.1.7/lib/rack/media_type.rb:33:in `params'",
  "/usr/local/lib64/ruby/gems/3.3.0/gems/rack-3.1.7/lib/rack/request.rb:451:in `media_type_params'",
  "/usr/local/lib64/ruby/gems/3.3.0/gems/rack-3.1.7/lib/rack/request.rb:459:in `content_charset'",
...
```

with the error `#<NoMethodError: No method 'start_with?' on instance of NilClass`

I'm happy to PR a fix to add guard for `nil` when calling `strip_doublequotes` here if that's all good!

Thanks!

dentarg:
How to reproduce?

I tried this, but I don't see any problem.

```ruby
$ cat rack3.ru
app = lambda do |env|
  req = Rack::Request.new(env)
  pp req.content_type
  [200, {}, ["OK"]]
end

run app
```

```ruby
$ cat rack3.gemfile.lock
GEM
  remote: https://rubygems.org/
  specs:
    nio4r (2.7.3)
    puma (6.4.3)
      nio4r (~> 2.0)
    rack (3.1.7)

PLATFORMS
  arm64-darwin-21
  ruby

DEPENDENCIES
[...]

PR #2263, opened 2024-11-23
fix: malformed charset param

AllyMarthaJ:
If a content type header is provided with a malformed part, there's no nice way of interpreting it since instead of being a key/value pair it's just a key.

I think the sensible option here is to assume, like in the case where a param is provided with an empty value, that the value is an empty string. I'm not super tied to this, though (for example, it could just as easily be omitted from the params entirely, or the value could simply be nil; each option are equally as valid).

issue: https://github.com/rack/rack/issues/2250

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file. For info on
 
 - Invalid cookie keys will now raise an error. ([#2193](https://github.com/rack/rack/pull/2193), [@ioquatix])
 - `Rack::MediaType#params` now handles empty strings. ([#2229](https://github.com/rack/rack/pull/2229), [@jeremyevans])
+- `Rack::MediaType#params` now handles parameters without values. ([#2263](https://github.com/rack/rack/pull/2263), [@AllyMarthaJ](https://github.com/AllyMarthaJ))
 
 ### Deprecated
 
--- a/lib/rack/media_type.rb
+++ b/lib/rack/media_type.rb
@@ -26,6 +26,11 @@ module Rack
       # provided.  e.g., when the CONTENT_TYPE is "text/plain;charset=utf-8",
       # this method responds with the following Hash:
       #   { 'charset' => 'utf-8' }
+      #
+      # This will pass back parameters with empty strings in the hash if they
+      # lack a value (e.g., "text/plain;charset=" will return { 'charset' => '' },
+      # and "text/plain;charset" will return { 'charset' => '' }, similarly to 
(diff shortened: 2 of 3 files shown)


jeremyevans (approved):
Having it be an empty string is consistent with what we do for parameters themselves with nothing after the `=` ( e.g. `?key=`), so that seems like the best way.

Could be useful to update CHANGELOG in this PR, but that can always be done later.

ioquatix:
Could you please add documentation somewhere explaining the behaviour?

AllyMarthaJ:
Done 🙂

ioquatix:
Thanks.

Just to be clear, this isn't the same as query parsing:

```ruby
> Rack::Utils.parse_query("charset")
=> {"charset"=>nil}
```

I think @jeremyevans prefers to use empty string, but this isn't consistent with the current implementation (which uses `nil`).

WDYT?

AllyMarthaJ:
Huh. Okay. It might be slightly cheaper to cleave off the `|| ''` and make it actually in line with query parsing 😛 

Does anyone in reality make the distinction between "present, but unspecified" and "present, but empty" for this kind of parameter? I...don't think the spec cares about it.

ioquatix:
I personally think it's good to be consistent with the query parser, but that's just me. @jeremyevans may have a different opinion.

jeremyevans:
The description in #2250 is `Headers with = in them without a trailing parameter result in a NoMethodError`. Therefore, I assumed this behavior was the `charset=` case and not the `charset` case, in which case the empty string would be consistent with query parsing.

Note that using an empty string for both cases (with or without `=`) for charset params seems fine to me.  Absent backwards compatibility issues, that is how we would treat query parsing, and this doesn't have the backwards compatibility issues.

AllyMarthaJ:
Yeah, apologies — the original issue's repro was wrong (I'll update that now), which is why it sat there for so long 😅
I'll update comments.

AllyMarthaJ:
Alrighty, I've opted to go with empty strings for both, and I've updated the doc.
Any chance we're moving to go with the same behaviour for the query parser in the future, or have we missed out on that one? 😛

ioquatix:
I don't think the behaviour of the query parser can be changed.

ioquatix:
Thanks for your contribution.

gstokkink:
@ioquatix can we get a release for this fix please? We're running into this on production 😄 Much appreciated and thanks for the hard work on this awesome gem!

gstokkink:
@ioquatix thanks for releasing, much appreciated! 🙇

PR #2268, opened 2024-12-11
`Rack::Lint.new(nil)` is invalid.

ioquatix:
Replace it with `Rack::Lint.new(valid_app)` and enforce this in `Rack::Lint#initialize(app)`.

### DIFF
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -18,10 +18,6 @@ module Rack
     REQUEST_PATH_AUTHORITY_FORM = /\A[^\/:]+:\d+\z/
     REQUEST_PATH_ASTERISK_FORM = '*'
 
-    def initialize(app)
-      @app = app
-    end
-
     # :stopdoc:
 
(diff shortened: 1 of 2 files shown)


lib/rack/lint.rb
```diff
@@ -39,6 +35,12 @@ class LintError < RuntimeError; end
     ##
     ## A Rack application is a Ruby object (not a class) that
     ## responds to +call+.
+    def initialize(app)
+      raise LintError, "app must respond to call" unless app.respond_to?(:call)
```
jeremyevans, line 39: You don't need to move the definition for things to work, though it's not a problem to move it, just more code churn.

jeremyevans (approved):
Looks good.  One minor suggestion, up to you.

lib/rack/lint.rb
ioquatix: I understand your concern. The reason for the move is so that the specification documentation and the actual check are proximal.

PR #2269, opened 2024-12-17
Remove autoloads for constants no longer shipped with rack

ccutrer:
Removes the following autoloads:

* Handler
* Server

Sort the remaining autoloads alphabetically.

Refs #2112


jeremyevans (approved):
JRuby failure is unrelated.  @ioquatix, are you aware of any backwards compatibility issues with this in regards to rackup?  `rack/handler` and `rack/server` are not in rack 3.0, so without rackup, referencing Rack::Handler or Rack::Server would fail anyway.  However, maybe this will break cases where rackup is used and the user is referencing Rack::Handler and Rack::Server?

ioquatix:
For the sake of clarity and focus, I would personally prefer not to backport cosmetic changes.

I don't see any problem with this, as those files don't exist anyway. Trying to use those constants would be an error.

It is true, those autoloads may inadvertently load files from another gem, e.g. older versions of the `rackup` gem had `rack/server` and so on. But they have emitted warnings for over a year now.

I'm good to merge, but @ccutrer can you consider only removing the relevant lines and avoiding cosmetic changes? If you feel strongly about it, just explain your reasoning and we can merge as is probably.

ccutrer:
> can you consider only removing the relevant lines and avoiding cosmetic changes?

You mean not re-sorting? I'm happy to do that. I just cherry-picked the commit to main (and resolved conflicts), so they came along. Just confirm that's what you mean, and I'm happy to do it.



> Trying to use those constants would be an error either way

That's the rub. If I'm using rack 3.0.11 with rackup 2.2.1, _and_ pact-mock_service 3.12.3, the latter [tries to be compatible with both rackup 2.2.1 and 2.1.x and older](https://github.com/pact-foundation/pact-mock_service/blob/v3.12.3/lib/rack/handler/webbrick.rb#L2), but as soon as it tries to _define_ `Rack::Handler` (on line 2), Ruby tries to autoload "rack/handler", which doesn't exist, and I'm stuck. Really that code is poorly written - the require doesn't need to be nested inside the module definition, only the assigning of the WEBrick constant. Perhaps I'll send them a PR as well...

ioquatix:
> You mean not re-sorting? I'm happy to do that. I just cherry-picked the commit to main (and resolved conflicts), so they came along. Just confirm that's what you mean, and I'm happy to do it.

Yes - however do you mind linking me to the cherry-picked commit too, for reference.

> Really that code is poorly written - the require doesn't need to be nested inside the module definition, only the assigning of the WEBrick constant. Perhaps I'll send them a PR as well...

I wrote this documentation which may help clarify how it should work: https://github.com/rack/rackup?tab=readme-ov-file#server-handler

ioquatix:
I checked https://github.com/jeremyevans/rack/commit/c9198625dbce22c840dfe6256f3e60db98991604 and see that the constants were also sorted. Therefore, since that ship has already sailed, I'm fine to merge this PR as it will improve the consistency of the stable branches.

ioquatix:
Thanks for your contribution, I'll push this out in a patch release (hopefully today).

PR #2273, opened 2025-01-06
Handle empty content-type parts

redox:
Hi,

we stumbled upon an issue with our Puma/rack stack where a content-type that seems to be valid (from a Cloudflare point-of-view at least - looks like it's not while looking at the [RFC](https://www.w3.org/Protocols/rfc1341/4_Content-Type.html)) ends up doing a `nil.downcase!`: `application/text;;charset=utf-8` (note the double `;`).

I'm unclear whether it's on `Rack` to validate the content-type, but at least we should not crash.


AllyMarthaJ:
Is this any different from https://github.com/rack/rack/pull/2263?

redox:
> Is this any different from #2263?

No, great to see another MR addressing the same issue has been merged!

AllyMarthaJ:
Wonderful :))

Fingers crossed on a release!

PR #2278, opened 2025-02-04
Make Rack::Deflater handle invalid last-modified header

### LINKED ISSUE #1737: Time header issue in deflater
I'm running into a weird bug with Rack::Deflater throwing this error:

ArgumentError not RFC 2616 compliant date: "Tue, 25 Feb 2020 11:51:53 +0000"

From

rack (2.2.3) lib/rack/deflater.rb in call at line 67

It appears to be related to this Ruby bug https://bugs.ruby-lang.org/issues/10733, with resolution to be to always force the header string to be UTF-8

"Mon, 19 Jan 2015 08:43:19 GMT".encode 'UTF-8'

It seems to only happen when requests are coming through the Cloudflare proxy/cdn. For now I've removed rack deflater from my stack. More details on the exception can be found in Sentry:

https://sentry.io/share/issue/b3ba8ace0b3a4143aea925f81fac47a6/

Let me know if you'd like me to test a fork of this with that encoding on the header to see if it resolves the issue.

jeremyevans:
Fixes #1737

### DIFF
--- a/lib/rack/deflater.rb
+++ b/lib/rack/deflater.rb
@@ -66,7 +66,13 @@ module Rack
         headers['content-encoding'] = "gzip"
         headers.delete(CONTENT_LENGTH)
         mtime = headers["last-modified"]
-        mtime = Time.httpdate(mtime).to_i if mtime
+        if mtime
+          begin
+            mtime = Time.httpdate(mtime).to_i
+          rescue ArgumentError
+            mtime = nil
+          end
+        end
         response[2] = GzipStream.new(body, mtime, @sync)
         response
       when "identity"
--- a/test/spec_deflater.rb
+++ b/test/spec_deflater.rb
@@ -75,8 +75,16 @@ describe Rack::Deflater do
         gz = Zlib::GzipReader.new(io)
         mtime = gz.mtime.to_i
         if last_mod = headers['last-modified']
-          Time.httpdate(last_mod).to_i.must_equal mtime
-        else
+          begin
+            header_mtime = Time.httpdate(last_mod).to_i
+          rescue ArgumentError
+            # nothing
+          else
+            header_mtime.must_equal mtime
+            mtime = nil
+          end
+        end
+        if mtime
           mtime.must_be(:<=, Time.now.to_i)
           mtime.must_be(:>=, start.to_i - 1)
         end
@@ -310,8 +318,27 @@ describe Rack::Deflater do
     closed.must_equal true
   end
 
-  it 'handle gzip response with last-modified header' do
+  it 'handle gzip response with valid last-modified header' do
     last_modified = Time.now.httpdate
+    options = {
+      'response_headers' => {
+        'content-type' => 'text/plain',
+        'last-modified' => last_modified
+      }
+    }
+
+    verify(200, 'Hello World!', 'gzip', options) do |status, headers, body|
+      headers.must_equal({
+        'content-encoding' => 'gzip',
+        'vary' => 'Accept-Encoding',
+        'last-modified' => last_modified,
+        'content-type' => 'text/plain'
+      })
+    end
+  end
+
+  it 'handle gzip response with invalid last-modified header' do
+    last_modified = "Xyz, 27 Dez 2024 18:26:57 GMT"
     options = {
       'response_headers' => {
         'content-type' => 'text/plain',
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ All notable changes to this project will be documented in this file. For info on
 ### Fixed
 
 - `Rack::RewindableInput::Middleware` no longer wraps a nil input. ([#2259](https://github.com/rack/rack/pull/2259), [@tt](https://github.com/tt))
+- `Rack::Deflater` no longer fails if the application returns an invalid last-modified header. ([#1737](https://github.com/rack/rack/issues/1737), [@jeremyevans])
 
 ## [3.1.9] - 2025-01-31
 


lib/rack/deflater.rb
```diff
@@ -66,7 +66,13 @@ def call(env)
         headers['content-encoding'] = "gzip"
         headers.delete(CONTENT_LENGTH)
         mtime = headers["last-modified"]
-        mtime = Time.httpdate(mtime).to_i if mtime
+        if mtime
```
thadeu, line 69: Is this conditional could be a new method or helper module?

```rb
module TimeHelper
  def httpdate(mtime = nil)
    return if mtime.nil?

    Time.httpdate(mtime) rescue nil
  end
end

mtime = TimeHelper.httpdate(headers['last-modified'])
```

test/spec_deflater.rb
```diff
@@ -75,8 +75,16 @@ def verify(expected_status, expected_body, accept_encoding, options = {}, &block
         gz = Zlib::GzipReader.new(io)
         mtime = gz.mtime.to_i
         if last_mod = headers['last-modified']
-          Time.httpdate(last_mod).to_i.must_equal mtime
-        else
+          begin
```
thadeu, line 78: TimeHelper could be use here?

ioquatix:
Does this silently hide bugs in application code?

thadeu:
> Does this silently hide bugs in application code?

@ioquatix I think we have 3 ways here.

1. Last-modified exists and it's valid
2. Last-modified exists but it's not valid
3. Last-modified not exist

In the all cases, we have tests to cover that branches. So, what do you mean when you say: `"silently hide bugs"` ?

lib/rack/deflater.rb
```diff
@@ -66,7 +66,13 @@ def call(env)
...
+          begin
+            mtime = Time.httpdate(mtime).to_i
+          rescue ArgumentError
+            mtime = nil
+          end
```
ioquatix, line 74: If we are going to do this (which I'm slightly against - isn't it an application bug that needs to be fixed?), we should at least log the error, otherwise it's silently ignoring it, which I'm strongly against.

thadeu: @ioquatix Ok, I'm with you in that point. The tracking error could be help the developer to discover where they code is wrong. But without raise, otherwise the fix doesnt make sense

ioquatix:
> In the all cases, we have tests to cover that branches. So, what do you mean when you say: "silently hide bugs" ?

If we are eating the exception without logging it, that is "silently hiding bugs" assuming that generating an invalid value is an application bug.

lib/rack/deflater.rb
jeremyevans: I'm generally against unnecessary indirection, so I do not think this would improve things.

jeremyevans:
This change does not modify the `last-modified` header value, if it was invalid before, it is still invalid.  It just will not set the gzip timestamp if it is invalid.

One option is certainly to tell the user that Deflater is doing the correct thing, and they should fix the application. An alternative approach would be to make this behavior opt-in, so only people affected by this issue can choose to ignore it.

I don't have strong feelings either way.  I lean towards the behavior in this PR, because I think it's Deflater's job to deflate the body, not to be pedantic about header value formats.  If Lint doesn't reject the response, I think Deflater should be able to deal with it.

ioquatix:
My point is not so much about the functionality, but that we are adding extra complexity which sets a precedent for other middleware. I would support adding a check to `Rack::Lint` which expects the format to be correct, according to the RFCs.

jeremyevans:
> My point is not so much about the functionality, but that we are adding extra complexity which sets a precedent for other middleware. I imagine there are other places where we try to parse response (or request) headers, and I believe it's unrealistic and probably undesirable to add error handling like this to all of them.

I'm not sure it's unrealistic or undesirable, but I do think either approach (handling invalid header values or not) is reasonable.

> I would support adding a check to `Rack::Lint` which expects the format to be correct, according to the RFCs.

I think that's an even slipperier slope than adding error handling to the middleware.  If you start checking last-modified format, where do you stop?

> The application is clearly in the wrong here AFAICT.

Agreed.  I think the question is really whether middleware should be robust against invalid application response header formats.

ioquatix:
> I think that's an even slipperier slope than adding error handling to the middleware. If you start checking last-modified format, where do you stop?

You enforce what the RFCs recommend (MUST and SHOULD).

I disagree that it's even slipperier. It's either in the `Rack::Lint` once, or spread throughout every middleware that needs to deal with it.

jeremyevans:
> > I think that's an even slipperier slope than adding error handling to the middleware. If you start checking last-modified format, where do you stop?
> 
> You enforce what the RFCs recommend (MUST and SHOULD).
> 
> I disagree that it's "even slipperier". It's either in the `Rack::Lint` once, or spread throughout every middleware that needs to deal with it. I'd prefer to enforce it in one place as a standard, rather than ad-hoc throughout the entire code base + other code bases, etc.

The slipperier part is that there are a lot of headers mentioned in the RFCs, many of which the middleware that ship with Rack do not use.  Checking and enforcing the format of all of them seems like a large job.  For some headers, it's likely challenging to do the checking correctly.  Rack::Lint would get much larger than currently it is to do that.

Be aware that in most cases where Rack::Lint is used, Rack::Lint wraps a lot of outer middleware (as you want to check that those middleware are compliant with the rack SPEC), so any middleware that is wrapped by Rack::Lint would not have Rack::Lint doing the enforcement.  Instead, for invalid headers like the one in this ticket, the middleware would raise an exception before Rack::Lint handled the response.

Anyway, if you don't think this approach is the right one, and I definitely do not think we want to make changes to Rack::Lint for this, I think the safest approach is to close this and #1737, and tell users it is on them to fix the application generating the invalid headers.  Are you OK with that?

ioquatix:
Yes, this is an application bug, the application should be fixed.

ioquatix:
> The slipperier part is that there are a lot of headers mentioned in the RFCs, many of which the middleware that ship with Rack do not use. Checking and enforcing the format of all of them seems like a large job. For some headers, it's likely challenging to do the checking correctly. Rack::Lint would get much larger than currently it is to do that.

I'd actually be okay with this.

thadeu:
> Yes, this is an application bug, the application should be fixed.

@ioquatix If Last-modified header is not present, we wont use it, but if the last-modified is invalid, we will to parse it the same way? I dont agree with it. The same way that we check if the value is present, we need to check if the value is valid before to use `Time.httpdate` to them.

Rack must be intelligent to handle it anytime.

Many of applications use the Rack::Deflater to compress data, I mean, is the all of applications is broken, if the Last-modified is empty?

jeremyevans:
> > Yes, this is an application bug, the application should be fixed.
> 
> @ioquatix If Last-modified header is not present, we wont use it, but if the last-modified is invalid, we will to parse it the same way? I dont agree with it. The same way that we check if the value is present, we need to check if the value is valid before to use `Time.httpdate` to them.
> 
> Rack must be intelligent to handle it anytime.
> 
> Many of applications use the Rack::Deflater to compress data, I mean, is the all of applications is broken, if the Last-modified is empty or invalid?

Any application that returns a `last-modified` that isn't in HTTP-date format is broken.  See RFC 7232 Section 2.2: https://datatracker.ietf.org/doc/html/rfc7232#section-2.2 .  I don't think there is any requirement that Rack work around application bugs.

> Would you like to stop down many applications because last-modified is invalid?

I'm guessing you meant "slow down".  Turning this question around, do you think it is reasonable for Rack to slow down all usage of Rack::Deflater for applications with correct behavior just because your application has incorrect behavior?

Anyway, based on the discussion here, I'm closing this.

PR #2287, opened 2025-02-21
Reduce unnecessary calls to Rack::Request.ip_filter

willbryant:
Previously this was typically invoked for every address found in REMOTE_ADDR/HTTP_FORWARDED/HTTP_X_FORWARDED_FOR. When ip_filter matches against several hundred IP ranges (eg. Cloudfront IP ranges), whole ms could be wasted.

Now it'll only be invoked until one acceptable address is found.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ All notable changes to this project will be documented in this file. For info on
 
 - Invalid cookie keys will now raise an error. ([#2193](https://github.com/rack/rack/pull/2193), [@ioquatix])
 - `Rack::MediaType#params` now handles empty strings. ([#2229](https://github.com/rack/rack/pull/2229), [@jeremyevans])
+- Avoid unnecessary calls to the `ip_filter` lambda to evaluate `Request#ip` ([#2287](https://github.com/rack/rack/pull/2287), [@willbryant])
 
 ### Deprecated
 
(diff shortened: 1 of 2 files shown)


jeremyevans (approved):
Thanks for working on this!

ioquatix:
@willbryant do you mind adding an entry to the changelog?

ioquatix (approved):
I'm okay with this. Of note, is that `rejected_trusted_ip_addresses` was removed, a public interface. Of course, it's no longer useful and was probably an implementation detail, and I checked GitHub code search and could not see any usage in the wild.

willbryant:
Yeah, I did wonder about that - and came to the same conclusion, but do you think I should restore it until a future major release to be on the safe side?

ioquatix:
One option is to restore it and add a deprecation note to it. It is the safest approach. However, it also makes more work for maintainers. So, I'm okay with either option. Let's see what @jeremyevans thinks.

jeremyevans:
> I'm okay with this. Of note, is that `rejected_trusted_ip_addresses` was removed, a public interface. Of course, it's no longer useful and was probably an implementation detail, and I checked GitHub code search and could not see any usage in the wild.

Are you sure it was public?  It comes after the `private` keyword.

ioquatix:
Ah yeah you are right, it's way up there haha.

PR #2291, opened 2025-02-22
Add `assert_required` to clarify spec assertions.

ioquatix:
These are the code changes sans documentation changes from <https://github.com/rack/rack/pull/2256>.

The addition of two assertion helpers, is designed to clearly communicate the expectation of whether something is required or optional.

1. All the existing checks are semantically the same.
2. There are a few new checks as described by the documentation but not explicitly validated by the lint.
3. There are some minor cosmetic changes to the code for consistency.

### DIFF
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -99,6 +99,12 @@ module Rack
         return [@status, @headers, self]
       end
 
+      private def assert_required(key)
+        raise LintError, "env missing required key #{key}" unless @env.include?(key)
+
+        return @env[key]
+      end
+
       ##
       ## == The Environment
       ##
@@ -154,6 +160,7 @@ module Rack
         ## <tt>QUERY_STRING</tt>:: The portion of the request URL that
         ##                         follows the <tt>?</tt>, if any. May be
         ##                         empty, but is always required!
+        assert_required(QUERY_STRING)
 
         ## <tt>SERVER_NAME</tt>:: When combined with <tt>SCRIPT_NAME</tt> and
         ##                        <tt>PATH_INFO</tt>, these variables can be
@@ -297,9 +304,6 @@ module Rack
         ## is reserved for use with the Rack core distribution and other
         ## accepted specifications and must not be used otherwise.
         ##
-        %w[REQUEST_METHOD SERVER_NAME QUERY_STRING SERVER_PROTOCOL rack.errors].each do |header|
-          raise LintError, "env missing required key #{header}" unless env.include? header
-        end
 
         ## The <tt>SERVER_PORT</tt> must be an Integer if set.
         server_port = env["SERVER_PORT"]
@@ -308,7 +312,8 @@ module Rack
         end
 
         ## The <tt>SERVER_NAME</tt> must be a valid authority as defined by RFC7540.
-        unless (URI.parse("http://#{env[SERVER_NAME]}/") rescue false)
+        server_name = assert_required(SERVER_NAME)
+        unless (URI.parse("http://#{server_name}/") rescue false)
           raise LintError, "#{env[SERVER_NAME]} must be a valid authority"
         end
 
@@ -318,7 +323,7 @@ module Rack
         end
 
         ## The <tt>SERVER_PROTOCOL</tt> must match the regexp <tt>HTTP/\d(\.\d)?</tt>.
-        server_protocol = env['SERVER_PROTOCOL']
+        server_protocol = assert_required(SERVER_PROTOCOL)
         unless %r{HTTP/\d(\.\d)?}.match?(server_protocol)
           raise LintError, "env[SERVER_PROTOCOL] does not match HTTP/\\d(\\.\\d)?"
         end
@@ -335,7 +340,8 @@ module Rack
         ## There are the following restrictions:
 
         ## * <tt>rack.url_scheme</tt> must either be +http+ or +https+.
-        unless ALLOWED_SCHEMES.include?(env[RACK_URL_SCHEME])
+        rack_url_scheme = assert_required(RACK_URL_SCHEME)
+        unless ALLOWED_SCHEMES.include?(rack_url_scheme)
           raise LintError, "rack.url_scheme unknown: #{env[RACK_URL_SCHEME].inspect}"
         end
 
@@ -346,7 +352,7 @@ module Rack
         end
 
         ## * There must be a valid error stream in <tt>rack.errors</tt>.
-        rack_errors = env[RACK_ERRORS]
+        rack_errors = assert_required(RACK_ERRORS)
         check_error_stream(rack_errors)
         @env[RACK_ERRORS] = ErrorWrapper.new(rack_errors)
 
@@ -356,7 +362,8 @@ module Rack
         check_early_hints env
 
         ## * The <tt>REQUEST_METHOD</tt> must be a valid token.
-        unless env[REQUEST_METHOD] =~ /\A[0-9A-Za-z!\#$%&'*+.^_`|~-]+\z/
+        request_method = assert_required(REQUEST_METHOD)
+        unless request_method =~ /\A[0-9A-Za-z!\#$%&'*+.^_`|~-]+\z/
           raise LintError, "REQUEST_METHOD unknown: #{env[REQUEST_METHOD].dump}"
         end
 
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -118,16 +118,19 @@ module Rack
         ## The CGI keys (named without a period) must have String values.
         ## If the string values for CGI keys contain non-ASCII characters,
         ## they should use ASCII-8BIT encoding.
-        env.each { |key, value|
-          next  if key.include? "."   # Skip extensions
+        env.each do |key, value|
+          next if key.include?(".") # Skip extensions
+          
           unless value.kind_of? String
             raise LintError, "env variable #{key} has non-string value #{value.inspect}"
           end
+
           next if value.encoding == Encoding::ASCII_8BIT
+
           unless value.b !~ /[\x80-\xff]/n
             raise LintError, "env variable #{key} has value containing non-ASCII characters and has non-ASCII-8BIT encoding #{value.inspect} encoding: #{value.encoding}"
           end
-        }
+        end
 
         ##
         ## The environment is required to include these variables
@@ -212,7 +215,7 @@ module Rack
         ##                          the protocols advertised by the client in
         ##                          the +upgrade+ header (HTTP/1) or the
         ##                          +:protocol+ pseudo-header (HTTP/2).
-        if protocols = @env['rack.protocol']
+        if protocols = env[RACK_PROTOCOL]
           unless protocols.is_a?(Array) && protocols.all?{|protocol| protocol.is_a?(String)}
             raise LintError, "rack.protocol must be an Array of Strings"
           end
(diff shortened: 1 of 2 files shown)


jeremyevans (changes requested):
I'm OK with `assert_required`, but I don't think `assert_optional` adds anything.  First, it is misnamed as it does not assert anything.  Second, I think it is an unnecessary abstraction.  While it does offer a consistent API with `assert_required`, I don't think that benefit is worth the cost of the abstraction. I'd rather have the `if` be explicit than implicit.

I do not think it is a good idea to mix code changes with style/whitespace changes in the same commit.  Style/whitespace changes should be a separate commit.

SPEC.rdoc
```diff
@@ -135,6 +135,7 @@ There are the following restrictions:
   * Only <tt>CONNECT</tt> requests may have <tt>PATH_INFO</tt> set to an authority (authority-form). Note that in HTTP/2+, the authority-form is not a valid request target.
   * <tt>CONNECT</tt> and <tt>OPTIONS</tt> requests must not have <tt>PATH_INFO</tt> set to a URI (absolute-form).
   * Otherwise, <tt>PATH_INFO</tt> must start with a <tt>/</tt> and must not include a fragment part starting with '#' (origin-form).
+* The <tt>CONTENT_TYPE</tt>, if given, must be a string.
```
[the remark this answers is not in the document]
ioquatix: Originally I was validating it should be a mime type, but IIRC you didn't like that, so it got reduced to this. You are correct of course, so I'll remove it.

ioquatix:
> I don't think assert_optional adds anything. First, it is misnamed as it does not assert anything. Second, I think it is an unnecessary abstraction. While it does offer a consistent API with assert_required, I don't think that benefit is worth the cost of the abstraction. I'd rather have the if be explicit than implicit.

I appreciate your feedback and understand your concerns regarding `assert_optional`. When I first reviewed the SPEC, I also found the distinction between required and optional fields somewhat unclear. The SPEC contains statements like:

```ruby
      ## The environment is required to include these variables
      ## (adopted from PEP333), except when they'd be empty, but see
      ## below.
```

followed by additional clarifications that aren’t directly adjacent to the relevant tests. While one can infer the details with enough effort, I believe our goal should be to make the intent explicit and consistent.

My aim has been to have a single, clear specification for each field, with its corresponding assertions placed right next to the documentation. This approach not only reduces ambiguity but also helps in maintaining and verifying the intent behind each field’s behavior.

Regarding `assert_optional`, I agree that its name might be a bit misleading since it doesn’t “assert” in the traditional sense. However, the assertion is encapsulated within the block, meaning it clearly indicates, "assert this key is optional with the following tests."  In other words, the assertion includes the execution of the block which contains further specifications.

On the performance front, I ran benchmarks comparing `assert_optional(env, y) do |z|` with a straightforward `if z = env[y]` check. The results were very similar:

- Baseline (without lint): ~240,000 r/s
- `assert_optional` lint: ~79,000 r/s
- `if ...` lint: ~80,000 r/s

Given that the performance difference is negligible, I believe the consistency and clarity provided by our current approach are worth retaining.

ioquatix:
> I don't think assert_optional adds anything. First, it is misnamed as it does not assert anything. Second, I think it is an unnecessary abstraction. While it does offer a consistent API with assert_required, I don't think that benefit is worth the cost of the abstraction. I'd rather have the if be explicit than implicit.

I appreciate your feedback and understand your concerns regarding `assert_optional`. When I first reviewed the SPEC, I found the distinction between required and optional fields somewhat unclear. The SPEC contains statements like:

```ruby
      ## The environment is required to include these variables
      ## (adopted from PEP333), except when they'd be empty, but see
      ## below.
```

followed by additional clarifications that aren’t directly adjacent to the relevant tests. While one can infer the details with enough effort, I believe our goal should be to make the intent explicit and consistent.

My aim has been to have a single, clear specification for each field, with its corresponding assertions placed right next to the documentation. This approach not only reduces ambiguity but also helps in maintaining and verifying the intent behind each field’s behavior.

Regarding `assert_optional`, I agree that its name might be a bit misleading since it doesn’t “assert” in the traditional sense. However, the assertion is encapsulated within the block, meaning it clearly indicates, "assert this key is optional with the following tests."  In other words, the assertion includes the execution of the block which contains further specifications.

On the performance front, I ran benchmarks comparing `assert_optional(env, y) do |z|` with a straightforward `if z = env[y]` check. The results were very similar:

- Baseline (without lint): ~240,000 r/s
- `assert_optional` lint: ~79,000 r/s
- `if ...` lint: ~80,000 r/s

Given that the performance difference is negligible, I believe the consistency and clarity provided by the proposed approach are worth introducing. We could also clarify the logic by making the `assert_optional` block mandatory. However, if despite the above reasoning, you are still against it, I'll remove that part of the change.

Once you let me know how you want to proceed, I can split this PR into one which introduces those changes and one which has all the other cosmetic changes if that's what you are after.

<details>
<summary>Benchmark Code</summary>

```ruby
#!/usr/bin/env ruby

require "benchmark"
require "rack"

builder = Rack::Builder.new
builder.use Rack::Lint
builder.run do |env|
  [200, {"content-type" => "text/plain"}, ["Hello World"]]
end

request = Rack::MockRequest.new(builder.to_app)
response = request.get("/")

Benchmark.bm do |x|
  x.report do
    10_000.times do
      request.get("/")
    end
  end
end
```
</details>

ioquatix:
I am fine with your proposed course of action.

However, let me clarify, the point of the block is to avoid leaking local variables and keep the assertions neatly organised/self contained. I prefer the block approach for this reason.

lib/rack/lint.rb
```diff
@@ -304,14 +317,12 @@ def check_environment(env)
         ## is reserved for use with the Rack core distribution and other
         ## accepted specifications and must not be used otherwise.
         ##
-        %w[REQUEST_METHOD SERVER_NAME QUERY_STRING SERVER_PROTOCOL rack.errors].each do |header|
-          raise LintError, "env missing required key #{header}" unless env.include? header
-        end
```
jeremyevans, line 302: This should be part of the first commit, since you are replacing this with separate calls to `assert_required`.

lib/rack/lint.rb
```diff
@@ -203,6 +206,11 @@ def check_environment(env)
 
         ## <tt>rack.hijack?</tt>:: See below, if present and true, indicates
         ##                         that the server supports partial hijacking.
+        if rack_is_hijack = env[RACK_IS_HIJACK]
+          unless rack_is_hijack == true || rack_is_hijack == false
+            raise LintError, "rack.hijack? must be true or false"
+          end
```
jeremyevans, line 212: This would be a SPEC change, since previously values other than true or false were accepted. Later in SPEC, it states `If +rack.hijack?+ is present in +env+ and truthy` (note `truthy` and not `true`).  So I don't think we should make this change, at least not in this PR.  If you would like to make this SPEC change, please submit a separate PR with SPEC changes and a justification with why you think we should be more strict for `rack.hijack?`.

lib/rack/lint.rb
```diff
@@ -304,14 +317,12 @@ def check_environment(env)
...
-        %w[REQUEST_METHOD SERVER_NAME QUERY_STRING SERVER_PROTOCOL rack.errors].each do |header|
-          raise LintError, "env missing required key #{header}" unless env.include? header
-        end
 
         ## The <tt>SERVER_PORT</tt> must be an Integer if set.
-        server_port = env["SERVER_PORT"]
-        unless server_port.nil? || (Integer(server_port) rescue false)
-          raise LintError, "env[SERVER_PORT] is not an Integer"
+        if server_port = env[SERVER_PORT]
+          unless (Integer(server_port) rescue false)
```
jeremyevans, line 323: As this was the previous code, I don't think you need to change it in this PR, but I don't think this check is correct.  SPEC says `must be an Integer`, not `must be convertible to an Integer`, so the correct check here is `server_port.is_a?(Integer)`.

jeremyevans (changes requested):

lib/rack/lint.rb
```diff
@@ -403,8 +418,10 @@ def check_environment(env)
         end
 
         ## * The <tt>CONTENT_LENGTH</tt>, if given, must consist of digits only.
-        if env.include?("CONTENT_LENGTH") && env["CONTENT_LENGTH"] !~ /\A\d+\z/
-          raise LintError, "Invalid CONTENT_LENGTH: #{env["CONTENT_LENGTH"]}"
+        if content_length = env["CONTENT_LENGTH"]
+          unless (Integer(content_length) rescue false)
```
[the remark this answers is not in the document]
ioquatix: Ah yes, good call. I agree.

lib/rack/lint.rb
ioquatix: This is my fault, in the original spec for `rack.hijack?` only `true` and `false` were allowed. The word `truthy` was introduced by me: https://github.com/rack/rack/commit/3012643ea6a89fefe8cc0c68d4992531c367c906

Previously it was only `true` or `false`, e.g. https://github.com/rack/rack/blob/f393916baf057bd613290b505150b7ae5da1b9c4/lib/rack/lint.rb#L550

I am okay to do this in a separate PR, but the union of the `if present and true` and `and truthy` only allows for `true` according to the spec, so I don't see this as a breaking change or even a big compatibility issue since this is the same behaviour as Rack 2.x.

lib/rack/lint.rb
ioquatix: Good call, done.

lib/rack/lint.rb
ioquatix: I intend to rebase and merge this PR (FF), so I'll add a fix for this as a separate commit.

ioquatix: Hmm, okay, I tried this but "The <tt>SERVER_PORT</tt> must be an Integer if set." directly conflicts with all values must be strings?

jeremyevans: Interesting.  I would assume in order to pass Lint, servers are providing a string?  Puma appears to pass string, so I think we should change SPEC to state `SERVER_PORT` should be all digits as well (just like `CONTENT_LENGTH`).

lib/rack/lint.rb
```diff
@@ -308,7 +312,8 @@ def check_environment(env)
         end
 
         ## The <tt>SERVER_NAME</tt> must be a valid authority as defined by RFC7540.
-        unless (URI.parse("http://#{env[SERVER_NAME]}/") rescue false)
+        server_name = assert_required(SERVER_NAME)
+        unless (URI.parse("http://#{server_name}/") rescue false)
           raise LintError, "#{env[SERVER_NAME]} must be a valid authority"
```
jeremyevans, line 331: Up to you, but you could use `server_name` here (with similar changes it other error messages.

lib/rack/lint.rb
```diff
@@ -318,8 +332,10 @@ def check_environment(env)
         end
 
         ## The <tt>HTTP_HOST</tt> must be a valid authority as defined by RFC7540.
-        unless (URI.parse("http://#{env[HTTP_HOST]}/") rescue false)
-          raise LintError, "#{env[HTTP_HOST]} must be a valid authority"
+        if http_host = env[HTTP_HOST]
```
jeremyevans, line 335: Not sure if this is used elsewhere.  Does this result in any verbose mode unused variable warnings?

lib/rack/lint.rb
jeremyevans: Let's use a separate PR for this.  As long as the common servers all provide `true`/`false` (and I assume they do for Rack 2 compliance), it should be fine.

jeremyevans (commented):
I think this looks OK, after the removal of the new `rack_is_hijack` validation.

lib/rack/lint.rb
ioquatix: Good idea, done.

lib/rack/lint.rb
ioquatix: I checked it does not appear to emit any warning.

jeremyevans: That's because you changed the code to use the variable in the error message (but not in the URI parse string for some reason).

When I reviewed this, the variable wasn't used: https://github.com/rack/rack/commit/9351e88bf4f55772309e391e336ad1f654df7c52#diff-2db485952a203ded9d5ae1301e35d82601b1c2410f914e767b02c4933b7e8a57R337

jeremyevans (approved):

ioquatix:
> That's because you changed the code to use the variable in the error message (but not in the URI parse string for some reason).

I was not paying attention to it, I'll fix it, thanks for pointing that out.

lib/rack/lint.rb
ioquatix: Thanks, fixed.

lib/rack/lint.rb
ioquatix: https://github.com/rack/rack/issues/2293

lib/rack/lint.rb
ioquatix: https://github.com/rack/rack/issues/2294

ioquatix:
I've also added this issue: https://github.com/rack/rack/issues/2295

Issue #2288, opened 2025-02-21
Can Request#ip be cached?


willbryant:
Following on from #2287, the other thing we noticed is that all the logic in `Rack::Request#ip` is run for each call.

This is often not noticed as it may only be hit once, typically by a request logger. But it's not uncommon for there to be many calls.

For example, in rack-attack, every individual `safelist_ip` call adds a filter like this:

```
      def safelist_ip(ip_address)
        @anonymous_safelists << Safelist.new { |request| IPAddr.new(ip_address).include?(IPAddr.new(request.ip)) }
      end
```

Note that each entry in the `Safelist` makes its own call to `request.ip` when invoked, which happens on every request. This significantly multiplies the call count and overhead. In our Rails app this added up to 7ms of overhead which was not counted in the request duration (since it was before request processing proper started) which was both surprising and hard to find.

rack-attack could be refactored to avoid this, but that's only going to help with the calls from rack-attack, not the combination of rack-attack and any other code using `Request#ip`.

So, I think the natural thing to do would be to cache the result of `Request#ip`. This seems like a fairly safe optimization - unless we have the expectation that some middleware will *change* something about the trusted proxy list (etc.) after calling `Request#ip` itself.

This seems to me kinda unlikely, and would be relying on internal implementation behavior, but I wanted to ask if we think that sort of thing might go on, before drafting a patch?

ioquatix:
I am fine with caching in an ivar.

mikegee:
#2292 resolved this, right?

jeremyevans:
@mikegee Correct, thanks for the reminder.

PR #2292, opened 2025-02-22
Avoid re-running logic to determine Request#ip once it's been calculated the first time

willbryant:
As mentioned in #2288, this reduces the amount of redundant re-calculation needed when gems like rack-attack call `Request#ip` more than once.

Feedback on the code factoring most welcome. This seemed like the least invasive way to do it; the usual simple `@ip ||= begin..end` pattern inside the method wouldn't work due to the `return` statements, but if we don't like the idea of the logic being in one place and the caching another, I can set the ivar in each return statement.

I did feel like doing it like this makes it very easy to add caching to other methods included from Helper without rewriting them, though, which could be desired over time?

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -61,7 +61,12 @@ module Rack
 
     def initialize(env)
       @env = env
+      @ip = nil
       @params = nil
+    end
+
+    def ip
+      @ip ||= super
     end
 
     def params
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ All notable changes to this project will be documented in this file. For info on
 - Invalid cookie keys will now raise an error. ([#2193](https://github.com/rack/rack/pull/2193), [@ioquatix])
 - `Rack::MediaType#params` now handles empty strings. ([#2229](https://github.com/rack/rack/pull/2229), [@jeremyevans])
 - Avoid unnecessary calls to the `ip_filter` lambda to evaluate `Request#ip` ([#2287](https://github.com/rack/rack/pull/2287), [@willbryant])
+- Only calculate `Request#ip` once per request ([#2292](https://github.com/rack/rack/pull/2292), [@willbryant])
 
 ### Deprecated
 


jeremyevans (commented):
This is certainly an noninvasive way to add caching, but `super` is not all that optimized, and often `ip` is only called once per request.  I'd prefer a conditional at the top of the current method, and setting the ivar in each explicit/implicit return.  However, I don't object to this approach, so let's wait for feedback from @ioquatix and see what he prefers.

ioquatix:
I am fine with this approach.

The alternative is to move it into the `Helper`. Currently that module is stateless, and is used by code outside of Rack. Introducing an instance variable cache makes `Helper` stateful, and may break downstream code, or the assumptions it makes about `Helper#ip`.

The only downside to this change is that caching is injected around the computation with no way to change the computation itself. Personally, when I introduce some kind of cache, I'd prefer to have two methods, e.g. `protected def compute_ip` and `def ip = @ip ||= compute_ip`. This way, sub-classes can modify `compute_ip` and still have the correct caching behaviour. In other words, if someone is sub-classing `Rack::Request` and overrides `def ip`, that override won't be subject to caching which I feel is a downside.

willbryant:
In general I like that approach (#ip/#compute_ip) too. However, this has the same issue you mention earlier - 

> Introducing an instance variable cache makes Helper stateful, and may break downstream code, or the assumptions it makes about Helper#ip.

As either the #ip/#compute_ip methods are both in Helper, and it becomes stateful; or, #compute_ip is in Helper and #ip is not, and we've changed the API to Helper as it no longer responds to `ip`.

ioquatix:
In principle, in `Helper`, you just have `def compute_ip; current logic; end; def ip = current_ip`

willbryant:
Yep seems reasonable, sort of to my first draft but avoids `super` and as you say, makes it easier to override.

I've force-pushed that. I made `compute_ip` private to keep the freedom to change mind later.

ioquatix:
I'm okay with this, and even thought I suggested it, I also don't mind the simpler approach either.

The benefit of this approach is that we removed the call to `super` (which is probably negligible), but more importantly, we still provide a way to correctly override the implementation of determining the `ip` without breaking existing code that uses `def ip; ... super` (thought it may not benefit as much from the cache).

Let's see what @jeremyevans has to say about it.

lib/rack/request.rb
```diff
@@ -755,6 +759,10 @@ def x_forwarded_proto_priority
 
     include Env
     include Helpers
+
+    def ip
```
jeremyevans, line 763: Can we move this near the `def params` definition (around line 62)?

jeremyevans (commented):
Looking at other Request methods, the caching we are adding here is most similar to `params`, so the way @willbryant was originally adding caching was consistent with the current code.  I didn't realize this at first, because of where the method was defined.

I worry that this changes behavior for applications that include a module in `Request` that overrides `ip`, because now their override is no longer used.  So in attempting to make caching of `ip` automatic for subclasses, we are breaking one use case.  That's not an automatic blocker, but we should consider whether that is an appropriate tradeoff.  If so, we should at least note this as a backwards incompatibility in the CHANGELOG.

I don't have objections to the separate `compute_ip` method, though we don't use that approach for `params`.

Note that `Helper` is already stateful depending on your definition, it just uses env caching instead of ivar caching for handling state changes.

Considering all that, my preference would be for @willbryant's original commit, with the method moved near the `def params` that does the same thing.  However, if we decide that automatic subclass caching is more important than backwards compatibility, I'm OK with this approach (still with the method moved).

willbryant:
Done, and added `@ip = nil` to match params, which I assume was done for the benefit of people running in higher warning levels.

jeremyevans (approved):

ioquatix:
Thanks for your contribution.

Issue #2293, opened 2025-02-24
`Rack::Lint` has inconsistent definition of `SERVER_PORT`.


ioquatix:
See <https://github.com/rack/rack/pull/2291#discussion_r1966665271> for context.

The `SERVER_PORT`, according to the Rack specification, must be a String and an Integer. We should reconcile this inconsistency.

ioquatix:
Basically, it looks like every server should be using a String, or it would fail `Rack::Lint`'s validations:

```
        ## The CGI keys (named without a period) must have String values.
        ## If the string values for CGI keys contain non-ASCII characters,
        ## they should use ASCII-8BIT encoding.
        env.each do |key, value|
          next if key.include?(".") # Skip extensions
          
          unless value.kind_of? String
            raise LintError, "env variable #{key} has non-string value #{value.inspect}"
          end
```

e.g. puma: <https://github.com/puma/puma/blob/ca201ef69757f8830b636251b0af7a51270eb68a/lib/puma/request.rb#L408-L423>.

ioquatix:
Fixed in #2296

PR #2296, opened 2025-02-24
Fix specification of `SERVER_PORT`.

ioquatix:
See <https://github.com/rack/rack/issues/2293> for context.


jeremyevans (approved):
Please add a CHANGELOG entry.  I know the enforcement has not changed (modulo `0x` prefix and similar), but since the wording has changed, I think we should note it.  You can put it under Bug Fixes instead of SPEC changes, since the previous SPEC wording is really a bug (since the value could not be an integer).

PR #2298, opened 2025-02-25
Better validation of `HTTP_HOST` and `SERVER_NAME` according to RFCs.

### LINKED ISSUE #2295: `Rack::Lint` `SERVER_NAME` and `HTTP_HOST` are insufficiently validated?
When we use `URI.parse("http://#{http_host}/")`, we assume that `http_host` holds a valid, sanitized hostname. However, `URI.parse` is fairly permissive in what it accepts. This can lead to unexpected (and potentially unsafe) behavior if `http_host` is user-controlled or not rigorously validated before parsing. Attackers can craft unusual or malicious values to manipulate the resulting URI in ways that are not obviously valid domain names. The same applies to `SERVER_NAME`.

Below is a condensed list of potentially unusual or invalid values that `URI.parse("http://#{string}/")` would still parse:

1. **Userinfo embedded**  
   - `user:pass@evil.com`  
   (Parses as `http://user:pass@evil.com/`)

2. **Port numbers and paths**  
   - `example.com:8080/path/to/endpoint`  
   (Parses as `http://example.com:8080/path/to/endpoint/`)

3. **Query or fragment**  
   - `example.com?foo=bar#frag`  
   (Parses as `http://example.com?foo=bar#frag/`)

ioquatix:
Fixes #2295

### DIFF
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
 require 'forwardable'
-require 'uri'
 
 require_relative 'constants'
 require_relative 'utils'
@@ -17,6 +16,35 @@ module Rack
     REQUEST_PATH_ABSOLUTE_FORM = /\A#{Utils::URI_PARSER.make_regexp}\z/
     REQUEST_PATH_AUTHORITY_FORM = /\A[^\/:]+:\d+\z/
     REQUEST_PATH_ASTERISK_FORM = '*'
+
+    # Match a host name, according to RFC3986.
+    # Copied from `URI::RFC3986_Parser::HOST` because older Ruby versions (< 3.3) don't expose it.
+    HOST_PATTERN = /
+      (?<IP-literal>\[(?:
+          (?<IPv6address>
+            (?:\h{1,4}:){6}
+            (?<ls32>\h{1,4}:\h{1,4}
+            | (?<IPv4address>(?<dec-octet>[1-9]\d|1\d{2}|2[0-4]\d|25[0-5]|\d)
+                \.\g<dec-octet>\.\g<dec-octet>\.\g<dec-octet>)
+            )
+          | ::(?:\h{1,4}:){5}\g<ls32>
+          | \h{1,4}?::(?:\h{1,4}:){4}\g<ls32>
+          | (?:(?:\h{1,4}:)?\h{1,4})?::(?:\h{1,4}:){3}\g<ls32>
+          | (?:(?:\h{1,4}:){,2}\h{1,4})?::(?:\h{1,4}:){2}\g<ls32>
+          | (?:(?:\h{1,4}:){,3}\h{1,4})?::\h{1,4}:\g<ls32>
+          | (?:(?:\h{1,4}:){,4}\h{1,4})?::\g<ls32>
+          | (?:(?:\h{1,4}:){,5}\h{1,4})?::\h{1,4}
+          | (?:(?:\h{1,4}:){,6}\h{1,4})?::
+          )
+        | (?<IPvFuture>v\h++\.[!$&-.0-9:;=A-Z_a-z~]++)
+        )\])
+    | \g<IPv4address>
+    | (?<reg-name>(?:%\h\h|[!$&-.0-9;=A-Z_a-z~])*+)
+    /x.freeze
+    SERVER_NAME_PATTERN = /\A#{HOST_PATTERN}\z/.freeze
+    HTTP_HOST_PATTERN = /\A#{HOST_PATTERN}(:\d*+)?\z/.freeze
+
+    private_constant :HOST_PATTERN, :SERVER_NAME_PATTERN, :HTTP_HOST_PATTERN
 
     # :stopdoc:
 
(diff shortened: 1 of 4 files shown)


ioquatix:
Found one issue (successfully):

```
Error:
HostAuthorizationTest#test_blocks_requests_with_invalid_hostnames:
Rack::Lint::LintError: env[HTTP_HOST] must be a valid authority: "attacker.com#x.example.com"
    /home/runner/work/rack/rack/lib/rack/lint.rb:359:in `check_environment'
    /home/runner/work/rack/rack/lib/rack/lint.rb:94:in `response'
    /home/runner/work/rack/rack/lib/rack/lint.rb:72:in `call'
    /opt/hostedtoolcache/Ruby/3.3.7/x64/lib/ruby/gems/3.3.0/gems/rack-test-2.2.0/lib/rack/test.rb:360:in `process_request'
    /opt/hostedtoolcache/Ruby/3.3.7/x64/lib/ruby/gems/3.3.0/gems/rack-test-2.2.0/lib/rack/test.rb:153:in `request'
    lib/action_dispatch/testing/integration.rb:297:in `process'
    lib/action_dispatch/testing/integration.rb:19:in `get'
    lib/action_dispatch/testing/integration.rb:388:in `get'
    test/dispatch/host_authorization_test.rb:448:in `block in <class:HostAuthorizationTest>'
```

cc @matthewd can we change this test in Rails to not use `Rack::Lint` as this is a malformed request according to the updated `Rack::Lint` specification.

lib/rack/lint.rb
```diff
@@ -18,6 +18,33 @@ class Lint
     REQUEST_PATH_AUTHORITY_FORM = /\A[^\/:]+:\d+\z/
     REQUEST_PATH_ASTERISK_FORM = '*'
 
+    # Match a host name, according to RFC3986.
+    # Copied from `URI::RFC3986_Parser::HOST` because older Ruby versions (< 3.3) don't expose it.
+    HOST_PATTERN = /
+      (?<IP-literal>\[(?:
```
jeremyevans, line 24: I think if we are copying it, we should drop the unused named captures.

lib/rack/lint.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 require 'forwardable'
-require 'uri'
+require 'uri/rfc3986_parser'
```
jeremyevans, line 4: Is this actually used now that we copied the implementation?

test/spec_lint.rb
```diff
@@ -191,12 +191,12 @@ def obj.fatal(*) end
     lambda {
       Rack::Lint.new(valid_app).call(env("SERVER_NAME" => "A Host"))
     }.must_raise(Rack::Lint::LintError).
-      message.must_equal "A Host must be a valid authority"
+      message.must_equal "env[SERVER_NAME] must be a valid host: \"A Host\""
```
jeremyevans, line 194: The error message wording is kind of awkward, though I guess less awkward than before.  Maybe something like (I think we should also quote the key):

```suggestion
      message.must_equal "env['SERVER_NAME'] is an invalid host: \"A Host\""
```

Similar for HTTP_HOST

jeremyevans (commented):
Looks pretty good.  I have a few comments, please let me know what you think.

lib/rack/lint.rb
ioquatix: We can drop it.

lib/rack/lint.rb
ioquatix: I think it serves as documentation. With `match?` it won't matter anyway, and it avoids any effort to copy an updated Regexp in the future (from `uri`).

test/spec_lint.rb
jeremyevans: > `SERVER_NAME` is a constant. I feel like the quotes just add extra noise and don't really add anything of value. It's also consistent with other errors. If you want to change that, I suggest we merge this as is and then update it in a separate PR.

`SERVER_NAME` isn't a valid constant.  `Rack::SERVER_NAME` is, but that isn't what is listed.  However, if this is consistent with other errors, I agree there is no need to fix it in this PR. I would still be in favor of fixing it later.

> Regarding the actual error, I think the `LintError` should specifically call out the requirement, as in, the server name must be a valid host. If you say "Server name is invalid host", then it begs the question "why does that matter?". When we say "Server name must be a valid host" we are clearly describing the requirement.

The issue with `must be a valid host` is that is awkward as it ends with the provided, invalid host, with no statement that it is invalid.  Users can probably infer it, but I don't think it is obvious.  I'm fine keeping `must be a valid host`, as long as additional wording is added.  Maybe 

```suggestion
      message.must_equal "env['SERVER_NAME'] must be a valid host. Invalid host provided: \"A Host\""
```

lib/rack/lint.rb
jeremyevans: Fair enough.  Let's make both constants private, though.

ioquatix: That makes sense. Maybe we should do that for the other constants too.

test/spec_lint.rb
ioquatix: I am okay with:

```
env[SERVER_NAME] must be a valid host, given \"A Host\"
```

I think there is value keeping the error message terse.

jeremyevans: > I think there is value keeping the error message terse.

Can you explain why terseness in this area is beneficial?  Given the choice between clarity and terseness, I would choose clarity for Lint's exception messages.  Even if users can probably infer correctly, I don't think it is beneficial to force the need for such inference on them, when we could instead make the error message more clear.

ioquatix: I think most people prefer terse error messages, myself included. I think this discussion and change is getting outside of the scope of this PR. This PR originally proposed error messages that are consistent with adjacent and similar error messages. Why don't we go with what was originally proposed (and consistent with others) and then if you feel strongly to reword all the error messages, open a new PR. I'm fine with making the messages clearer.

jeremyevans: Looking at the adjacent error messages, I think it would be consistent if you dropped the display of the invalid data.  So just: `env[SERVER_NAME] must be a valid authority` (and similar for HTTP_HOST).  We don't show invalid protocols or ports, so I guess there is no reason we must show invalid hosts or authorities.

ioquatix: Okay, I will do that.

jeremyevans (approved):
Looks good, but please wait until the issue is fixed in Rails so the external tests pass on master.

PR #2301, opened 2025-03-07
Add documentation for Rack::Multipart::UploadedFile

jeremyevans:
The only code change is switching from overriding respond_to? to overridding respond_to_missing?.  We dropped Ruby 1.8 support a while back.

### DIFF
--- a/lib/rack/multipart/uploaded_file.rb
+++ b/lib/rack/multipart/uploaded_file.rb
@@ -69,8 +69,8 @@ module Rack
       alias_method :local_path, :path
 
       # Return true if the tempfile responds to the method.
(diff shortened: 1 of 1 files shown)


ioquatix (approved):

ioquatix:
I think according to our past discussions, it's better if code and documentation changes are done in separate PRs.

jeremyevans:
> I think according to our past discussions, it's better if code and documentation changes are done in separate PRs.

Fair enough.  I split the code change into a separate commit.  If you would prefer a separate PR for that, please let me know.

ioquatix:
I'm fine with how it was, and this is even better.

PR #2313, opened 2025-03-20
Add documentation workflow.

ioquatix:
Generate the RDoc and push it to GitHub pages.

https://rack.github.io/rack/


Gemfile
```diff
@@ -9,10 +9,9 @@ group :maintenance, optional: true do
   gem "rubocop-packaging", require: false
 end
 
-unless ENV['CI']
-  group :doc do
-    gem 'rdoc'
-  end
+group :doc do
+  gem "rdoc", git: "https://github.com/ioquatix/rdoc.git"
```
tenderlove, line 13: Why do we need to use a fork?

tenderlove (commented):
I think this is fine but it would be nice if we didn't need to use a fork of RDoc.

Gemfile
samuel-williams-shopify: Because the official RDoc does not handle images with alt text correctly and we have a nice logo, but we don't want the logo in the "Table of Contents". There is an upstream PR to fix this, once it's merged I'll remove the use of my fork.

https://github.com/ruby/rdoc/pull/1320

I'll share a screen shot.

ioquatix: image

See the sliced off image on the LHS. We don't have that using the fork:

image

jeremyevans: I think you should be able to work around the issue by hiding the image in the table of contents using CSS.  That seems preferable to using a fork, to me at least.

jeremyevans (commented):
I think this looks good.  Thanks for working on it.  I have one requested change, and then I'm OK merging this.

Gemfile
ioquatix: I considered that option and it is not feasible.

jeremyevans: This appears to work and not be infeasible:

```css
div.nav-section img { display: none; }
```

The image is purely decorative, I'm not sure why the lack of alt text would matter.  From looking at `README.md`, the alt text just says "Rack", correct?

If you do not like using that one line of CSS, then use the rdoc gem at version 6.10.0.  That's the version that ships with Ruby 3.4, and it does not have the issue (I think it avoids the issue by not having a table of contents).

ioquatix: If you hide the image, then there is no title at all.

ioquatix: Maybe you can explain why using a fork is a problem? I'm not sure what the issue is here.

tenderlove: It looks like your change was merged upstream, so if we can switch to using upstream then I am fine with this.

I would hope the downsides of depending on a fork were self evident, but I'll try to spell it out.  Using a fork is problematic because now the Rack repository depends on you, personally. If you decide to stop maintaining the fork (for whatever reason) then the maintenance burden of your fork falls on the Rack team. This brings us a host of maintenance burdens, like watching upstream for security issues that we'd have to backport.  Maintaining a fork of _any_ repository (even yours) is not a responsibility I'm willing to take, and is also not good for the Rack project overall.

ioquatix: Unfortunately we can't use upstream as we are blocked on https://github.com/ruby/rdoc/pull/1113

My fork specifically works around that by including the generated files.

At this point, it sounds like we'd rather just wait for an official release, which I'm fine with.

ioquatix: Okay, RDoc was released with my changes, so we can use the official release.

jeremyevans (commented):
Assuming I understand this correctly, this means `rack.github.io` will reflect the `main` branch.  I don't think that's a good idea.  I think we should change it to reflect the most recent release (3.1.12).  The vast majority of our users are using rack via gems, not running the main branch, and the documentation should be targeted at them.  We would not want a user running the latest release version to look at the documentation, try something based on it, and find out it doesn't work because the feature isn't in any release.

We could consider subdirectories per version (e.g. `2.2`, `3.0`, `3.1`), but that's probably more difficult to setup.

samuel-williams-shopify:
Unfortunately github pages is per repository and there is no per-branch feature. It would make situations like this simpler.

I normally run my documentation from `main` but (try to) list the version availability in the documentation, e.g. we have a change log which lists features which are as yet unreleased.

While I understand your point, I'm against running the documentation workflow on 3-1-stable branches. No matter what, the issue you suggest will be a problem (half our user base is still on Rack 2.2). In addition, we don't backport documentation fixes/changes, so I think you've under-estimated the amount of work required to make the documentation look good on 3-1-stable.

A better solution is for us to clearly document the intended version support in the code, e.g. "This class was introduced in Rack 2.2" or "This method was introduced in Rack 3.1". But that is a lot of work, but also has significant benefit for users. Another solution would be to simply say "This documentation is for the main branch of Rack, some features may not be released yet". I'm happy to add that to the readme so that users at least are made aware.

jeremyevans:
I'm not in favor of adding documentation to methods explaining when they were introduced. Can we generate the documentation for `main` in `/main`?  For now, we could add a redirect from `/`/`/index.html` to `/main`.  That will allow us to easily support versioned documentation at `/3.1`, `/3.0`, `/2.2`, etc. in the future, and have `/index.html` show links to the available versions.

jeremyevans (approved):

PR #2314, opened 2025-03-21
Change SPEC to require request environment keys to be strings

jeremyevans:
Lint previously would raise for most non-String values, since it called:

```ruby
  key.include?(".")
```

on each environment key.  However, it's best to state this explicitly and add a specific Lint check for it.

### DIFF
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -134,9 +134,14 @@ module Rack
       ##
       ## Incoming HTTP requests are represented using an environment. \
       def check_environment(env)
-        ## The environment must be an unfrozen instance of +Hash+. The Rack application is free to modify the environment, but the modified environment should also comply with this specification.
+        ## The environment must be an unfrozen instance of +Hash+. The Rack application is free to modify the environment, but the modified environment should also comply with this specification. \
         raise LintError, "env #{env.inspect} is not a Hash, but #{env.class}" unless env.kind_of? Hash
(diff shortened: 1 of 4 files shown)


lib/rack/lint.rb
```diff
@@ -134,10 +134,15 @@ def response
...
       ## Incoming HTTP requests are represented using an environment. \
       def check_environment(env)
-        ## The environment must be an unfrozen instance of +Hash+. The Rack application is free to modify the environment, but the modified environment should also comply with this specification.
+        ## The environment must be an unfrozen instance of +Hash+. The Rack application is free to modify the environment, but the modified environment should also comply with this specification. \
         raise LintError, "env #{env.inspect} is not a Hash, but #{env.class}" unless env.kind_of? Hash
         raise LintError, "env should not be frozen, but is" if env.frozen?
 
+        ## All environment keys must be strings.
+        unless env.keys.all?(String)
+          raise LintError, "env contains non-string keys"
```
ioquatix, line 143: This error message does not report which keys are problematic. As a user, I could imagine it would be hard to find which key is not a String. Can we log that?

ioquatix: What about something like:

```ruby
keys = env.keys.reject{|key| String === key}
if keys.any?
  raise LintError, "env contains non-string keys: #{keys.inspect}"
end
```

jeremyevans: Pushed another commit for that.

ioquatix (approved):

PR #2315, opened 2025-03-21
Allow rake spec task to work when environment does not default to UTF-8

jeremyevans:
You can reproduce the failure by doing:

```
  LANG= rake spec
```

Explicitly using UTF-8 encoding when opening the SPEC and lint.rb files, as well as marking the regexp as UTF-8, fixes the issue.


ioquatix (approved):
LGTM.

PR #2318, opened 2025-03-24
Add `nil` as a valid return from a Response `body.to_path`

### LINKED ISSUE #2317: Rack::Lint warning for invalid to_path might not work
There is supposed to be a check for a body responding to `to_path` but not returning a valid path.
Pipes return `nil` for `to_path` naturally, which can cause servers to crash.

I can't get this check from Rack::Lint triggered though

```
app = Rack::Builder.new do
  use Rack::Lint
  run ->(env) {
    r, w = IO.pipe
    [200, {'content-type' => 'text/html'}, r]
  }
end

run app
```

This example causes a crash within puma

MSP-Greg:
Currently, a Response's `body.to_path` is only allowed to return a `String`.  Ruby's `IO.to_path` can be `nil` or a `String`.

Allow `nil` to be returned, add tests, update SPEC.

Closes #2317

### DIFF
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -837,8 +837,12 @@ module Rack
         ##
         ## If the Body responds to +to_path+, it must return a +String+ path for the local file system whose contents are identical to that produced by calling +each+; this may be used by the server as an alternative, possibly more efficient way to transport the response. The +to_path+ method does not consume the body.
         if @body.respond_to?(:to_path)
-          unless ::File.exist? @body.to_path
-            raise LintError, "The file identified by body.to_path does not exist"
+          optional_path = @body.to_path
+
+          if optional_path != nil
+            unless optional_path.is_a?(String) && ::File.exist?(optional_path)
+              raise LintError, "The file identified by body.to_path does not exist"
+            end
           end
         end
       end
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -841,7 +841,7 @@ module Rack
 
           if optional_path != nil
             unless optional_path.is_a?(String) && ::File.exist?(optional_path)
-              raise LintError, "The file identified by body.to_path does not exist"
+              raise LintError, "body.to_path must be nil or a path to an existing file"
             end
           end
         end
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -835,7 +835,7 @@ module Rack
 
       def verify_to_path
         ##
-        ## If the Body responds to +to_path+, it must return a +String+ path for the local file system whose contents are identical to that produced by calling +each+; this may be used by the server as an alternative, possibly more efficient way to transport the response. The +to_path+ method does not consume the body.
+        ## If the Body responds to +to_path+, it must return either +nil+ or a +String+. If a +String+ is returned, it must be a path for the local file system whose contents are identical to that produced by calling +each+; this may be used by the server as an alternative, possibly more efficient way to transport the response. The +to_path+ method does not consume the body.
         if @body.respond_to?(:to_path)
           optional_path = @body.to_path
 
--- a/test/spec_lint.rb
+++ b/test/spec_lint.rb
@@ -607,6 +607,17 @@ describe Rack::Lint do
     body.to_path.must_equal __FILE__
   end
 
+  it "handles body.to_path returning nil" do
+    body = Object.new
+    def body.each; end
+    def body.to_path; nil end
+    app = lambda { |env| [200, {}, body] }
+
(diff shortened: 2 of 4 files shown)


lib/rack/lint.rb
```diff
@@ -828,10 +828,11 @@ def close
 
       def verify_to_path
         ##
-        ## If the Body responds to +to_path+, it must return a +String+ path for the local file system whose contents are identical to that produced by calling +each+; this may be used by the server as an alternative, possibly more efficient way to transport the response. The +to_path+ method does not consume the body.
+        ## If the Body responds to +to_path+, it must return either +nil+ or a +String+.  If a +String+ is returned, it must be a path for the local file system whose contents are identical to that produced by calling +each+; this may be used by the server as an alternative, possibly more efficient way to transport the response.  The +to_path+ method does not consume the body.
         if @body.respond_to?(:to_path)
-          unless ::File.exist? @body.to_path
-            raise LintError, "The file identified by body.to_path does not exist"
+          temp = @body.to_path
+          unless temp.nil? || temp.is_a?(::String) &&::File.exist?(temp)
```
jeremyevans, line 834:
```suggestion
          unless temp.nil? || (temp.is_a?(::String) && ::File.exist?(temp))
```

I recommend avoiding mixing `||` and `&&` without parentheses (even in cases like this, where the behavior is correct).

test/spec_lint.rb
```diff
@@ -604,6 +604,17 @@ def body.to_path; __FILE__ end
     body.to_path.must_equal __FILE__
   end
 
+  it "responds to nil to_path" do
```
jeremyevans, line 607:
```suggestion
  it "handles body.to_path returning nil" do
```

jeremyevans (approved):
Looks good to me, though I have a couple recommendations.

lib/rack/lint.rb
MSP-Greg: Updated

test/spec_lint.rb
MSP-Greg: Thanks for the review.  Updated.

Re the above, I've looked at Ruby's [Precedence](https://msp-greg.github.io/ruby_trunk/file.precedence.html) too often to forget it.  Others may be otherwise, so paren's are a good suggestion...

ioquatix:
Looking good, can you please add a section to the changelog. See https://github.com/rack/rack/pull/2314 as a reference for the kind of changelog entry required.

lib/rack/lint.rb
```diff
@@ -828,10 +828,11 @@ def close
 
       def verify_to_path
         ##
-        ## If the Body responds to +to_path+, it must return a +String+ path for the local file system whose contents are identical to that produced by calling +each+; this may be used by the server as an alternative, possibly more efficient way to transport the response. The +to_path+ method does not consume the body.
+        ## If the Body responds to +to_path+, it must return either +nil+ or a +String+.  If a +String+ is returned, it must be a path for the local file system whose contents are identical to that produced by calling +each+; this may be used by the server as an alternative, possibly more efficient way to transport the response.  The +to_path+ method does not consume the body.
         if @body.respond_to?(:to_path)
-          unless ::File.exist? @body.to_path
-            raise LintError, "The file identified by body.to_path does not exist"
+          temp = @body.to_path
```
ioquatix, line 833: I find the current code a bit hard to read, and `temp` does not mean anything.

```suggestion
          optional_path = @body.to_path
          
          if optional_path != nil
            unless optional_path.is_a?(String) && ::File.exist?(optional_path)
              raise LintError, "body.to_path must be nil or be a path to an existing file: #{optional_path.inspect}"
            end
          end
```

MSP-Greg: Updated and added the CHANGELOG entry.  Rebased twice also...

lib/rack/lint.rb
```diff
@@ -835,10 +835,14 @@ def close
 
       def verify_to_path
         ##
-        ## If the Body responds to +to_path+, it must return a +String+ path for the local file system whose contents are identical to that produced by calling +each+; this may be used by the server as an alternative, possibly more efficient way to transport the response. The +to_path+ method does not consume the body.
+        ## If the Body responds to +to_path+, it must return either +nil+ or a +String+.  If a +String+ is returned, it must be a path for the local file system whose contents are identical to that produced by calling +each+; this may be used by the server as an alternative, possibly more efficient way to transport the response.  The +to_path+ method does not consume the body.
```
ioquatix, line 838: For consistency please do not use double space. Other than that, this PR looks great.

MSP-Greg: @ioquatix sorry for the delay, fixed and rebased.

PR #2322, opened 2025-04-11
Avoid modifying a potentially frozen string

tomhughes:
Ruby 4 is expected to make constant strings frozen by default and ruby 3.4 adds a feature which will warn about attempts to modify such strings when `--enable-frozen-string-literal` is used.

Some testing frameworks are now enabling that be default to allow future problems to be discovered and that let me to discover that this case was trying to append to a constant string - specifically if a mock request is created with a constant string as the input we wind up trying to modify it when `set_encoding` is called on the `StringIO` object.

### DIFF
--- a/lib/rack/mock_request.rb
+++ b/lib/rack/mock_request.rb
@@ -141,7 +141,7 @@ module Rack
 
       rack_input = opts[:input]
       if String === rack_input
-        rack_input = StringIO.new(rack_input)
+        rack_input = StringIO.new(rack_input, "r")
       end
 
       if rack_input
--- a/test/spec_mock_request_unfrozen.rb
+++ b/test/spec_mock_request_unfrozen.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: false
+
+require_relative 'helper'
+require 'yaml'
+require_relative 'psych_fix'
+
+separate_testing do
+  require_relative '../lib/rack/mock_request'
+  require_relative '../lib/rack/lint'
+  require_relative '../lib/rack/request'
+  require_relative '../lib/rack/body_proxy'
+end
+
+app = Rack::Lint.new(lambda { |env|
+  req = Rack::Request.new(env)
+
+  if input = env["rack.input"]
+    env["mock.postdata"] = input.read
+  end
+
+  if req.GET["error"]
+    env["rack.errors"].puts req.GET["error"]
+    env["rack.errors"].flush
+  end
+
+  body = req.head? ? "" : env.to_yaml
+  response = Rack::Response.new(
+    body,
+    req.GET["status"] || 200,
+    "content-type" => "text/yaml"
+  )
+  response.set_cookie("session_test", { value: "session_test", domain: "test.com", path: "/" })
+  response.set_cookie("secure_test", { value: "secure_test", domain: "test.com",  path: "/", secure: true })
+  response.set_cookie("persistent_test", { value: "persistent_test", max_age: 15552000, path: "/" })
+  response.set_cookie("persistent_with_expires_test", { value: "persistent_with_expires_test", expires: Time.httpdate("Thu, 31 Oct 2021 07:28:00 GMT"), path: "/" })
+  response.set_cookie("expires_and_max-age_test", { value: "expires_and_max-age_test", expires: Time.now + 15552000 * 2, max_age: 15552000, path: "/" })
+  response.finish
+})
+
+describe Rack::MockRequest do
+  it "doesn't warn when POST is given an unfrozen constant string" do
+    Warning[:deprecated] = true
+    capture_warnings(Warning) do |warnings|
+      env = Rack::MockRequest.env_for("/foo", method: :post, input: "test")
+      warnings.must_be :empty?
+      env["REQUEST_METHOD"].must_equal "POST"
+      env["QUERY_STRING"].must_equal ""
+      env["PATH_INFO"].must_equal "/foo"
+      env["rack.input"].string.must_equal "test"
+    end
+    Warning[:deprecated] = false
+  end
+end


jeremyevans:
Can you please include a test that shows what is actually breaking?  Providing frozen input to a StringIO doesn't freeze the StringIO, and setting the StringIO encoding does not modify the input string:

```ruby
x = "a"
x.freeze
s = StringIO.new(x)
s.set_encoding(Encoding::BINARY)
x.encoding # => #<Encoding:UTF-8>
```

Maybe there is a problem here, but if so, it isn't obvious to me.  If you can add a failing test, that will make it much easier to determine whether this is a problem in Rack.

tomhughes:
I can have a go but I was debugging this from warnings coming out of a rails test so there was quite a lot of distance and it might be tricky to reproduce... It was definitely warning about modifying a frozen string on the `set_encoding` line and adding that `dup` call definitely fixes it.

I'll see what I can do but it likely won't be for a couple of days.

ioquatix:
Even just the output of the failing test would be useful.

tomhughes:
I did attempt a test yesterday and it did indeed prove hard to reproduce the problem..

Going back to the original problem in the rails+minitest environment what I have is that this test action:

```ruby
xml = "<osm><changeset>" \
      "<tag k='created_by' v='osm test suite checking changesets'/>" \
      "</changeset></osm>"
assert_difference "Changeset.count", 1 do
  post api_changesets_path, :params => xml, :headers => auth_header
end
assert_response :success
```

triggers this warning:

```
/home/tom/.bundle/ruby/3.4.0/gems/rack-3.1.12/lib/rack/mock_request.rb:148: warning: literal string will be frozen in the future (run with --debug-frozen-string-literal for more information)
```

Adding some logging into `mock_request.rb` the arguments passed to `env_for` appear to be:

```ruby
uri = "http://www.example.com/api/0.6/changesets"
opts = {
  "rack.test" => true,
  "REMOTE_ADDR" => "127.0.0.1",
  "SERVER_PROTOCOL" => "HTTP/1.0",
  method: :post,
  "SERVER_NAME" => "www.example.com",
  "SERVER_PORT" => "80", "HTTPS" => "off",
  "rack.url_scheme" => "http",
  "REQUEST_URI" => "/api/0.6/changesets",
  "HTTP_HOST" => "www.example.com",
  "HTTP_ACCEPT" => "text/xml,application/xml,application/xhtml+xml,text/html;q=0.9,text/plain;q=0.8,image/png,*/*;q=0.5",
  "HTTP_AUTHORIZATION" => "Bearer 439Dq7XN3OG93otZPtoc5VJ5oHVmno3hQMBHFKgUIDo",
  "REQUEST_METHOD" => "POST",
  "CONTENT_TYPE" => "application/x-www-form-urlencoded",
  input: "<osm><changeset><tag k='created_by' v='osm test suite checking changesets'/></changeset></osm>"
}
```

and the call trace is:

```
/home/tom/.bundle/ruby/3.4.0/gems/rack-test-2.2.0/lib/rack/test.rb:336:in 'Rack::Test::Session#env_for'
/home/tom/.bundle/ruby/3.4.0/gems/rack-test-2.2.0/lib/rack/test.rb:152:in 'Rack::Test::Session#request'
/home/tom/.bundle/ruby/3.4.0/gems/actionpack-8.0.2/lib/action_dispatch/testing/integration.rb:297:in 'ActionDispatch::Integration::Session#process'
/home/tom/.bundle/ruby/3.4.0/gems/actionpack-8.0.2/lib/action_dispatch/testing/integration.rb:25:in 'ActionDispatch::Integration::RequestHelpers#post'
/home/tom/.bundle/ruby/3.4.0/gems/actionpack-8.0.2/lib/action_dispatch/testing/integration.rb:388:in 'ActionDispatch::Integration::Runner#post'
/home/tom/.bundle/ruby/3.4.0/gems/rails-controller-testing-1.0.5/lib/rails/controller/testing/integration.rb:16:in 'block (2 levels) in <module:Integration>'
/home/tom/src/osm/rails/test/controllers/api/changesets_controller_test.rb:994:in 'block (2 levels) in Api::ChangesetsControllerTest#test_repeated_changeset_create'
/home/tom/.bundle/ruby/3.4.0/gems/activesupport-8.0.2/lib/active_support/testing/assertions.rb:49:in 'ActiveSupport::Testing::Assertions#assert_nothing_raised'
/home/tom/.bundle/ruby/3.4.0/gems/activesupport-8.0.2/lib/active_support/testing/assertions.rb:287:in 'ActiveSupport::Testing::Assertions#_assert_nothing_raised_or_warn'
/home/tom/.bundle/ruby/3.4.0/gems/activesupport-8.0.2/lib/active_support/testing/assertions.rb:117:in 'ActiveSupport::Testing::Assertions#assert_difference'
/home/tom/src/osm/rails/test/controllers/api/changesets_controller_test.rb:993:in 'block in Api::ChangesetsControllerTest#test_repeated_changeset_create'
```

jeremyevans:
I did some research, and old versions of Ruby (before 2.3) had a StringIO that could exhibit this problem:

```
$ ruby23 -r stringio 
x = "a"
x.freeze
s = StringIO.new(x)
s.set_encoding(Encoding::BINARY)
x.encoding # => #<Encoding:UTF-8>

$ ruby22 -r stringio 
x = "a"
x.freeze
s = StringIO.new(x)
s.set_encoding(Encoding::BINARY)
x.encoding # => #<Encoding:UTF-8>
-:4:in `set_encoding': can't modify frozen String (RuntimeError)
        from -:4:in `<main>'
```

However, Rack doesn't support < Ruby 2.4. You also aren't running an old version of Ruby.

I also tried the most recent stringio release (3.1.6), it doesn't raise an error for the above code.

What version of stringio are you using?

tomhughes:
It seems to be 3.1.6 which is the current version:

```
% bundle exec rails c       
Loading development environment (Rails 8.0.2)
open-street-map(dev)> StringIO::VERSION
=> "3.1.6"
```

ioquatix:
I know this is tricky, but what we really need is a failing test case. We've both had a go at trying to figure it out but could not reproduce - doesn't mean it's not an issue, but we need to know the root cause before trying to figure out how to address it correctly.

tomhughes:
Right, so I do now understand what's really going on here, and it's a product of a couple of things:

1. My original comment about `--enable-frozen-string-literal` is wrong - the real issue is that minitest does `Warning[:deprecated] = true` so the string is not actually frozen but ruby 3.4 warns about what would happen if it was frozen.
2. The behaviour of `StringIO` changes when it is constructed on a frozen string - if the string is frozen then it defaults to creating a read-only IO object which in turn stops `set_encoding` from trying to modify the string.

The result is what you can see here:

```
% irb
irb(main):001> StringIO.new("").set_encoding(Encoding::BINARY)
=> #<StringIO:0x00007fc65a6ac070>
irb(main):002> Warning[:deprecated] = true
=> true
irb(main):003> StringIO.new("").set_encoding(Encoding::BINARY)
(irb):3: warning: literal string will be frozen in the future (run with --debug-frozen-string-literal for more information)
=> #<StringIO:0x00007fc6599de3c0>
irb(main):004> StringIO.new("", "r").set_encoding(Encoding::BINARY)
=> #<StringIO:0x00007fc65a3926f0>
irb(main):005> StringIO.new("".freeze).set_encoding(Encoding::BINARY)
=> #<StringIO:0x00007fc65afac148>
```

So I now think the correct solution here is to always construct the StringIO as readonly given it's an input stream but I still need to get a test working.

tomhughes:
I've updated this now, though the test is a bit nasty...

jeremyevans:
Thank you for the analysis.  If `StringIO` handles a frozen string correctly, an unfrozen string correctly, and warns incorrectly for a string that will be frozen in the future Ruby version, the issue is in `StringIO` and should be fixed there.  I'll work on submitting an upstream patch.

jeremyevans:
StringIO pull request submitted: https://github.com/ruby/stringio/pull/128

ioquatix:
Should we close this issue?

jeremyevans:
My upstream patch to stringio was accepted.  Please wait until the next stringio gem release, and then update to that.  In the meantime, you could probably work around the issue by doing `string = +string` before passing it to `StringIO.new`.

ioquatix:
I think the next release should be out soon, according to what I read on the issue tracker.

Issue #2323, opened 2025-04-12
Mounted Grape App -> Rack::Lint::LintError: Response body is already closed


ericproulx:
Hi 👋,

I'm working on a feature to add `linting` in [Grape](https://github.com/ruby-grape/grape) by adding `Rack::Lint` in the stack. It works well but I'm encountering a `Rack::Lint::LintError: Response body is already closed` when a Grape API is mounted in a Rails app.

When calling it, Rails's middlewares stack will deal with a `Rack::Lint::Wrapper` body instead of the real one. This ain't the issue but if we look at `Rack::Etag`, it might call [body.to_ary](https://github.com/rack/rack/blob/4f15e7b814922af79605be4b02c5b7c3044ba206/lib/rack/etag.rb#L32). Indeed, the `to_ary` function will be the [one](https://github.com/rack/rack/blob/4f15e7b814922af79605be4b02c5b7c3044ba206/lib/rack/lint.rb#L906) in `Rack::Lint::Wrapper` and it will `close` the body. Unfortunately, when calling `each` on it, it will raise a `Rack::Lint::LintError` even though our original linting was valid.

I've looked into `Rack::Lint` and somehow, it was [setting](https://github.com/rack/rack/blob/df6c47357f6c6bec2d585f45f417285d813d9b3a/lib/rack/lint.rb#L734) `@closed = false` when calling `each` in Rack 2, but this is not the case anymore.

I've open an [issue](https://github.com/ruby-grape/grape/issues/2559) on our side if you want to take a look at my example.

Thanks!

ioquatix:
Thanks for your report. Wow, this bug has existed for a long time. Introduced in https://github.com/rack/rack/pull/1887

Looks like we do all the right things except we don't return the right body. Whoops.

PR #2324, opened 2025-04-12
Ensure `Rack::ETag` correctly updates response body.

ioquatix:
Fixes <https://github.com/rack/rack/issues/2323>.


ericproulx:
@ioquatix I've tested it locally and it works perfectly :)

jeremyevans (approved):

PR #2327, opened 2025-05-09
Replace usage of `CGI::Cookie`, fix ruby 3.5 compatibility

Earlopain:
In Ruby 3.5, `cgi` will only contain functions related to escaping/unescaping.

https://bugs.ruby-lang.org/issues/21258

This is not an exact replicate of course, (`CGI::Cookie`) has some validations and coerces on setters but considering for what purpose this is, they don't seem necessary? During construction of the object rack already does conversions as necessary and setters don't make much sense, and aren't documented/tested for.

Although, for improved backwards compatibility, it wouldn't be much effort to make them `attr_accesor` instead.

### DIFF
--- a/lib/rack/mock_response.rb
+++ b/lib/rack/mock_response.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-require 'cgi/cookie'
 require 'time'
 
 require_relative 'response'
@@ -11,6 +10,35 @@ module Rack
   # MockRequest.
 
   class MockResponse < Rack::Response
+    if RUBY_VERSION >= '3.5'
+      class Cookie
+        attr_reader :name, :value, :path, :domain, :expires, :secure
+
+        def initialize(args)
+          @name = args["name"]
+          @value = args["value"]
+          @path = args["path"]
+          @domain = args["domain"]
+          @expires = args["expires"]
+          @secure = args["secure"]
+        end
+
+        def method_missing(method_name, *args, &block)
+          @value.send(method_name, *args, &block)
+        end
+        # :nocov:
+        ruby2_keywords(:method_missing) if respond_to?(:ruby2_keywords, true)
+        # :nocov:
+
+        def respond_to_missing?(method_name, include_all = false)
+          @value.respond_to?(method_name, include_all) || super
+        end
+      end
+    else
+      require 'cgi/cookie'
+      Cookie = CGI::Cookie
+    end
+
     class << self
       alias [] new
     end
@@ -83,7 +111,7 @@ module Rack
         Array(set_cookie_header).each do |cookie|
           cookie_name, cookie_filling = cookie.split('=', 2)
           cookie_attributes = identify_cookie_attributes cookie_filling
-          parsed_cookie = CGI::Cookie.new(
+          parsed_cookie = Cookie.new(
             'name' => cookie_name.strip,
             'value' => cookie_attributes.fetch('value'),
             'path' => cookie_attributes.fetch('path', nil),
@@ -100,7 +128,7 @@ module Rack
     def identify_cookie_attributes(cookie_filling)
       cookie_bits = cookie_filling.split(';')
       cookie_attributes = Hash.new
-      cookie_attributes.store('value', cookie_bits[0].strip)
+      cookie_attributes.store('value', Array(cookie_bits[0].strip))
       cookie_bits.drop(1).each do |bit|
         if bit.include? '='
           cookie_attribute, attribute_value = bit.split('=', 2)
--- a/test/spec_mock_response.rb
+++ b/test/spec_mock_response.rb
@@ -84,6 +84,7 @@ describe Rack::MockResponse do
   it "provides access to session cookies" do
     res = Rack::MockRequest.new(app).get("")
     session_cookie = res.cookie("session_test")
+    session_cookie[0].must_equal "session_test"
     session_cookie.value[0].must_equal "session_test"
     session_cookie.domain.must_equal "test.com"
     session_cookie.path.must_equal "/"
--- a/test/spec_request.rb
+++ b/test/spec_request.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 require_relative 'helper'
-require 'cgi'
+require 'cgi/escape'
 require 'forwardable'
 require 'securerandom'
 


lib/rack/mock_cookie.rb
```diff
@@ -0,0 +1,26 @@
+module Rack
+  # Compatibility around CGI::Cookie
+  class MockCookie # :nodoc:
```
jeremyevans, line 3: I think it would be better to have this as `Rack::MockResponse::Cookie`, and move it to the `mock_response.rb` file, since `MockResponse` seems to be the only place it is used.

ioquatix:
This looks okay to me. @jeremyevans wdyt?

lib/rack/mock_response.rb
```diff
@@ -83,13 +83,13 @@ def parse_cookies_from_header
...
           cookie_name, cookie_filling = cookie.split('=', 2)
           cookie_attributes = identify_cookie_attributes cookie_filling
-          parsed_cookie = CGI::Cookie.new(
-            'name' => cookie_name.strip,
-            'value' => cookie_attributes.fetch('value'),
-            'path' => cookie_attributes.fetch('path', nil),
-            'domain' => cookie_attributes.fetch('domain', nil),
-            'expires' => cookie_attributes.fetch('expires', nil),
-            'secure' => cookie_attributes.fetch('secure', false)
+          parsed_cookie = MockCookie.new(
```
jeremyevans, line 86: For compatibility, I think we should use `CGI::Cookie` on Ruby < 3.5, and the new cookie implementation on Ruby 3.5+.

lib/rack/mock_cookie.rb
Earlopain: Sure, that seems fine. I added private_constat too.

lib/rack/mock_response.rb
Earlopain: Yes, good idea. I've done as much

Earlopain:
I've added `ruby-head` to CI to make sure it's green (or at least not the fault of rack if not). Will remove in a bit again if it looks ok.

jeremyevans:
> I've added `ruby-head` to CI to make sure it's green (or at least not the fault of rack if not). Will remove in a bit again if it looks ok.
> 
> Edit: Does ok on protocol-rack and rails tests but then chokes on roda because of something in rdoc that was extracted. Just wanted to make sure I didn't mess something up.

I heard rdoc in ruby-head has some issue with psych that could be causing problems.  Roda may also need changes for cgi removal.

Earlopain:
> I heard rdoc in ruby-head has some issue with psych that could be causing problems. Roda may also need changes for cgi removal.

To pass tests, a bit more is required: `tilt` for example expects on rdoc but doesn't declare it. Since ruby 3.5, it is just gone and you need to explicitly depend on it
> $ RBENV_VERSION=ruby-dev bundle exec ruby -ve 'require "tilt/rdoc"'
ruby 3.5.0dev (2025-05-09T14:24:38Z master f30f0f0a22) +PRISM [x86_64-linux]
/home/user/.rbenv/versions/ruby-dev/lib/ruby/gems/3.5.0+0/gems/tilt-2.6.0/lib/tilt/rdoc.rb:3: warning: rdoc used to be loaded from the standard library, but is not part of the default gems since Ruby 3.5.0.
You can add rdoc to your Gemfile or gemspec to fix this error.
/home/user/.rbenv/versions/ruby-dev/lib/ruby/3.5.0+0/bundled_gems.rb:59:in 'Kernel.require': cannot load such file -- rdoc (LoadError)

OTOH, it does not depends on no template engine right now, so it seems be the responsibility of the consumer to add it to their gemfile instead. Does that sound right?

In regards to cgi, I think it's good. It does already just require `cgi/escape`, which is fine.

lib/rack/mock_response.rb
```diff
@@ -11,6 +10,41 @@ module Rack
...
+      end
+      # :nocov:
+      ruby2_keywords(:method_missing) if respond_to?(:ruby2_keywords, true)
+      # :nocov:
+
+      def respond_to_missing?(method_name, include_all = false)
+        @value.respond_to?(method_name, include_all) || super
+      end
+    end
+    private_constant :Cookie
```
jeremyevans, line 37: I think it is a bad idea to use `private_constant` for this class when you could be returning instances of it.  I would only define the class on Ruby 3.5+, and have `Rack::MockResponse::Cookie` be this class on Ruby 3.5+, and `CGI::Cookie` on earlier Ruby versions.

Earlopain: Makes it a bit nicer actually, thanks. I ran external tests on my fork and grape fails because it defines `Rack::MockResponse::Cookie` already (in test only!), so I openend https://github.com/ruby-grape/grape/pull/2565 for that.

byroot:
For the 2.6.0 failure you can bump `psych`.

Also could we get this moving? It's preventing testing ruby-head against anything that uses `rack` and given the recent large changes in ruby I'd like to be able to test ruby-head to find potential bugs.

Earlopain:
Probably I can just bust the cache with `cache-version` from `ruby/setup-ruby` to pick up the new version if CI needs to be green to merge. Someone let me know if I should do that.

jeremyevans (approved):

jeremyevans:
I dropped the related cache entry, but now there is a different problem in psych.  Instead of:

```
NoMethodError: undefined method `const_source_location' for Object:Class
  /home/runner/work/rack/rack/vendor/bundle/ruby/2.6.0/gems/psych-5.2.4/lib/psych/core_ext.rb:23:in `<top (required)>'
```

We now get:

```
NoMethodError: undefined method `members' for #<StringIO:0x00005641dd5f2880>
    vendor/bundle/ruby/2.6.0/gems/psych-5.2.5/lib/psych/visitors/yaml_tree.rb:170:in `visit_Data'
```

This is probably because:

```
$ ruby26 -r stringio -e 'p StringIO.superclass'
Data
```

Psych needs to be updated to only assume a `Struct`-like `Data` on Ruby 3.2+.

Earlopain:
https://github.com/ruby/psych/pull/729 Unfortunate series of events (https://bugs.ruby-lang.org/issues/3072)

byroot:
Psych `5.2.6` is out with the fix.

jeremyevans:
Deleted the cache entry and reran the job.  Specs are now passing, so I'll merge shortly.  Thank you @Earlopain for fixing the issue in rack, and @byroot for fixing the issue in psych.

byroot:
> and @byroot for fixing the issue in psych.

Well, credit to @Earlopain here to, I just merged.

MSP-Greg:
I hate to ask, and I'm not sure what the 'blast radius' of this is...

Puma CI hasn't run for a few days, and I just ran it, and this caused all the ruby head builds to fail.  I can add `cgi` to our Gemfile, but wondering if this may warrant a new release?

jeremyevans:
When testing with ruby-head, it would be best to also be testing with rack/main.  We'll make sure to release versions of rack with this change before the final release of Ruby 3.5.0.

byroot:
> before the final release of Ruby 3.5.0.

Not to command you or anything, but ideally compatibility fixes are released quickly so they don't hide subsequent issues.

There was a number of big changes in ruby-head recently, and lots of ruby-head CIs are uterly broken making it hard to notice when a new compatibility issue in introduced.

MSP-Greg:
> When testing with ruby-head, it would be best to also be testing with rack/main.

Thanks.  In the past, Puma CI has been stable, but (obviously) significant Ruby head changes can affect things.

For now, I'm changing all of Puma's Ruby head CI jobs so they use rack/main, and they're passing.

> There was a number of big changes in ruby-head recently, and lots of ruby-head CIs are utterly broken

When the Ruby 'infrastructure' for GHA was stable, I was often an 'advocate' for testing with Windows and also Ruby head.  There was a lot of friction about it, but I believe it's common now.  I hope people realize that it's needed to quickly determine the true 'blast radius' of changes, or better to fix it now rather than wait for the December release...

jeremyevans:
If you are testing the master branch of ruby, and not the latest release, I don't think it is unreasonable to expect you to test with the master branch of other ruby projects, and not the latest release (at least when necessary).  Expecting all ruby projects to release new gem versions quickly for all changes in ruby-head that affect them is expecting too much, IMO.

That being said, I don't handle Rack releases.  We should probably backport this to Rack 3.1, 3.0, and 2.2, since we'll want them to be compatible with Ruby 3.5.  If another Rack committer wants to release new versions with the fixes shortly after that, I have no objections.

hsbt:
Thanks all 🙏

PR #2328, opened 2025-05-12
[3-1-stable] Replace usage of `CGI::Cookie`

Earlopain:
It's https://github.com/rack/rack/pull/2327, but for the latest stable version (:

cc @byroot

### DIFF
--- a/lib/rack/mock_response.rb
+++ b/lib/rack/mock_response.rb
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 
-require 'cgi/cookie'
 require 'time'
 
 require_relative 'response'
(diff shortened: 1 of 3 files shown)


jeremyevans (approved):

Earlopain:
@byroot I wasn't thinking about that, pelase do.

lib/rack/mock_response.rb
```diff
@@ -100,7 +128,7 @@ def parse_cookies_from_header
     def identify_cookie_attributes(cookie_filling)
       cookie_bits = cookie_filling.split(';')
       cookie_attributes = Hash.new
-      cookie_attributes.store('value', cookie_bits[0].strip)
+      cookie_attributes.store('value', Array(cookie_bits[0].strip))
```
ioquatix, line 131: Do you mind explaining why this line was changed?

Earlopain: `CGI::Cookie` does this in its constructor (it's also a subclass of `Array`). It's required for the wrapper so `value[0]` doesn't just return a single char.

PR #2330, opened 2025-05-12
[2.2] Update test suite for Ruby 3.5

byroot:
`Hash#inspect` changed a bit which is causing spec to fail, but it's quite trivial to make code compatible.


jeremyevans (approved):

ioquatix:
Thanks for sorting this out.

PR #2331, opened 2025-05-16
[3-1-stable] Test on Ruby v3.4.

Earlopain:
jruby-head (or rather jruby-10) is now targeting ruby 3.4, which means that test changes from https://github.com/rack/rack/pull/2272 are required for CI to be green.


ioquatix:
Thanks. Merged this on a 🚆.

PR #2332, opened 2025-05-17
Drop dependency on `CGI::Cookie`.

ioquatix:
I don't think we need to bother with compatibility in 3.2+ and this simplifies the code and dependencies for us.

cc @Earlopain


jeremyevans (approved):
I suppose for a minor bump, the potential backwards compatibility issues are acceptable.  Definitely needs a note in the changelog, because the behavior is not identical between the two.

PR #2333, opened 2025-05-17
Feature detect `CGI::Cookie`.

ioquatix:
I would prefer to feature detect `CGI::Cookie` rather than version check.

If we approve this, I want to backport to 2.2 and 3.0.

cc @Earlopain

This is a follow up to <https://github.com/rack/rack/pull/2327>.

### DIFF
--- a/lib/rack/mock_response.rb
+++ b/lib/rack/mock_response.rb
@@ -10,7 +10,11 @@ module Rack
   # MockRequest.
 
   class MockResponse < Rack::Response
-    if RUBY_VERSION >= '3.5'
+    begin
+      # Recent versions of the CGI gem may not provide `CGI::Cookie`.
+      require 'cgi/cookie'
+      Cookie = CGI::Cookie
+    rescue LoadError
       class Cookie
         attr_reader :name, :value, :path, :domain, :expires, :secure
 
@@ -34,9 +38,6 @@ module Rack
           @value.respond_to?(method_name, include_all) || super
         end
       end
-    else
-      require 'cgi/cookie'
-      Cookie = CGI::Cookie
     end
 
     class << self
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 All notable changes to this project will be documented in this file. For info on how to format all future additions to this file please reference [Keep A Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [3.1.15] - 2025-05-18
+
+- Optional support for `CGI::Cookie` if not available. ([#2327](https://github.com/rack/rack/pull/2327), [#2333](https://github.com/rack/rack/pull/2333), [@earlopain])
+
 ## [3.1.14] - 2025-05-06
 
 ### Security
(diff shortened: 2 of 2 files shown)


lib/rack/mock_response.rb
```diff
@@ -128,7 +129,7 @@ def parse_cookies_from_header
     def identify_cookie_attributes(cookie_filling)
       cookie_bits = cookie_filling.split(';')
       cookie_attributes = Hash.new
-      cookie_attributes.store('value', Array(cookie_bits[0].strip))
+      cookie_attributes.store('value', cookie_bits[0].strip)
```
Earlopain, line 132: This is needed. rack doesn't test against ruby-dev or previews so it doesn't show but here are the failures:

```sh
Finished in 3.234368s, 340.0974 runs/s, 1473.8583 assertions/s.

  1) Failure:
Rack::MockResponse#test_0007_parses cookies giving max-age precedence over expires [test/spec_mock_response.rb:120]:
Expected: "expires_and_max-age_test"
  Actual: "e"

  2) Failure:
Rack::MockResponse#test_0004_provides access to session cookies [test/spec_mock_response.rb:87]:
Expected: "session_test"
  Actual: "s"

  3) Failure:
Rack::MockResponse#test_0006_provides access to persistent cookies set with expires [test/spec_mock_response.rb:109]:
Expected: "persistent_with_expires_test"
  Actual: "p"

  4) Failure:
Rack::MockResponse#test_0009_parses cookie headers with equals sign at the end [test/spec_mock_response.rb:138]:
--- expected
+++ actual
@@ -1 +1 @@
-"_somebase64encodedstringwithequalsatthened="
+"_"


  5) Failure:
Rack::MockResponse#test_0005_provides access to persistent cookies set with max-age [test/spec_mock_response.rb:98]:
Expected: "persistent_test"
  Actual: "p"

  6) Failure:
Rack::MockResponse#test_0012_parses multiple set-cookie headers provided as hash with array value [test/spec_mock_response.rb:155]:
Expected: "awesome"
  Actual: "a"

  7) Failure:
Rack::MockResponse#test_0008_provides access to secure cookies [test/spec_mock_response.rb:128]:
Expected: "secure_test"
  Actual: "s"
```

Earlopain (commented):
I guess this is fine. But, users could end up with `cgi` in their gemfile one way or another in which case the version check is more predictable.

lib/rack/mock_response.rb
ioquatix: Agreed, did not mean to commit this change.

jeremyevans (approved):
This allows a Ruby 3.5 user to opt-in to the cgi/cookie support by installing the gem.  It seems a little at-odds with #2332, in that this increases the cases where cgi/cookie is used, while #2332 eliminates the cgi/cookie usage completely.  However, I guess that makes sense, as this allows full backwards compatibility if the cgi gem is manually installed, which is probably what we want in a stable branch.

ioquatix:
@Earlopain What is the name you prefer to use on the changelog?

```
[@earlopain]: https://github.com/earlopain "earlopain"
```

Or is it "Earlopain" or "Earl O'Pain"?

Earlopain:
It's good like this, I don't care about capitalization, both is good. Thanks for handling the releases!

PR #2336, opened 2025-05-23
Test against ruby-head

Earlopain:
This should give more confidence that changes for unreleased ruby versions are correct.

I wonder if there is interest in this. I briefly looked into history and didn't find anything about this (but maybe I missed stuff). `ruby-head` are surprisingly stable from my experience but there may occasionally be legit failures that are entirely the fault of ruby.


ioquatix (approved):

ioquatix:
I think this is probably a good idea. Thanks.

PR #2343, opened 2025-06-17
Remove Rake::Files from middleware in README

### LINKED ISSUE #2342: Rack::Files is not a middleware
You can't `use Rack::Files`
```ruby 
class Rack::Files
   # lack parameter app here
   def initialize(root, headers = {}, default_mime = 'text/plain')
      @root = (::File.expand_path(root) if root)
      @headers = headers
      @default_mime = default_mime
      @head = Rack::Head.new(lambda { |env| get env })
    end
end

class Builder
      if @map
        mapping, @map = @map, nil
        @use << proc { |app| generate_map(app, mapping) }
      end
    end
end
```

DormancyWang:
fix #2342

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,7 +142,6 @@ middleware:
 * `Rack::ETag` for setting `etag` header on bodies that can be buffered.
 * `Rack::Events` for providing easy hooks when a request is received and when
   the response is sent.
-* `Rack::Files` for serving static files.
 * `Rack::Head` for returning an empty body for HEAD requests.
 * `Rack::Lint` for checking conformance to the [Rack Specification].
 * `Rack::Lock` for serializing requests using a mutex.
@@ -159,7 +158,7 @@ middleware:
   a nice and helpful way with clickable backtrace.
 * `Rack::ShowStatus` for using nice error pages for empty client error
   responses.
-* `Rack::Static` for more configurable serving of static files.
+* `Rack::Static` for configurable serving of static files.
 * `Rack::TempfileReaper` for removing temporary files creating during a request.
 
 All these components use the same interface, which is described in detail in the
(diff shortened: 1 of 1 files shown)


ioquatix:
@jeremyevans I checked this PR and wondered about moving `Rack::Files` -> `Rack::Static::Files`? Does it have any other purpose?

ioquatix:
@DormancyWang Thanks for this change.

What about also rewriting:

```
* `Rack::Static` for more configurable serving of static files.
```

as saying "for more configurable serving of static files" no longer makes sense with your proposed change.

jeremyevans:
> @jeremyevans I checked this PR and wondered about moving `Rack::Files` -> `Rack::Static::Files`? Does it have any other purpose?

It's used directly by other software.  Multiple Roda plugins use `Rack::Files`.  So I am against moving it.

dentarg:
So does Sinatra ([src](https://github.com/sinatra/sinatra/blob/5e1598501eb23a8673d61034df7be7d50c228400/lib/sinatra/base.rb#L441), [src](https://github.com/sinatra/sinatra/blob/5e1598501eb23a8673d61034df7be7d50c228400/lib/sinatra/base.rb#L300)).

DormancyWang:
@ioquatix I removed 'more' and kept 'configurable'.  
`Rack::Directory` is already under **Convenience Interfaces** section, so `Rack::Files` should be easy to find too.  
Do you think we should add a description for it?

ioquatix:
> Rack::Directory is already under Convenience Interfaces section, so Rack::Files should be easy to find too.
> Do you think we should add a description for it?

That seems reasonable to me.

ioquatix (approved):

Issue #2326, opened 2025-05-07
Raised `Invalid request parameters` errors using `application/x-www-form-urlencoded` as content type


kemuridama:
We updated `rack` to `v2.2.14` to correspond a security issue of CVE-2025-46727. And then we occured `ActionController::BadRequest: Invalid request parameters: total number of query parameters (xxxx) exceeds limit (4096)` errors.

Our service uses `application/x-www-form-urlencoded` as a content  type in POST requests. [According to test cases in `rack/test/spec_request.rb`](https://github.com/rack/rack/blob/79ab0b9794a529f26a9a5eba10077f5ee6a097ce/test/spec_request.rb#L744-L754), I think that `rack` will parse a request body of `application/x-www-form-urlencoded` as `params`. So `rack` raises that error by limiting a length of params is over.

Is this behavior intentional?

ioquatix:
Yes, are you expecting more than 4000 parameters?

kemuridama:
Thank you for your answer.

Yes, I thought `rack` parses a POST request body of `application/x-www-form-urlencoded` as same as other content types.

jeremyevans:
If you are using more than 4096 parameters, you need to configure a higher limit. the easiest way to do so is by setting the `RACK_QUERY_PARSER_PARAMS_LIMIT` environment variable to a higher value before loading Rack.

I apologize that this negatively affected your application, but in order to avoid denial of service, we had to pick a limit, and your application is apparently over the limit.

mrsimo:
We updated today and started getting these errors as well. We didn't expect a change like this to come in a patch version bump.

I can sort of understand where this is coming from, but I feel like this is something that would usually happen at a higher proxy level, no? We have configured an explicit `client_max_body_size` in nginx already, is there an additional concern in Rack? Particularly for the error on number of parameters (as opposed to the bytesize), we're getting errors when the body isn't even 500Kb.

jeremyevans:
Forget to account for the `&` to separate parameters, so it's only a maximum 16x increase in process memory per request body byte.   So potentially 64MB process memory required per request with Rack's default limits.  Remaining points are the same.

jeremyevans:
My two previous comments implied that you could actually cause that much memory usage after the patch, but that isn't accurate.  The reason we added the limit on the number of parameters is to prevent that much memory usage.  Please take my 2 previous comments as what would be possible with just a bytesize limit and no limit on the number of parameters.

mrsimo:
[...]

PR #2344, opened 2025-06-18
CHANGELOG.md: 3.1.14 breaking change

runephilosof-abtion:
References: https://github.com/rack/rack/issues/2326

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,11 +52,11 @@ All notable changes to this project will be documented in this file. For info on
 
 ## [3.1.14] - 2025-05-06
 
-:warning: **This release includes a security fix that may cause certain routes in previously working applications to fail if query parameters exceed 4,096 in count or 4 MB in total size.**
+:warning: **This release includes a security fix that may cause certain routes in previously working applications to fail if query parameters exceed 4,096 in count or 4 MB in total size. See <https://github.com/rack/rack/discussions/2356> for more details.**
 
 ### Security
 
-- [CVE-2025-46727](https://github.com/rack/rack/security/advisories/GHSA-gjh7-p2fx-99vx) Unbounded parameter parsing in `Rack::QueryParser` can lead to memory exhaustion. This was addressed by introducing two limits, `bytesize_limit` (total query size) and `params_limit` (total number of query parameters). Applications that expect to parse very large queries or queries with a very large number of parameters may need to increase the default limits.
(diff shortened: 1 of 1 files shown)


ioquatix:
@jeremyevans Can you please review and either merge or close as you see fit.

CHANGELOG.md
```diff
@@ -52,6 +52,9 @@ All notable changes to this project will be documented in this file. For info on
 
 ## [3.1.14] - 2025-05-06
 
+:warning: **This release includes a breaking change.**
+It introduces two limits `bytesize_limit` and `params_limit`. Which limits "total query size" and "total number of query parameters".
+
 ### Security
 
 - [CVE-2025-46727](https://github.com/rack/rack/security/advisories/GHSA-gjh7-p2fx-99vx) Unbounded parameter parsing in `Rack::QueryParser` can lead to memory exhaustion.
```
jeremyevans, line 60:
```suggestion
- [CVE-2025-46727](https://github.com/rack/rack/security/advisories/GHSA-gjh7-p2fx-99vx) Unbounded parameter parsing in `Rack::QueryParser` can lead to memory exhaustion. This was addressed by introducing two limits, `bytesize_limit` (total query size) and `params_limit` (total number of query parameters). Applications that expect to parse very large queries or queries with a very large number of parameters may need to increase the default limits.
```

CHANGELOG.md
```diff
@@ -52,6 +52,9 @@ All notable changes to this project will be documented in this file. For info on
 
 ## [3.1.14] - 2025-05-06
 
+:warning: **This release includes a breaking change.**
+It introduces two limits `bytesize_limit` and `params_limit`. Which limits "total query size" and "total number of query parameters".
+
```
jeremyevans, line 57:
```suggestion
```

jeremyevans (changes requested):
We should move this text to the related security change.

CHANGELOG.md
jeremyevans: For the vast majority of applications, it's not a breaking change, as the applications are under the default limits. For a small percentage of applications that expect requests that exceed the default limits, it is breaking.

We previously weren't mentioning the limits and that they may need to be updated at all in CHANGELOG. I do think it is worth mentioning the limits and that some applications may need to increase them, but in the same section as the security change.

My changes are deliberate and considered. If you think it should be displayed differently, I have no objections. Can you please either merge or close as **you** see fit?

ioquatix: Sure, no problem.

CHANGELOG.md
ioquatix:
```suggestion
:warning: **This release includes a security fix that may cause certain routes in previously working applications to fail if query parameters exceed 4,096 in count or 4 MB in total size.**
```

PR #2348, opened 2025-07-12
Add changelog for version 2.2.17

extern-c:
The chagelog for version 2.2.17 is missing on the main branch. 

See https://github.com/rack/rack/blob/2-2-stable/CHANGELOG.md and https://github.com/rack/rack/blob/main/CHANGELOG.md.


ioquatix (approved):
LGTM

PR #2349, opened 2025-07-16
Add Rack::Request#query_parser=

jeremyevans:
This makes it simpler to override the query parser on a per-request basis, instead of having to override the query_parser method.

See #2326 for some background on why this may be desired.


ioquatix (approved):
LGTM, but please add details to the changelog and squash merge.

matthewd:
It's a pattern that was already effectively encouraged via method override, but do we need to worry that a simple accessor increases the risk of cache contamination here? Whichever `Rack::Request` instance['s `query_parser`] decides the values that get cached in the env, which will then be used by other instances. If that's just a disagreement on limits that seems fine, but it could be Bad News if this was set to something that's not a `Rack::QueryParser`?

ioquatix:
I agree and I suppose there are two paths forward:

- `Rack::Request` should persist all state to `env` e.g. `env["rack.query_parser"]`, or
- https://github.com/rack/rack/issues/1879

PR #2350, opened 2025-07-24
Fix make_delete_cookie_header when domain and path values contain regex-unsafe characters

NARKOZ:
When deleting cookies, the `domain` and `path` values are interpolated directly into regular expression without escaping, triggering a `RegexpError`. 
This pull request fixes that by escaping `domain` and `path` with `Regexp.escape`.

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -292,8 +292,8 @@ module Rack
       end
 
       key = escape(key)
-      domain = value[:domain]
-      path = value[:path]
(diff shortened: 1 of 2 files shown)


ioquatix:
Yeah, didn't I fix this issue in 3.x?

ioquatix:
Thanks for your PR but it looks like we decided not to merge for the reasons given above. If you think this assessment is wrong, please feel free to continue the discussion.

Issue #2341, opened 2025-06-09
Rack::QueryParser#parse_nested_query discards duplicate values


doriantaylor:
Is this supposed to happen?

```ruby
require 'rack'

qp = Rack::Utils.default_query_parser

qs = 'foo=1&foo=2'

# what I need
t1 = qp.parse_query qs, ?&
# { "foo" => ["1", "2"] }

# what I get instead (called from Rack::Request#POST)
t2 = qp.parse_nested_query qs, ?&
# { "foo" => "2" }
```

If so, is there a way to make it use the other call when parsing form input e.g. via `Rack::Request#POST`?

jeremyevans:
This is expected behavior.  If you want an array of values, you need to use a `[]` parameter name suffix: `qs = 'foo[]=1&foo[]=2'`

doriantaylor:
So I'm currently feeding `POST` data into [a protocol implementation](https://github.com/doriantaylor/rb-rdf-kv) that [embeds graph manipulation statements](https://doriantaylor.com/rdf-kv) into ordinary key-value pairs, so that key syntax is a no-go.

This is what I found I had to do to work around it:

```ruby
    post = req.POST # to actually parse
    if post = req.env['rack.request.form_pairs']
      post = post.reduce({}) do |hash, pair|
        (hash[pair.first] ||= []) << pair.last
        hash
      end
    elsif post = req.env['rack.request.form_vars']
      post = Rack::Utils.default_query_parser.parse_query post
    else
      post = {}
    end
```

what if there was like `req.POST raw: true` or something that ran `parse_query` instead of `parse_nested_query`?

ioquatix:
What about exposing `req.form_pairs` (if we don't already)?

cc @matthewd

matthewd:
> What about exposing req.form_pairs (if we don't already)?

Yeah, I had meant to do that as a follow-up (particularly to avoid the need for the floating `req.POST` call just to populate the env value)... but haven't done so [yet]. 🤦🏻‍♂️

doriantaylor:
`env['rack.request.form_pairs']` doesn't mix query string and message body, does it? I didn't check. If it does, would it be possible to have a parameter that can isolate one or the other?

jeremyevans:
`env['rack.request.form_pairs']` is only body params.  There is no equivalent for query params as there is no need, because you can always reparse the query string (you cannot rely on reparsing the body as the body is not rewindable by default in Rack 3).

PR #2351, opened 2025-07-24
Add `Rack::Request#form_pairs`

matthewd:
Instead of hiding the population of `RACK_REQUEST_FORM_PAIRS` inside `POST`, this extracts it as a separately accessible public method. To make it properly useful _as_ a public method, unlike the existing behaviour where a pairs-wanting caller must handle the non-multipart cases for themselves, here it's unified: if we have pairs available in any form, `form_pairs` will return them.

`POST` is thus reduced to the more specific job of converting the pair-stream to structured hash format only, slightly more similar to `GET`. (No behaviour change, of course -- just that the "where could form data be coming from" stuff is now delegated to this new method.)

This is the follow-up originally promised in https://github.com/rack/rack/pull/2088#issuecomment-1637445845, and more recently mentioned in https://github.com/rack/rack/issues/2341#issuecomment-3008205735

### DIFF
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -91,6 +91,25 @@ module Rack
       return params.to_h
     end
 
+    # Parses a query string by breaking it up at the '&', returning all key-value
+    # pairs as an array of [key, value] arrays. Unlike parse_query, this preserves
+    # all duplicate keys rather than collapsing them.
+    def parse_query_pairs(qs, separator = nil, &unescaper)
+      unescaper ||= method(:unescape)
+
+      pairs = []
+
+      check_query_string(qs, separator).split(separator ? (COMMON_SEP[separator] || /[#{separator}] */n) : DEFAULT_SEP).each do |p|
+        next if p.empty?
+        k, v = p.split('=', 2).map!(&unescaper)
+        pairs << [k, v]
+      end
+
+      return pairs
+    rescue ArgumentError => e
+      raise InvalidParameterError, e.message, e.backtrace
+    end
+
     # parse_nested_query expands a query string into structural types. Supported
     # types are Arrays, Hashes and basic value types. It is possible to supply
     # query strings with parameters of conflicting types, in this case a
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -492,6 +492,46 @@ module Rack
         get_header(RACK_REQUEST_QUERY_HASH) || set_header(RACK_REQUEST_QUERY_HASH, parse_query(query_string, '&'))
       end
 
+      # Returns the form data pairs received in the request body.
+      #
+      # This method support both application/x-www-form-urlencoded and
+      # multipart/form-data.
+      def form_pairs
+        if pairs = get_header(RACK_REQUEST_FORM_PAIRS)
+          return pairs
+        elsif error = get_header(RACK_REQUEST_FORM_ERROR)
+          raise error.class, error.message, cause: error.cause
+        end
+
+        begin
+          rack_input = get_header(RACK_INPUT)
+
+          # Otherwise, figure out how to parse the input:
+          if rack_input.nil?
+            set_header(RACK_REQUEST_FORM_PAIRS, [])
+          elsif form_data? || parseable_data?
+            if pairs = Rack::Multipart.parse_multipart(env, Rack::Multipart::ParamList)
+              set_header RACK_REQUEST_FORM_PAIRS, pairs
+            else
+              form_vars = get_header(RACK_INPUT).read
+
+              # Fix for Safari Ajax postings that always append \0
+              # form_vars.sub!(/\0\z/, '') # performance replacement:
+              form_vars.slice!(-1) if form_vars.end_with?("\0")
+
+              set_header RACK_REQUEST_FORM_VARS, form_vars
+              pairs = query_parser.parse_query_pairs(form_vars, '&')
+              set_header(RACK_REQUEST_FORM_PAIRS, pairs)
+            end
+          else
+            set_header(RACK_REQUEST_FORM_PAIRS, [])
+          end
+        rescue => error
+          set_header(RACK_REQUEST_FORM_ERROR, error)
+          raise
+        end
+      end
+
       # Returns the data received in the request body.
       #
       # This method support both application/x-www-form-urlencoded and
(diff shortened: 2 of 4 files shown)


lib/rack/query_parser.rb
```diff
@@ -91,6 +91,25 @@ def parse_query(qs, separator = nil, &unescaper)
...
+
+      check_query_string(qs, separator).split(separator ? (COMMON_SEP[separator] || /[#{separator}] */n) : DEFAULT_SEP).each do |p|
+        next if p.empty?
+        k, v = p.split('=', 2).map!(&unescaper)
+        pairs << [k, v]
+      end
+
+      return pairs
+    rescue ArgumentError => e
+      raise InvalidParameterError, e.message, e.backtrace
```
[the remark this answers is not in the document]
matthewd: Yeah, I added this in response to a test failure. TBH I didn't look closely at what was going on, and just accepted that while I'd based my implementation on `parse_query`, it was true that the call I was replacing was to `parse_nested_query` which had this construct... but it's a string-encoding issue, so I suspect it's actually something that `parse_query` is arguably missing (and is just less exercised in our tests, because most internals prefer `parse_nested_query`).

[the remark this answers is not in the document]
matthewd: I paid more attention this time: the exception being wrapped here comes from `unescape`, and results from an invalid `%`-encoded value. I do think that's more correctly applied to all QS-parsing operations, so I've moved the rescue onto my new private method, where it will now apply to all three operations. That seems safe for now-affected `parse_query`, because `InvalidParameterError < ArgumentError`.

lib/rack/query_parser.rb
```diff
@@ -215,20 +211,37 @@ def params_hash_has_key?(hash, key)
...
-          raise QueryLimitError, "total number of query parameters (#{param_count+1}) exceeds limit (#{@params_limit})"
-        end
+      return if !qs || qs.empty?
 
-        qs
+      if qs.bytesize > @bytesize_limit
+        raise QueryLimitError, "total query size (#{qs.bytesize}) exceeds limit (#{@bytesize_limit})"
+      end
+
+      pairs = qs.split(separator ? (COMMON_SEP[separator] || /[#{separator}] */n) : DEFAULT_SEP, @params_limit + 1)
```
matthewd, line 223: As this had become the only caller, I've inlined `check_query_string` (probably uncontroversial), and in doing so also modified how it works (maybe more controversial?), by integrating the parameter count check with the `split`. Happy to unwind if we're concerned it might be less safe.

lib/rack/query_parser.rb
```diff
@@ -215,20 +211,37 @@ def params_hash_has_key?(hash, key)
...
+
+      if pairs.size > @params_limit
+        param_count = pairs.size + pairs.last.count(separator || "&")
+        raise QueryLimitError, "total number of query parameters (#{param_count}) exceeds limit (#{@params_limit})"
+      end
+
+      if unescaper
```
matthewd, line 230: I've gone for some small duplication here, so that the default-`unescaper` happy-path can avoid the whole `method(:unescape)` indirection

matthewd (commented):
I think we've been squash-merging anyway, so I've kept my further refactoring as a separate commit for ease of review.

lib/rack/query_parser.rb
```diff
@@ -118,17 +101,11 @@ def parse_query_pairs(qs, separator = nil, &unescaper)
     def parse_nested_query(qs, separator = nil)
       params = make_params
 
-      unless qs.nil? || qs.empty?
-        check_query_string(qs, separator).split(separator ? (COMMON_SEP[separator] || /[#{separator}] */n) : DEFAULT_SEP).each do |p|
-          k, v = p.split('=', 2).map! { |s| unescape(s) }
-
-          _normalize_params(params, k, v, 0)
-        end
+      each_query_pair(qs, separator) do |k, v|
```
ioquatix, line 104: I'd like to avoid abbreviations.
```suggestion
      each_query_pair(query_separator, separator) do |key, value|
```

Abbreviations make it harder to understand code.

jeremyevans: `qs` is for `query_string`, not `query_separator`. While I don't object to using longer variable names, I don't see the need, and it wouldn't necessarily be consistent. In this method, we are calling `make_params` and `_normalize_params` methods, and all existing variables other than `separator` are already abbreviated and, IMO, easy to understand.

Abbreviating variables makes it easier to read code, as long as the abbreviated variable names are understandable. It's common to use `k` and `v` instead of `key` and `value`, just like it is common to use `i` instead of `index`.  Unnecessary verbosity leads to increased fatigue when reading the code.

lib/rack/query_parser.rb
```diff
@@ -234,20 +211,37 @@ def params_hash_has_key?(hash, key)
       true
     end
 
-    def check_query_string(qs, sep)
-      if qs
-        if qs.bytesize > @bytesize_limit
-          raise QueryLimitError, "total query size (#{qs.bytesize}) exceeds limit (#{@bytesize_limit})"
-        end
+    def each_query_pair(qs, separator, unescaper = nil)
+      return enum_for(:each_query_pair, qs, separator, unescaper) unless block_given?
```
jeremyevans, line 215: I know this is a pattern used by a lot of enumerable methods, but I'm not a fan. This is a private method, can we remove this? I'll include a suggestion for fixing the single affected caller.

```suggestion
```

lib/rack/query_parser.rb
```diff
@@ -94,20 +89,8 @@ def parse_query(qs, separator = nil, &unescaper)
...
-        next if p.empty?
-        k, v = p.split('=', 2).map!(&unescaper)
-        pairs << [k, v]
-      end
-
-      return pairs
-    rescue ArgumentError => e
-      raise InvalidParameterError, e.message, e.backtrace
+    def parse_query_pairs(qs, separator = nil)
+      each_query_pair(qs, separator).to_a
```
jeremyevans, line 93: Related to the removal of `enum_for`:

```suggestion
      a = []
      each_query_pair(qs, separator) do |k, v|
        a << [k, v]
      end
      a
```

lib/rack/query_parser.rb
jeremyevans: This makes the common case more efficient, with the trade-off of increased work in the pathological case.  However, as we only split up to the limit, I agree this is a better approach.

lib/rack/query_parser.rb
jeremyevans: I'm fine with this. What are your thoughts on deprecating the block argument to `parse_query` and the optional argument to this method, so we can eventually drop the duplicate code?

jeremyevans (approved):
Looks good, though I would like to remove the added `enum_for` before merging.

lib/rack/query_parser.rb
ioquatix: I basically disagree, and the fact that I didn't even know what the expansion was proves my point haha.

matthewd: None of those three variables are newly introduced here; if we did want to change any/all of them it feels like that would be better done all at once, and probably not in a commit that's also making other changes?

ioquatix:
Please make sure to add release notes and to squash merge.

ioquatix (approved):
LGTM.

ioquatix:
Thanks!

PR #2352, opened 2025-07-28
`lint.rb` consistency improvements as suggested by Claude.

### LINKED ISSUE #2303: Formatting `<tt>` or `+` in `lint.rb`.
@jeremyevans:

> I believe `+` is preferred in cases where it works. `<tt>` should be used in cases where + does not generate the appropriate markup.

For reference, I'd like to use `+` for class names, Ruby constants, and so on, and `<tt>` for general code.

We should ensure our usage in `lint.rb` is consistent with the guidelines of RDoc.

samuel-williams-shopify:
Fixes #2303.


lib/rack/lint.rb
```diff
@@ -808,11 +812,11 @@ def check_rack_protocol_header(status, headers)
         end
       end
       ##
-      ## Setting this value informs the server that it should perform a connection upgrade. In HTTP/1, this is done using the +upgrade+ header. In HTTP/2, this is done by accepting the request.
+      ## Setting this value informs the server that it should perform a connection upgrade. In <tt>HTTP/1</tt>, this is done using the +upgrade+ header. In <tt>HTTP/2</tt>, this is done by accepting the request.
       ##
       ## === The Body
       ##
-      ## The Body is typically an +Array+ of +String+ values, an enumerable that yields +String+ values, a +Proc+ instance, or a File-like object.
+      ## The Body is typically an +Array+ of +String+ values, an enumerable that yields +String+ values, a +Proc+, or a +File+-like object.
```
ioquatix, line 819:
```suggestion
      ## The Body is typically an +Array+ of +String+ values, an enumerable that yields +String+ values, a +Proc+, or an +IO+-like object.
```

SPEC.rdoc
```diff
@@ -34,8 +34,8 @@ The remainder of the request URL's "path", designating the virtual "location" of
 
 The <tt>PATH_INFO</tt>, if provided, must be a valid request target or an empty string, as defined by {RFC9110}[https://datatracker.ietf.org/doc/html/rfc9110#target.resource].
 * Only <tt>OPTIONS</tt> requests may have <tt>PATH_INFO</tt> set to <tt>*</tt> (asterisk-form).
-* Only <tt>CONNECT</tt> requests may have <tt>PATH_INFO</tt> set to an authority (authority-form). Note that in HTTP/2+, the authority-form is not a valid request target.
-* <tt>CONNECT</tt> and <tt>OPTIONS</tt> requests must not have <tt>PATH_INFO</tt> set to a URI (absolute-form).
+* Only <tt>CONNECT</tt> requests may have <tt>PATH_INFO</tt> set to an authority (authority-form). Note that in <tt>HTTP/2+</tt>, the authority-form is not a valid request target.
+* <tt>CONNECT</tt> and <tt>OPTIONS</tt> requests must not have <tt>PATH_INFO</tt> set to a <tt>URI</tt> (absolute-form).
```
jeremyevans, line 38: URI in this case doesn't not refer to the Ruby URI class, so I don't think we should use `<tt>` for it.

SPEC.rdoc
```diff
@@ -34,8 +34,8 @@ The remainder of the request URL's "path", designating the virtual "location" of
 
 The <tt>PATH_INFO</tt>, if provided, must be a valid request target or an empty string, as defined by {RFC9110}[https://datatracker.ietf.org/doc/html/rfc9110#target.resource].
 * Only <tt>OPTIONS</tt> requests may have <tt>PATH_INFO</tt> set to <tt>*</tt> (asterisk-form).
-* Only <tt>CONNECT</tt> requests may have <tt>PATH_INFO</tt> set to an authority (authority-form). Note that in HTTP/2+, the authority-form is not a valid request target.
-* <tt>CONNECT</tt> and <tt>OPTIONS</tt> requests must not have <tt>PATH_INFO</tt> set to a URI (absolute-form).
+* Only <tt>CONNECT</tt> requests may have <tt>PATH_INFO</tt> set to an authority (authority-form). Note that in <tt>HTTP/2+</tt>, the authority-form is not a valid request target.
```
jeremyevans, line 37: I don't think we should use `<tt>` here as HTTP/2+ doesn't refer to a specific symbol or code.  Possibly without the `+`, it would make sense, though I wouldn't encourage it even in that case.  Similar for other references in this file.

SPEC.rdoc
```diff
@@ -140,18 +140,18 @@ The error stream. See the section below on the error stream for more information
...
 
 === The Input Stream
 
-The input stream is an +IO+-like object which contains the raw HTTP request data. When applicable, its external encoding must be "ASCII-8BIT" and it must be opened in binary mode. The input stream must respond to +gets+, +each+, and +read+:
-* +gets+ must be called without arguments and return a string, or +nil+ on EOF.
+The input stream is an +IO+-like object which contains the raw HTTP request data. When applicable, its external encoding must be <tt>ASCII-8BIT</tt> and it must be opened in binary mode. The input stream must respond to +gets+, +each+, and +read+:
+* +gets+ must be called without arguments and return a +String+, or +nil+ on <tt>EOF</tt>.
```
jeremyevans, line 148: I don't think we should use `<tt>` for EOF. EOF in this sense is a concept, not a specific identifier/code. Same for elsewhere in this file.

SPEC.rdoc
```diff
@@ -202,30 +202,30 @@ This is an HTTP status. It must be an Integer greater than or equal to 100.
...
+The headers must be an unfrozen +Hash+. The header keys must be +String+ values. Special headers starting <tt>rack.</tt> are for communicating with the server, and must not be sent back to the client.
 
 * The headers must not contain a <tt>"status"</tt> key.
-* Header keys must conform to RFC7230 token specification, i.e. cannot contain non-printable ASCII, DQUOTE or <tt>(),/:;<=>?@[\]{}</tt>.
-* Header keys must not contain uppercase ASCII characters (A-Z).
-* Header values must be either a +String+ value, or an +Array+ of +String+ values, such that each +String+ value must not contain characters with an ASCII ordinal below 040 (32).
+* Header keys must conform to {RFC7230}[https://tools.ietf.org/html/rfc7230] token specification, i.e. cannot contain non-printable <tt>ASCII</tt>, <tt>DQUOTE</tt> or <tt>(),/:;<=>?@[\]{}</tt>.
+* Header keys must not contain uppercase <tt>ASCII</tt> characters (A-Z).
+* Header values must be either a +String+, or an +Array+ of +String+ values, such that each +String+ must not contain characters with an <tt>ASCII</tt> ordinal below <tt>040</tt> (32).
```
jeremyevans, line 210: Seems odd to use `<tt>` for 040 but not for 32, as they are both numbers. I don't think it needs to be used for either, but it should be consistent.

SPEC.rdoc
```diff
@@ -202,30 +202,30 @@ This is an HTTP status. It must be an Integer greater than or equal to 100.
...
 === The Headers
 
-The headers must be a unfrozen +Hash+. The header keys must be +String+ objects. Special headers starting <tt>rack.</tt> are for communicating with the server, and must not be sent back to the client.
+The headers must be an unfrozen +Hash+. The header keys must be +String+ values. Special headers starting <tt>rack.</tt> are for communicating with the server, and must not be sent back to the client.
 
 * The headers must not contain a <tt>"status"</tt> key.
-* Header keys must conform to RFC7230 token specification, i.e. cannot contain non-printable ASCII, DQUOTE or <tt>(),/:;<=>?@[\]{}</tt>.
-* Header keys must not contain uppercase ASCII characters (A-Z).
-* Header values must be either a +String+ value, or an +Array+ of +String+ values, such that each +String+ value must not contain characters with an ASCII ordinal below 040 (32).
+* Header keys must conform to {RFC7230}[https://tools.ietf.org/html/rfc7230] token specification, i.e. cannot contain non-printable <tt>ASCII</tt>, <tt>DQUOTE</tt> or <tt>(),/:;<=>?@[\]{}</tt>.
```
jeremyevans, line 208: DQUOTE directly names the identifier used in RFC7230, so it makes sense to use `<tt>` for that.  I don't think it makes sense to do so for ASCII.

SPEC.rdoc
```diff
@@ -239,17 +239,17 @@ If the Body responds to +to_path+, it must return either +nil+ or a +String+. If
 
 ==== Enumerable Body
 
-The Enumerable Body must respond to +each+. It must only be called once. It must not be called after being closed, and must only yield +String+ values.
+The Enumerable Body must respond to +each+. It must only be consumed once (calling this method consumes the body). It must not be called after being closed, and must only yield +String+ values.
```
jeremyevans, line 242:
```suggestion
The Enumerable Body must respond to +each+, which must not be called more than once, and must not be called after being closed. The +each+ method must only yield +String+ values.
```

SPEC.rdoc
```diff
@@ -239,17 +239,17 @@ If the Body responds to +to_path+, it must return either +nil+ or a +String+. If
...
 
 Middleware must not call +each+ directly on the Body. Instead, middleware can return a new Body that calls +each+ on the original Body, yielding at least once per iteration.
 
 If the Body responds to +to_ary+, it must return an +Array+ whose contents are identical to that produced by calling +each+. Middleware may call +to_ary+ directly on the Body and return a new Body in its place. In other words, middleware can only process the Body directly if it responds to +to_ary+. If the Body responds to both +to_ary+ and +close+, its implementation of +to_ary+ must call +close+.
 
 ==== Streaming Body
 
-The Streaming Body must respond to +call+. It must only be called once. It must not be called after being closed. It takes a +stream+ argument.
+The Streaming Body must respond to +call+. It must only be consumed once (calling this method consumes the body). It must not be called after being closed. It takes a +stream+ argument.
```
jeremyevans, line 250:
```suggestion
The Streaming Body must respond to +call+, which must not be called more than once, and must not be called after being closed. The +call+ method must accept a +stream+ argument.
```

lib/rack/lint.rb
```diff
@@ -55,6 +55,10 @@ def call(env = nil)
 
     # N.B. The empty `##` comments creates paragraphs in the output. A trailing "\" is used to escape the newline character, which combines the comments into a single paragraph.
     #
+    # Formatting conventions for documentation:
+    # - Use <tt> for: literal values, constants, method signatures, environment keys, technical terms, BNF symbols.
+    # - Use + for: Ruby types, concepts, method names, class references.
```
jeremyevans, line 60: rdoc does not format `<tt>` differently than `+`, so I don't think this is a good convention. IMO, `+` should be preferred over `<tt>` in any case where you can use it, as `<tt>` is significantly more verbose.

I'm not sure what the implied difference is between a Ruby type and a class reference in this context (in CRuby, you have T_HASH vs Hash, but that's not what is meant here), and both are generally also top-level constants.

I don't think we should use this monospace formatting for technical terms or concepts. For literal values, it depends. I don't think we need to use it for status codes or other numbers.

I agree we should use monospace formatting for Ruby constants, method names/signatures, environment keys, and BNF symbols.

jeremyevans (commented):
Many of the changes look good, but I think this uses monospace formatting in cases where it would be best not to. If you have a strong reason for wanting monospace in some of these areas, please explain why so I can reconsider.

lib/rack/lint.rb
samuel-williams-shopify: After I tried modifying `lint.rb` to follow your proposed model, I found a lot of surprises. I reviewed the RDoc source code and found some important distinctions that affect the practical choice between `+...+` and `<tt>...</tt>`.

You're absolutely right that:
- **Output formatting is identical** - both render as monospace/teletype - when they are parsed correctly.
- **`+...+` is more concise** when it works.

However, there's a crucial parsing difference that affects when we **can** use `+...+`:

### RDoc parsing test:

```ruby
✅ PARSED: Use +method+ here         # Simple identifiers work
✅ PARSED: The +Class+ name          # Class names work  
❌ NO MARKUP: Call +Array#<<+ method # Method signatures fail
❌ NO MARKUP: Use +String#[]+ accessor # Operators fail
❌ NO MARKUP: Handle +<script>+ tags  # HTML/XML fails
❌ NO MARKUP: Run +rdoc --help+ command # Commands with args fail

# But <tt>...</tt> handles all of these:
✅ PARSED: Call <tt>Array#<<</tt> method
✅ PARSED: Use <tt>String#[]</tt> accessor  
✅ PARSED: Handle <tt><script></tt> tags
✅ PARSED: Run <tt>rdoc --help</tt> command
```

#### The Design Rationale

Through code analysis, I found that `+...+` was designed for **simple, word-like identifiers** due to word boundary requirements in the parsing regex. It can't handle:
- Characters like `#`, `<`, `>`, `@` in method signatures.
- Spaces in command examples.
- Complex punctuation in code expressions.

> IMO, + should be preferred over <tt> in any case where you can use it, as <tt> is significantly more verbose.

It turns out that knowing when to do this is unintuitive and can lead to mistakes. I don't have a strong opinion about how to update `lint.rb`, but `<tt>` seems like a safer default even if it's more verbose, due to the subtle differences in parsing.

```ruby
require 'rdoc'
am = RDoc::Markup::AttributeManager.new

# Test cases that demonstrate the limitations
test_cases = [
  # Cases where +...+ should work
  'Use +method+ here',
  'The +Class+ name', 
  'Set +DEBUG+ flag',
  
  # Cases where +...+ fails due to character restrictions  
  'Call +Array#<<+ method',
  'Use +String#[]+ accessor',
  'Handle +<script>+ tags',
  'Run +rdoc --help+ command',
  'Set +user@domain+ email',
  'See +#method+ for more details', 
  
  # Same content with <tt>...</tt>
  'Call <tt>Array#<<</tt> method',
  'Use <tt>String#[]</tt> accessor', 
  'Handle <tt><script></tt> tags',
  'Run <tt>rdoc --help</tt> command',
  'Set <tt>user@domain</tt> email'
]

puts '=== Testing Current RDoc Parsing Behavior ==='
puts

test_cases.each do |text|
  result = am.flow(text)
  
  # Check if any TT markup was applied
  has_tt = result.any? { |node| 
    node.respond_to?(:turn_on) && 
    am.attributes.as_string(node.turn_on) == 'TT'
  }
  
  status = has_tt ? '✅ PARSED' : '❌ NO MARKUP'
  puts "#{status}: #{text} -> #{result.inspect}"
end
```

SPEC.rdoc
ioquatix:
```suggestion
* <tt>CONNECT</tt> and <tt>OPTIONS</tt> requests must not have <tt>PATH_INFO</tt> set to a URI (absolute-form).
```

SPEC.rdoc
ioquatix:
```suggestion
* Only <tt>CONNECT</tt> requests may have <tt>PATH_INFO</tt> set to an authority (authority-form). Note that in HTTP/2+, the authority-form is not a valid request target.
```

SPEC.rdoc
ioquatix:
```suggestion
* +gets+ must be called without arguments and return a +String+, or +nil+ on EOF (end-of-file).
```

SPEC.rdoc
ioquatix:
```suggestion
* Header values must be either a +String+, or an +Array+ of +String+ values, such that each +String+ must not contain characters with an ASCII ordinal below 040 (32).
```

SPEC.rdoc
ioquatix:
```suggestion
* Header keys must conform to {RFC7230}[https://tools.ietf.org/html/rfc7230] token specification, i.e. cannot contain non-printable ASCII, <tt>DQUOTE</tt> or <tt>(),/:;<=>?@[\]{}</tt>.
```

ioquatix:
Thanks for the feedback I believe most of it has been actioned. I made a few minor changes to the response body specification wording. Tricky.

SPEC.rdoc
```diff
@@ -202,30 +202,30 @@ This is an HTTP status. It must be an Integer greater than or equal to 100.
...
 
-There must not be a <tt>content-length</tt> header key when the status is 1xx, 204, or 304.
+There must not be a <tt>content-length</tt> header key when the status is <tt>1xx</tt>, <tt>204</tt>, or <tt>304</tt>.
 
 ==== The <tt>rack.protocol</tt> Header
 
 If the <tt>rack.protocol</tt> header is present, it must be a +String+, and must be one of the values from the <tt>rack.protocol</tt> array from the environment.
 
-Setting this value informs the server that it should perform a connection upgrade. In HTTP/1, this is done using the +upgrade+ header. In HTTP/2, this is done by accepting the request.
+Setting this value informs the server that it should perform a connection upgrade. In HTTP/1, this is done using the +upgrade+ header. In HTTP/2+, this is done by accepting the request.
```
jeremyevans, line 224:
```suggestion
Setting this value informs the server that it should perform a connection upgrade. In HTTP/1, this is done using the +upgrade+ header. In HTTP/2, this is done by accepting the request.
```

jeremyevans (approved):
Looks good, other than one extra `+`

SPEC.rdoc
jeremyevans: Sorry, initially thought this was leftover formatting. No need for changes.

PR #2354, opened 2025-07-28
Relax `check_header_value`.

### LINKED ISSUE #2308: Should `Rack::Lint#check_header_value` be more or less restrictive around response header values?
> [Response] Header values must be either a `String` instance, or an `Array` of `String` values, such that each `String` instance must not contain characters with an ASCII ordinal value below 040/32/0x20.

According to RFC7230:

```
     header-field   = field-name ":" OWS field-value OWS

     field-name     = token
     field-value    = *( field-content / obs-fold )
     field-content  = field-vchar [ 1*( SP / HTAB ) field-vchar ]
     field-vchar    = VCHAR / obs-text
```

However, `Rack::Lint` will fail if a response header contains a tab character:

```ruby
require "rack"

lint = Rack::Lint.new(proc{|env| [200, {"foo" => "bar\xFFbaz"}, []]})

lint.call({"HTTP_USER_AGENT" => "foo\tbar", "REQUEST_METHOD" => "GET", "SERVER_NAME" => "localhost", "QUERY_STRING" => "", "SERVER_PROTOCOL" => "HTTP/1.1", "rack.errors" => $stderr, "rack.url_scheme" => "http", "PATH_INFO" => "/"})
=> /Users/samuel/.gem/ruby/3.4.2/gems/rack-3.1.12/lib/rack/lint.rb:734:in 'Rack::Lint::Wrapper#check_header_value': invalid header value foo: "bar\tbaz" (Rack::Lint::LintError)
```

Should we loosen this part of the SPEC as this is technically allowed as per the RFC?

In addition, characters above ASCII value 126 are not permitted, but we allow them. Should that be an error?

```ruby
require "rack"

lint = Rack::Lint.new(proc{|env| [200, {"foo" => "bar\xFFbaz".b}, []]})

lint.call({"HTTP_USER_AGENT" => "foo\tbar", "REQUEST_METHOD" => "GET", "SERVER_NAME" => "localhost", "QUERY_STRING" => "", "SERVER_PROTOCOL" => "HTTP/1.1", "rack.errors" => $stderr, "rack.url_scheme" => "http", "PATH_INFO" => "/"})
=> 
[200,
 {"foo" => "bar\xFFbaz"},
 #<Rack::Lint::Wrapper:0x0000000120551740
  @app=#<Proc:0x00000001222ea158 (irb):3>,
  @body=[],
  @closed=false,
  @content_length=nil,
  @env=
   {"HTTP_USER_AGENT" => "foo\tbar",
    "REQUEST_METHOD" => "GET",
    "SERVER_NAME" => "localhost",
    "QUERY_STRING" => "",
    "SERVER_PROTOCOL" => "HTTP/1.1",
    "rack.errors" => #<Rack::Lint::Wrapper::ErrorWrapper:0x000000012429cdc0 @error=#<IO:<STDERR>>>,
    "rack.url_scheme" => "http",
    "PATH_INFO" => "/",
    "rack.lint" => [#<Rack::Lint::Wrapper:0x0000000120551740 ...>],
    "rack.lint.body_iteration" => 0},
  @head_request=false,
  @headers={"foo" => "bar\xFFbaz"},
  @invoked=nil,
  @lint=[#<Rack::Lint::Wrapper:0x0000000120551740 ...>],
  @response=[200, {"foo" => "bar\xFFbaz"}, []],
  @size=0,
  @status=200>]
```

(note the use of `\xFF` which is technically, invalid).

ioquatix:
As has been previously discussed, `Rack::Lint` is not for validating user data or the correctness of data, just that the `env` has the right shape and that the interfaces are used correctly.

Fixes #2308.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ All notable changes to this project will be documented in this file. For info on
 
 - Request environment keys must now be strings. ([#2310](https://github.com/rack/rack/issues/2310), [@jeremyevans])
 - Add `nil` as a valid return from a Response `body.to_path` ([#2318](https://github.com/rack/rack/pull/2318), [@MSP-Greg])
+- `#check_header_value` is relaxed, only disallowing CR/LF/NUL characters. ([#2354](https://github.com/rack/rack/pull/2354), [@ioquatix])
 
 ### Added
 
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -743,7 +743,7 @@ module Rack
 
       def check_header_value(key, value)
         ## such that each +String+ value must not contain <tt>NUL</tt> (<tt>\0</tt>), <tt>CR</tt> (<tt>\r</tt>), or <tt>LF</tt> (<tt>\n</tt>).
-        if value =~ /[\x00\x0A\x0D]/
+        if value.match?(/[\x00\x0A\x0D]/)
           raise LintError, "invalid header value #{key}: #{value.inspect}"
         end
       end
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -742,8 +742,8 @@ module Rack
       end
 
       def check_header_value(key, value)
-        ## such that each +String+ value must not contain characters with an ASCII ordinal below 040 (32).
-        if value =~ /[\000-\037]/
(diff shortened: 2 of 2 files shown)


ioquatix:
Thanks Jeremy.

Yes, basically this is enforcing the bare minimum, similar to various servers I evaluated: https://github.com/rack/rack/issues/2308#issuecomment-2741538390

I actually tried following the RFCs more closely and broke a production environment, so this (and similar) checks as described by the RFCs are simply not sufficient for real world usage. Therefore, making this check more strict would probably break real world applications, and honestly, I'm not sure there is any real advantage.

So, this change seems pragmatic to me.

PR #2355, opened 2025-07-29
`Rack::Builder` `#use`, `#run` and `#map` all return `nil`.

ioquatix:
See <https://github.com/rack/rack/pull/2346> for more context.

Alternatively, it could return `nil`, but I think returning `self` is slightly more useful. However, it does lock in the return value (`nil` might give us more flexibility in the future).

In principle with this change, one could write:

```
Rack::Builder.new.use(...).use(...).run(...).to_app
```

I'm open to either idea. WDYT.


jeremyevans (commented):
I agree we should be explicit in the return value. My preference is to have the methods return `nil`, so we can potentially use a different return value in the future. If we return `self`, we are basically stuck with that going forward, even if later we find it would be better to return something else. I don't think the minor benefit of supporting method chaining is worth potentially limiting ourselves in the future. If we choose to return `self` and not `nil`, we should add specs for the behavior we want to allow.

ioquatix:
Thanks, I'm fine with either, so I'll change it to `nil`.

Just for future reference, apart from `nil` and `self` would there be any other realistic option?

jeremyevans:
> Just for future reference, apart from `nil` and `self` would there be any other realistic option?

I'm not sure. I don't have any ideas yet. But since there's no burning need to support method chaining for this, I'd rather leave the option open.

PR #2363, opened 2025-08-01
Add response_finished support to TempfileReaper

skipkayhil:
When supported by the server, this delays closing any tempfiles until after the server socket has been closed (assuming it will be), and additionally removes an allocation (the BodyProxy).

(hiding whitespace from the diff makes the changes more clear)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file. For info on
 
 ## Unreleased
 
+### SPEC Changes
+
+- Define `rack.response_finished` callback arguments more strictly. ([#2365](https://github.com/rack/rack/pull/2365), [@skipkayhil](https://github.com/skipkayhil))
+
 ### Added
 
 - Add support for `rack.response_finished` to `Rack::TempfileReaper`. ([#2363](https://github.com/rack/rack/pull/2363), [@skipkayhil](https://github.com/skipkayhil))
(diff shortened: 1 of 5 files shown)


lib/rack/tempfile_reaper.rb
```diff
@@ -9,25 +9,35 @@ module Rack
...
       env[RACK_TEMPFILES] ||= []
 
-      begin
-        _, _, body = response = @app.call(env)
-      rescue Exception
-        env[RACK_TEMPFILES]&.each(&:close!)
-        raise
-      end
+      if response_finished = env[RACK_RESPONSE_FINISHED]
+        response_finished << RESPONSE_FINISHED_HANDLER
```
ioquatix, line 24:
```suggestion
        tempfiles = env[RACK_TEMPFILES] ||= []

      if response_finished = env[RACK_RESPONSE_FINISHED]
        response_finished << proc do
          tempfiles.each(&:close!)
        end
```

I'm not in favour of my proposed approach, but I thought I'd mention it.

ioquatix:
I was also wondering if for whatever reason, we had multiple `TempfileReaper` middleware how would they interact.

For example would close be better off like this:

```
while io = tempfiles.pop
  io.close!
end
```

Otherwise, the array is never cleared. Does it matter?

skipkayhil:
> Does it matter?

```irb
irb(main):001> t = Tempfile.new
=> #<File:/tmp/20250801-253946-bs40nb>
irb(main):002> t.close!
=> true
irb(main):003> t.close!
=> nil
```

I was also curious, but it seems like it would work fine (But also if someone is running duplicate middleware that seems more like an application error)

PR #2365, opened 2025-08-06
Add response_finished arity to Rack::Lint

skipkayhil:
Ref rails/rails#55425
Ref rails/rails#55458

If these callbacks _must_ be called with four arguments, then this will help ensure that happens.


lib/rack/lint.rb
```diff
@@ -462,15 +462,33 @@ def check_environment(env)
...
       end
 
+      class ResponseFinishedWrapper
+        def initialize(callback)
+          @callback = callback
+        end
+
+        ## The callables are called with <tt>environment, status, headers, error</tt> arguments and should not raise any exceptions. The callables would typically be called after sending the response to the client, but it could also be called if an error occurs while generating the response or sending the response (in that case, the +error+ argument will be a kind of +Exception+). The callables will be called in reverse order.
+        def call(*args)
+          unless args.length == 4
```
jeremyevans, line 484: I agree with adding this check, but it may be better to check the types of each argument instead of just checking the arity.

skipkayhil:
Note: 

Pitchfork currently just `.call`s when the `callable.arity == 0` (because that's how `after_reply` was implemented), so this would disallow that

https://github.com/Shopify/pitchfork/blob/18869d2f02549a54d7b2db6e0351e7fa71e95546/lib/pitchfork/http_server.rb#L897-L899

The open PR to `puma` would not need any changes with this extra validation:

https://github.com/puma/puma/pull/3681

lib/rack/lint.rb
skipkayhil: I think that's a good idea as well, I have to run for now but I'll work on it

jeremyevans:
> Pitchfork currently just `.call`s when the `callable.arity == 0` (because that's how `after_reply` was implemented), so this would disallow that
> 
> https://github.com/Shopify/pitchfork/blob/18869d2f02549a54d7b2db6e0351e7fa71e95546/lib/pitchfork/http_server.rb#L897-L899
> 
> The open PR to `puma` would not need any changes with this extra validation:
> 
> [puma/puma#3681](https://github.com/puma/puma/pull/3681)

I think it's a good time for this type of change. We just released Rack 3.2, so Rack 3.3 is not coming any time soon, and servers will have plenty of time to adjust before we release Rack 3.3.

lib/rack/lint.rb
```diff
@@ -462,13 +462,47 @@ def check_environment(env)
...
+          ## * <tt>headers</tt> valid Rack response headers
+          headers.nil? || @wrapper.check_headers(headers)
+
+          ## * <tt>error</t> an Exception
+          unless error.nil? || Exception === error
+            raise LintError, "rack.response_finished callable's fourth argument must be an Exception or nil, got #{error}"
           end
         end
+
+        ## and should not raise any exceptions.
```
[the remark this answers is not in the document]
ioquatix: I don't think those two goals are mutually exclusive - this spec is about the application code.

A test like this would be more suitable for https://github.com/socketry/rack-conform and I think it would be a good idea to confirm servers work correctly in the face of errors.

ioquatix:
```
Error:
ExecutorTest#test_complete_callbacks_are_called_on_rack_response_finished:
Rack::Lint::LintError: One of SCRIPT_NAME or PATH_INFO must be set (make PATH_INFO '/' if SCRIPT_NAME is empty)
    /home/runner/work/rack/rack/lib/rack/lint.rb:200:in 'Rack::Lint::Wrapper#check_environment'
    /home/runner/work/rack/rack/lib/rack/lint.rb:491:in 'Rack::Lint::Wrapper::ResponseFinishedWrapper#call'
    test/dispatch/executor_test.rb:183:in 'ExecutorTest#test_complete_callbacks_are_called_on_rack_response_finished'

bin/rails test /home/runner/work/rack/rack/external/rails/actionpack/test/dispatch/executor_test.rb:173

..E

Error:
ExecutorTest#test_complete_callbacks_are_called_once_on_rack_response_finished_when_exception_is_raised:
Rack::Lint::LintError: One of SCRIPT_NAME or PATH_INFO must be set (make PATH_INFO '/' if SCRIPT_NAME is empty)
    /home/runner/work/rack/rack/lib/rack/lint.rb:200:in 'Rack::Lint::Wrapper#check_environment'
    /home/runner/work/rack/rack/lib/rack/lint.rb:491:in 'Rack::Lint::Wrapper::ResponseFinishedWrapper#call'
    test/dispatch/executor_test.rb:202:in 'ExecutorTest#test_complete_callbacks_are_called_once_on_rack_response_finished_when_exception_is_raised'
```

ioquatix:
@adrianna-chang-shopify are you able to fix the Rails failures?

skipkayhil:
I opened a PR to fix Rails^ and it also found a bug in my last commit, so I updated the implementation/tests a bit more too

lib/rack/lint.rb
```diff
@@ -462,15 +462,51 @@ def check_environment(env)
...
 
+      class ResponseFinishedWrapper
+        def initialize(wrapper, callback)
+          @wrapper = wrapper
+          @callback = callback
+        end
+
+        def call(env, status, headers, error)
+          ## * <tt>env</tt> a valid Rack environment
+          env.nil? || @wrapper.check_environment(env)
```
jeremyevans, line 491: I don't think the environment should ever be nil, it must always be valid.  Other values can be nil. In general, status and headers should be set if an error is not given, but there could be potentially be cases where that isn't true (non-local, non-exception exit such as `throw`).

jeremyevans (commented):
Looks good, except for one tweak I think we should make.

ioquatix:
Can you please update the changelog.

PR #2368, opened 2025-08-22
Use SimpleCov.add_filter block instead of string

jeremyevans:
Using a string is almost always a bad idea, because if any part of the path contains the string, it will filter out the path. Switch to using a block and doing a match on the entire path prefix.


test/helper.rb
```diff
@@ -5,7 +5,7 @@
 
   SimpleCov.start do
     enable_coverage :branch
-    add_filter "/test/"
+    add_filter{|f| f.filename.match?(%r{\A#{Regexp.escape(__dir__)}/})}
```
ioquatix, line 8:
```suggestion
    add_filter{|f| f.filename.start_with?(__dir__)}
```

Is this easier to read? Otherwise, I don't have much opinion.

ioquatix (approved):

test/helper.rb
jeremyevans: it misses the following slash.  It's not a big deal, but if we ever added something that started with `test`, it would filter that out too.

I guess we could do:

`add_filter{|f| f.filename.start_with?(__dir__ + "/")}`

ioquatix:
LGTM.

PR #2369, opened 2025-08-23
Explicitly mark request environment variables as required or optional

### LINKED ISSUE #2309: Re-order `rack.`
@jeremyevans

> Can we put the required `rack.` variables first under a heading, and list the optional `rack.` variables second under a different heading? If you are OK with it, but it sounds like too much work, I can work on a PR for that after this is merged.

I am okay with this. However, there are many dimensions on which we could organise this information. Maybe just clearly defining which fields are required or not is sufficient?

jeremyevans:
While you could read the description for each variable to determine whether it is required or optional, having it in the header for the variable makes it more clear.

Fixes #2309

### DIFF
--- a/SPEC.rdoc
+++ b/SPEC.rdoc
@@ -20,7 +20,7 @@ The server and application can store their own data in the environment, too. The
 
 ==== <tt>REQUEST_METHOD</tt>
 
-The HTTP request method, such as "GET" or "POST". This cannot ever be an empty string, and so is always required.
+The HTTP request method, such as "GET" or "POST".
 
 ==== <tt>SCRIPT_NAME</tt>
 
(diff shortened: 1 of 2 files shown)


ioquatix:
I don't think we should do this in the section title because IIUC:

1. It's going to mess with permanent links to sections.
2. It's going to make the sidebar table of contents hard to read.

jeremyevans:
Fair enough, what do you propose instead?

ioquatix:
I have no idea actually. I actually reviewed the same issue, considered the same option and came to the same conclusion I mentioned, that the ToC would become a mess, then didn't know how to proceed so moved on to other things.

jeremyevans:
I tried a different approach that adds a list of required keys and a list of optional keys, without modifying the related headers.

ioquatix:
After discussion, we decided to close this.

Issue #2373, opened 2025-08-28
`Rack::Events` breaks streaming bodies


unflxw:
When `Rack::Events` is in the middleware chain, it causes responses with streaming bodies (where the body responds to `#call` and not `#each`) to fail with a `NoMethodError`.

## Root cause analysis

The `Rack::Events` middleware allows one or more handlers to react to events related to a request. One of these events is `on_send`, which is triggered when the response is sent.

In order to implement `on_send`, `Rack::Events` wraps the body that is passed down the middleware chain in a `Rack::Events::EventedBodyProxy`, a subclass of `BodyProxy` with [a custom `#each` implementation](https://github.com/rack/rack/blob/main/lib/rack/events.rb#L89-L92) that calls the `on_send` method on the handlers and calls `#each` on the proxied body.

According to [the spec](https://github.com/rack/rack/blob/main/SPEC.rdoc#label-The+Body), given a body that implements both `#each` and `#call`, a conforming Rack server must call `#each`. Streaming bodies must therefore only implement `#call`:

> A Body that responds to both `each` and `call` must be treated as an Enumerable Body, not a Streaming Body. If it responds to `each`, you must call `each` and not `call`. If the Body doesn't respond to `each`, then you can assume it responds to `call`.

Due to `EventedBodyProxy`'s custom `#each` instrumentation, the body will always respond to `#each`, and therefore will always be treated as an enumerable body, and not as a streaming body. The Rack server will call `#each` on the `EventedBodyProxy`, which will in turn call `#each` on the body it proxies -- **even when the proxied body does not implement `#each`**.

## Quick "fix"

In `EventedBodyProxy`'s implementation:

```ruby
def respond_to?(method, include_all=false)
  return @body.respond_to?(:each, include_all) if method == :each
  super
end
```

This should cause it to only respond to `#each` if the body it proxies also responds to `#each`, preventing it from breaking streaming bodies' requests.

**This would not be a proper fix, as it does not implement the `on_send` event for streaming bodies.** `EventedBodyProxy` does not implement `#call`, and so it will proxy it directly to the proxied body (as per `BodyProxy`'s `method_missing`, which it inherits from) without ever calling `on_send` on the handlers. A proper fix, I believe, would need to call `on_send` when `close` is called on the stream, probably by wrapping it in some sort of `EventedStreamProxy`.

## Reproduction case

```ruby
# config.ru
require "rack/events"

class StreamingBody
  def call(stream)
    stream.write "Hello world!"
    stream.close
[...]

PR #2375, opened 2025-09-01
Rake events streaming bodies

unflxw:
Fixes rack/rack#2373.

---

## [Fix streaming bodies when using Rack::Events](https://github.com/rack/rack/commit/65317d91a3a443828e76b5a9c7ff8aec59a38e4d)

When `Rack::Events` is used, it would use `EventedBodyProxy` to wrap
the body. Unlike `BodyProxy`, `EventedBodyProxy` always implements
`#each`.

As per the spec, streaming bodies must both implement `#call` and not
implement `#each`, and if both are present, the server must call
`#each`, not `#call`.

When a streaming body was wrapped by an `EventedBodyProxy`, the proxy
will nonetheless `.respond_to?(:each)`. This would cause the server to
call `#each` on the proxy. The proxy would then forward the call to
`#each` to the body that it proxies, which does not implement `#each`,
raising a `NoMethodError`.

To fix this, `EventedBodyProxy` will only `.respond_to?(:each)` if
the body it proxies responds to `#each`.

Note that, while this fixes the `NoMethodError`, this means that the
`on_send` handler will never be invoked for streaming bodies.

## [Call on_send handler for streaming bodies](https://github.com/rack/rack/commit/fcc4c679bcccdc1e186148991b3710f5cd6461a7)

When using the `Rack::Events` middleware, call `on_send` on the
middleware's handlers when `#call` is invoked, in the same way it
is done when `#each` is invoked.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ All notable changes to this project will be documented in this file. For info on
 
 - Raise before exceeding a part limit, not after. ([#2362](https://github.com/rack/rack/pull/2362), [@matthew-puku](https://github.com/matthew-puku))
 
+### Fixed
+
+- Fix an issue where a `NoMethodError` would be raised when using `Rack::Events` with streaming bodies. ([#2375](github.com/rack/rack/pull/2375), [@unflxw](https://github.com/unflxw))
+
 ## [3.2.0] - 2025-07-31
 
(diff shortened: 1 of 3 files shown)


ioquatix (approved):

ioquatix:
Thanks for your contribution.

As this is a bug fix, we could consider backporting. WDYT @jeremyevans?

jeremyevans:
No objections to backporting this fix to 3.2.

PR #2377, opened 2025-09-09
Allow dynamic headers for Rack::Files

codergeek121:
This PR adds support for dynamic headers in Rack::Files via the headers configuration.

For example the following configuration add the header "x-dynamic" with a random integer between 1 and 10 for each request handled by Rack::Files:

```
Rack::Files.new(".", {
  "x-dynamic" => ->(request) { rand(1..10).to_s }
})
```

This feature could be used to provide different cache durations for different files. E.g. a longer duration for files with a contenthash and shorter duration for files without a contenthash. This would allow solving issues like
https://github.com/rails/rails/pull/55249 within Rack::Files directly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ All notable changes to this project will be documented in this file. For info on
 
 ### Added
 
+- Add support for dynamic headers with procs to `Rack::Files`. ([#2377](https://github.com/rack/rack/pull/2377), [@codergeek121](https://github.com/codergeek121))
 - Add support for `rack.response_finished` to `Rack::TempfileReaper`. ([#2363](https://github.com/rack/rack/pull/2363), [@skipkayhil](https://github.com/skipkayhil))
 - Add support for streaming bodies when using `Rack::Events`. ([#2375](github.com/rack/rack/pull/2375), [@unflxw](https://github.com/unflxw))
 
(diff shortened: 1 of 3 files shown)


jeremyevans (commented):
I'm against adding this, as I don't think the complexity is worth the benefit. I'd be OK adding a `set_headers` method that is passed the base headers and request, and runs ` headers.merge!(@headers) if @headers` by default. You could then subclass `Rack::Files` and override the method to add the behavior you want.

codergeek121:
I think this is a great suggestion. This keeps `Rack::Files` simple and would still allow to add a similar feature to Rails. I'll adapt this PR 👍

codergeek121:
I adapted the PR to provide a public `set_headers` method and added a simple test for it 👍

ioquatix:
This looks good to me but I have a slight concern about the name `set_headers`, since it sounds like it might set `@headers` or something similar. Ruby has some similar conventions like `Exception#set_backtrace`.

Perhaps `assign_headers` is better? @jeremyevans do you have any thoughts?

jeremyevans:
> This looks good to me but I have a slight concern about the name `set_headers`, since it sounds like it might set `@headers` or something similar. Ruby has some similar conventions like `Exception#set_backtrace`.
> 
> Perhaps `assign_headers` is better? @jeremyevans do you have any thoughts?

I'm fine with either, so if you would prefer `assign_headers`, let's go with that.

ioquatix:
@codergeek121 wdyt?

codergeek121:
I'd be fine with both names as well. I think your points made sense though and I've renamed the method 👍

ioquatix (approved):

PR #2389, opened 2025-10-09
Runtime in Server-Timing header

lewispb:
[Server-Timing](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Server-Timing) is the standard way to report server performance to the user agent nowadays. I like that it's standardized around reporting milliseconds, and this is presented meaningfully in web browsers dev tools. 

This PR introduces `Rack::ServerTiming` to set (or append to) the Server-Timing response header, with a duration measurement value, similar to the existing `Rack::Runtime` middleware.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@ All notable changes to this project will be documented in this file. For info on
 - Add `Rack::Request#prefetch?` for identifying requests with `Sec-Purpose: prefetch` header set. ([#2405](https://github.com/rack/rack/pull/2405), [@glaszig](https://github.com/glaszig))
 - Add `rack.request.trusted_proxy` environment key to indicate whether the request is coming from a trusted proxy.
 - Add `Rack::Request#headers` for simpler access to request headers by header name. ([#1881](https://github.com/rack/rack/pull/1881), [@jeremyevans](https://github.com/jeremyevans))
+- Introduce `Rack::ServerTiming` middleware for instrumenting runtime in part or all of an application, and reporting via the [Server-Timing](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Server-Timing) header. ([#2375](github.com/rack/rack/pull/2389), [@lewispb](https://github.com/lewispb))
 
 ### Changed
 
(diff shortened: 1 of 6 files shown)


jeremyevans (commented):
I can't think of a situation where I'd want both headers, one with seconds and one with milliseconds.

In terms of implementation, this overrides an existing `server-timing` header, when it should probably append to one that already exists. And when using milliseconds, we'd probably want `%0.3f` instead of `%0.6f` (most examples I see in the Mozilla/W3C use at most 2 digits of precision).

Maybe it would be better to add a `Rack::ServerTiming` middleware that only handles `server-timing`?

lewispb:
> I can't think of a situation where I'd want both headers, one with seconds and one with milliseconds.

That's fair. Happy to split this out into `Rack::ServerTiming`.

> this overrides an existing server-timing header

I don't think my implementation does, does it?

jeremyevans:
> > this overrides an existing server-timing header
> 
> I don't think my implementation does, does it?

No, it doesn't, sorry about that.

lewispb:
Thanks for your initial review @jeremyevans. I've addressed it with a new `Rack::ServerTiming` middleware here.

lewispb:
It looks like the external test failure is unrelated.

<details>
<summary>Failure message</summary>

```
  1) Rack::Sendfile when calling sendfile contains Sendfile headers
     Failure/Error: expect(headers).to include('X-Accel-Redirect')
       expected {"cache-control" => "no-cache", "content-type" => "application/json"} to include "X-Accel-Redirect"
       Diff:
       @@ -1 +1,2 @@
       -["X-Accel-Redirect"]
       +"cache-control" => "no-cache",
       +"content-type" => "application/json",
     # ./spec/grape/integration/rack_sendfile_spec.rb:34:in `block (3 levels) in <top (required)>'
```
</details>

jeremyevans (approved):
Looks good, but needs to be rebased to fix conflicts. Rebasing should fix the external tests as well.

lewispb:
@jeremyevans rebased!

ioquatix:
@copilot resolve the merge conflicts in this pull request

ioquatix:
I'm slightly against adding more middleware to Rack. Probably better in `rack-contrib`.

PR #2396, opened 2025-10-18
Reject line folding in multipart headers

wtn:
Commit 4795831a0a310c2d31102749e551b38faab6401f enables obsolete line folding (obs-fold) contrary to [RFC 7230 §3.2.4](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.4).

So we revert 4795831a0a310c2d31102749e551b38faab6401f entirely, and provide a replacement fix with regression tests.

RFC 7230 §3.2.4 ¶5 states:
> A server that receives an obs-fold in a request message that is not within a message/http container MUST either reject the message by sending a 400 (Bad Request), preferably with a representation explaining that obsolete line folding is unacceptable, or replace each received obs-fold with one or more SP octets prior to interpreting the field value or forwarding the message downstream.

Rejecting is the correct choice, as there is no benefit to processing such requests.

### DIFF
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -27,14 +27,19 @@ module Rack
       include BadRequest
     end
 
+    # Line folding not allowed in multipart headers
+    class ObsoleteFoldingError < StandardError
+      include BadRequest
+    end
+
     # Prefer to use the BoundaryTooLongError class or Rack::BadRequest.
     Error = BoundaryTooLongError
 
     EOL = "\r\n"
     MULTIPART = %r|\Amultipart/.*boundary=\"?([^\";,]+)\"?|ni
-    MULTIPART_CONTENT_TYPE = /Content-Type:[ \t]*(.*)#{EOL}/ni
-    MULTIPART_CONTENT_DISPOSITION = /Content-Disposition:(.*)(?=#{EOL}(\S|\z))/ni
-    MULTIPART_CONTENT_ID = /Content-ID:[ \t]*([^#{EOL}]*)/ni
+    MULTIPART_CONTENT_TYPE = /^Content-Type:[ \t]*(.*)#{EOL}/ni
+    MULTIPART_CONTENT_DISPOSITION = /^Content-Disposition:(.*)(?=#{EOL}(\S|\z))/ni
+    MULTIPART_CONTENT_ID = /^Content-ID:[ \t]*([^#{EOL}]*)/ni
 
     # Rack::Multipart::Parser handles parsing of multipart/form-data requests.
     #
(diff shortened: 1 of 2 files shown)


matthewd:
RFC 7230 describes how a server should behave during parsing of HTTP headers; this code is concerned with parsing MIME headers, which I believe is currently defined by [RFC 5322 §2.2.3](https://datatracker.ietf.org/doc/html/rfc5322#section-2.2.3).

wtn:
RFC 7230 concerns processing HTTP messages, inclusive of headers and bodies.

RFC 7230 has been superseded by [RFC 9112](https://datatracker.ietf.org/doc/html/rfc9112) but I would keep the references as-is.

Suggest to check out and run the tests for each commit in sequence.

jeremyevans:
@matthewd is correct. RFC 7230 Section 3 is about HTTP messages (i.e. requests or responses). Section 3.2 applies to parsing header fields for those messages. It does not apply to MIME headers inside request bodies. RFC 5322 covers the MIME header parsing and requires handling folding.

One way to think about this difference is to consider HTTP 2. With HTTP 2, RFC 7230/9112 does not apply (RFC 7540/9113 does). However, HTTP 1/2/3 only changes the HTTP message parsing, it does not change the format of request bodies for multipart messages. The multipart request body format remains the same regardless of HTTP version.

lib/rack/multipart/parser.rb
```diff
@@ -27,16 +27,19 @@ class BoundaryTooLongError < StandardError
...
     EOL = "\r\n"
-    FWS = /[ \t]+(?:\r\n[ \t]+)?/ # whitespace with optional folding
-    HEADER_VALUE = "(?:[^\r\n]|\r\n[ \t])*" # anything but a non-folding CRLF
     MULTIPART = %r|\Amultipart/.*boundary=\"?([^\";,]+)\"?|ni
-    MULTIPART_CONTENT_TYPE = /^Content-Type:#{FWS}?(#{HEADER_VALUE})/ni
-    MULTIPART_CONTENT_DISPOSITION = /^Content-Disposition:#{FWS}?(#{HEADER_VALUE})/ni
-    MULTIPART_CONTENT_ID = /^Content-ID:#{FWS}?(#{HEADER_VALUE})/ni
+    MULTIPART_CONTENT_TYPE = /^Content-Type:[ \t]*(.*)#{EOL}/ni
+    MULTIPART_CONTENT_DISPOSITION = /^Content-Disposition:(.*)(?=#{EOL}(\S|\z))/ni
+    MULTIPART_CONTENT_ID = /^Content-ID:[ \t]*([^#{EOL}]*)/ni
```
jhawthorn: I don't have an opinion on treatment of line-folding, but this introduces an obvious ReDoS and we must not merge it as-is!

jhawthorn (changes requested):

wtn:
@jhawthorn The patch meant to fix the ReDoS is in the following commit, and can be squashed if necessary. The commits are separate now for legibility.

wtn:
Commits b26a314 and bc74e65 resolve this issue.

Issue #2395, opened 2025-10-17
Wrong proxy order of X-Forwarded-Host


sdalu:
Code for `forwarded_authority` takes the last element of the proxy chain
https://github.com/rack/rack/blob/d8d50702429265c28b84968e43caafc4bffd82ef/lib/rack/request.rb#L406

Mozilla documentation and Apache behaviour is to add a new proxy at the end, so the  `forwarded_authority`  should take the first element
https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For

(I haven't check if the problem exists for other x-forwarded-*)

wtn:
The intent for `forwarded_authority` is to return the last (most recent trusted proxy) value for security reasons.

It sounds like what you actually want is the client IP address.

ioquatix:
@wtn is correct. There is past discussion, e.g. https://github.com/rack/rack/issues/1746

sdalu:
I don't want the client IP, but I do want the hostname/scheme/proto used to make the initial web request. Otherwise, Sinatra doesn't build the correct URL (`uri` method) when passing through multiple proxies.

Issue #1746, opened 2021-04-21
Selecting last ip address in the X-Forwarded-For Header


ahoglund:
👋🏼 Friendly Rack Maintainers!

We've encountered a bit of an issue with the return value of the `ip` method in the case where the `Rack::Request.ip_filter` has been set to a custom lambda. In the scenario where `Rack::Request.ip_filter` has been set to not exclude all private ip address ranges, this line:

https://github.com/rack/rack/blob/master/lib/rack/request.rb#L368

would return a proxy address rather than a client ip.  At the very least, it has more _potential_ to do so since it's selecting the `last` ip in the list rather than the first.  And it is my understanding that the `X-Forwarded-For` header should be populated with `client, proxy1, proxy1`.

Example:

```ruby
   # override the ip filter to only include 127.0.0.1, ::1, and localhost
    Rack::Request.ip_filter = lambda { |ip| /\A127\.0\.0\.1\z|\A::1\z|\Alocalhost\z/i.match(ip)  }
```

Doing this type of override may be needed in the case where a request is coming from some other server in a private cloud or datacenter, but is still considered an "external" service.

```shell
$ curl -H "X-Forwarded-For: 10.5.4.3, 10.2.2.2" http://service:3000
```

This request would have its `ip` set as `10.2.2.2` since neither of these ip's were filtered out. But according to how `X-Forwarded-For` works (and [this comment](https://github.com/rack/rack/blob/master/lib/rack/request.rb#L364-L367)) it seems that in this case we would want to default to grabbing the first ip in the list in this scenario.

If I take the standard scenario, where the default Rack ip_filter is used selecting the `first` ip in the list would still work as desired.

```shell
$ curl -H "X-Forwarded-For: 88.2.3.4, 10.2.2.2" http://service:3000
```

In this case, the `10.2.2.2` would be filtered by the default [Rack ip_filter](https://github.com/rack/rack/blob/master/lib/rack/request.rb#L19).  So whether or not you grab `first` or `last` is irrelevant.

And in the case where it filters out everything:
```shell
$ curl -H "X-Forwarded-For: 10.5.4.3, 10.2.2.2" http://service:3000
```

It would grab the `first` item in this list anyways.

I am proposing we change [this code](https://github.com/rack/rack/blob/master/lib/rack/request.rb#L368) to call `first` in both scenarios.

I would be happy to open a PR to change this but wanted to ask this question first because maybe there are other scenarios in which getting `last` is totally valid.

Thanks!

[...]

PR #2397, opened 2025-10-20
Correct IP address order

alexanderadam:
What Stéphane pointed out in #2395 sounds right to me. [Mozilla Doc](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For#directives) says

> If a request goes through multiple proxies, the IP addresses of each successive proxy are listed. This means that the rightmost IP address is the IP address of the most recent proxy and the leftmost IP address is the address of the originating client (assuming well-behaved client and proxies).

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -403,7 +403,7 @@ module Rack
               return forwarded.last
             end
           when :x_forwarded
-            if (value = get_header(HTTP_X_FORWARDED_HOST)) && (x_forwarded_host = split_header(value).last)
+            if (value = get_header(HTTP_X_FORWARDED_HOST)) && (x_forwarded_host = split_header(value).first)
               return wrap_ipv6(x_forwarded_host)
             end
           end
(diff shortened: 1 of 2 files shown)


wtn:
The [security concerns](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For#security_and_privacy_concerns) section of the MDN document says:

> Leftmost (untrusted) values must only be used for cases where there is no negative impact from using spoofed values.

The rightmost value is the only one we always know isn't spoofed, that's why it's the authority. `:-)`

So the proposed change looks wrong to me, and cannot be accepted.

ioquatix:
@wtn is correct. There is past discussion, e.g. https://github.com/rack/rack/issues/1746

wtn:
@ioquatix Can you reopen this discussion, please.

There is some confusion, as the docs quoted above refer to `x-forwarded-for`.

But the patch changes `x-forwarded-host` and the change may be what we want.

The tests for `x-forwarded-host` might more appropriately use domain names instead of IP addresses, to clarify their purpose.

ioquatix:
It looks like I'm unable to re-open it because the fork was deleted. Do you want to re-open a PR?

alexanderadam:
Yes, I thought that it is not needed any more. I guess I was too fast with cleaning up my forks. :wink: 

The PR is available at #2400 again. That's an easy PR id to remember for sure.

ioquatix:
Thanks and I apologise for not paying enough attention to this PR, it has been an extremely busy past 2 weeks.

PR #2400, opened 2025-10-23
Allow to configure the selected value for `x-forwarded-host`

alexanderadam:
This is basically PR #2397. I just deleted the fork because I though that it is not needed any more :laughing: 

/CC @wtn @ioquatix @jeremyevans 
___

What Stéphane pointed out in #2395 sounds right to me. [Mozilla Doc](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/X-Forwarded-For#directives) says

> If a request goes through multiple proxies, the IP addresses of each successive proxy are listed. This means that the rightmost IP address is the IP address of the most recent proxy and the leftmost IP address is the address of the originating client (assuming well-behaved client and proxies).

I couldn't find a spec that defines the rules for `x-forwarded-host` but I'd _assume_ that they apply there as well.
Also [the GitLab folks say](https://gitlab.com/gitlab-org/gitlab-pages/-/merge_requests/386):

> When multiple reverse-proxies are stacked, the X-Forwarded-Host header
contains all hops, separated by ", ".
> We take the first one, i.e. probably the fqdn set by the browser.

Howeveeeeer, [another comment](https://gitlab.com/gitlab-org/gitlab-pages/-/merge_requests/386#note_535448323) in that very issue states that Apache might _prepend_ hosts. I don't read it that way [in Apache's documentation](https://httpd.apache.org/docs/current/mod/mod_proxy.html#x-headers) though.
Is Apache still using SVN? :scream: 
If so, then [this might still be the current code](https://svn.apache.org/repos/asf/httpd/httpd/tags/2.4.9/modules/proxy/proxy_util.c):

```c
    if (dconf->add_forwarded_headers) {
        if (PROXYREQ_REVERSE == r->proxyreq) {
            const char *buf;

            /* Add X-Forwarded-For: so that the upstream has a chance to
             * determine, where the original request came from.
             */
            apr_table_mergen(r->headers_in, "X-Forwarded-For",
                             r->useragent_ip);

            /* Add X-Forwarded-Host: so that upstream knows what the
             * original request hostname was.
             */
            if ((buf = apr_table_get(r->headers_in, "Host"))) {
                apr_table_mergen(r->headers_in, "X-Forwarded-Host", buf);
            }

            /* Add X-Forwarded-Server: so that upstream knows what the
             * name of this proxy server is (if there are more than one)
             * XXX: This duplicates Via: - do we strictly need it?
             */
            apr_table_mergen(r->headers_in, "X-Forwarded-Server",
                             r->server->server_hostname);
        }
    }
```

But it's too late at my place to make any sense of that right now.
I will go to :bed: bed now. :wink: 

**PS:** I'm looking for a new adventure in case anybody is looking to hire someone or in case someone would like to work with a PO/PM or Ruby/Rails/Crystal dev

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -412,7 +412,7 @@ module Rack
               return forwarded.last
             end
           when :x_forwarded
-            if (value = get_header(HTTP_X_FORWARDED_HOST)) && (x_forwarded_host = split_header(value).last)
+            if (value = get_header(HTTP_X_FORWARDED_HOST)) && (x_forwarded_host = split_header(value).first)
               return wrap_ipv6(x_forwarded_host)
             end
           end
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -38,10 +38,21 @@ module Rack
       # similar to <tt>[:scheme, :proto]</tt>.  You can remove either or
       # both of the entries in array to ignore that respective header.
       attr_accessor :x_forwarded_proto_priority
+
+      # Whether to use the first (leftmost) value from the
+      # <tt>X-Forwarded-Host</tt> header. The default is +false+,
+      # which returns the last (rightmost) value from the most
+      # recent (and typically trusted) proxy. Set to +true+ to
+      # use the first value, which is the original client-supplied
+      # host, useful for generating URLs and redirects. Note that
+      # earlier values can be spoofed by clients, so only set to
+      # +true+ if you trust your proxy chain to sanitize headers.
+      attr_accessor :x_forwarded_host_first
     end
 
     @forwarded_priority = [:forwarded, :x_forwarded]
     @x_forwarded_proto_priority = [:proto, :scheme]
+    @x_forwarded_host_first = false
 
     valid_ipv4_octet = /\.(25[0-5]|2[0-4][0-9]|[01]?[0-9]?[0-9])/
 
@@ -412,7 +423,8 @@ module Rack
               return forwarded.last
             end
           when :x_forwarded
-            if (value = get_header(HTTP_X_FORWARDED_HOST)) && (x_forwarded_host = split_header(value).first)
+            if (value = get_header(HTTP_X_FORWARDED_HOST)) && (hosts = split_header(value)) && !hosts.empty?
+              x_forwarded_host = x_forwarded_host_first ? hosts.first : hosts.last
               return wrap_ipv6(x_forwarded_host)
             end
           end
@@ -786,6 +798,10 @@ module Rack
 
       def x_forwarded_proto_priority
         Request.x_forwarded_proto_priority
+      end
+
+      def x_forwarded_host_first
+        Request.x_forwarded_host_first
       end
     end
 
--- a/test/spec_request.rb
+++ b/test/spec_request.rb
@@ -410,8 +410,8 @@ class RackRequestTest < Minitest::Spec
         forwarded_port.must_equal [2345]
 
       req("HTTP_FORWARDED"=>"host=1.2.3.4, host=3.4.5.6",
-        "HTTP_X_FORWARDED_HOST" => "2.3.4.5,4.5.6.7").
-        forwarded_authority.must_equal '4.5.6.7'
+        "HTTP_X_FORWARDED_HOST" => "example.com,cdn.vade.io,proxy.internal").
+        forwarded_authority.must_equal 'example.com'
 
       req("HTTP_FORWARDED"=>"proto=https",
         "HTTP_X_FORWARDED_PROTO" => "ws",
--- a/test/spec_request.rb
+++ b/test/spec_request.rb
@@ -411,7 +411,7 @@ class RackRequestTest < Minitest::Spec
 
       req("HTTP_FORWARDED"=>"host=1.2.3.4, host=3.4.5.6",
         "HTTP_X_FORWARDED_HOST" => "example.com,cdn.vade.io,proxy.internal").
-        forwarded_authority.must_equal 'example.com'
+        forwarded_authority.must_equal 'proxy.internal'
 
       req("HTTP_FORWARDED"=>"proto=https",
         "HTTP_X_FORWARDED_PROTO" => "ws",
@@ -539,6 +539,31 @@ class RackRequestTest < Minitest::Spec
     ensure
       Rack::Request.forwarded_priority = default_priority
       Rack::Request.x_forwarded_proto_priority = default_proto_priority
+    end
+  end
+
+  # Also read the following comment for more context:
+  # https://github.com/rack/rack/pull/2400#issuecomment-3477891669
+  it "have forwarded_authority respect x_forwarded_host_first" do
+    begin
+      def self.req(headers)
+        make_request(Rack::MockRequest.env_for("/", headers))
+      end
+
+      # default behavior uses last (rightmost, most recent proxy)
+      req("HTTP_X_FORWARDED_HOST" => "client.example.com,proxy.internal").
+        forwarded_authority.must_equal 'proxy.internal'
+
+      Rack::Request.x_forwarded_host_first = true
+
+      # when true, uses first (leftmost, client-supplied)
+      req("HTTP_X_FORWARDED_HOST" => "client.example.com,proxy.internal").
+        forwarded_authority.must_equal 'client.example.com'
+
+      req("HTTP_X_FORWARDED_HOST" => "example.com,vade.io,proxy.internal").
+        forwarded_authority.must_equal 'example.com'
+    ensure
+      Rack::Request.x_forwarded_host_first = false
     end
   end
 
@@ -1368,7 +1393,7 @@ EOF
       'rack.input' => StringIO.new('invalid=bar&invalid[foo]=bar'),
       'HTTP_CONTENT_TYPE' => "application/x-www-form-urlencoded",
     }
-    
+
     2.times do
       # The actual exception type here is unimportant - just that it fails.
       assert_raises(Rack::Utils::ParameterTypeError) do


alexanderadam:
Okay, so I had some time with it again and the way I understand it, is that [`apr_table_mergen` _appends_ the new value](https://apr.apache.org/docs/apr/trunk/group__apr__tables.html#ga1d50805448114c476cfcd00d5ee3e3a8) to an existing header.

This means that Apache is actually doing the right thing like it's expected to.
And as far as I understand, this also means, that rack took the wrong host before.

However, I absolutely don't understand why this was not pointed out before.
That makes it a little suspicious and maybe there's something else that we overlook here?
Or maybe `x-forwarded-host`'s impact isn't _that_ big?
I mean, there are so many applications out there using rack. Did that never pop up somewhere?
I struggle to believe that.

**PS:** I think that [William's proposal to replace the IPs in the test with hostnames](https://github.com/rack/rack/pull/2397#issuecomment-3437716527), makes this far clearer to be honest.
Good thinking.

ioquatix:
This PR changes `X-Forwarded-Host` header handling from `.last` (rightmost/most recent proxy) to `.first` (leftmost/original client) in the `forwarded_authority` method. The git history reveals this behavior has changed multiple times with different rationales.

## Complete Timeline

### 0. **Oct 8, 2015** - Commit [2f782ce0](https://github.com/rack/rack/commit/2f782ce0)
**"add Rack::Request#authority"** by @tenderlove
- **Initial implementation**: Added `authority` method (for HTTP/2 pushes)
- **Code**: Simple `SERVER_NAME + ':' + SERVER_PORT`
- **Note**: No forwarded header handling at this point

### 1. **Before Nov 2019** - Original behavior
- `X-Forwarded-Host` used **`.first`** (leftmost/original client)
- This was the behavior for several years

### 2. **Nov 29, 2019** - Commit [58eed301](https://github.com/rack/rack/commit/58eed301)
**"Support RFC 7239: HTTP Forwarded header"**
- **Changed to**: `.last` (rightmost value)
- **Rationale**: Consistency with RFC 7239 `Forwarded` header standard
- **Code**: Both RFC 7239 `Forwarded` and `X-Forwarded-Host` used `.last`
- **Impact**: Broke from the multi-year precedent of using `.first`

### 3. **Feb 6, 2020** - Commit [290523f6](https://github.com/rack/rack/commit/290523f6)
**"Improve `Rack::Request#authority` and related methods"**
- **Changed back to**: `.first` (leftmost value)
- **Rationale**: Not explicitly stated, but likely recognized the issue with URL generation
- **Code**: Refactored into separate `forwarded_authority` method using `.first`
- **Note**: Simplified to only handle `X-Forwarded-Host` (removed RFC 7239 from this method)
- **Duration**: This `.first` behavior lasted only ~2 years

### 4. **March 17, 2022** - Commit [b87d1828](https://github.com/rack/rack/commit/b87d1828)
**"Allow configuring priorities for Forwarded and X-Forwarded-*"** by @jeremyevans
- **Changed back to**: `.last` (rightmost value) - **CURRENT BEHAVIOR**
- **Rationale**: **"earlier forwarded authorities can be forged by the client"** (security concern)
- **Related Issues**: 
  - Fixes #1809
  - Fixes #1829
  - Implements #1423
  - Implements #1832
- **Key quote from commit message**:
  > "Make Request#forwarded_authority use the last forwarded authority instead of the first forwarded authority, since earlier forwarded authorities can be forged by the client."
- **Duration**: Current behavior for ~3.5 years

## The Core Conflict

### Security Argument (`.last` - current behavior since March 2022)
- Client can forge the leftmost values in `X-Forwarded-Host`.
- Only the rightmost (most recent proxy) value is trustworthy.
- Prevents client spoofing attacks.
- **This is the current behavior and was an explicit security fix**.

### URL Generation Argument (`.first` - proposed in this PR #2400)
- Applications need to know what hostname the **client originally requested**.
- Using `.last` breaks URL generation (issue #2395).
- The original host is needed for redirects, asset URLs, etc.
- **This was the original behavior for many years (2015-2019) and briefly again (2020-2022)**.

## The Real Question

The 2022 commit explicitly chose `.last` for security reasons, arguing that clients can forge earlier values. However, this creates a practical problem: **how does an application know what hostname the client actually requested?**

**Historical Context**: 
- The **original behavior from 2015-2019 was `.first`** - this worked for years.
- Changed to `.last` briefly in Nov 2019 for RFC 7239 consistency.
- Reverted to `.first` in Feb 2020 (likely due to practical issues).
- Changed back to `.last` in March 2022 specifically for security.

This is similar to the `X-Forwarded-For` IP address handling (issue #1746), where Rack ultimately chose `.first` because applications need the original client IP, not the most recent proxy.

**Key consideration**: If we can't trust the client-provided host header, then:
- How should applications generate correct URLs?
- Should this be configurable (like the 2022 commit made other forwarded headers)?
- Is there a different method applications should use for URL generation?
- Why does `X-Forwarded-For` use `.first` for IP but `X-Forwarded-Host` use `.last` for hostname?

## Recommendation

This needs careful consideration of:

1. **Historical precedent**: `.first` was the behavior for ~4 years (2015-2019), then again for 2 years (2020-2022)
2. **Security implications**: Can allowing `.first` enable client spoofing attacks? What specific threat model?
3. **Practical usage**: How are applications supposed to generate URLs if they can't trust the original host?
4. **Consistency**: `X-Forwarded-For` uses `.first` for the client IP - should host work the same way?
5. **Configuration**: Should this be configurable like other forwarded header priorities added in 2022?
6. **Real-world impact**: The 2022 change has been in place for 3.5 years - what actual problems has it caused? If we did decide to change this, what would break?

The 2022 commit by @jeremyevans specifically addressed this with security in mind and fixed multiple issues (#1809, #1829). Any change should either:
- Provide strong justification why the security concern is not valid, OR
- Make this behavior configurable (like `forwarded_priority`), OR  
- Document an alternative approach for applications needing the original host for URL generation, OR
- Explain why the security reasoning that applied to `X-Forwarded-Host` doesn't apply to `X-Forwarded-For`

**Note**: This would be the **third reversal** of this behavior, suggesting either:
- The trade-offs are genuinely difficult and context-dependent (arguing for configurability).
- There's a deeper architectural issue about how forwarded headers should be handled.

Given all that, it may be easier to have a separate method for "client_authority" (or argument, or configuration...). This research was mostly just to understand the background, not what the most appropriate solution is going forward.

alexanderadam:
Soooo, I woke up from my hibernation and I hope that you had a great start into 2026.

I changed the behaviour to be configurable like @ioquatix suggested.
And I think that risks and implications mainly derive from the ecosystem the respective Rack app is deployed to.

`.first` is dangerous if the client sends `X-Forwarded-Host: very-evil-website.com` and the proxy doesn't sanitize it.
IMHO it should be the reverse proxie's responsibility though.
`.last` might cause URL generation issues because it's using the internal host names.

Therefore both are "wrong" in a way — depending on the implications being made.

Making it configurable might be the best solution.

…although having `.first` as the default sounds "more" correct to me and this would be as I understand the specification.
However, this also brings risks. In most cases one would use a reverse proxy and hopefully have it configured to overwrite or strip `X-Forwarded-Host`. Given the risks I changed the default back to `.last` though (as it already is right now). This comes with with "wrong host interpretation bug" risk. But I think that most frameworks allow to set the "real" domain separately anyway, so we're kinda good, I guess?
I hope that the comment in the code makes all of that clear.
It boils down to "correct but insecure" or "incorrect but less risky". One could argue that it's not Rack's responsibility to make sure that users have reverse proxies in place and that they have the securely configured and that Rack should follow the standard as expected instead. But with what I know now, I think that it would be better to have the "wrong" default and making it the users responsibility to make sure that the Rack app is in a safe environment and have them manually switch to the "correct" behaviour.

What do you think?

**TL;DR:** I changed the default back to `.last` but made it configurable.

jeremyevans (commented):
Using `.first` is going to introduce security issues in applications that use `Request#host` assuming it is a valid URL for the host, because it's possible for an attacker to forge it. This seems like a footgun to me.

I think we should discourage users from using Request methods to determine the domain of the request. We should encourage them to configure the valid domain(s) for their application explicitly. I don't think there is a way to use Request methods securely and correctly in all deployment environments. You either use the `host` header ( or `:authority` pseudo-header in HTTP/2), which may be incorrect in reverse proxy cases and insecure if your application can receive requests with arbitrary `host` headers, or you consider `forwarded`/`x-forwarded-for` headers, which will be insecure and/or incorrect in many cases. One way to do this is to support a `rack.request.host` env key that is used in preference to request-supplied information, and encourage the use of the `Rack::Config` middleware to set that.

FWIW, having `host` use `forwarded_authority` before `host_authority` seems like a security issue for any application that accepts requests without a reverse proxy in front, since that would be vulnerable to the same issue as using `first` instead of `last` for forwarded information. @ioquatix what are your thoughts regarding that?

alexanderadam:
> We should encourage them to configure the valid domain(s) for their application explicitly.

I agree and I the projects that I'm aware of are doing things like this anyway for many cases.

> I don't think there is a way to use Request methods securely and correctly in all deployment environments.

Exactly. Both feel wrong.

tenderlove:
> I think we should discourage users from using Request methods to determine the domain of the request. We should encourage them to configure the valid domain(s) for their application explicitly. I don't think there is a way to use Request methods securely and correctly in all deployment environments. You either use the `host` header ( or `:authority` pseudo-header in HTTP/2), which may be incorrect in reverse proxy cases and insecure if your application can receive requests with arbitrary `host` headers, or you consider `forwarded`/`x-forwarded-for` headers, which will be insecure and/or incorrect in many cases. One way to do this is to support a `rack.request.host` env key that is used in preference to request-supplied information, and encourage the use of the `Rack::Config` middleware to set that.

💯 

I don't think Rack can make good security guarantees regarding headers because we can't know how users configure their production applications (nor should we have an opinion about it).

I think adding an env key that users should configure makes sense.

jeremyevans:
> I think adding an env key that users should configure makes sense.

I've submitted #2493, which adds support for `env['rack.request.config'] = {authority: "example.com"}`

ioquatix:
@alexanderadam Can close this one #2493 lands?

alexanderadam:
Superseded by #2493

PR #2405, opened 2025-10-29
add Rack::Request#prefetch? helper

glaszig:
allows to effortlessly identify prefetch request as defined by WICG. inspects the `Sec-Purpose` header as defined in specification [0].

the usefulness is probably best explained by the following quote from the original fetch draft [1]:

"The server can use this to adjust the caching expiry for prefetches, to disallow the prefetch, or to treat it differently when counting page visits."

[0]: https://wicg.github.io/nav-speculation/prefetch.html#sec-purpose-header
[1]: https://fetch.spec.whatwg.org/#http-sec-purpose
also: https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Sec-Purpose

this header has [quite a history][history]. i thought about supporting more permutations but settled on what's in the spec.

[history]: https://lionralfs.dev/blog/exploring-the-usage-of-prefetch-headers/


glaszig:
reasonable. can be mapped in a middleware. i removed the `x-` version.

for reference, the reason turbo uses `x-sec-purpose` is that brower's fetch() implementations don't allow to override `sec-purpose`: hotwired/turbo#1107.

rafaelfranca (approved):

glaszig:
@rafaelfranca to get this working with turbo's `x-sec-purpose` where would be a good place?

- rails itself via `AD::Request#prefetch?` override?
- turbo-rails gem via new helper `turbo_prefetch?` which exclusively checks `x-sec-purpose`?
- turbo-rails gem via mapping to `sec-purpose` in some middleware?

i personally sympathize with option 1 or 2.

ioquatix:
Can you please add a note to the changelog.

glaszig:
> add a note to the changelog

done.

Issue #2414, opened 2025-12-11
Encoding::CompatibilityError in Rack::MethodOverride


bquorning:
When a multipart request is handled by `Rack::MethodOverride`, and one of the parts is using an encoding that is not compatible with ASCII (as per `Encoding#ascii_compatible?`), an `Encoding::CompatibilityError` exceptionis raised.

I wrote the following script to show which encodings cause the error:
```rb
gem "rack", "3.2.4"

require "rack/mock_request"
require "rack/method_override"

APP = Rack::MethodOverride.new(
  proc do
    [200, { "content-type" => "text/plain" }, []]
  end
)

def multipart_request_with_encoding(encoding)
  input = <<~EOF
    --AaB03x\r
    content-type: text/plain; charset=#{encoding.name}\r
    \r
    contents\r
    --AaB03x--\r
  EOF

  env = Rack::MockRequest.env_for(
    "/",
    "CONTENT_TYPE" => "multipart/form-data; boundary=AaB03x",
    "CONTENT_LENGTH" => input.size.to_s,
    :method => "POST",
    :input => input
  )
  APP.call(env)

  return env
end

puts "\n== ASCII compatible"
Encoding.list.select(&:ascii_compatible?).each do |encoding|
  response = multipart_request_with_encoding(encoding)
  puts "Success: #{encoding}"
rescue
  puts "Failed: #{encoding}"
end

puts "\n== ASCII incompatible"
Encoding.list.reject(&:ascii_compatible?).each do |encoding|
  multipart_request_with_encoding(encoding)
  puts "Success: #{encoding}"
rescue
  puts "Failed: #{encoding}"
end

puts "\n== Show the backtrace"
begin
  multipart_request_with_encoding(Encoding::UTF_16LE)
rescue => e
  puts e.backtrace.map { |line| line.sub(%r{^.*/gems/}, ".../gems/") }
end
```

The script’s output, ending with the full stacktrace:

```
== ASCII compatible
Success: ASCII-8BIT
Success: UTF-8
Success: US-ASCII
Success: UTF8-MAC
Success: EUC-JP
Success: Windows-31J
Success: Big5
Success: Big5-HKSCS
Success: Big5-UAO
Success: CESU-8
Success: CP949
Success: Emacs-Mule
Success: EUC-KR
Success: EUC-TW
Success: GB18030
Success: GBK
Success: ISO-8859-1
Success: ISO-8859-2
Success: ISO-8859-3
Success: ISO-8859-4
[...]

PR #2415, opened 2025-12-12
fix: keep multipart names UTF-8 & transcode bodies

alexanderadam:
Soooooo, the way I understand it, the name/field could stay UTF-8 all the time anyway.
While ASCII is _recommended_, UTF-8 is supported if both parties speak it.

> _Some commonly deployed systems use multipart/form-data with file names directly encoded including octets outside the US-ASCII range._
> _The encoding used for the file names is typically UTF-8, although HTML forms will use the charset associated with the form._
> […]
> _If non-ASCII field names are unavoidable, form or application creators SHOULD use UTF-8 uniformly._
> _This will minimize interoperability problems._

So if the client wants to speak UTF-8, so can we. And from ASCII to UTF-8 we have compatibility built-in if I remember correctly.
And we'd not make that special jump from `ISO_2022_JP` to UTF-8 for the field(!) again if we keep UTF-8.
Hence we can save some transcoding work here.

Now, the body is a different story obviously.
And it could be that I'm misunderstanding stuff here.
But I assume that we want to make sense of the body whenever possible.
Thus I'd always _try_ to transcode it into something that we understand.
I know there's a performance impact. But I'd assume that in most of the cases the body will be something that we can handle and we have an early return for that.

But the dummy encodings might include things that will still not transcode into UTF-8. For these things I'd suggest using binary. This way folks who _reeeeally_ want to tackle this can still do whatever they want, without being bothered by an internal Rack exception.

refs #2414

**EDIT:** Why is it _always_ TruffleRuby? :laughing: 
Wow, `ArgumentError: UTF-16 string byte length (51) is not a multiple of 2` - even the error message doesn't look like an `ArgumentError`. But I think encoding-wise there were a few strange things outside of UTF-8 in TruffleRuby. It even rings a bell.

Anyway, I'd like to postpone the fix until it's clear that the PR would be something that we'd like to continue with.

**PS:** I'm looking for a new adventure in case anybody is looking to hire someone or in case someone would like to work with a Ruby/Rails/Crystal dev

### DIFF
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -46,7 +46,7 @@ module Rack
     # information about uploaded files. For +file+ parameters, the hash includes:
     #
     # * +:filename+ - The original filename, already URL decoded by the parser
-    # * +:type+ - The content type of the uploaded file  
+    # * +:type+ - The content type of the uploaded file
     # * +:name+ - The parameter name from the form
     # * +:tempfile+ - A Tempfile object containing the uploaded data
     # * +:head+ - The raw header content for this part
@@ -542,7 +542,7 @@ module Rack
           end
         end
 
-        name.force_encoding(encoding)
+        # We're doing the encoding intentionally ONLY on the body and NOT the name so that it stays UTF-8
         body.force_encoding(encoding)
       end
 
@@ -555,22 +555,22 @@ module Rack
         Encoding::BINARY
       end
 
-      REENCODE_DUMMY_ENCODINGS = {
-        # ISO-2022-JP is a legacy but still widely used encoding in Japan
-        # Here we convert ISO-2022-JP to UTF-8 so that it can be handled.
-        Encoding::ISO_2022_JP => true
-
-        # Other dummy encodings are rarely used and have not been supported yet.
-        # Adding support for them will require careful considerations.
-      }
-
       def handle_dummy_encoding(name, body)
+        # body is a String for text fields and a Hash for file uploads
+        return name, body if !body.is_a?(String) || !body.encoding.dummy?
+
         # A string object with a 'dummy' encoding does not have full functionality and can cause errors.
-        # So here we covert it to UTF-8 so that it can be handled properly.
-        if name.encoding.dummy? && REENCODE_DUMMY_ENCODINGS[name.encoding]
-          name = name.encode(Encoding::UTF_8)
+        # Dummy encodings like UTF-16LE, UTF-16BE, UTF-32LE, UTF-32BE and ISO-2022-JP are ASCII-incompatible
+        # They will raise Encoding::CompatibilityError when used in string operations with ASCII/UTF-8 strings.
+        # See also issue #2414 for details. We convert them so that will be handled properly by the query parser.
+        begin
           body = body.encode(Encoding::UTF_8)
+        rescue Encoding::UndefinedConversionError, Encoding::InvalidByteSequenceError
+          # If it fails we will use binary encoding to avoid crashes.
+          # This would avoid dangerous interpretation and also unnecessary exceptions.
+          body = body.encode(Encoding::BINARY, invalid: :replace, undef: :replace)
         end
+
         return name, body
       end
 
--- a/test/spec_method_override.rb
+++ b/test/spec_method_override.rb
@@ -130,4 +130,23 @@ EOF
     env["REQUEST_METHOD"].must_equal "POST"
     env["rack.request.form_input"].must_be_nil
   end
+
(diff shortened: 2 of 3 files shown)


lib/rack/multipart/parser.rb
```diff
@@ -555,22 +555,22 @@ def find_encoding(enc)
...
-          name = name.encode(Encoding::UTF_8)
+        # Dummy encodings like UTF-16LE, UTF-16BE, UTF-32LE, UTF-32BE and ISO-2022-JP are ASCII-incompatible
+        # They will raise Encoding::CompatibilityError when used in string operations with ASCII/UTF-8 strings.
+        # See also issue #2414 for details. We convert them so that will be handled properly by the query parser.
+        begin
           body = body.encode(Encoding::UTF_8)
+        rescue Encoding::UndefinedConversionError, Encoding::InvalidByteSequenceError
+          # If it fails we will use binary encoding to avoid crashes.
+          # This would avoid dangerous interpretation and also unnecessary exceptions.
+          body = body.encode(Encoding::BINARY, invalid: :replace, undef: :replace)
```
jeremyevans: I would think that when converting to binary encoding it's impossible to have invalid or undef characters. Plus, Ruby has an existing shortcut for this:

```suggestion
          body = body.b
```

However, considering we've already forced encoding on the body earlier (I think), maybe:

```suggestion
          body.force_encoding(Encoding::BINARY)
```

jeremyevans (commented):
This mixes not encoding the parameter name with transcoding the body to UTF-8 for all dummy encodings, which seem like two separate (albeit somewhat related) concerns.

This breaks backwards compatibility for users who expect parameter names to be encoded a certain way. It also changes how other dummy encodings are handled, which I don't think should be done without careful analysis of each dummy encoding.

Is there a problem with encoding parameter names for ASCII-compatible non-UTF-8 encodings? I would guess not, though I didn't do detailed testing:

```ruby
("[" + "\xf0".force_encoding("ISO-8859-1")).encoding
=> #<Encoding:ISO-8859-1>
```

I understand that Rack's behavior is causing production exceptions in your application for requests that you don't want, but I don't think this is the correct approach to solving that.

alexanderadam:
Hi Jeremy, sorry for the late response. I hope that you had a great start into the new year.

> This mixes not encoding the parameter name with transcoding the body to UTF-8 for all dummy encodings, which seem like two separate (albeit somewhat related) concerns.

I see your point. Should I rather create two different PRs here?

In my opinion the parameter name is the easy case at least:
In the beginning of the method we're making sure that the [parameter name is always UTF-8](https://github.com/rack/rack/blob/f2795ed356bcc2ae29414d49e3d701d2386284e1/lib/rack/multipart/parser.rb#L523).
And I guess that's what we want anyway, since this is what is likely to be handled anyway.
To me it reads that the standard expects ASCII with potential to UTF-8 but nothing else.
So once we have the names in UFT-8, we should be good from a Ruby perspective, since that what we're mostly working anyway, right?
Going from $whatever_encoding to UTF-8 and _then_ to whatever we got in `find_encoding` (which might be US_ASCII in most cases [thus having a low impact since UTF-8 is a superset] but it would be binary in the worst case).
Removing that line sounds like a win to me: keeping UTF-8 should make things easy, and if we already had that, we will just save a few CPU cycles (and I know how much you love performance :wink:).

> This breaks backwards compatibility for users who expect parameter names to be encoded a certain way.

I understand and agree with that, although I'm not really sure whether the impact would be realistically huge.
Do you know of any projects/code that make certain assumptions? Expecting US_ASCII is understandable but changing to UTF-8 should be fine'ish.
Would you recommend doing something particular for this very case?

> It also changes how other dummy encodings are handled, which I don't think should be done without careful analysis of each dummy encoding.

All dummy encodings could cause the same exception though. No?
Would you like me to extend the test suite for certain edge cases?

> Is there a problem with encoding parameter names for ASCII-compatible non-UTF-8 encodings?

I guess problems might mainly appear once we mix these names with other strings in Ruby (which are usually UTF-8).
Since the RFC is fine with UTF-8, we should be fine here in most cases I guess.

Unless someone is really doing `if foo.encoding == Encoding::ISO_8859_1`.

> I understand that Rack's behavior is causing production exceptions in your application for requests that you don't want, but I don't think this is the correct approach to solving that.

I don't have strong opinions here and I'm also fine with dropping the PR.
Are both changes (parameter name and body) not the correct approaches in your opinion?

> Ruby has an existing shortcut for this (`body.b`)

Super cool! Thank you!

alexanderadam:
Ah, nevermind. 

It's superseeded by #2416

PR #2417, opened 2026-01-09
Drop Ruby v3.2 from external tests as Rails head no longer supports it

bquorning:
Ruby 3.2 support was removed in https://github.com/rails/rails/commit/fc22a8fc62d5240c0d09f1078800900ea2


bquorning:
I have a couple of Rails tests failing when I add Ruby 4.0.

PR #2420, opened 2026-01-23
Fix MockResponse#body when the body is a Proc

tavianator:
Previously, we would see errors like

    NoMethodError: undefined method 'each' for an instance of Proc

in Rails integration tests when using streaming response bodies.


ioquatix:
I think this is a good step in the right direction, but we might want to also consider how a user might test actual bi-directional communication.

PR #2425, opened 2026-02-08
Fix 'Upgrade Guide' link to 3.0-3.2 pages

juneira:
The link "Upgrade Guide" is broken on documentation of 3.0 to 3.2 versions.

Example:
[Upgrade Guide - 3.2 documentation](https://rack.github.io/rack/3.2/UPGRADE-GUIDE_md.html)

image

image

### DIFF
--- a/.github/workflows/documentation.yaml
+++ b/.github/workflows/documentation.yaml
@@ -40,6 +40,8 @@ jobs:
         gem unpack rack -v "< 3.3"
         cd rack-3.2*
         cp ../.rdoc_options .
+        cp ../UPGRADE-GUIDE.md .
+        cp ../SECURITY.md .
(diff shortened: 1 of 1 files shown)


ioquatix:
This won't fix the issue unless we back port it right?

juneira:
> This won't fix the issue unless we back port it right?

yes

juneira:
The link "Security Policy" is broken too, I hadn't see this before.

juneira:
I change the workflow to include the _md files_ instead of include this on `gemspec`.
So this will fix without back port it.

I tested this on my fork:
[Github Page Rack](https://juneira.github.io/rack/)

@jeremyevans  @ioquatix

jeremyevans (approved):

ioquatix:
Thanks!

PR #2428, opened 2026-02-22
Freeze more internal constants and improve Ractor compatibility

jhawthorn:
This freezes a bunch of constants and replaces some unshareable procs with plain classes. This moves us towards being able to use `Rack::Request` and `Rack::Response` from within a Ractor, as well as can help prevent thread-safety issues.

I tried to be very conservative in limiting this to constants users should not be modifying. For example `FORM_DATA_MEDIA_TYPES`, `PARSEABLE_DATA_MEDIA_TYPES`, `MIME_TYPES`, `Rack::Request.forwarded_priority`, and `Rack::Request.x_forwarded_proto_priority` are documented (or seem likely) to be modifiable by users so they were left unchanged. I also didn't change any optional middleware (some of which have constants that could be frozen) other than `Lint` to keep this focused.

The two procs here could be made safe using Ruby 4.0's `Ractor.shareable_proc`, but it's simpler (and compatible with all versions) to just create a `call`able class.

If any of these we're worried about or are contentious I can back it out and try separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ All notable changes to this project will be documented in this file. For info on
 - Multipart parser drops support for RFC 2231 `filename*` parameter (prohibited by RFC 7578) and now properly handles UTF-8 encoded filenames via percent-encoding and direct UTF-8 bytes. ([#2398](https://github.com/rack/rack/pull/2398), [@wtn](https://github.com/wtn))
 - The query parser now raises `Rack::QueryParser::IncompatibleEncodingError` if we try to parse params that are not ASCII compatible. ([#2416](https://github.com/rack/rack/pull/2416), [@bquorning](https://github.com/bquorning))
 - The mime type for `.pem` files has been changed from `application/x-x509-ca-cert` to `application/x-pem-file`. ([#2435](https://github.com/rack/rack/pull/2435), [@jeremyevans](https://github.com/jeremyevans))
+- Freeze `Rack::Auth::AbstractRequest::AUTHORIZATION_KEYS`, `Rack::Utils::STATUS_WITH_NO_ENTITY_BODY`, `Rack::Multipart::Parser::EMPTY`, `Rack::Utils.default_query_parser`, and internal constants in `Rack::Lint`. Also make the multipart parser's default Tempfile factory Ractor-compatible. ([#2428](https://github.com/rack/rack/pull/2428), [@jhawthorn](https://github.com/jhawthorn))
 
 ### Fixed
 
--- a/lib/rack/auth/abstract/request.rb
+++ b/lib/rack/auth/abstract/request.rb
@@ -37,7 +37,7 @@ module Rack
 
       private
 
-      AUTHORIZATION_KEYS = ['HTTP_AUTHORIZATION', 'X-HTTP_AUTHORIZATION', 'X_HTTP_AUTHORIZATION']
+      AUTHORIZATION_KEYS = ['HTTP_AUTHORIZATION', 'X-HTTP_AUTHORIZATION', 'X_HTTP_AUTHORIZATION'].freeze
 
       def authorization_key
         @authorization_key ||= AUTHORIZATION_KEYS.detect { |key| @env.has_key?(key) }
--- a/lib/rack/lint.rb
+++ b/lib/rack/lint.rb
@@ -922,7 +922,7 @@ module Rack
         verify_to_path
       end
 
-      BODY_METHODS = {to_ary: true, each: true, call: true, to_path: true}
+      BODY_METHODS = {to_ary: true, each: true, call: true, to_path: true}.freeze
 
       def to_path
         @body.to_path
@@ -976,7 +976,7 @@ module Rack
         REQUIRED_METHODS = [
           :read, :write, :<<, :flush, :close,
           :close_read, :close_write, :closed?
-        ]
+        ].freeze
 
         def_delegators :@stream, *REQUIRED_METHODS
 
--- a/lib/rack/multipart/parser.rb
+++ b/lib/rack/multipart/parser.rb
@@ -117,7 +117,7 @@ module Rack
       end
 
       MultipartInfo = Struct.new :params, :tmp_files
-      EMPTY         = MultipartInfo.new(nil, [])
+      EMPTY         = MultipartInfo.new(nil, [].freeze).freeze
 
       def self.parse_boundary(content_type)
         return unless content_type
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -68,7 +68,6 @@ module Rack
       @param_depth_limit = param_depth_limit
       @bytesize_limit = bytesize_limit
       @params_limit = params_limit
-      freeze
     end
 
     # Stolen from Mongrel, with some small modifications:
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -68,6 +68,7 @@ module Rack
       @param_depth_limit = param_depth_limit
       @bytesize_limit = bytesize_limit
       @params_limit = params_limit
+      freeze
     end
 
     # Stolen from Mongrel, with some small modifications:
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -32,7 +32,7 @@ module Rack
     # The default amount of nesting to allowed by hash parameters.
     # This helps prevent a rogue client from triggering a possible stack overflow
     # when parsing parameters.
-    self.default_query_parser = QueryParser.make_default(32)
+    self.default_query_parser = QueryParser.make_default(32).freeze
 
     module_function
 
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -651,7 +651,7 @@ module Rack
     }
 
     # Responses with HTTP status codes that should not have an entity body
-    STATUS_WITH_NO_ENTITY_BODY = Hash[((100..199).to_a << 204 << 304).product([true])]
+    STATUS_WITH_NO_ENTITY_BODY = Hash[((100..199).to_a << 204 << 304).product([true])].freeze
 
     SYMBOL_TO_STATUS_CODE = Hash[*HTTP_STATUS_CODES.map { |code, message|
       [message.downcase.gsub(/\s|-/, '_').to_sym, code]


jeremyevans (commented):
The constant freezing would break any application that tries to modify the constants. If we want to move to frozen constants, we need to do proper deprecation, IMO. That would entail:

1. Release 3.3 (assuming we included the change in 3.3):
   1. Add new private constant name for same object as existing constant
   2. Deprecate current public constant
   3. Switch internal access to private constant
2. Release 4.0 (major release where we can drop backwards compatibility)
   1. Remove current public constant
   2. Freeze private constant

Alternative approaches for allowing the frozen constants earlier than the next major version:

1. When loading the files, do not freeze by default, only when enabled via an option
   1. Environment variable
   2. `Ractor.count > 1`
   3. Something else?
2. Keep current code the same
   1. Require `rack/ractor_support` or similar that does the freezing

Personally, I dislike the approach of introducing classes where we are using procs. If anything is calling methods other than `call` that they would expect to work on a Proc (e.g. `arity`, `parameters`, `source_location`), their code will break.  If you use the `rack/ractor_support` approach, that file could call `Ractor.shareable_proc`.

jhawthorn:
> The constant freezing would break any application that tries to modify the constants. If we want to move to frozen constants, we need to do proper deprecation, IMO. That would entail:

Are there constants here you think users are going to modify, or are we blanket rejecting any changes to constants in Rack (I would think not based on #2374)? As I wrote in the description I was careful to select those that make no sense for the user to modify and if I've missed a use case I'll happily remove it from this PR. If it's a blanket rejection I'll close this, but I think it would be good for all users to freeze constants where it makes sense.

Looking through these one-by-one:
* `Rack::Auth::AbstractRequest::AUTHORIZATION_KEYS` - This really looks internal, in fact it's inside a `private` block, which of course does not make the constant private, but the code predates Ruby's `private_constant`. Though I suppose it's possible a user could modify it.
* `Multipart::Parser::EMPTY` - There is no reasonable use case for a user to modify this.
* `REENCODE_DUMMY_ENCODINGS` - Maybe there's a case for this being modified, but the comment suggests any modifications are unsupported and any additions need careful consideration
* `KNOWN_HEADERS` - Afaict, this is just an optimization. Users have no reason to modify it as there's a fallback that does the same
* `Rack::Lint` - This is `Rack::Lint`, users should not modify it
* `COMMON_SEP` - Afaict, again this is just an optimization with a fallback, so users have no reason to modify it.
* `DEFAULT_PORTS` - I could see this being modified (for wss:// maybe?)
* `STATUS_WITH_NO_ENTITY_BODY` - This is a pretty exhaustive list that should be defined rigidly by the HTTP spec, so I have a hard time imagining a use case to modify it
* `HTTP_STATUS_CODES`/`SYMBOL_TO_STATUS_CODE` - I could see these being modified, but both have to be modified together to make sense. I generally take a constant being read at boot time as a strong signal that it isn't meant to be mutated.


> Personally, I dislike the approach of introducing classes where we are using procs. If anything is calling methods other than `call` that they would expect to work on a Proc (e.g. `arity`, `parameters`, `source_location`), their code will break. If you use the `rack/ractor_support` approach, that file could call `Ractor.shareable_proc`.

Are you imagining users calling these proc methods or Rack calling them internally in the future? Take `TEMPFILE_FACTORY`, for example. That's a constant accessed by multipart only when a different tempfile factory wasn't specified by the env, and a user should never look at it (it maybe could have been a private_constant).

In Rack's spec we tend to be very explicit about which method names we expect given objects to respond to. I'd think we'd want to apply the same principles to where we allow configuration of the library. [In fact it seems like we did exactly that with the tempfile factory](https://github.com/rack/rack/blob/main/lib/rack/lint.rb#L406-L424).

That said, I'm happy to change to a `shareable_proc` approach if that's what's necessary.

> * When loading the files, do not freeze by default, only when enabled via an option
>   
>   1. Environment variable
>   2. `Ractor.count > 1`
>   3. Something else?
> 
> * Keep current code the same
>   
>   1. Require `rack/ractor_support` or similar that does the freezing

We will need an approach something like that for the constants I mentioned that users are likely to modify and this PR did not change (FORM_DATA_MEDIA_TYPES, PARSEABLE_DATA_MEDIA_TYPES, MIME_TYPES, Rack::Request.forwarded_priority, Rack::Request.x_forwarded_proto_priority, `Rack::Mime::MIME_TYPES`). I'd like something adjacent to the freeze_app options you introduced in #1149. I'd hoped to do this as a first pass to simplify that work and bring thread safety benefits.

jeremyevans:
> > The constant freezing would break any application that tries to modify the constants. If we want to move to frozen constants, we need to do proper deprecation, IMO. That would entail:
> 
> Are there constants here you think users are going to modify, or are we blanket rejecting any changes to constants in Rack (I would think not based on #2374)? As I wrote in the description I was careful to select those that make no sense for the user to modify and if I've missed a use case I'll happily remove it from this PR. If it's a blanket rejection I'll close this, but I think it would be good for all users to freeze constants where it makes sense.

This isn't a blanket rejection, and I'm only one committer, so I don't have the authority to reject PRs by myself.

#2374 was a change to freeze regexps, and it's much less likely that a user would attempt to modify a regexp constant than an array/hash constant.

> Looking through these one-by-one:

I appreciate you providing a justification for each change here.

> * `Rack::Auth::AbstractRequest::AUTHORIZATION_KEYS` - This really looks internal, in fact it's inside a `private` block, which of course does not make the constant private, but the code predates Ruby's `private_constant`. Though I suppose it's possible a user could modify it.

It's possible a user would want to modify it to support a different header. FWIW, the existing `'X-HTTP_AUTHORIZATION'` and `'X_HTTP_AUTHORIZATION'` entries in the array makes no sense, as those aren't valid env keys for an HTTP header. The valid key for an `X-Authorization` header would be `HTTP_X_AUTHORIZATION`. Still, I agree this it a bit of a longshot. 

> * `Multipart::Parser::EMPTY` - There is no reasonable use case for a user to modify this.

Agreed.

> * `REENCODE_DUMMY_ENCODINGS` - Maybe there's a case for this being modified, but the comment suggests any modifications are unsupported and any additions need careful consideration

This is a case where I think a user may want to modify it. We were deliberately conservative and only did this for one encoding, but I think this is deliberately left mutable for a user to support their own. If we didn't want the user modifying it, we would have hard coded the default encoding and not used a hash. This was added fairly recently.

> * `KNOWN_HEADERS` - Afaict, this is just an optimization. Users have no reason to modify it as there's a fallback that does the same

The fallback is slower. The reason to modify it is so that users can optimize their custom headers in the same manner.

> * `Rack::Lint` - This is `Rack::Lint`, users should not modify it

Agreed.

> * `COMMON_SEP` - Afaict, again this is just an optimization with a fallback, so users have no reason to modify it.

Again, users could potentially want to modify it to optimize use of custom separators.

> * `DEFAULT_PORTS` - I could see this being modified (for wss:// maybe?)

Agreed.

> * `STATUS_WITH_NO_ENTITY_BODY` - This is a pretty exhaustive list that should be defined rigidly by the HTTP spec, so I have a hard time imagining a use case to modify it

Seems reasonable, though I imagine there may be cases where users are dealing with a non-RFC compliant application and may need to delete a key. Still, that's probably remote enough that freezing by default is acceptable.

> * `HTTP_STATUS_CODES`/`SYMBOL_TO_STATUS_CODE` - I could see these being modified, but both have to be modified together to make sense. I generally take a constant being read at boot time as a strong signal that it isn't meant to be mutated.

I think both of these are deliberately mutable so that users can modify them as they see fit.

Note that starting in Ruby 4.0, it makes more sense for `REENCODE_DUMMY_ENCODINGS` and  `STATUS_WITH_NO_ENTITY_BODY` to be sets instead of hashes. Not saying we should make that change by itself, but it we going to change the constant name anyway, it may be worth a switch.

> > Personally, I dislike the approach of introducing classes where we are using procs. If anything is calling methods other than `call` that they would expect to work on a Proc (e.g. `arity`, `parameters`, `source_location`), their code will break. If you use the `rack/ractor_support` approach, that file could call `Ractor.shareable_proc`.
> 
> Are you imagining users calling these proc methods or Rack calling them internally in the future? Take `TEMPFILE_FACTORY`, for example. That's a constant accessed by multipart only when a different tempfile factory wasn't specified by the env, and a user should never look at it (it maybe could have been a private_constant).

I wouldn't expect many applications to do it. However, I wouldn't be suprised if there are some that do.

> In Rack's spec we tend to be very explicit about which method names we expect given objects to respond to. I'd think we'd want to apply the same principles to where we allow configuration of the library. [In fact it seems like we did exactly that with the tempfile factory](https://github.com/rack/rack/blob/main/lib/rack/lint.rb#L406-L424).

The rack SPEC is often type based (e.g. env must be hash, response must be array, status must be integer, response headers must be hash). However, there are certainly cases where it specifies methods to give calling code more flexibility. Note that doesn't apply to many of these changes, as they are not directly related to SPEC.

> That said, I'm happy to change to a `shareable_proc` approach if that's what's necessary.

My preference would be to keep these as-is, and only use `Ractor.shareable_proc` on an opt-in basis.

> > * When loading the files, do not freeze by default, only when enabled via an option
> >   
> >   1. Environment variable
> >   2. `Ractor.count > 1`
> >   3. Something else?
> > * Keep current code the same
> >   
> >   1. Require `rack/ractor_support` or similar that does the freezing
> 
> We will need an approach something like that for the constants I mentioned that users are likely to modify and this PR did not change (FORM_DATA_MEDIA_TYPES, PARSEABLE_DATA_MEDIA_TYPES, MIME_TYPES, Rack::Request.forwarded_priority, Rack::Request.x_forwarded_proto_priority, `Rack::Mime::MIME_TYPES`). I'd like something adjacent to the freeze_app options you introduced in #1149. I'd hoped to do this as a first pass to simplify that work and bring thread safety benefits.

Based on the above discussion, I'm OK with freezing some of the constants in this PR without deprecation. The remaining ones should probably be treated similarly to the ones you think the users are likely to modify.

However, you may want to wait for opinions from other committers. I tend to be the most conservative of the committers when it comes to potentially backwards incompatible changes.

ioquatix:
If we freeze constants, users will find out very quickly (usually at boot). So there is that.

Long term, for things we expect users to modify, I'd prefer we don't have users modifying global state and instead inject it like we are doing with `rack.request.trusted_proxy`: https://github.com/rack/rack/pull/2424

jhawthorn:
Updated per Jeremy's feedback to the set of uncontentious (or at least less contentious) changes

- Removed .freeze from KNOWN_HEADERS, COMMON_SEP, DEFAULT_PORTS, HTTP_STATUS_CODES, SYMBOL_TO_STATUS_CODE, REENCODE_DUMMY_ENCODINGS
- Reverted both proc -> class changes
- Kept .freeze on AUTHORIZATION_KEYS, STATUS_WITH_NO_ENTITY_BODY, and EMPTY
- Kept .freeze inside Rack::Lint

jeremyevans (commented):
I'm OK with this except for one change we didn't discuss previously.

jeremyevans (approved):

tenderlove (approved):
This seems good to me.

> > Require rack/ractor_support or similar that does the freezing
> We will need an approach something like that for the constants I mentioned that users are likely to modify and this PR did not change (FORM_DATA_MEDIA_TYPES, PARSEABLE_DATA_MEDIA_TYPES, MIME_TYPES, Rack::Request.forwarded_priority, Rack::Request.x_forwarded_proto_priority, Rack::Mime::MIME_TYPES). I'd like something adjacent to the freeze_app options you introduced in https://github.com/rack/rack/pull/1149. I'd hoped to do this as a first pass to simplify that work and bring thread safety benefits.

+1 here.  I think "boot app, webserver asks app to make itself Ractor compatible" is likely the direction Ractor based webserver should go.

ioquatix:
@jhawthorn do you mind adding an entry to the change log? Then, I think we are good to merge.

PR #2434, opened 2026-03-13
Consider forwarded_priority and x_forwarded_proto_priority for x-forwarded-ssl

jeremyevans:
Previously, X-Forwarded-SSL was considered even if forwarded_priority was an empty array. This adds :ssl as the first element to x_forwarded_proto_priority, so by default, it will be considered in preference to x-forwarded-proto and x-forwarded-scheme, but it will only be considered if the forwarded header is not present. This will not consider x-forwarded-ssl header if :x_forwarded is not in forwarded_priority or if :ssl is not in x_forwarded_proto_priority.

Fixes #2433

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -40,7 +40,7 @@ module Rack
     end
 
     @forwarded_priority = [:forwarded, :x_forwarded]
-    @x_forwarded_proto_priority = [:proto, :scheme]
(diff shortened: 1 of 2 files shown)


ioquatix (approved):

ioquatix:
Thanks, LGTM.

PR #2435, opened 2026-03-13
Fix the mime type for .pem

jeremyevans:
From my research, application/x-x509-ca-cert is not good as it implies a single CA certificate encoded in DER format, whereas a pem file could include a non-CA certificate, a private key, or something else.

RFC 8894 Section 6.1 states the encoding of
application/x-x509-ca-cert is binary, while pem files are ASCII encoded (https://datatracker.ietf.org/doc/html/rfc8894#section-6.1-1.10)

IANA reflects this in there official documentation https://www.iana.org/assignments/media-types/application/x-x509-ca-cert

application/x-pem-file appears to be the recommended mime type for .pem (https://pki-tutorial.readthedocs.io/en/latest/mime.html).

Linux associates .pem with both application/x-x509-ca-cert and application/x-pem-file.

Even places that include application/x-x509-ca-cert and don't include application/x-pem-file specify that
application/x-x509-ca-cert should be used for .crt and .der, not .pem:

* Apache: https://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types
* https://mimetype.io/all-types

However, both Nginx and OpenBSD associate .pem with application/x-x509-ca-cert in their default mime type list.


ioquatix (approved):
LGTM. We should probably add a note in the change log.

PR #2438, opened 2026-03-25
Add getting started with Rack guide

ayushn21:
An initial stab at writing a "Getting started with Rack" guide.

It covers the topics:

* Creating a basic rack app and running it
* Handing routes and HTTP verbs
* Reading request bodies
* Streaming response bodies
* Creating and adding middleware

The list of topics differs a bit from the issue ticket. As I was writing the guide, I felt this structure suited the narrative better. I excluded details about `Rack::Builder` because I don't think I could improve on what's already there in the RDoc. As such I've just linked to the RDoc page.

I haven't mentioned any frameworks, but I did have to write the guide using a specific server, so I chose Puma as I believe it is the most widely used Rack server. I've signposted alternatives to try and be as agnostic as possible.

This is just an initial draft so obviously open to any and all feedback and I will make any requested changes!

Closes https://github.com/rack/rack/issues/2437

### DIFF
--- a/GETTING-STARTED.md
+++ b/GETTING-STARTED.md
@@ -4,44 +4,33 @@ This guide demonstrates how to create a basic Rack application and run it using 
 
 ## Creating a Rack Application
 
-A Rack app is a class which implements a `call` method. It is passed an [`env`](./SPEC_rdoc.html#the-request-environment) hash, known as the Rack environment, which is created by parsing the incoming HTTP request.
+A Rack app is an object which implements a `call` method. It is passed an [`env`](./SPEC_rdoc.html#the-request-environment) hash, known as the Rack environment.
 
 ```ruby
-class App
-  def call(env)
-    [200, { "content-type" => "text/plain" }, ["Hello World"]]
-  end
+rack_app = lambda do |env|
+  [200, { "content-type" => "text/plain" }, ["Hello World"]]
 end
 
-run App.new
+run rack_app
 ```
 
-When an HTTP request is made, the Rack-compliant web server parses it to create the `env` hash, and forwards it to the application's `call` method. This method must return an array representing the HTTP response.
+When an HTTP request is made, the Rack-compliant web server parses it to create the `env` hash, and calls the application with `env`. The `call` method must return an array with exactly three elements, representing the HTTP response:
 
-The first element is the HTTP response code, in this case `200`. The second element is a hash containing any Rack and HTTP response headers we wish to send. And the last element is an array of strings representing the response body.
+1. The HTTP response code (`200` in the above example).
+2. A hash containing any HTTP response headers we wish to send.
+3. An enumerable object that yields strings, representing the response body.
 
-You can run this app by organizing it into a folder:
+Rack applications are generally run using the web server's command line program, with the entry point for the application being stored in a `config.ru` file:
 
 ```bash
-$ mkdir rack-demo
-$ cd rack-demo
-$ bundle init
-$ bundle add rack puma
 $ cat > config.ru << APP
-class App
-  def call(env)
-    [200, { "content-type" => "text/plain" }, ["Hello World"]]
-  end
+rack_app = lambda do |env|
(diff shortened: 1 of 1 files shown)


jeremyevans (commented):
Thank you for your submission. I have a lot of suggested changes, but in general I think the guide is useful and I am in favor of merging it after the changes.

GETTING-STARTED.md
```diff
@@ -0,0 +1,282 @@
...
+    else
+      [200, { "content-type" => "text/plain" }, ["Responding to all other verbs"]]
+    end
+  end
+end
+
+run App.new
+```
+
+Rack provides a [`Rack::Request`](./Rack/Request.html) utility class which implements a convenient interface to the Rack environment. It is stateless, meaning the `env` passed to the constructor will be directly accessed and modified.
```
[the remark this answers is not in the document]
ayushn21: Ahh right thanks, I'd pretty much lifted that from the RDoc.

ayushn21 (commented):
Thanks for the thorough feedback @jeremyevans, I learned a lot there ... especially with how my slightly clumsy wording could result in technical inaccuracies. Something to keep in mind for the future!

I think I've addressed all your feedback. Since we're doing all the demos using a `lamda`, I've added a new snippet at the start of the guide demonstrating a Rack app using a `class`. While it might be overkill since we say a Rack app can be any object, I think it's important to explicitly demonstrate to beginners that a `class` may also be used.

Even after 10+ years as a professional programmer, my brain confuses "class" and "object" from time to time. As such, I just wanted to be completely clear that a `class` can be used.

Let me know if you'd like any further changes!

GETTING-STARTED.md
```diff
@@ -0,0 +1,253 @@
...
+
+```ruby
+app = lambda do |env|
+  body = \
+    case env["PATH_INFO"]
+    when "/admin"
+      ["Hello Admin"]
+    else
+      ["Hello World"]
+    end
```
jeremyevans: I would reformat this, but it's up to you. At least the `\` can be removed.

```suggestion
  body = case env["PATH_INFO"]
  when "/admin"
    ["Hello Admin"]
  else
    ["Hello World"]
  end
```

jeremyevans (commented):
Most issues fixed, thank you! I only have a few minor suggestions.

ayushn21:
I think I've addressed everything @jeremyevans, let me know if there's anything else!

I can squash all the commits into one if you wish? Might be cleaner that way but leaving as is for now to make it easier to review.

Also, the links to other RDoc pages in this guide use relative paths, so they won't work if someone's reading the Markdown on GitHub itself, rather than on the docs site. In other guides such as the [Upgrade Guide](https://rack.github.io/rack/main/UPGRADE-GUIDE_md.html), the links are not defined in the Markdown, and RDoc handles the linking automatically at build time. 

I'll remove all explicit RDoc links from this guide to follow that pattern. When linking to another document like the `Spec` or `Readme`, I'll link to GitHub instead of the docs. Let me know if you prefer another method!

GETTING-STARTED.md
```diff
@@ -0,0 +1,260 @@
+# Getting Started With Rack
+
+This guide demonstrates how to create a basic Rack application and run it using a Rack-compliant web server. We'll use [Puma](https://puma.io) in this tutorial — alternative web servers can be found in the [Readme](https://github.com/rack/rack?tab=readme-ov-file#supported-web-servers).
```
jeremyevans: We should probably avoid em-dashes, as it makes it look like this was written by an LLM.

```suggestion
This guide demonstrates how to create a basic Rack application and run it using a Rack-compliant web server. We'll use [Puma](https://puma.io) in this tutorial (alternative web servers can be found in the [Readme](https://github.com/rack/rack?tab=readme-ov-file#supported-web-servers)).
```

jeremyevans (approved):

ioquatix:
Thanks for your efforts.

PR #2444, opened 2026-04-01
Clarify timing of patch releases and advisories

nightpool:
This aligns the rack security policy with other modern security policies, including Ruby's, Rails', and RubyGems', and the model security guidelines from the [Open Source Security Foundation](https://github.com/ossf/oss-vulnerability-guide/blob/main/maintainer-guide.md#publish-your-vulnerability-management-process:~:text=9%20Cut%20a%20release%20and%20publicly%20disclose%20the%20issue)

https://guides.rubygems.org/security/#reporting-security-vulnerabilities:~:text=Reporting%20a%20security%20vulnerability%20with%20your%20own%20gem

Thank you for your hard work in handling Rack's security vulnerabilities and reports! Unlike the Feb 19th release, my team was not notified of any issues with Rack through our normal process, we had to rely on word of mouth from someone who happened to be in the invite-only Slack channel[1] where the release was announced. I suspect very few other people using Rack are in this slack channel (the security announcement only had 6 users reacting to it).

Due to these vulnerabilities not being published publicly, advisory databases that defenders rely on to notify them of security issues, like https://github.com/rubysec/ruby-advisory-db, https://security.snyk.io/package/rubygems/rack and https://github.com/dependabot were not able to pick up this security advisory and notify team members. We got lucky by having an employee active in the development community for Bundler, but most teams do not have that luxury.

[1] At least, I was not able to find a public invite for it anywhere. It seems like the invite for the community has been removed from the Bundler website as of February

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -17,6 +17,6 @@ If you believe you have found a security issue in this project, please do not op
 1. Security report is received and reviewed by the Rack maintainers.
 2. The problem is confirmed and a list of all affected versions is determined. Code is audited to find any potential similar problems.
 3. Fixes are prepared for all releases which are still supported.
-4. Patches are released, new gem versions are published to RubyGems, and security advisories are published.
+4. Patches are released, new gem versions are published to RubyGems, and security advisories are published simultaneously.
 
 In cases where coordination with other projects or distributions is necessary, we may implement an embargo period before public disclosure. However, for most security issues, we aim to release fixes and advisories as quickly as possible.


nightpool:
> 9 Cut a release and publicly disclose the issue
> 
> On the day of public disclosure, publish your disclosure announcement ([see templates](https://github.com/ossf/oss-vulnerability-guide/blob/main/maintainer-guide.md#communication-templates)). If using GitHub Security Advisories, "publishing" your private Security Advisory will add it to the "Security" tab. If you are not using GitHub Security Advisories, publish the announcement to your release notes or security bulletins. If you have CVE ids for the vulnerabilities fixed, include the CVE id(s) of the vulnerabilities that were fixed in this release.
> 
> It's also recommended to send the announcement to appropriate mailing lists for your community (i.e., a security-announce@ list and even a general mailing list for high-impact vulnerabilities).
> 
> 10 Disclose quickly
> 
> You might choose to briefly withhold details about how the vulnerability can be exploited, hoping that this will give users a little more time to update before attackers begin exploiting the vulnerability. This only makes sense if it's not obvious to attackers how the vulnerability can be exploited, and in most cases, attackers will find it obvious. In addition, attackers can usually review changes made to software (in source or executable form) and easily determine an attack. Thus, withholding detailed information can only be helpful for a few days at most, even in the few cases where it helps at all.

In this case, the exploits were easily reconstructable from the patches available. Once I was aware of the issue, I was able to have develop a testing tool for our website to see if any of the issues were remediated by our WAF in only a few minutes per vulnerability. Especially since the commit messages gave detailed information about the vulnerabilities, the delay only served to prevent notifications for defenders.

jeremyevans:
Requiring simultaneous publishing for security advisories can result in delaying the publication of fixes, especially in the case where a release fixes many security issues (as the recent releases did). Security advisory publishing and gem publishing actually being simultaneous is impossible, those are two separate processes. If you want to make an accurate and meaningful change here, you would need to list a specific maximum amount of time between the separate parts.

I'm not against having a specific maximum time defined, but I'm also not the one that is responsible for publishing the security advisories and handling the releases.  I'm fine leaving the decision to the person actually doing the work. For disclosure, @ioquatix and I discussed this, and I agreed we could delay publishing the security advisories in order to get the releases out.

nightpool:
That's totally understandable, but I think that having even a very brief/placeholder security advisory and then updating it later with more details is much preferable to having a no security advisory at all. Right now all of these security advisories exist in one state or another, but they're all unpublished. 

Even just the amount of information that is currently in the changelog (for example: `- [CVE-2026-34827](https://github.com/advisories/GHSA-v6x5-cg8r-vv6x) Quadratic-time multipart header parsing allows denial of service via escape-heavy quoted parameters.`) would be very very helpful to have.

nightpool:
> Security advisory publishing and gem publishing actually being simultaneous is impossible, those are two separate processes. If you want to make an accurate and meaningful change here, you would need to list a specific maximum amount of time between the separate parts.

I don't think anyone is saying that they need to be exactly simultaneous down to the second. But they should be done around the same time. What my change is intended to convey is that _intentionally_ delaying the security advisory publishing to "give users a window to upgrade" (as seen in the quote below) is a policy that I feel like disservices me as a defender and only serves to help attackers.

This was the quote from @ioquatix on Slack that motivated the change:

> Our view is that the most important action is updating. Publishing detailed advisories immediately doesn’t necessarily help defenders act faster, but it does make exploitation easier once the issues are widely indexed.
Releasing fixes first gives users a window to upgrade before full details are broadly distributed.

Defenders don't benefit from releases that they can't know about. I'm not picky on the *content* of the announcement, but my experience as a defender tells me that some form of public announcement is necessary for this process to work

ioquatix:
I'm still of the opinion that the release with the change log is sufficient public disclosure for someone to know they need to update. I understand the CVE system helps spread the word further, but we also may have internal (non-disclosed) obligations.

@tenderlove you did work on security in the past, do you have any opinion about this change?

ioquatix:
I'm still of the opinion that the release with the change log is sufficient public disclosure for someone to know they need to update. I understand the CVE system helps spread the word further, but we also may have internal (non-disclosed) obligations.

The note to various Slacks about the releases is an early heads up, since I know there are people there who run critical systems.

I will release the security advisories tonight which is within 24 hours of the original releases.

@tenderlove you did work on security in the past, do you have any opinion about this change?

tenderlove:
I think they should be released simultaneously (or as close as is possible).  Typically for the Rails project, we'll (after patches are created etc):

1. Prepare CVE reports on GitHub
2. Request CVEs
3. Rewrite commit messages to include CVE (not everyone does this, but I do)
4. Fix change logs to include them
5. Ship gems
6. Click the publish button on GitHub for the security report

The time between step 5 and 6 is usually a few minutes (maybe an hour if we mess something up or forget).  Is this accurate @jhawthorn?

@nightpool I think this process mostly satisfies what you're looking for? Is that accurate?

ioquatix:
After considering the feedback, I've changed my opinion, and think we should update the process. I deleted my previous comments which don't add to the discussion.

I have published all the advisories now. Thanks for raising this issue.

jhawthorn:
> The time between step 5 and 6 is usually a few minutes (maybe an hour if we mess something up or forget). Is this accurate @jhawthorn?

That's about right, I used to be very concerned about everything being as immediate as possible, but now think it's best to be more relaxed. I think within an hour of the releases is a very reasonable target.

Rails did switch to using the GHSA system a while ago (which it looks like is in use here 👍). It's a good system (in particular because of the exactness of the version numbers, though they are super annoying to input). That data when published will directly feed dependabot, the push alerts, and some other downstream consumers. After publishing the GHSA (which publishes the CVE as well) I run an automated script to mirror the announcement to our forum.

OP also mentions ruby-advisory-db, I can't speak for the project but I _believe_ it's hand edited (or at least the previous Rails release I made ended up hand-edited incorrect versions there 😅), so _I believe_ you are relying on the timing of a 3rd party volunteer there.

samuel-williams-shopify:
> though they are super annoying to input

I had to do this ~3 x 13 advisories :p

nightpool:
I think ruby-advisory-db has a ghsa import script but I don't know the
cadence.

Thank you for accepting the PR! Yes, to @tenderlove's point, I definitely
based it on the Rails security policy (although simpler because there's no
distro embargo period)

PR #2454, opened 2026-04-27
Fix "+" Encoding/Decoding in the cookie value

andreavocado:
See https://github.com/rack/rack-session/issues/60#issuecomment-4316776527

### DIFF
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -334,7 +334,7 @@ module Rack
         # RFC 6265: cookie values are opaque strings. Only decode percent-encoded
         # sequences (%XX) — do NOT use decode_www_form_component which would
         # silently convert + to space (form-encoding semantics have no place here).
-        cookies[key] = (value.gsub(/%([0-9A-Fa-f]{2})/) { $1.to_i(16).chr(Encoding::UTF_8) } rescue value) unless cookies.key?(key)
+        cookies[key] = value.gsub(/%([0-9A-Fa-f]{2})/) { [$1.to_i(16)].pack('C') }.force_encoding(Encoding::UTF_8) unless cookies.key?(key)
       end
     end
 
@@ -408,7 +408,9 @@ module Rack
 
       value = [value] unless Array === value
 
-      return "#{key}=#{value.map { |v| escape v }.join('&')}#{domain}" \
+      # Use percent-encoding with %20 for spaces (not +) since cookie values
+      # are not form-encoded. This ensures round-tripping with parse_cookies_header.
+      return "#{key}=#{value.map { |v| escape(v).gsub('+', '%20') }.join('&')}#{domain}" \
         "#{path}#{max_age}#{expires}#{secure}#{httponly}#{same_site}#{partitioned}"
     end
 
--- a/lib/rack/utils.rb
+++ b/lib/rack/utils.rb
@@ -331,7 +331,10 @@ module Rack
       value.split(/; */n).each_with_object({}) do |cookie, cookies|
         next if cookie.empty?
         key, value = cookie.split('=', 2)
-        cookies[key] = (unescape(value) rescue value) unless cookies.key?(key)
+        # RFC 6265: cookie values are opaque strings. Only decode percent-encoded
+        # sequences (%XX) — do NOT use decode_www_form_component which would
+        # silently convert + to space (form-encoding semantics have no place here).
+        cookies[key] = (value.gsub(/%([0-9A-Fa-f]{2})/) { $1.to_i(16).chr(Encoding::UTF_8) } rescue value) unless cookies.key?(key)
       end
     end
 
--- a/test/spec_utils.rb
+++ b/test/spec_utils.rb
@@ -647,12 +647,45 @@ describe Rack::Utils, "cookies" do
     # Percent-encoded sequences should still be decoded.
     env = Rack::MockRequest.env_for("", "HTTP_COOKIE" => "foo=hello%20world")
     Rack::Utils.parse_cookies(env).must_equal({ "foo" => "hello world" })
+
+    # High-byte percent sequences produce raw bytes force-encoded as UTF-8.
+    env = Rack::MockRequest.env_for("", "HTTP_COOKIE" => "foo=%FF")
+    Rack::Utils.parse_cookies(env)["foo"].bytes.must_equal [0xFF]
+    Rack::Utils.parse_cookies(env)["foo"].encoding.must_equal Encoding::UTF_8
   end
 
   it "generates appropriate cookie header value" do
     Rack::Utils.set_cookie_header('name', 'value').must_equal 'name=value'
     Rack::Utils.set_cookie_header('name', %w[value]).must_equal 'name=value'
     Rack::Utils.set_cookie_header('name', %w[va ue]).must_equal 'name=va&ue'
+    # Spaces are encoded as %20, not + (RFC 6265: cookie values are not form-encoded)
+    Rack::Utils.set_cookie_header('name', 'a b').must_equal 'name=a%20b'
+    # + is percent-encoded to preserve literal +
+    Rack::Utils.set_cookie_header('name', 'a+b').must_equal 'name=a%2Bb'
+  end
+
+  it "round-trips cookie values through set_cookie_header and parse_cookies" do
+    test_values = [
+      "simple",
+      "hello world",
+      "a+b",
+      "a+b/c=d",
+      "abc+def/ghi==",     # base64-like
+      "café",              # multibyte UTF-8
+      "a&b",
+      "100%",
+      "key=val",
+      "semi;colon",
+      "a b+c/d=e&f",
+    ]
+
+    test_values.each do |original|
+      header = Rack::Utils.set_cookie_header("test", original)
+      cookie_str = header.split(";", 2).first  # strip attributes
+      env = Rack::MockRequest.env_for("", "HTTP_COOKIE" => cookie_str)
+      parsed = Rack::Utils.parse_cookies(env)["test"]
+      parsed.must_equal original, "Round-trip failed for #{original.inspect}: got #{parsed.inspect}"
+    end
   end
 
   it "sets and deletes cookies in header hash" do
--- a/test/spec_utils.rb
+++ b/test/spec_utils.rb
@@ -636,6 +636,17 @@ describe Rack::Utils, "cookies" do
     env = Rack::MockRequest.env_for("", "HTTP_COOKIE" => "%66oo=baz;foo=bar")
     cookies = Rack::Utils.parse_cookies(env)
     cookies.must_equal({ "%66oo" => "baz", "foo" => "bar" })
+
+    # RFC 6265: + in a cookie value is a literal +, not a space.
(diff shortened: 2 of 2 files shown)


jcmfernandes:
Chiming in as I came across https://github.com/rack/rack-session/pull/61.

My understanding is that this bug is real, that is, we aren't compliant with RFC 6265. However, everything works as expected as long as the cookies are **written and read** by rack as it calls `URI.encode_www_form_component` on the value to write. That is, round-trips work as expected, something that can be confirmed by adding

```ruby
  it "round-trips cookie values containing '+' between Response and Request" do
    res = Rack::Response.new
    res.set_cookie "foo", "a+b"
    req = make_request \
      Rack::MockRequest.env_for("", "HTTP_COOKIE" => res["set-cookie"])
    req.cookies.must_equal "foo" => "a+b"
  end
```

to `test/spec_request.rb`. This test passes because `res.set_cookie "foo", "a+b"` actually writes `foo=a%2Bb`. You can also just open your browser inspector, refresh this exact page, and see the `%2B` in `set-cookie _gh_sess=...` as GitHub is known for being rack-powered.

So, I suggest reverting https://github.com/rack/rack-session/pull/61 as there's no need to use URL-safe base64 encoding. Ruby's URL-safe base64 implementation is (approximately) a `String#tr` on the return from `Base64.strict_encode64` which is a penalty we don't need to pay. Furthermore, https://github.com/rack/rack-session/pull/61 will break most existing V2-encrypted sessions.

I have to admit that my heart kind of stopped for a second as I was going over https://github.com/rack/rack-session/pull/61 :sweat_smile: how could the V2 encryptor be broken for this long without any other reports? How could rack be broken like this for years? Round-trips work, and that's the explanation.

EDIT:
Stating somewhat the obvious, but this fix, as is, will corrupt all the cookies in the wild that have a URL-encoded `+` symbol. rack-sessions will undoubtedly suffer.

lib/rack/utils.rb
```diff
@@ -331,7 +331,10 @@ def parse_cookies_header(value)
       value.split(/; */n).each_with_object({}) do |cookie, cookies|
         next if cookie.empty?
         key, value = cookie.split('=', 2)
-        cookies[key] = (unescape(value) rescue value) unless cookies.key?(key)
+        # RFC 6265: cookie values are opaque strings. Only decode percent-encoded
+        # sequences (%XX) — do NOT use decode_www_form_component which would
+        # silently convert + to space (form-encoding semantics have no place here).
+        cookies[key] = (value.gsub(/%([0-9A-Fa-f]{2})/) { $1.to_i(16).chr(Encoding::UTF_8) } rescue value) unless cookies.key?(key)
```
jeremyevans: I don't believe the  `Encoding::UTF_8` behavior is correct here. You turn `%FF` into `"\u00FF"` (`"\xC3\xBF"` in binary). Is that really what we want? If you were going to encode UTF-8 data into a percent encoded cookie value, you would encode each byte in the UTF-8 string that needed to be percent encoded. You wouldn't encode each character as a single percent encoded value, because that would not allow you to encode unicode characters above codepoint 255.  This approach is also backwards incompatible, as `URI.decode_www_form_component("%FF")` is `"\xFF".force_encoding(Encoding::UTF_8)`. A better approach would be to force encode the entire string after decoding to `Encoding::UTF_8`, even if it results in an invalid encoding, as that preserves backwards compatibility.

The `rescue` seems unnecessary. Do you have evidence it can be hit? `(0..255).each{it.to_s(16).to_i(16).chr(Encoding::UTF_8)}` does not raise.

jeremyevans (commented):
Thanks for submitting this. I agree we should remove the `+` decoding to ` `.

However, I think this is incomplete. While this prevents decoding `+` to ` `, it does nothing to stop ` ` from being encoded as `+`. So this appears to break round tripping. I think you need to at least change `set_cookie_header` to encode ` ` as `%20` instead of `+`.

We probably need to add cookie round trip specs, testing that `parse_cookies` correctly parses `set_cookie_header` output for a wide range of characters.

This will break cookies that use `+` during a Rack upgrade. So we'll have to decide whether to put this in Rack 3.3, or whether it will have to wait for Rack 4.

andreavocado:
> Chiming in as I came across [rack/rack-session#61](https://github.com/rack/rack-session/pull/61).
> 
> My understanding is that this bug is real, that is, we aren't compliant with RFC 6265. However, everything works as expected as long as the cookies are **written and read** by rack as it calls `URI.encode_www_form_component` on the value to write. That is, round-trips work as expected, something that can be confirmed by adding
> 
> ```ruby
>   it "round-trips cookie values containing '+' between Response and Request" do
>     res = Rack::Response.new
>     res.set_cookie "foo", "a+b"
>     req = make_request \
>       Rack::MockRequest.env_for("", "HTTP_COOKIE" => res["set-cookie"])
>     req.cookies.must_equal "foo" => "a+b"
>   end
> ```
> 
> to `test/spec_request.rb`. This test passes because `res.set_cookie "foo", "a+b"` actually writes `foo=a%2Bb`. You can also just open your browser inspector, refresh this exact page, and see the `%2B` in `set-cookie _gh_sess=...` as GitHub is known for being rack-powered.
> 
> So, I suggest reverting [rack/rack-session#61](https://github.com/rack/rack-session/pull/61) as there's no need to use URL-safe base64 encoding. Ruby's URL-safe base64 implementation is (approximately) a `String#tr` on the return from `Base64.strict_encode64` which is a penalty we don't need to pay. Furthermore, [rack/rack-session#61](https://github.com/rack/rack-session/pull/61) will break most existing V2-encrypted sessions.
> 
> I have to admit that my heart kind of stopped for a second as I was going over [rack/rack-session#61](https://github.com/rack/rack-session/pull/61) 😅 how could the V2 encryptor be broken for this long without any other reports? How could rack be broken like this for years? Round-trips work, and that's the explanation.
> 
> EDIT: Stating somewhat the obvious, but this fix, as is, will corrupt all the cookies in the wild that have a URL-encoded `+` symbol. rack-sessions will undoubtedly suffer.

You're right about:

 - Round-trips within Rack work because set_cookie encodes + as %2B via encode_www_form_component, so read-back is correct.
 - rack-session#61's switch to URL-safe base64 is indeed unnecessary if cookies are always written and read by Rack.

On the other hand:

 - The bug matters for interoperability. If a non-Rack client (JavaScript's document.cookie, another framework, a CDN edge worker) sets a cookie with a literal +, Rack will silently corrupt it to a space. RFC 6265 treats cookie values as opaque — + is literal, not form-encoded.
 - Existing Rack-set cookies use %2B for +, which the new code decodes identically. The only change is that literal + (from non-Rack writers) is now correctly preserved instead of being mangled.

lib/rack/utils.rb
andreavocado: Thanks for your detailed review! Based on your feedback I adjusted the code. Please have a look.

jcmfernandes:
> > Chiming in as I came across [rack/rack-session#61](https://github.com/rack/rack-session/pull/61).
> > 
> > My understanding is that this bug is real, that is, we aren't compliant with RFC 6265. However, everything works as expected as long as the cookies are **written and read** by rack as it calls `URI.encode_www_form_component` on the value to write. That is, round-trips work as expected, something that can be confirmed by adding
> > 
> > ```ruby
> >   it "round-trips cookie values containing '+' between Response and Request" do
> >     res = Rack::Response.new
> >     res.set_cookie "foo", "a+b"
> >     req = make_request \
> >       Rack::MockRequest.env_for("", "HTTP_COOKIE" => res["set-cookie"])
> >     req.cookies.must_equal "foo" => "a+b"
> >   end
> > ```
> > 
> > to `test/spec_request.rb`. This test passes because `res.set_cookie "foo", "a+b"` actually writes `foo=a%2Bb`. You can also just open your browser inspector, refresh this exact page, and see the `%2B` in `set-cookie _gh_sess=...` as GitHub is known for being rack-powered.
> > 
> > So, I suggest reverting [rack/rack-session#61](https://github.com/rack/rack-session/pull/61) as there's no need to use URL-safe base64 encoding. Ruby's URL-safe base64 implementation is (approximately) a `String#tr` on the return from `Base64.strict_encode64` which is a penalty we don't need to pay. Furthermore, [rack/rack-session#61](https://github.com/rack/rack-session/pull/61) will break most existing V2-encrypted sessions.
> > 
> > I have to admit that my heart kind of stopped for a second as I was going over [rack/rack-session#61](https://github.com/rack/rack-session/pull/61) 😅 how could the V2 encryptor be broken for this long without any other reports? How could rack be broken like this for years? Round-trips work, and that's the explanation.
> > 
> > EDIT: Stating somewhat the obvious, but this fix, as is, will corrupt all the cookies in the wild that have a URL-encoded `+` symbol. rack-sessions will undoubtedly suffer.
> 
> You're right about:
> 
>  - Round-trips within Rack work because set_cookie encodes + as %2B via encode_www_form_component, so read-back is correct.
>  - rack-session#61's switch to URL-safe base64 is indeed unnecessary if cookies are always written and read by Rack.
> 
> On the other hand:
> 
>  - The bug matters for interoperability. If a non-Rack client (JavaScript's document.cookie, another framework, a CDN edge worker) sets a cookie with a literal +, Rack will silently corrupt it to a space. RFC 6265 treats cookie values as opaque — + is literal, not form-encoded.
>  - Existing Rack-set cookies use %2B for +, which the new code decodes identically. The only change is that literal + (from non-Rack writers) is now correctly preserved instead of being mangled.

As stated before, I agree that this must be fixed, but:

1. There's objectively no bug in rack-session, so we should revert your PR as, furthermore, it will break most existing sessions. 
2. Fixing this without introducing a migration path will break **all** cookies out in the wild that have a + char, and that's a huge set of established cookies. This would cause, for example, mass logouts from sites like GitHub. 

EDIT:

I replied from my phone and I clearly missed some of what you wrote. Apologies for that.

My admittedly shallow understanding of RFC 6265 has me convinced that we must interpret cookie values as opaque octet sequences to be compliant. If I'm right, then continuing to decode `%2B` (or any other octet sequence) keeps rack non-compliant.

Now, if we want to become 6265-compliant (IMHO, we should), then we must make this a configuration/option that if disabled by default for now - that is, we continue form encoding/decoding cookie values. This is something that could be revisited for the next major version since it's a breaking change. We'll then need to discuss how to handle invalid cookie values in both requests and responses. I suggest taking a look at how golang's stdlib behaves: it ignores invalid cookie values in requests and removes invalid octets from responses,logging a warning noting that it did it.

jcmfernandes:
I took a stab at this: https://github.com/jcmfernandes/rack/tree/rfc6265-compliant-cookies

I don't intend to step on your work @andreavocado so please check it out and let me know what you think. I took inspiration from golang's approach to the problem at hand.

andreavocado:
> I don't intend to step on your work @andreavocado so please check it out and let me know what you think. I took inspiration from golang's approach to the problem at hand.

Great! I appreciate your effort to fix this issue and I'm happy you take over this 👍 So I'd close this PR and continue with your PR...

The code looks good to me. About the migration plan: Maybe we should extend this with some examples or more detailed instructions: Common questions may be:
1. What does auditing mean? Do I have to change percent-encoded octets (%2B) before writing the cookie? Do I have to encode/decode on my own?
2. How to rotate cookies?

PR #2467, opened 2026-05-14
Improve builder docs

at9596:


lib/rack/builder.rb
```diff
@@ -269,7 +269,7 @@ def freeze_app
     # Return the Rack application generated by this instance.
     def to_app
       app = @map ? generate_map(@run, @map) : @run
-      fail "missing run or map statement" unless app
+      fail "missing run or map statement in config.ru" unless app
```
jeremyevans: It's possible to use Builder with a different filename, or with no filename at all. So I don't think we can accept this as we cannot be sure it is correct. I don't think we can make it correct currently, as we don't pass down the path used into the Builder initialization. It doesn't seem worth it to me to make the changes that would be required to do this correctly.

jeremyevans (commented):
Thank you for submitting a pull request.

As a general note, it's best to not send unrelated changes in the same pull request. Personally, I don't think the churn for Rubocop is worth it, but other committers may feel differently.

lib/rack/builder.rb
at9596: Thanks for the clarification! That make sense. I mistakenly assumed that ``` Builder class ``` can not be used outside the ``` config.ru ```. I will revert my changes and keep the PR focused.

PR #2468, opened 2026-05-20
Rack::RewindableInput: add `seek` method

byroot:
Since it support `rewind` I don't see why it shouldn't support `seek`. It has come up in `marcel`: https://github.com/rails/marcel/pull/144


jeremyevans (approved):

ioquatix (approved):
LGTM - can you add release notes.

PR #2469, opened 2026-05-28
Make Rack::Files ractor safe.

andrewn617:
I am exploring getting a Rails app running in a Ractor and ran into an issue with `Rack::Files`.

The lambda we store in `@head` needs to be shareable. But it needs to reference the `Rack::Files` instance as `self`, so we can't make it shareable with `Ractor.lamda_shareable` since we haven't finished initializing self yet.

So instead I made a small callbale object which can be frozen later when the application is made shareable.

### DIFF
--- a/lib/rack/files.rb
+++ b/lib/rack/files.rb
@@ -29,11 +29,21 @@ module Rack
 
     attr_reader :root
 
+    class Callable
+      def initialize(files)
+        @files = files
+      end
+
+      def call(env)
+        @files.get env
+      end
+    end
+
     def initialize(root, headers = {}, default_mime = 'text/plain')
       @root = (::File.expand_path(root) if root)
       @headers = headers
       @default_mime = default_mime
-      @head = Rack::Head.new(lambda { |env| get env })
+      @head = Rack::Head.new(Callable.new(self))
     end
 
     def call(env)
--- a/test/spec_files.rb
+++ b/test/spec_files.rb
@@ -382,4 +382,14 @@ content-range: bytes 60-80/209\r
     res.must_be :not_found?
     res.body.must_be :empty?
   end
+
+  if defined?(Ractor)
+    it "can be made ractor shareable" do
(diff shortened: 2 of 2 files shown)


jeremyevans (commented):
I'm against making changes that make the code more complicated and less understandable just to allow for ractor support. Other committers may feel differently, though.

tenderlove:
> I'm against making changes that make the code more complicated and less understandable just to allow for ractor support. Other committers may feel differently, though.

Rack is supposed to provide a set of APIs that have no opinion about web server implementation. This includes the concurrency mechanisms any particular web server chooses.  IOW I would not reject PRs for thread/fiber/process safety issues, so I expect the same for Ractors too.

tenderlove (approved):

ioquatix:
Are we not just working around issues in Ractor with less than ideal code?

tenderlove:
> Are we not just working around issues in Ractor

I don't think this is any more of a "work around" than adding a mutex or concurrent data structures to ensure thread/fiber safety.

>  with less than ideal code?

Honestly I don't think this is "less than ideal" either. It introduces a class that can actually be tested (if we wanted).

jeremyevans:
> > Are we not just working around issues in Ractor
> 
> I don't think this is any more of a "work around" than adding a mutex or concurrent data structures to ensure thread/fiber safety.

Adding a mutex or concurrent data structure is generally necessary for thread safety on any Ruby implementation of threads. This change is only necessary due to ractor limitations on CRuby. What benefit does this change have on JRuby or TruffleRuby? They can already do parallel execution without this change.
 
> > with less than ideal code?
> 
> Honestly I don't think this is "less than ideal" either. It introduces a class that can actually be tested (if we wanted).

I disagree. Without ractors there is no way we would accept this, it duplicates the behavior of a trivial lambda and is unnecessary complexity.

FWIW, I'm not strongly against this particular change (IMO it makes the code worse, but only slightly), but it leads to a slippery slope.

PR #2471, opened 2026-06-02
Support streaming bodies in Rack::Deflater

### LINKED ISSUE #2470: Rack::Deflater crashes (NoMethodError: each) on Rack 3 streaming/callable bodies
## Summary

`Rack::Deflater` raises `NoMethodError: undefined method 'each'` when it wraps a
Rack 3 streaming body, i.e. a body that responds to `#call(stream)` and not to
`#each`. Deflater wraps the body in `GzipStream` and iterates it with
`@body.each`, but a streaming body only implements `#call`. The client receives a
broken, truncated gzip stream (just the gzip header/trailer that `Zlib::GzipWriter`
emits before the exception unwinds) and the server logs the `NoMethodError`.

Reproduced on Rack 3.2.6.

## Where this happens in the source

In `lib/rack/deflater.rb`, the gzip branch unconditionally wraps the body:

```ruby
when "gzip"
  response[2] = GzipStream.new(body, GZIP_MTIME, @sync)
  response
```

and `GzipStream#each` iterates with `@body.each` for everything that is not a `File`:

```ruby
def each(&block)
  @writer = block
  gzip.mtime = @mtime if @mtime
  if @body.is_a? ::File # XXX: Should probably be ::IO
    while part = @body.read(BUFFER_LENGTH)
      ...
  else
    @body.each { |part|   # <- streaming (call-only) body raises NoMethodError here
      ...
```

There is no `respond_to?(:each)` / `respond_to?(:call)` branch anywhere in
`Deflater#call`, `should_deflate?`, or `GzipStream` (the only special-case is
`@body.is_a? ::File`). So a Rack 3 streaming body falls into the `else` branch and
crashes.

This is the same defect class that was fixed for `Rack::Events` in #2375 (forwarding
`#each` to a streaming body that only implements `#call`). The equivalent fix was
never applied to `Rack::Deflater`.

The SPEC seems to make this Deflater's obligation rather than a "by design" skip:
SPEC.rdoc states "The Body must respond to `each` or `call`... If the Body doesn't
respond to `each`, then you can assume it responds to `call`" and "Middleware must
not call `each` directly on the Body", and the UPGRADE-GUIDE says middleware "must
now check if the body responds to `#each` or not to determine if it is an enumerable
or streaming body." Please correct me if I am reading this wrong.

## Minimal reproduction (Rack only, no Sinatra)

`config.ru`:

```ruby
# frozen_string_literal: true
require "rack"
require "rack/deflater"

# A Rack 3 streaming body: responds to #call(stream), not #each.
class StreamingBody
  def call(stream)
    3.times { |i| stream.write("chunk #{i}\n") }
  ensure
    stream.close
  end
end

use Rack::Deflater
run ->(env) { [200, { "content-type" => "text/plain" }, StreamingBody.new] }
```

Run it under a Rack 3 server and request with gzip accepted:

```
$ rackup config.ru
$ curl -s --compressed http://localhost:9292/ ; echo "exit=$?"
```

## Expected vs actual

**Expected:** either a correctly gzip-encoded streaming response, or (acceptable
fallback) the streaming body passed through uncompressed. No exception.

**Actual:** `NoMethodError: undefined method 'each' for an instance of
StreamingBody` raised from `Rack::Deflater::GzipStream#each`, and the client
receives a broken ~20-byte gzip stream (gzip header/trailer only, no payload).

## Proposed direction (deferential, happy to be redirected)

Assuming I have read `deflater.rb` correctly, two options:

1. **Support streaming bodies (preferred, matches the callable-body intent and the
   `Rack::Events` precedent in #2375).** When the body responds to `#call` and not
   `#each`, return a callable gzip body that responds to `#call(stream)`: it calls
   `body.call` with a wrapping stream that gzips each `write`/`<<` and flushes
   through to the real stream, so SSE and other long-lived streams stay compressed.
   It may be worth skipping compression for `text/event-stream` specifically, since
   buffering/flushing gzip on SSE can hurt the latency that SSE exists to provide.

2. **Detect and skip (minimal, safe stopgap).** Before wrapping, if the body does
   not respond to `#each`, pass it through uncompressed instead of wrapping it in
   `GzipStream`. This avoids the crash but gives up compression on exactly the
   long-lived responses where it helps most.

I would lean toward option 1 to match #2375 and the SPEC's explicit `call`-body
support, with option 2 as the conservative fallback if you prefer to land a fix
quickly.

## PR offer

Happy to open a PR if this is wanted, with a regression test in
`test/spec_deflater.rb` (minitest spec style) and a `CHANGELOG.md` entry under
`### Fixed`. Please let me know which direction you would prefer (full streaming
support vs. skip) and I will follow Rack's conventions (`bundle exec rake test`).

SAY-5:
Deflater wraps the response body in `GzipStream` and consumes it with `@body.each`, but a Rack 3 streaming body only responds to `call(stream)`, so it raises `NoMethodError: undefined method 'each'` and the client gets a truncated gzip stream. This adds a `call` path to `GzipStream` that compresses what the body writes to the stream, and delegates `respond_to?(:each/:call)` to the wrapped body so a server treats the result as enumerable or streaming on the same basis as the original. Same idea as the streaming support added for `Rack::Events` in #2375.

Fixes #2470.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,7 @@ All notable changes to this project will be documented in this file. For info on
 - `Rack::Multipart::UploadedFile` now delegates keyword arguments to the wrapped tempfile. Calls such as `uploaded_file.readlines(chomp: true)` raised `TypeError` on Ruby 3.0+. ([#2481](https://github.com/rack/rack/issues/2481), [#2499](https://github.com/rack/rack/pull/2499), [@SeanLF](https://github.com/SeanLF))
 - Multipart parser: limit MIME header size check to the unread buffer region to avoid false `multipart mime part header too large` errors when previously read data accumulates in the scan buffer. ([#2392](https://github.com/rack/rack/pull/2392), [@alpaca-tc](https://github.com/alpaca-tc), [@willnet](https://github.com/willnet), [@krororo](https://github.com/krororo))
 - Multipart parser: add nil guards to prevent `NoMethodError` crashes when handling `Content-Disposition` without parameters and `Content-Type` parameters without '='. ([@haruki0409](https://github.com/haruki0409))
+- `Rack::Deflater` now gzips streaming (call-only) bodies instead of raising `NoMethodError: undefined method 'each'`. ([#2470](https://github.com/rack/rack/issues/2470), [@SAY-5](https://github.com/SAY-5))
 
 ## [3.2.6] - 2026-04-01
 
--- a/lib/rack/deflater.rb
+++ b/lib/rack/deflater.rb
@@ -197,6 +197,9 @@ module Rack
         # and with sync enabled every write already flushes.
         @gzip.flush unless @flushed
         @flushed = true
+        # Propagate to the wrapped stream so a write-then-flush (e.g. SSE)
+        # is not left sitting in a buffering server-side wrapper.
+        @stream.flush if @stream.respond_to?(:flush)
         self
       end
 
--- a/lib/rack/deflater.rb
+++ b/lib/rack/deflater.rb
@@ -134,6 +134,29 @@ module Rack
         gzip.finish
       end
 
+      # Compress a streaming (call-only) body on the fly, wrapping the
+      # server-provided stream so the body writes plain data and the client
+      # receives gzip.
+      def call(stream)
+        @writer = stream.method(:write)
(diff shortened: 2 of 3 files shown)


SAY-5:
Bumping this, the flush propagation fix from 61f7864 has been sitting for a few weeks now if there's time for another look.

PR #2472, opened 2026-06-20
Skip ETag generation for Cache-Control: no-store responses

islue:
## Summary

`Rack::ETag` adds a weak `ETag` to any bufferable (`to_ary`) 200/201 response regardless of `Cache-Control`. This PR makes it skip ETag generation when the response is marked `Cache-Control: no-store`.

## Why

`Cache-Control: no-store` directs that a cache "MUST NOT store any part of either the immediate request or the response" (RFC 9111 §5.2.2.5). An `ETag` is "an opaque validator for distinguishing between multiple representations of the same resource" (RFC 9110 §8.8.3); a cache uses it to revalidate a *stored* representation via a conditional request. If a response may never be stored, there is nothing for a cache to revalidate, so the weak `ETag` that `Rack::ETag` generates can never be used for that purpose — it is dead metadata on a response that has explicitly opted out of caching.

**This is not a revert of #1416.** That change was correct: `no-cache` permits storage but requires revalidation before reuse (RFC 9111 §5.2.2.4), so a validator is genuinely useful there. `no-store` is the opposite — it forbids storage outright. When both directives are present, `no-store` dominates: once nothing may be stored, `no-cache`'s "revalidate before reuse" never comes into play, so the `ETag` is skipped. The existing "set etag even if no-cache is given" behavior is unchanged, and a new test documents the `no-cache, no-store` precedence.

## Scope and motivation

To be upfront: I haven't traced this to a concrete failure — it's a correctness/tidiness change, not a bug report. For a fully conformant client an auto `ETag` on a `no-store` response is harmless, because the response is never stored and the validator is never exercised. But it is equally never *useful*, and it does hand a content-derived validator to clients or intermediaries that mishandle `no-store`: one that stored the response and later revalidated would get a `304 Not Modified` (empty body) back from `Rack::ConditionalGet`, for a response that was never supposed to be stored. Suppressing the validator costs conformant clients nothing — it also skips computing a body digest that would only be discarded — and keeps `Rack::ETag`'s output consistent with the response's own caching intent. It seemed worth keeping clean.

## Tests

- `not set etag if no-store is given`
- `not set etag even if no-store is combined with no-cache` (documents that `no-store` wins over `no-cache`)

### DIFF
--- a/lib/rack/etag.rb
+++ b/lib/rack/etag.rb
@@ -8,9 +8,11 @@ require_relative 'utils'
 module Rack
   # Automatically sets the etag header on all String bodies.
   #
-  # The etag header is skipped if etag or last-modified headers are sent or if
+  # The etag header is skipped if etag or last-modified headers are sent, if
   # a sendfile body (body.responds_to :to_path) is given (since such cases
-  # should be handled by apache/nginx).
+  # should be handled by apache/nginx), or if the response is marked
+  # cache-control: no-store (since such responses must not be stored, an
+  # entity-tag validator would never be usable).
   #
   # On initialization, you can pass two parameters: a cache-control directive
   # used when etag is absent and a directive when it is present. The first
@@ -55,6 +57,11 @@ module Rack
       end
 
       def skip_caching?(headers)
+        if cache_control = headers[CACHE_CONTROL]
+          # Skip ETag generation if caching is explicitly denied:
+          return true if cache_control.match?(/\bno-store\b/)
+        end
+
         headers.key?(ETAG_STRING) || headers.key?('last-modified')
       end
 
--- a/lib/rack/etag.rb
+++ b/lib/rack/etag.rb
@@ -7,9 +7,11 @@ require_relative 'constants'
 module Rack
   # Automatically sets the etag header on all String bodies.
   #
-  # The etag header is skipped if etag or last-modified headers are sent or if
+  # The etag header is skipped if etag or last-modified headers are sent, if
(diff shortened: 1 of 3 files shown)


jeremyevans (approved):
This seems reasonable. Potentially, client software could be using the ETag values for things other than caching (such as for detecting file changes instead of doing their own hashing). However, that seems like an unlikely scenario, and shouldn't preclude us from making this change, IMO.

lib/rack/etag.rb
```diff
@@ -55,7 +57,8 @@ def etag_status?(status)
       end
 
       def skip_caching?(headers)
-        headers.key?(ETAG_STRING) || headers.key?('last-modified')
+        headers.key?(ETAG_STRING) || headers.key?('last-modified') ||
+          /\bno-store\b/.match?(headers[CACHE_CONTROL].to_s)
```
ioquatix: I would prefer we expanded this and added comments explaining the behaviour, e.g.

```ruby
if cache_control = headers[CACHE_CONTROL]
  # Skip ETag generation if caching is explicitly denied:
  return false if cache_control.match?(/\bno-store\b/)
```

or something to that effect.

islue: Good call — done.
The matcher is borrowed from Deflater#should_deflate?; I used return true to match skip_caching?'s semantics (it returns true to skip).

islue:
Hi, resolved the conflict in CHANGELOG.md. Any further fix request?

PR #2474, opened 2026-07-06
Add support for HTTP QUERY method

seanpdoyle:
As of June 2026, [RFC 10008][] is a proposed standard.

From the RFC Abstract:

> This specification defines the QUERY method for HTTP. A QUERY requests
> that the request target process the enclosed content in a safe and
> idempotent manner and then respond with the result of that processing.
> This is similar to POST requests, but QUERY requests can be
> automatically repeated or restarted without concern for partial state
> changes.

This commit proposes support for the HTTP Query method by way of the `Rack::QUERY` constant and the corresponding `#query?` methods for both `Rack::Request` and `Rack::MockRequest` classes.

[RFC 10008]: https://www.rfc-editor.org/rfc/rfc10008.html


jeremyevans (approved):
This seems reasonable to add. net/http recently added support for QUERY. We might want to wait until there are Ruby web servers supporting it before merging, though.

seanpdoyle:
> We might want to wait until there are Ruby web servers supporting it before merging, though.

That is reasonable. For reference, https://github.com/puma/puma/pull/3973 adds support to Puma.

seanpdoyle:
@jeremyevans https://github.com/puma/puma/pull/3973 has been merged, adding `QUERY` support to Puma.

jeremyevans:
I've already approved this, there is no need to rerequest review. Might want to rebase on main to get the JRuby tests passing. We'll see what other committers think, though I expect they will be fine with it.

ioquatix:
Thanks.

PR #2475, opened 2026-07-06
 Enable configuring MethodOverride Ractor safely

hmcguire-shopify:
Currently, using HTTP_METHODS or ALLOWED_METHODS from a non-main Ractor
will raise RactorIsolationErrors because they are mutable arrays. Since
the constants are public, they _could_ be modified by users to change
the behavior of MethodOverride.

This commit provides an alternative API for modifying the lists of
headers used by MethodOverride: passing them into initialize. By storing
the values in instance variables, calling `Ractor.make_shareable` on the
middleware instance can freeze values passed in (whether they were
already frozen or not).

Additionally, the public constants are deprecated. Once the deprecation
period ends, the constants themselves could potentially be renamed
and/or frozen.

### DIFF
--- a/lib/rack/method_override.rb
+++ b/lib/rack/method_override.rb
@@ -8,18 +8,28 @@ module Rack
   class MethodOverride
     HTTP_METHODS = %w[GET HEAD PUT POST DELETE OPTIONS PATCH LINK UNLINK]
 
+    PRIV_HTTP_METHODS = HTTP_METHODS
+    private_constant :PRIV_HTTP_METHODS
+    deprecate_constant :HTTP_METHODS
+
     METHOD_OVERRIDE_PARAM_KEY = "_method"
     HTTP_METHOD_OVERRIDE_HEADER = "HTTP_X_HTTP_METHOD_OVERRIDE"
     ALLOWED_METHODS = %w[POST]
 
-    def initialize(app)
+    PRIV_ALLOWED_METHODS = ALLOWED_METHODS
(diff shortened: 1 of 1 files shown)


jeremyevans (commented):
I don't think we can freeze these without deprecation warnings, because users could be modifying them to change the supported methods. We would need to deprecate them in the next minor release, and then we could remove/make private/freeze them them in the following minor release.

hmcguire-shopify:
Updated, I believe the new approach should be fully backwards compatible (although more code, temporarily)

jeremyevans (approved):
Thank you, this looks good. Irrespective of ractor compatibility, unfrozen public constants are generally a bad idea, and supporting per-instance configuration is better than supporting global configuration.

jeremyevans:
JRuby CI is failing due to an rbs issue, as rbs 4+ requires a native extension. Probably should limit rbs to < 4 temporarily on JRuby until the issue is fixed. External tests are failing in Roda's tests due to backwards incompatible changes in rdoc. I'll work on fixes for those.

jeremyevans:
JRuby CI fix: #2476 

External test failure is due to tilt's rdoc template (roda uses tilt's rdoc support in it's own specs). I fixed that (https://github.com/jeremyevans/tilt/commit/95afeb0633746455b10778967c9ea95c06daa506), just need to put out a release with the fix.

jeremyevans:
Tilt 2.8.0 has been released with the fix, so these tests should be green after #2476 is merged.

PR #2480, opened 2026-07-09
Add opt-in Ractor compat via rack/ractorize

hmcguire-shopify:
Ref https://github.com/rack/rack/pull/2428

Rack still has many constants and class variables in the public API which aren't frozen, and thus can't be referenced from the non-main Ractor. We can potentially deprecate the public constant modification, but it will take a few versions before we could actually freeze the values (and only then use within non-main Ractors).

This commit takes an opt-in approach: applications can `require "rack/ractorize"` and it will freeze all of the unfrozen constants/class variables for files that have been previously required. If public constant modification is deprecated as described above, then this file could eventually be simplified to only those constants that specifically require `Ractor.make_shareable`.

### DIFF
--- a/lib/rack/ractorize.rb
+++ b/lib/rack/ractorize.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+module Rack
+  file_loaded = ->(const) { const_defined?(const, false) && !autoload?(const) }
+
+  if file_loaded.(:Headers)
+    Headers::KNOWN_HEADERS.freeze
+  end
+
+  if file_loaded.(:Multipart)
+    Ractor.make_shareable(Multipart::Parser::TEMPFILE_FACTORY)
+    Multipart::Parser::REENCODE_DUMMY_ENCODINGS.freeze
+  end
+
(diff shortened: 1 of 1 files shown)


jeremyevans (approved):
I think this is definitely the approach we should take to add ractor support.

tenderlove:
I don't like that requiring a file has different behavior depending on the order in which it was required.  I think we should be making constants immutable in general, but the reality is that we have to deal with backwards compatibility.  I think this is a good compromise until we can reach "immutable constant" nirvana.

matthewd:
Should we just force-load everything here, instead of using the conditionals? 

The cost of a ractorizing app loading a handful of extra files/objects it might not use, while non-zero, strikes me as lower than them being obliged to pre-touch the ones they do need: it essentially gives up the abstraction if you have to know which parts of Rack need preloading.

hmcguire-shopify:
https://github.com/rack/rack/pull/2488

Removed the conditions so that the behavior is the same no matter the order 👍

PR #2482, opened 2026-07-27
Delegate keyword arguments in UploadedFile#method_missing

### LINKED ISSUE #2481: UploadedFile#method_missing drops keyword arguments on Ruby 3.0+
`Rack::Multipart::UploadedFile` forwards missing methods to the wrapped tempfile, but the delegator is not flagged with `ruby2_keywords`. On Ruby 3.0+, where keywords are separated from a trailing Hash, the keywords are rebuilt as a positional Hash before reaching the tempfile.

```ruby
require 'rack'
require 'tempfile'

f = Tempfile.new('x')
f.write("a\nb\n")
f.rewind
f.close

uf = Rack::Multipart::UploadedFile.new(f.path)

File.open(f.path) { |io| io.readlines(chomp: true) }  # => ["a", "b"]
uf.readlines(chomp: true)                             # => TypeError
uf.gets(chomp: true)                                  # => TypeError
```

```
TypeError: no implicit conversion of Hash into Integer
    delegate.rb:427:in 'IO#readlines'
    lib/rack/multipart/uploaded_file.rb:78:in 'Rack::Multipart::UploadedFile#method_missing'
```

Ruby 2.4 through 2.6 are unaffected, since they do not distinguish a trailing Hash from keywords. That is presumably why this went unnoticed when the other delegators were fixed in 2020.

`BodyProxy#method_missing` ([4804192](https://github.com/rack/rack/commit/4804192aaba4319f8a019043eeed84982c2515e7)) and `MockResponse::Cookie#method_missing` are both flagged; this one appears to have been missed.

Verified on Ruby 2.4.10, 2.6.10 and 3.4.10.

SeanLF:
Fixes #2481.

`Rack::Multipart::UploadedFile#method_missing` delegates to the wrapped tempfile but was never flagged with `ruby2_keywords`, so keyword arguments are rebuilt as a positional Hash on Ruby 3.0+:

```ruby
uf.readlines(chomp: true)  # => TypeError: no implicit conversion of Hash into Integer
uf.gets(chomp: true)       # => TypeError: no implicit conversion of Hash into Integer
```

This flags the delegator, matching `BodyProxy#method_missing` and `MockResponse::Cookie#method_missing`.

Ruby 2.4 to 2.6 are unaffected either way, since they do not distinguish a trailing Hash from keywords. The `respond_to?` guard is kept because `ruby2_keywords` does not exist before 2.7.

The added test fails without the change (`TypeError`) and passes with it. Verified on Ruby 2.4.10, 2.6.10, 3.0.7 and 4.0.6.

### DIFF
--- a/lib/rack/multipart/uploaded_file.rb
+++ b/lib/rack/multipart/uploaded_file.rb
@@ -77,9 +77,7 @@ module Rack
       def method_missing(method_name, *args, &block) #:nodoc:
         @tempfile.__send__(method_name, *args, &block)
       end
-      # :nocov:
       ruby2_keywords(:method_missing) if respond_to?(:ruby2_keywords, true)
-      # :nocov:
     end
   end
 end
--- a/lib/rack/multipart/uploaded_file.rb
+++ b/lib/rack/multipart/uploaded_file.rb
@@ -77,6 +77,9 @@ module Rack
       def method_missing(method_name, *args, &block) #:nodoc:
         @tempfile.__send__(method_name, *args, &block)
       end
+      # :nocov:
+      ruby2_keywords(:method_missing) if respond_to?(:ruby2_keywords, true)
+      # :nocov:
     end
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ All notable changes to this project will be documented in this file. For info on
 
 ### Fixed
 
+- `Rack::Multipart::UploadedFile` now delegates keyword arguments to the wrapped tempfile. Calls such as `uploaded_file.readlines(chomp: true)` raised `TypeError` on Ruby 3.0+. ([#2482](https://github.com/rack/rack/pull/2482), [@SeanLF](https://github.com/SeanLF))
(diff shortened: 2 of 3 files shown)


jeremyevans (approved):
Looks good, thank you for the patch.

In the future, if you have a fix prepared, it's best to submit just the pull request and not file a separate GitHub issue.

SeanLF:
Understood, thanks. Follow-up in #2484 (draft, stacked on this one).

lib/rack/multipart/uploaded_file.rb
```diff
@@ -77,6 +77,9 @@ def respond_to_missing?(*args)
       def method_missing(method_name, *args, &block) #:nodoc:
         @tempfile.__send__(method_name, *args, &block)
       end
+      # :nocov:
```
ioquatix: I'm not sure `:nocov:` is required here. the line itself will be evaluated due to the trailing conditional, and if we care about coverage, we should be testing multiple Ruby versions that cover this line anyway.

jeremyevans: `:nocov:` is used around all `ruby2_keywords` calls because we do branch coverage. We don't currently coverage test using multiple Ruby versions (if we start, then we can drop `:nocov:` around Ruby-version specific branches).

ioquatix (approved):
LGTM.

samuel-williams-shopify:
Merged in https://github.com/rack/rack/pull/2499

PR #2483, opened 2026-07-27
Use ruby2_keywords for Rack::Multipart::UploadedFile#method_missing

### LINKED ISSUE #2481: UploadedFile#method_missing drops keyword arguments on Ruby 3.0+

jeremyevans:
This allows correct forwarding of keyword arguments.

Fixes #2481

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ All notable changes to this project will be documented in this file. For info on
 
 - Multipart parser: limit MIME header size check to the unread buffer region to avoid false `multipart mime part header too large` errors when previously read data accumulates in the scan buffer. ([#2392](https://github.com/rack/rack/pull/2392), [@alpaca-tc](https://github.com/alpaca-tc), [@willnet](https://github.com/willnet), [@krororo](https://github.com/krororo))
 - Multipart parser: add nil guards to prevent `NoMethodError` crashes when handling `Content-Disposition` without parameters and `Content-Type` parameters without '='. ([@haruki0409](https://github.com/haruki0409))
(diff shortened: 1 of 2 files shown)


jeremyevans:
Sorry, looks like a duplicate of #2482.

Issue #1601, opened 2020-02-12
Rack 3.0 minimum Ruby version


ioquatix:
By the time we release Rack 3.0, I believe Ruby 2.4 will be EOL.

There are a few functions we would prefer to use, including `#tranform_keys`.

Do we want to support EOL Rubies? What would be a reasonable policy?

matthewd:
It's certainly not the deciding factor, but FWIW, the Rails 6.x series requires Ruby 2.5. 

If anything, that's mostly relevant in the other direction: we're ~going to~ _expecting_ to support 2.5 for a new release in 2021, so we'd prefer new Rack releases to also maintain support for that version until then (because otherwise we won't be able to use them).

That desire aside, I think it's generally reasonable to drop support for EOLed Ruby versions during major version bumps, even for deep infrastructure level projects like Rack. The disruption of the library upgrade will dwarf any disruption from a corresponding Ruby upgrade; it's not reasonable for people to expect us to help them be able to do the hard thing while skipping the easy one.

simi:
Just to remind, there are some stats available at http://ecosystem.rubytogether.org/ and `2.3` seems still widely used for some reason.

natematykiewicz:
As of today, Ruby 2.4 support has ended. https://twitter.com/rubylangorg/status/1249125876492615681?s=21

jeremyevans:
IMO, central infrastructure such as rack should only drop old ruby versions when doing so becomes a maintenance burden.  There is some Ruby <2.4 compatibility code we could remove (dropping 2.3 support), but I don't see any reason to drop Ruby 2.4 support just because it is no longer supported by ruby-core.

natematykiewicz:
I just mentioned it, since the originally post mentioned Ruby 2.4.

ioquatix:
We have dropped support for Ruby 2.3 and I believe we will continue to be pragmatic about it.

PR #2484, opened 2026-07-27
Use explicit keyword arguments instead of ruby2_keywords

SeanLF:
Draft, stacked on #2482. The first commit here is that PR; the diff reduces to a single commit once it merges.

Rack calls `ruby2_keywords` in four places, all behind `respond_to?` guards. Ruby proposes deprecating and removing it ([Feature #22205](https://bugs.ruby-lang.org/issues/22205), currently Open, no target version). Because of the guards, removal would not raise. It would silently skip, and middleware and delegators taking keyword arguments would start receiving a positional Hash.

```
with ruby2_keywords:   args=[:positional]                kwargs={answer: 42}
without:               args=[:positional, {answer: 42}]  kwargs={}
```

This forwards keywords explicitly on Ruby 3.0+ in `Builder#use`, `BodyProxy#method_missing`, `MockResponse::Cookie#method_missing` and `UploadedFile#method_missing`. **Argument forwarding is unchanged on every supported Ruby.**

### Why the definitions are version-gated

Adding `**kwargs` outright regresses Ruby 2.4 to 2.7, where a trailing positional Hash is not distinguished from keywords:

| call | current | naive `**kwargs` |
|---|---|---|
| `use M, opts` where `opts == {}` | `{}` | `ArgumentError (given 1, expected 2)` |
| `use M, {"a" => 1, :b => 2}` | `{"a"=>1, :b=>2}` | `ArgumentError` |

An empty options Hash is dropped entirely; a Hash mixing Symbol and non-Symbol keys is split in two. So the pre-3.0 branch keeps `ruby2_keywords`, which exists on every Ruby that branch can run on. Follows the `if`/`else` version-gate style of `Utils#clock_time` and `Utils#escape_html`.

<details>
<summary>If Rack ever drops Ruby &lt; 3.0</summary>

Every gate collapses and each site becomes a single definition with no duplication and no `:nocov:`. Not proposing that here; #1601 settled the policy and this patch deliberately keeps 2.4 working.
</details>

### Verification

Compared the values actually received by middleware and delegated objects against unpatched `main` on **2.4, 2.6, 2.7, 3.0 and 3.4**: byte-identical for keyword arguments, positional Hash, empty Hash, mixed-key Hash, no-argument middleware and splat middleware. Same for `BodyProxy` (including `close`-on-raise, `to_str`, and `method(:x).call`), for `UploadedFile`, and for consumers that subclass `Builder` or prepend a module over `use`. `Builder#map` flush ordering verified unchanged across four `map`/`use` interleavings.

Three tests added. They fail against a naive `**kwargs` version (2 errors on 2.6) and against the `ruby2_keywords`-removed scenario.

### Notes

- `Builder#use.parameters` gains `[:keyrest, :kwargs]` on 3.0+. Only observable API change.
- `Builder#use` now delegates to a private `add_middleware`, a pure extraction of the existing `@map`/`@use` logic, so the two version branches stay one line each.
- Subclasses overriding `use` with a bare splat already lose keywords on Ruby 3.0+. Pre-existing, unchanged here.
- No documentation changes: `SPEC.rdoc` covers the server/application protocol, not the Builder DSL, and the call syntax is unchanged.

No urgency, since #22205 is still Open. Happy to hold this until it is accepted, or to split it per file if that reviews more easily.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ All notable changes to this project will be documented in this file. For info on
 - Multipart parser drops support for RFC 2231 `filename*` parameter (prohibited by RFC 7578) and now properly handles UTF-8 encoded filenames via percent-encoding and direct UTF-8 bytes. ([#2398](https://github.com/rack/rack/pull/2398), [@wtn](https://github.com/wtn))
 - The query parser now raises `Rack::QueryParser::IncompatibleEncodingError` if we try to parse params that are not ASCII compatible. ([#2416](https://github.com/rack/rack/pull/2416), [@bquorning](https://github.com/bquorning))
 - The mime type for `.pem` files has been changed from `application/x-x509-ca-cert` to `application/x-pem-file`. ([#2435](https://github.com/rack/rack/pull/2435), [@jeremyevans](https://github.com/jeremyevans))
+- Use explicit keyword arguments instead of `ruby2_keywords` on Ruby 3.0+, which Ruby proposes to deprecate and remove ([Feature #22205](https://bugs.ruby-lang.org/issues/22205)). Argument forwarding is unchanged on every supported Ruby. ([@SeanLF](https://github.com/SeanLF))
 - Freeze `Rack::Auth::AbstractRequest::AUTHORIZATION_KEYS`, `Rack::Utils::STATUS_WITH_NO_ENTITY_BODY`, `Rack::Multipart::Parser::EMPTY`, `Rack::Utils.default_query_parser`, and internal constants in `Rack::Lint`. ([#2428](https://github.com/rack/rack/pull/2428), [@jhawthorn](https://github.com/jhawthorn))
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ All notable changes to this project will be documented in this file. For info on
 
 ### Fixed
 
+- `Rack::Multipart::UploadedFile` now delegates keyword arguments to the wrapped tempfile. Calls such as `uploaded_file.readlines(chomp: true)` raised `TypeError` on Ruby 3.0+. ([#2482](https://github.com/rack/rack/pull/2482), [@SeanLF](https://github.com/SeanLF))
 - Multipart parser: limit MIME header size check to the unread buffer region to avoid false `multipart mime part header too large` errors when previously read data accumulates in the scan buffer. ([#2392](https://github.com/rack/rack/pull/2392), [@alpaca-tc](https://github.com/alpaca-tc), [@willnet](https://github.com/willnet), [@krororo](https://github.com/krororo))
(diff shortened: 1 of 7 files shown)


jeremyevans:
Thank you for the pull request.

I think if Ruby decides to deprecate `ruby2_keywords`, we should drop support for Ruby < 3.0 (maybe 2.7 if `...` will work), as opposed to duplicating method definitions and keeping support. However, I'm not sure how the other committers feel about that.

SeanLF:
Yes, `...` works on 2.7, including for `Builder#use`.

It's the same code at either floor, so 2.7 vs 3.0 stays a policy call rather than a technical one. Explicit `*args, **kwargs` would force 3.0: on 2.7 it still mangles a Hash mixing Symbol and non-Symbol keys.

Cleaner diff too: no `ruby2_keywords`, no version gates, no duplicated definitions. Happy to convert this PR to it, or to park it until #22205 progresses.

<details>
<summary>Verification</summary>

`Dots` uses `...`, `Kw` uses explicit keywords, `Flag` is the current approach. `Builder#use` stores its arguments in a proc rather than forwarding immediately, so that shape is what's exercised. `Pos` takes a required positional options Hash, which is where they diverge.

```sh
for v in 2.7 3.0 3.4; do printf "ruby:%-5s " "$v"; docker run --rm ruby:$v ruby -e '
class Dots; def use(m, ...); @p = proc { |app| m.new(app, ...) }; end; def to_app(a); @p[a]; end; end
class Kw;   def use(m, *a, **k, &b); @p = proc { |app| m.new(app, *a, **k, &b) }; end; def to_app(a); @p[a]; end; end
class Flag; def use(m, *a, &b); @p = proc { |app| m.new(app, *a, &b) }; end; ruby2_keywords(:use); def to_app(a); @p[a]; end; end
class Pos; def initialize(app, options); $out << options; end; end
o = {}
[Dots, Kw, Flag].each { |c| $out = []
  [->(b){ b.use(Pos, o) }, ->(b){ b.use(Pos, {"s" => 1, sym: 2}) }, ->(b){ b.use(Pos, {a: 1}) }].each do |fn|
    b = c.new; begin; fn.call(b); b.to_app(:app); rescue ArgumentError; $out << :ArgumentError; end
  end
  puts "#{c}: #{$out.inspect}" }
'; echo; done
```

```
ruby:2.7   Dots: [{}, {"s"=>1, :sym=>2}, {:a=>1}]
           Kw:   [{}, :ArgumentError, {:a=>1}]
           Flag: [{}, {"s"=>1, :sym=>2}, {:a=>1}]
ruby:3.0   all three identical
```

2.4-2.6 raise SyntaxError on `...`. JRuby 9.4 and TruffleRuby both match.

Also checked against unpatched `main`: `BodyProxy` delegation including `close`-on-raise and `method(:x).call`, `UploadedFile` delegation, `Builder#map` flush ordering, and consumers that subclass `Builder` or prepend a module over `use`. All identical.
</details>

PR #2485, opened 2026-07-29
Apply header_rules to the same canonical path used to serve the file

anir0y:
Follow-up to `GHSA-q4qf-9j86-f5mh` as requested in GHSA-74qm-qhv6-h593 — sending it as a normal PR since the security implications are minimal.

`Rack::Static#applicable_rules` canonicalises with `unescape_path` only, while the servability decision and the file server both also apply `clean_path_info`:

| step | canonicalisation | decides |
|---|---|---|
| `Static#call` → `can_serve` | `clean_path_info(unescape_path(path))` | whether to serve |
| `Rack::Files#call` | `clean_path_info(unescape_path(path_info))` | which file is served |
| `Static#applicable_rules` | `unescape_path(path)` | which headers are applied |

So a path with dot-segments normalises onto a rule-matched path and is served, while the rule is tested against the un-normalised string and fails to match — the file comes back without its configured headers.

Affected shape is prefix matching: `String` rules and `Regexp`s anchored with `\A`. Extension-anchored (`:fonts`, `Array`, `/\.woff\z/`) and unanchored rules are unaffected, since the suffix survives normalisation.

### Change

```ruby
path = ::Rack::Utils.clean_path_info(::Rack::Utils.unescape_path(path))
```

the same pair the other two call sites already use.

### Tests

Three added to `test/spec_static.rb`, covering a dot-segment path, its percent-encoded form, and a `\A`-anchored `Regexp` rule. Verified in both directions — they fail on unpatched `lib/rack/static.rb` (`max-age=100` instead of `400`, and `nil` for the Regexp case) and pass with the change. `spec_static.rb` 34 runs / 0 failures; `spec_files.rb` and `spec_utils.rb` unaffected.

A longer-term option, if you'd prefer it: compute the canonical path once in `Static#call` and pass it to both `can_serve` and `applicable_rules`, so the two cannot drift again. Happy to redo it that way.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ All notable changes to this project will be documented in this file. For info on
 
 - Multipart parser: limit MIME header size check to the unread buffer region to avoid false `multipart mime part header too large` errors when previously read data accumulates in the scan buffer. ([#2392](https://github.com/rack/rack/pull/2392), [@alpaca-tc](https://github.com/alpaca-tc), [@willnet](https://github.com/willnet), [@krororo](https://github.com/krororo))
 - Multipart parser: add nil guards to prevent `NoMethodError` crashes when handling `Content-Disposition` without parameters and `Content-Type` parameters without '='. ([@haruki0409](https://github.com/haruki0409))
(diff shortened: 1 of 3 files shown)


jeremyevans (approved):
I would remove all comments added in this pull request as I don't think they add value. Otherwise, this looks good.

PR #2486, opened 2026-07-29
Backport OBS unfolding for multipart requests to 3-1-stable

anir0y:
Requested in GHSA-mpw4-4qx9-2625. Backport of d50c4d3d ("Implement OBS unfolding for multipart requests per RFC 5322 2.2.3") from `main`.

`GHSA-rx22-g9mx-qrhv` records its affected range as `>= 3.2.0, < 3.2.6`, so this branch is outside it — but the parser here accepts folded multipart headers and preserves the fold, leaving the CRLF embedded in parsed parameter values:

```
rack 3.1.21   Content-Disposition: ...; filename="a<CRLF> b.txt"   ->  filename == "a\r\n b.txt"
rack 3.2.6    (same input, with the fix)                           ->  filename == "a b.txt"
```

Applies cleanly here — same `handle_mime_head` structure as `main`, so it is the upstream diff unmodified: the `OBS_UNFOLD` constant plus one `gsub!` on each of the Content-Type and Content-Disposition strings.

### Tests

The regression test from the same upstream commit, ported verbatim. Verified in both directions: it fails on this branch without the change and passes with it.

```
without:  96 runs, 323 assertions, 1 failures
with:     96 runs, 324 assertions, 0 failures
```

I appreciate `CONTRIBUTING.md` reserves backports for security patches — sending this because you asked for it on the advisory, and framing it as the correctness fix you described rather than a security one. Happy to close if you would rather it only land on `main`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 All notable changes to this project will be documented in this file. For info on how to format all future additions to this file please reference [Keep A Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## Unreleased
+
+### Security
+
+- [CVE-2026-26962](https://github.com/advisories/GHSA-rx22-g9mx-qrhv) Improper unfolding of folded multipart headers preserves CRLF in parsed parameter values.
+
 ## [3.1.21] - 2026-04-01
 
 ### Security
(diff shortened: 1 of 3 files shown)


jeremyevans (approved):
As we backported this to 3.2, and 3.1 is affected, it seems reasonable to backport this to 3.1

jeremyevans:
Please fix the authorship of the original commit when backporting, though.

PR #2490, opened 2026-08-03
Freeze ShowExceptions::TEMPLATE

hmcguire-shopify:
Freezing an ERB instance should be all that's necessary to make it shareable across Ractors.

This commit freezes the ERB instance inline since its below the :stopdoc: marker, meaning it should already not be getting modified by any consumers.

### DIFF
--- a/lib/rack/show_exceptions.rb
+++ b/lib/rack/show_exceptions.rb
@@ -133,7 +133,7 @@ module Rack
     # Copyright (c) Django Software Foundation and individual contributors.
     # Used under the modified BSD license:
     # http://www.xfree86.org/3.3.6/COPYRIGHT2.html#5
-    TEMPLATE = ERB.new(<<~'HTML')
+    TEMPLATE = ERB.new(<<~'HTML').freeze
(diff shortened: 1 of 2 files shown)


lib/rack/show_exceptions.rb
hmcguire-shopify, line 130: Maybe instead of `:stopdoc:` we should `private_constant` it?

jeremyevans: `private_constant` can break code currently accessing it to modify the template used. We can set a new internal `private_constant` to the same value, and use `deprecate_constant` on the current constant. Then drop the deprecated constant later, and after that, we can add the `freeze` and remove the `Ractor.make_shareable`.

jeremyevans (changes requested):

lib/rack/show_exceptions.rb
hmcguire-shopify: Why is the `stopdoc` not indicative of "safe to change"? I assumed that "documented" was the contract for backwards compatibility.

jeremyevans: `private` (method or constant) indicates it is internal and safe to change. `stopdoc` only indicates it is not documented.

PR #2491, opened 2026-08-12
Normalize path before matching Rack::Static header_rules

anir0y:
`Rack::Static` decides whether a request is servable (`#call`), and `Rack::Files` locates the file on disk, using the same normalization pair:

```ruby
Utils.clean_path_info(Utils.unescape_path(path))
```

`applicable_rules` only applied `unescape_path`:

| step | transformation applied | decides |
| --- | --- | --- |
| `Static#call` → `can_serve` | `clean_path_info(unescape_path(path))` | whether to serve |
| `Rack::Files#call` | `clean_path_info(unescape_path(path_info))` | which file is served |
| `Static#applicable_rules` | `unescape_path(path)` only | which headers are applied |

So a request whose path normalizes onto a rule-matched path is served, while the rule is tested against the un-normalized string and fails to match — the file comes back without its configured headers.

Prefix-matching rules are the affected shape, because dot segments change the prefix while normalization restores the served path. Measured before this change, with `header_rules` configured for a `/fonts` directory and a request for `/foo/../fonts/test.woff`:

| rule | result |
| --- | --- |
| `String '/fonts'` | headers missing |
| `String 'fonts'` | headers missing |
| `Regexp %r{\A/fonts/}` | headers missing |
| `Regexp %r{/fonts/}` (unanchored) | headers applied |
| `Regexp %r{\.woff\z}` | headers applied |
| `Array ['woff']`, `:fonts`, `:all` | headers applied |

Suffix-matching and match-anywhere rules are unaffected because the suffix survives the traversal. A `String` rule is the documented way to attach headers to a directory of static assets, so the affected shape is the common configuration.

## The change

Apply the same normalization pair in `applicable_rules`, so all three agree.

I kept the normalization local to the method rather than passing the already-computed `actual_path` down from `#call`. `#call` deliberately re-reads `PATH_INFO` at the end (after the `:urls` hash mapping and the gzip `.gz` handling may have rewritten it), and header rules should keep matching the path actually served rather than the original request path. Passing `actual_path` would silently change behaviour for the `:urls` mapping case.

## Tests

Regression coverage for each affected shape — `String` with and without a leading slash, `\A`-anchored `Regexp`, and duplicate slashes — plus an extension-anchored case as a control for the shapes that already worked.

All five new prefix-rule tests fail without the change (they fall through to the `:all` rule) and pass with it; the extension-anchored control passes either way. `spec_static.rb`, `spec_files.rb` and `spec_utils.rb` are green.

Opening this as a normal pull request as requested.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ All notable changes to this project will be documented in this file. For info on
 
 - Multipart parser: limit MIME header size check to the unread buffer region to avoid false `multipart mime part header too large` errors when previously read data accumulates in the scan buffer. ([#2392](https://github.com/rack/rack/pull/2392), [@alpaca-tc](https://github.com/alpaca-tc), [@willnet](https://github.com/willnet), [@krororo](https://github.com/krororo))
 - Multipart parser: add nil guards to prevent `NoMethodError` crashes when handling `Content-Disposition` without parameters and `Content-Type` parameters without '='. ([@haruki0409](https://github.com/haruki0409))
+- `Rack::Static`: match `header_rules` against the same normalized path used to decide whether the file is served, so requests containing dot segments or duplicate slashes receive their configured headers. ([@anir0y](https://github.com/anir0y))
 
 ## [3.2.6] - 2026-04-01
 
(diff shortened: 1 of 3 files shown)


anir0y:
Closing as a duplicate — I already have #2485 open against `main` with the same `applicable_rules` change, and I missed it before opening this. Sorry for the noise.

I'll consolidate on #2485 and move the extra regression cases and the CHANGELOG entry from here over to it.

PR #2492, opened 2026-08-13
Allow disabling the query bytesize limit via nil

alpaca-tc:
## Background

`Request#POST` reads the request body using:

```ruby
get_header(RACK_INPUT).read(query_parser.bytesize_limit + 2)
```

If `bytesize_limit` is configured close to the OS's `read(2)` size ceiling (`2**32` on many platforms), the `+ 2` padding pushes the requested read size past that ceiling. The resulting behavior is OS-dependent: some platforms raise `Errno::EINVAL`, others silently truncate the read.

Working around this by lowering `RACK_QUERY_PARSER_BYTESIZE_LIMIT` a few bytes below the OS ceiling is fragile: callers have to know about the internal `+ 2` and back it out of a config value, and any change to that offset (or a different OS ceiling) breaks the workaround again.

Some deployments also enforce a request size limit upstream (e.g. in a reverse proxy), making rack's own `bytesize_limit` redundant. Those users currently have no way to fully disable the check; they still have to pick *some* number small enough to stay under the OS read limit, effectively managing the same limit twice.

## What changed

`QueryParser#bytesize_limit` and `QueryParser#params_limit` now accept `nil` to mean "no limit":

- `each_query_pair` skips the bytesize check entirely when `@bytesize_limit` is `nil`, and skips the params-count check entirely when `@params_limit` is `nil`.
- `Request#POST` reads the body with an unbounded `read` (no length argument) when `bytesize_limit` is `nil`, instead of `bytesize_limit + 2`.
- `RACK_QUERY_PARSER_BYTESIZE_LIMIT` and `RACK_QUERY_PARSER_PARAMS_LIMIT` become `nil` (no limit) when set to a **negative** value.

This lets callers who already bound request size elsewhere opt out of the check with `bytesize_limit: nil` / `params_limit: nil`, without having to reason about rack's internal read padding relative to the OS's read size limit.

### Backwards compatibility for `0`

Unlike `Rack::Multipart::Parser::PARSER_BYTESIZE_LIMIT` (where a non-positive `RACK_MULTIPART_PARSER_BYTESIZE_LIMIT` is treated as `nil`), a value of `0` is **not** treated specially here, for either the environment variables or the `bytesize_limit:`/`params_limit:` keyword arguments:

- Before this change, `RACK_QUERY_PARSER_BYTESIZE_LIMIT=0` (or `bytesize_limit: 0`) resulted in a limit of `0`, which effectively rejected every non-empty query string/body. Some deployments may rely on this to reject all query parameters.
- Converting `0` to "no limit" would silently invert that behavior for existing users, so `0` continues to be used as a literal limit of zero. Only a negative value (env var) or explicit `nil` (keyword argument) disables the check.

### DIFF
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -52,11 +52,10 @@ module Rack
         end
       end
 
-      val
+      val if val >= 0
     end
 
-    bytesize_limit = env_int.call("RACK_QUERY_PARSER_BYTESIZE_LIMIT", 4194304)
-    BYTESIZE_LIMIT = bytesize_limit > 0 ? bytesize_limit : nil
+    BYTESIZE_LIMIT = env_int.call("RACK_QUERY_PARSER_BYTESIZE_LIMIT", 4194304)
     private_constant :BYTESIZE_LIMIT
 
     PARAMS_LIMIT = env_int.call("RACK_QUERY_PARSER_PARAMS_LIMIT", 4096)
@@ -233,9 +232,10 @@ module Rack
         raise QueryLimitError, "total query size exceeds limit (#{@bytesize_limit})"
       end
 
-      pairs = qs.split(separator ? (COMMON_SEP[separator] || /[#{separator}] */n) : DEFAULT_SEP, @params_limit + 1)
+      sep = separator ? (COMMON_SEP[separator] || /[#{separator}] */n) : DEFAULT_SEP
+      pairs = @params_limit ? qs.split(sep, @params_limit + 1) : qs.split(sep)
 
-      if pairs.size > @params_limit
+      if @params_limit && pairs.size > @params_limit
         param_count = pairs.size + pairs.last.count(separator || "&")
         raise QueryLimitError, "total number of query parameters (#{param_count}) exceeds limit (#{@params_limit})"
       end
--- a/lib/rack/query_parser.rb
+++ b/lib/rack/query_parser.rb
@@ -55,7 +55,8 @@ module Rack
       val
     end
 
-    BYTESIZE_LIMIT = env_int.call("RACK_QUERY_PARSER_BYTESIZE_LIMIT", 4194304)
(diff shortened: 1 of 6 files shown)


lib/rack/query_parser.rb
```diff
@@ -55,7 +55,8 @@ def self.make_default(param_depth_limit, **options)
       val
```
jeremyevans: Preserve backwards compatibility for a `0` value. Be consistent for both `RACK_QUERY_PARSER_BYTESIZE_LIMIT` and `RACK_QUERY_PARSER_PARAMS_LIMIT`:

```suggestion
      val if val >= 0
```

lib/rack/query_parser.rb
```diff
@@ -55,7 +55,8 @@ def self.make_default(param_depth_limit, **options)
...
     end
 
-    BYTESIZE_LIMIT = env_int.call("RACK_QUERY_PARSER_BYTESIZE_LIMIT", 4194304)
+    bytesize_limit = env_int.call("RACK_QUERY_PARSER_BYTESIZE_LIMIT", 4194304)
+    BYTESIZE_LIMIT = bytesize_limit > 0 ? bytesize_limit : nil
```
jeremyevans:
```suggestion
    BYTESIZE_LIMIT = env_int.call("RACK_QUERY_PARSER_BYTESIZE_LIMIT", 4194304)
```

jeremyevans (commented):
Thank you for the pull request.

There is a backwards compatibility issue with the proposed behavior. I've suggested a couple fixes to improve that.

`params_limit` should get the same treatment, for consistency.

Please also update the documentation to explain how to disable the limits, as well as the changelog.

ioquatix:
I don't have a strong opinion about this but in principle LGTM.

alpaca-tc:
@jeremyevans  I've updated the code and the PR description, and rebased and force-pushed to resolve the conflicts 😄  🙏

jeremyevans (approved):
Changes look good, thank you!

PR #2493, opened 2026-08-13
Support rack.request.config :authority key

jeremyevans:
This makes Rack::Request#authority use the value of the key if set. This affects the following methods:

* authority
* host
* hostname
* host_with_port
* port

This should be used in any case where you know what the host should be. It can prevent user-submitted overrides via Forwarded or X-Forwarded-Host headers, or the Host header (if the server can be reached by arbitrary hostnames).


ioquatix (approved):
Seems reasonable to me.

PR #2494, opened 2026-08-13
Remove Rack::Request#parse_multipart

jeremyevans:
This is a private method that has not been used since 3855d1d2638af92193c0636107247bc31c98b2f1. It was deprecated in 1ba8ce1f6bb2843485b853c331f8be0ab1c96bd6, before Rack 3.2.0, so it is safe to remove.


ioquatix (approved):
@matthewd Please let us know if this breaks Rails some how.

PR #2496, opened 2026-08-13
Enable warnings for test:regular

jeremyevans:
test:separate already enables warnings, and in general I think its best to run tests with warnings enabled.


ioquatix (approved):
LGTM.

PR #2497, opened 2026-08-13
Support rack.request.config :multipart_parser

jeremyevans:
This allows configuring the multipart parser for a request similar to how the query parser is configured, via rack.request.config.

To enable this, add some class methods to Rack::Multipart::Parser:

* parse_from_env
* default_tempfile_factory
* default_bufsize
* default_query_parser

Most of the logic from Rack::Multipart.parse_multipart is moved to Rack::Multipart::Parser.parse_from_env.

Rack::Request#multipart_parser is added to return the multipart parser to use. The default is a subclass of Rack::Multipart::Parser that will use Rack::Multipart::ParamList as the default query parser.

Rack::Request#POST now calls multipart_parser.parse_from_env(env) to perform the multipart parse.

In the future, we can move more multipart parsing settings (e.g. limits) into class or instance methods of Rack::Multipart::Parser, so that more settings can be configured via rack.request.config :multipart_parser.

### DIFF
--- a/lib/rack/request.rb
+++ b/lib/rack/request.rb
@@ -599,7 +599,7 @@ module Rack
           if rack_input.nil?
             set_header(RACK_REQUEST_FORM_PAIRS, [])
           elsif form_data? || parseable_data?
-            if pairs = Rack::Multipart.parse_multipart(env, Rack::Multipart::ParamList)
+            if pairs = multipart_parser.parse_from_env(env)
               set_header RACK_REQUEST_FORM_PAIRS, pairs
             else
               # Add 2 bytes. One to check whether it is over the limit, and a second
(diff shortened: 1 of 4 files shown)


lib/rack/request.rb
```diff
@@ -778,6 +778,18 @@ def query_parser
         @query_parser || config_value(:query_parser) || Utils.default_query_parser
       end
 
+      # The default multipart parser
+      class MultipartParser < Multipart::Parser
```
ioquatix, line 775: Why don't we just add `def self.default_query_parser` to `Multipart::Parser`?

jeremyevans: Because I want to override `default_query_parser` so that `parse_from_env` doesn't need the `query_parser` argument passed. This allows for configuring the query parser used for multipart parsing (which by default is different that the one used for parsing query strings).  Without this, if we want to support customizing the multipart query parser, we would have to do so via a separate option (e.g. `:multipart_query_parser`?). If you prefer the separate option approach, I can switch to that.

jeremyevans: It's possible I misread your question. Note that `Multipart::Parser.default_query_parser` is already defined, this is overriding the method to use the multipart query parser that `Rack::Request` uses by default (`Rack::Multipart::ParamList`), while the `Multipart::Parser` default is `Rack::Utils.default_query_parser`.

samuel-williams-shopify: Thanks, I was asking why the Request-specific subclass was necessary. The distinction between the standalone parser default and Request’s ParamList default clarifies it. This approach makes sense to me.

samuel-williams-shopify (approved):

PR #2499, opened 2026-08-13
Require Ruby 2.7.3 and replace ruby2_keywords

samuel-williams-shopify:
Ruby Feature [#22205](https://bugs.ruby-lang.org/issues/22205) records agreement to deprecate `ruby2_keywords`, with warnings beginning in Ruby 4.2 and removal in Ruby 4.4. Therefore, we should remove our usage of `ruby2_keywords`, otherwise Rack 3.3 (tentatively) would break on Ruby 4.4. As it stands, all released versions of Rack will break on Ruby 4.4.

- require Ruby 2.7.3 or newer
- replace `ruby2_keywords` with `...` argument forwarding in `Rack::Builder` and Rack's delegators
- test both Ruby 2.7.3 exactly and the latest Ruby 2.7 patch release
- remove compatibility branches for older Ruby versions
- add coverage for forwarding positional hashes and keyword arguments together

## Why Ruby 2.7.3?

Ruby 2.7 introduced `...` argument forwarding. Support for leading arguments such as `def method_missing(name, ...)`, which Rack needs, was backported in Ruby 2.7.3. Ruby 2.7.0 through 2.7.2 therefore cannot parse all of the forwarding definitions used here.

Using 2.7.3 as the minimum preserves compatibility with the final Ruby 2.7 release line while allowing Rack to remove every `ruby2_keywords` call without version-gated forwarding implementations. CI includes both exact Ruby 2.7.3 and `2.7`, which resolves to the latest patch release.

A Ruby 3.0 minimum would provide a cleaner keyword-separation boundary, but is not technically required when `...` is used. The trade-off is retaining support for Ruby 2.7's transitional keyword semantics; the forwarding sites in this PR avoid that ambiguity by forwarding the original arguments unchanged.

`Rack::Headers#except` remains conditional using feature detection because Ruby before 3.0 does not provide `Hash#except`.

## Related work

This is stacked on #2482 so `Rack::Multipart::UploadedFile` is migrated directly to `...` instead of introducing another `ruby2_keywords` dependency. The original commit and authorship from @SeanLF are preserved; if #2482 merges first, that commit will drop out of this PR.

This takes the minimum-version approach discussed in #2484 instead of retaining older Ruby support through version-gated method definitions.

## Verification

- `bundle exec rake`
- targeted forwarding tests for `Rack::Builder`, `Rack::BodyProxy`, `Rack::MockResponse::Cookie`, and `Rack::Multipart::UploadedFile`
- built the gem and verified its `required_ruby_version`


lib/rack/mock_response.rb
```diff
@@ -23,12 +23,9 @@ def initialize(args)
         @secure = args["secure"]
       end
 
-      def method_missing(method_name, *args, &block)
-        @value.send(method_name, *args, &block)
+      def method_missing(method_name, ...)
+        @value.send(method_name, ...)
```
jeremyevans: Not an issue with this PR, but no reason to use a separate argument here, this should change to `def method_missing(...)` to benefit from the generic argument forwarding optimization.

.github/workflows/test.yaml
```diff
@@ -13,10 +13,6 @@ jobs:
         os:
           - ubuntu-latest
         ruby:
-          - '2.4'
-          - '2.5'
-          - '2.6'
-          - '2.7'
```
jeremyevans, line 19: Let's keep 2.7 unless there is a reason that it won't work with `...` argument forwarding.

lib/rack/multipart/uploaded_file.rb
```diff
@@ -74,12 +74,9 @@ def respond_to_missing?(*args)
       end
 
       # Delegate method missing calls to the tempfile.
-      def method_missing(method_name, *args, &block) #:nodoc:
-        @tempfile.__send__(method_name, *args, &block)
+      def method_missing(method_name, ...) #:nodoc:
```
jeremyevans: Also use `def method_missing(...)` here.

jeremyevans (changes requested):

jeremyevans:
Note that Matz agreed to the deprecation (4.2) and removal (4.4) of `ruby2_keywords`, so it's beyond just a proposal.

.github/workflows/test.yaml
ioquatix: I am fine with that, but in that case, the minimum Ruby version should be 2.7.3 and we should test both `2.7` and `2.7.3` to ensure we don't regress.

lib/rack/mock_response.rb
samuel-williams-shopify: Updated to use generic `def method_missing(...)` forwarding.

.github/workflows/test.yaml
samuel-williams-shopify: Updated the minimum to Ruby 2.7.3. CI now tests both exact `2.7.3` and `2.7` (the latest patch release).

lib/rack/multipart/uploaded_file.rb
samuel-williams-shopify: Updated this delegator to use generic `def method_missing(...)` forwarding as well.

jeremyevans (approved):

eregon:
> Why Ruby 2.7.3?

Indeed `def foo(name, ...)` is 2.7.3+, I got the version wrong in https://bugs.ruby-lang.org/issues/22205#note-14 and now updated that :)
(I checked with `docker run -it ghcr.io/ruby/all-ruby /all-ruby/all-ruby -e 'class A; def method_missing(name, ...); name; end; end'`)

PR #2504, opened 2026-09-02
Read the q weight regardless of parameter ordering

dylanpulver:
`Rack::Utils.q_values` matches `/\Aq=/` against the whole parameter string, so a weight is only seen when `q` is the first parameter. RFC 9110 Section 12.5.1: *"Recipients SHOULD process any parameter named `q` as weight, regardless of parameter ordering."*

Chrome's default navigation Accept header ends with `application/signed-exchange;v=b3;q=0.7` (the string Chromium's own `content/browser/network/accept_header_browsertest.cc` asserts). Rack reads that entry as `1.0`:

```ruby
chrome = "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7"
Rack::Utils.best_q_match(chrome, %w[application/signed-exchange text/html])
# main => "application/signed-exchange", expected "text/html"
```

The fix splits the parameter list and applies the same anchored pattern per parameter.

- **The ReDoS fence holds.** 934ed4fb deliberately replaced the regex splits here with literal ones. This keeps that, and I measured it linear out to 64k parameters.
- **Dropping the `\A` anchor does not work.** It reads `text/plain;seq=5` as `q=5.0` and lets it override a real later weight.
- **Blast radius.** Of 635 generated headers, 259 change and every one has a parameter before the weight. All 292 with `q` first or no `q` are byte-identical.
- **Still not quoted-string aware** (`foo="a;b"` splits inside the quotes), which is pre-existing.

`bundle exec rake test`: 2486 runs, 12146 assertions, 0 failures, 4 skips.

Drafted with Claude Opus 5 (`claude-opus-5`); the RFC reading, the measurements and the review are mine.


lib/rack/utils.rb
```diff
@@ -134,12 +134,20 @@ def build_nested_query(value, prefix = nil)
...
+    # looked for in every parameter rather than only the first one. Parameters
+    # are split on a literal ';' and matched with an anchored pattern to keep
+    # this linear in the length of the header.
     def q_values(q_value_header)
       q_value_header.to_s.split(',').map do |part|
         value, parameters = part.split(';', 2).map(&:strip)
         quality = 1.0
-        if parameters && (md = /\Aq=([\d.]+)/.match(parameters))
-          quality = md[1].to_f
+        parameters&.split(';')&.each do |parameter|
```
jeremyevans, line 143: I think either we should do this properly and actually parse the parameters (including quote handling, similar to `forwarded_values`), or maybe just change the regexp to `/(\A|;\s*)q=([\d.]+)/`.

jeremyevans (commented):
Note that we have similar code in `Request#parse_http_accept_header`, which we should change at the same time.

dylanpulver:
Done — `Request#parse_http_accept_header` now does the same, in that method's own idiom. Also rebased on main; the only conflict was the CHANGELOG entry, and the line now names both methods.

Test coverage went into the existing `parse Accept-Encoding correctly` and `parse Accept-Language correctly` blocks rather than new ones: `"gzip ; deflate ; q=0.5"` and `"fr ; x=y ; q=0.5"`. Reverting just `request.rb` makes exactly those two fail (four failures, since the file runs them under both `RackRequestTest` and `TestProxyRequest`), so they pin the change rather than passing either way. `rake test` is green.

PR #2505, opened 2026-09-06
Rescue Rack::BadRequest in Rack::MethodOverride

navidemad:
`Rack::MethodOverride` calls `req.POST` on every `POST` request to look for the `_method` parameter, before the application and any exception-handling middleware run. It only rescued an explicit list of parsing errors (`InvalidParameterError`, `ParameterTypeError`, `ParamsTooDeepError`, `IncompatibleEncodingError`, `EOFError`), so the other errors tagged with `Rack::BadRequest` — `Rack::Multipart::BoundaryTooLongError`, `MultipartPartLimitError`, `MultipartTotalPartLimitError` — escaped the middleware stack and turned a malformed client request into a 500.

We hit this in production with `BoundaryTooLongError`: a client sending a `multipart/form-data` boundary longer than 70 characters gets a bare 500 from the app server instead of a 400, and the error lands in error tracking as an unhandled exception.

### Change

- `rescue BadRequest` replaces the explicit list, so every parsing error Rack already tags as a bad request is handled the same way. The `Rack::BadRequest` marker module was introduced for exactly this purpose.
- `rescue EOFError` stays first: `Rack::Multipart::EmptyContentError` inherits from `EOFError` and also includes `BadRequest`, so it must keep its own `"Bad request content body"` message (an existing test covers this).

### Tests

Two new specs in `test/spec_method_override.rb` (boundary of 71 characters, `multipart_part_limit` exceeded). Both raise through the middleware on `main` and pass with this change. Full `rake test`: 2488 runs, 0 failures.

Companion Rails change so that `ActionDispatch::Request#POST` also maps these errors to `ActionController::BadRequest`: https://github.com/rails/rails/pull/58677.


CHANGELOG.md
```diff
@@ -31,6 +31,7 @@ All notable changes to this project will be documented in this file. For info on
 
 ### Fixed
 
+- `Rack::MethodOverride` now rescues every `Rack::BadRequest` error raised while parsing the request body (including `Rack::Multipart::BoundaryTooLongError`, `MultipartPartLimitError` and `MultipartTotalPartLimitError`) instead of an explicit list, so a malformed multipart body no longer escapes the middleware as a 500. ([#2505](https://github.com/rack/rack/pull/2505), [@navidemad](https://github.com/navidemad))
```
jeremyevans:
```suggestion
- `Rack::MethodOverride` now rescues `Rack::BadRequest` to avoid the middleware raising an exception for invalid request bodies. ([#2505](https://github.com/rack/rack/pull/2505), [@navidemad](https://github.com/navidemad))
```

jeremyevans (approved):
Looks good to me other than the CHANGELOG entry should be shorter.

ioquatix (approved):

PR #2508, opened 2026-09-09
Fix NoMethodError on Content-Type with an empty parameter segment

### LINKED ISSUE #2274: NoMethodError in Rack::MediaType.params with incorrect Content-Type
# Context
We integrated the rack-utf8_sanitizer gem into our project. However, requests with incorrect Content-Type headers are now resulting in a 500 error.

The issue arises because rack-utf8_sanitizer calls the content_charset method in the `Rack::Request::Helpers`, which leads to the following stack trace:
```
/path/to/ruby/gems/rack-3.1.8/lib/rack/media_type.rb:44:in `strip_doublequotes'
/path/to/ruby/gems/rack-3.1.8/lib/rack/media_type.rb:37:in `block in params'
/path/to/ruby/gems/rack-3.1.8/lib/rack/media_type.rb:33:in `each'
/path/to/ruby/gems/rack-3.1.8/lib/rack/media_type.rb:33:in `each_with_object'
/path/to/ruby/gems/rack-3.1.8/lib/rack/media_type.rb:33:in `params'
/path/to/ruby/gems/rack-3.1.8/lib/rack/request.rb:451:in `media_type_params'
/path/to/ruby/gems/rack-3.1.8/lib/rack/request.rb:459:in `content_charset'
```

# Steps to Reproduce:
1. Run the following command in an IRB session:
```
Rack::MediaType.params("application/vnd.api+json, something")
```
2. Observe the NoMethodError.
```
full_message': undefined method `start_with?' for nil (NoMethodError)

          (str.start_with?('"') && str.end_with?('"')) ? str[1..-2] : str
          ^^^^^^^^^^^^
 ```

# Suggested fix
Ensure the method handles unexpected input gracefully by returning a hash with the key and an empty string value, like `{"something" => ""}`, instead of raising an error.

### PR DESCRIPTION
**Pull request by youdie006**
## Problem

`Rack::MediaType.params` raises `NoMethodError` on a Content-Type that contains an empty parameter segment:

```ruby
Rack::MediaType.params('application/text;;charset=utf-8')
# => NoMethodError: undefined method `downcase!' for nil:NilClass
#    lib/rack/media_type.rb:40:in `block in params'
```

`"".split('=', 2)` returns `[]`, so `k, v = s.split('=', 2)` leaves `k` nil and `k.downcase!` raises.

This was reported before. #2273 reported exactly `application/text;;charset=utf-8` and was closed as a duplicate of #2263, and #2274 was closed by its reporter with "Apparently it was already fixed!". But #2263 (05e3997a) only made the **value** side tolerant, via `strip_doublequotes`. The key side is still unguarded, so both reports are still reproducible on `main` today (a9833c8f).

There is a second, independent instance of the same shape in the multipart parser, and this one is reachable through the public API:

```ruby
body = "--b\r\nContent-Disposition: form-data; name=\"f\"\r\nContent-Type: text/plain;;charset=utf-8\r\n\r\nhello\r\n--b--\r\n"
env  = Rack::MockRequest.env_for('/', 'CONTENT_TYPE' => 'multipart/form-data; boundary=b',
                                      'CONTENT_LENGTH' => body.bytesize.to_s, :input => StringIO.new(body))
Rack::Multipart.parse_multipart(env)
# => NoMethodError: undefined method `strip!' for nil:NilClass
#    lib/rack/multipart/parser.rb:579:in `block in tag_multipart_encoding'
```

So a request body from the network crashes the parser.

## Fix

Two lines.

**`lib/rack/media_type.rb`** - skip segments that carry no key at all. That is the behaviour @AllyMarthaJ preferred in the #2263 review ("I'm actually thinking it might be better to omit it from the hash because actually we don't know whether it's a key, value, neither, or both"), and it leaves the existing `';=utf-8' => { '' => 'utf-8' }` behaviour untouched - only the no-key case is skipped.

**`lib/rack/multipart/parser.rb`** - `tag_multipart_encoding` already has the right guard. 83fff582 added `next unless v`, but placed it one line *below* `k.strip!`, so it never gets the chance to fire. Moving it up one line is behaviour-preserving: `k` is nil only when `v` is nil too, and `String#split` returns fresh strings, so the in-place `strip!` was never observable by the caller.

I also added the documentation @ioquatix asked for in the #2263 review, describing what happens to a segment with no key.

## Verification

Ruby 3.2.3, base a9833c8f.

- Red: the new `spec_media_type.rb` table gives `30 runs, 0 failures, 5 errors`, each `NoMethodError` at `media_type.rb:40`. Green: `30 runs, 30 assertions, 0 failures, 0 errors`.
- Red on the multipart side: the standalone repro above raises at `parser.rb:579`; after the fix it returns `{"field"=>"hello"}`.
- Mutation, media_type - deleting `next if k.nil?` fails exactly the 5 interior-empty-segment rows and leaves `application/text;=utf-8` passing; over-guarding to `next if k.nil? || k.empty?` fails exactly `application/text;=utf-8` and nothing else. Disjoint, so the table pins both sides of the boundary rather than just one.
- Mutation, multipart - reverting the reorder fails only the new empty-segment test; deleting `k.strip!` entirely fails the three pre-existing charset tests (`spec_multipart.rb`) that the reorder must not disturb.
- Full suite: `rake test:regular` -> `1263 runs, 5418 assertions, 0 failures, 0 errors, 2 skips` (2 skips pre-existing; baseline before my multipart test was 1262/5416).

Closes #2274. Also covers the case #2273 reported.

---

Disclosure: I used an AI assistant while preparing this change. I read, ran and verified everything above myself, and I am happy to adjust the approach - in particular, if you would rather `MediaType.params` keep an empty-string key for `;;` instead of skipping the segment, that is a one-word change.


lib/rack/media_type.rb
```diff
@@ -31,12 +31,16 @@ def type(content_type)
...
+      #
+      # Segments that carry no key at all (e.g., the empty segment in
+      # "text/plain;;charset=utf-8") are skipped.
       def params(content_type)
         return {} if content_type.nil? || content_type.empty?
 
         content_type.split(SPLIT_PATTERN)[1..-1].each_with_object({}) do |s, hsh|
           s.strip!
           k, v = s.split('=', 2)
+          next if k.nil?
```
jeremyevans: No need to worry about nil/false difference here, so let's use the more idiomatic:

```suggestion
          next unless k
```

lib/rack/media_type.rb
```diff
@@ -31,12 +31,16 @@ def type(content_type)
       # lack a value (e.g., "text/plain;charset=" will return { 'charset' => '' },
       # and "text/plain;charset" will return { 'charset' => '' }, similarly to 
       # the query params parser (barring the latter case, which returns nil instead)).
+      #
+      # Segments that carry no key at all (e.g., the empty segment in
+      # "text/plain;;charset=utf-8") are skipped.
```
jeremyevans: Not worth documenting.

```suggestion
```

test/spec_media_type.rb
```diff
@@ -80,4 +80,27 @@
       Rack::MediaType.params(@content_type)['charset'].must_equal ''
     end
   end
+
+  # Empty parameter segments (";;", "; ;", ",,") have no key at all.
```
jeremyevans:
```suggestion
```

test/spec_multipart.rb
```diff
@@ -1442,6 +1442,19 @@ def initialize(*)
...
   end
 
+  it "handles Content-Type with an empty parameter segment" do
+    body = "--boundary\r\nContent-Disposition: form-data; name=\"field\"\r\nContent-Type: text/plain;;charset=us-ascii\r\n\r\ndata\r\n--boundary--\r\n"
+    env = Rack::MockRequest.env_for("/", {
+      "CONTENT_TYPE" => "multipart/form-data; boundary=boundary",
+      "CONTENT_LENGTH" => body.bytesize.to_s,
+      :input => StringIO.new(body)
+    })
+    # The charset after the empty segment must still be applied.
```
jeremyevans:
```suggestion
```

jeremyevans (approved):
This looks good to me after a few minor tweaks.

youdie006:
All four applied — `next unless k`, and the three comments dropped.

Re-ran both affected specs after the change: `spec_media_type.rb` 30 runs / 30 assertions and `spec_multipart.rb` 103 runs / 331 assertions, 0 failures and 0 errors on each.

PR #2511, opened 2026-09-19
Fix Rack::Deflater crash on Rack 3 streaming (call-only) bodies

### LINKED ISSUE #2470: Rack::Deflater crashes (NoMethodError: each) on Rack 3 streaming/callable bodies

hikmetba-bit:
## Summary

Fixes #2470.

`Rack::Deflater` unconditionally wrapped the response body in `GzipStream`
when negotiating gzip, then `GzipStream#each` called `@body.each`. A Rack 3
**Streaming Body** (one that responds to `#call(stream)` but not `#each`,
per `SPEC.rdoc`) has no `#each`, so this raised `NoMethodError` once the
wrapped body was consumed, and the client received a truncated ~20-byte
gzip stream (just the `GzipWriter` header/trailer) before the exception
unwound.

## Fix

This implements the conservative "detect and skip" option from the issue:
only wrap the body in `GzipStream` when it responds to `#each` (or is a
`File`); otherwise pass the streaming body through unchanged and skip the
`content-encoding` header, instead of crashing.

A fuller fix that actually gzip-compresses Streaming Bodies (option 1 in
the issue, matching the `Rack::Events` precedent in #2375) is a larger,
separate change — happy to follow up with that if maintainers prefer it,
this PR just stops the crash.

## Testing

- Added a regression test in `test/spec_deflater.rb` using a `#call`-only
  body and asserting: (a) no exception, (b) `content-encoding` is not set
  to `gzip`, (c) the original body is returned untouched and still callable.
- Verified the new test fails (`Expected "gzip" to be nil`) against the
  code without this fix, and passes with it.
- Ran the full test suite (`bundle exec rake test`, minus `test/spec_multipart.rb`,
  which this Windows sandbox's antivirus intermittently quarantines on load —
  unrelated to this change): 1143 runs, 9 failures / 4 errors, all pre-existing
  on unmodified `main` too (Windows-specific `Rack::Files`/`Rack::Directory`/
  encoding tests, unrelated to `Deflater`) — no new failures introduced.
- `bundle exec ruby -Ilib -Itest test/spec_deflater.rb`: 30 runs, 114
  assertions, 0 failures.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


lib/rack/deflater.rb
```diff
@@ -73,9 +73,15 @@ def call(env)
 
       case encoding
       when "gzip"
-        headers['content-encoding'] = "gzip"
-        headers.delete(CONTENT_LENGTH)
-        response[2] = GzipStream.new(body, GZIP_MTIME, @sync)
+        # GzipStream only knows how to consume an Enumerable Body (#each) or
+        # a File. A Streaming Body (#call-only, see SPEC.rdoc) has no #each
+        # to iterate, so leave it untouched rather than crash inside
+        # GzipStream#each with a NoMethodError.
```
jeremyevans, line 79: This belongs in the commit message, not a code comment.

```suggestion
```

CHANGELOG.md
```diff
@@ -35,6 +35,7 @@ All notable changes to this project will be documented in this file. For info on
 - `Rack::Multipart::UploadedFile` now delegates keyword arguments to the wrapped tempfile. Calls such as `uploaded_file.readlines(chomp: true)` raised `TypeError` on Ruby 3.0+. ([#2481](https://github.com/rack/rack/issues/2481), [#2499](https://github.com/rack/rack/pull/2499), [@SeanLF](https://github.com/SeanLF))
 - Multipart parser: limit MIME header size check to the unread buffer region to avoid false `multipart mime part header too large` errors when previously read data accumulates in the scan buffer. ([#2392](https://github.com/rack/rack/pull/2392), [@alpaca-tc](https://github.com/alpaca-tc), [@willnet](https://github.com/willnet), [@krororo](https://github.com/krororo))
 - Multipart parser: add nil guards to prevent `NoMethodError` crashes when handling `Content-Disposition` without parameters and `Content-Type` parameters without '='. ([@haruki0409](https://github.com/haruki0409))
+- `Rack::Deflater` no longer raises `NoMethodError` when wrapping a Streaming Body (one that responds to `#call` but not `#each`); it now leaves such bodies uncompressed instead of crashing. ([#2470](https://github.com/rack/rack/issues/2470))
```
jeremyevans, line 38:
```suggestion
- Make `Rack::Deflater` no longer raise exception for streaming bodies. ([#2470](https://github.com/rack/rack/issues/2470))
```

